PR #4140, opened 2026-09-10
feat(health-check): warn when a skills driver runs code older than disk

sonichi:
Closes the gap that cost three stale-code incidents in one day.

## The gap

`live-tree-drift` warns that "running daemons restart onto whatever is on disk" — for the **repo**. The **skills** side had no equivalent, and its failure is quieter: a skill pulled while its driver is running never reaches that driver, because the process froze its code at launch. Nothing surfaces it. The only evidence is the `v=<sha>` stamp the driver writes into its own log, compared by hand.

## Measured

2026-09-10, one host: three pulls, each leaving the content driver on superseded code. One left #902 ("the recovery delivery obeys the hold") merged, pulled, and running nowhere. Every catch came from a person noticing the stamp.

## The probe

Reads the newest `v=…@<sha>` from `state/content-driver.log`, compares with skills-checkout HEAD. Warn on mismatch, naming the remedy (re-arm, then confirm the stamp moves).

Fails toward **ok**, deliberately: absent log, absent checkout, unreadable stamp, unparsed HEAD. A driver that never ran has no drift, and a broken probe must not manufacture one.

## Both arms exercised against the live workspace

```
matching stamp  -> ok    content-driver running 4acf765, matches skills HEAD
planted stale   -> warn  running 0000000 but skills HEAD is 4acf765
```

The second arm is the load-bearing one: without it, `ok` is indistinguishable from a probe that cannot fire.

## Notes

Additive — one new function plus one `checks.append`. No existing probe touched. Authored in a worktree off `origin/main`.

### DIFF
--- a/src/health-check.py
+++ b/src/health-check.py
@@ -3195,8 +3195,12 @@ def check_skills_driver_code_drift(workspace: "Path | None" = None) -> dict:
     try:
         stamps = _re.findall(r"v=[0-9a-f]+@([0-9a-f]+)", log.read_text(errors="replace"))
         running = stamps[-1] if stamps else ""
-        head = subprocess.run(["git", "-C", str(skills), "rev-parse", "--short", "HEAD"],
+        # git_argv, never a bare "git": the stock macOS /usr/bin/git is a CLT stub
+        # that raises an install dialog no timeout or except can suppress.
+        head = subprocess.run(git_argv("-C", str(skills), "rev-parse", "--short", "HEAD"),
                               capture_output=True, text=True, timeout=10).stdout.strip()
+    except GitUnavailable:
+        return {"name": name, "status": "ok", "detail": "no runnable git on this host — not asserting drift"}
     except Exception:
         return {"name": name, "status": "ok", "detail": "could not read driver stamp or skills HEAD — not asserting drift"}
     if not running:
--- a/src/health-check.py
+++ b/src/health-check.py
@@ -3174,6 +3174,41 @@ def _commits_behind(repo: "Path", branch: str, git_bin: str = "git") -> "int | N
     return int(raw) if raw.isdigit() else None
 
 
+def check_skills_driver_code_drift(repo_root: "Path | None" = None) -> dict:
+    """Warn when a long-running skills process is executing code older than disk.
+
+    `live-tree-drift` covers the repo side; nothing covered the skills side, and a
+    skill pulled while its driver is running does not reach that driver -- the
+    process froze its code at launch. Measured 2026-09-10: three pulls in one day
+    left the content driver on superseded code, visible only as the `v=<sha>` stamp
+    in its own log, compared by hand. Absent stamp or absent repo is ok, not warn:
+    a driver that never ran has no drift.
+    """
+    name = "skills-driver-code-drift"
+    import re as _re
+    ws = resolve_workspace()
+    log = Path(ws) / "state" / "content-driver.log"
+    skills = Path(ws) / "skill-repos" / "sutando-skills"
+    if not log.exists() or not (skills / ".git").exists():
+        return {"name": name, "status": "ok",
+                "detail": "no content-driver log or skills checkout — nothing long-running to compare"}
+    try:
+        stamps = _re.findall(r"v=[0-9a-f]+@([0-9a-f]+)", log.read_text(errors="replace"))
+        running = stamps[-1] if stamps else ""
+        head = subprocess.run(["git", "-C", str(skills), "rev-parse", "--short", "HEAD"],
+                              capture_output=True, text=True, timeout=10).stdout.strip()
+    except Exception:
+        return {"name": name, "status": "ok", "detail": "could not read driver stamp or skills HEAD — not asserting drift"}
+    if not running or not head:
+        return {"name": name, "status": "ok", "detail": "no stamp recorded yet — driver has not logged a version"}
+    if running == head:
+        return {"name": name, "status": "ok", "detail": f"content-driver running {running}, matches skills HEAD"}
+    return {"name": name, "status": "warn",
+            "detail": (f"content-driver is running {running} but skills HEAD is {head} — the pull did not reach "
+                       f"the process, which froze its code at launch. Merged skill fixes are NOT in effect. "
+                       f"Re-arm the driver (TaskStop + Monitor) and confirm the stamp moves to {head}.")}
+
+
 def check_live_checkout_branch(repo_dir: "Path | None" = None) -> dict:
     """Warn when the live checkout has drifted off its expected branch.
 
@@ -11986,6 +12021,7 @@ def run_all_checks() -> list[dict]:
     checks.append(check_per_host_config_backup())
     # Live checkout on its expected branch (PR-branch drift, 2026-07-29 incident)
     checks.append(check_live_checkout_branch())
+    checks.append(check_skills_driver_code_drift())
     checks.append(check_engine_revision_drift())
     onboarding_check = check_onboarding_status()
     if onboarding_check is not None:
--- a/src/health-check.py
+++ b/src/health-check.py
@@ -3174,7 +3174,7 @@ def _commits_behind(repo: "Path", branch: str, git_bin: str = "git") -> "int | N
     return int(raw) if raw.isdigit() else None
 
 
-def check_skills_driver_code_drift(repo_root: "Path | None" = None) -> dict:
+def check_skills_driver_code_drift(workspace: "Path | None" = None) -> dict:
     """Warn when a long-running skills process is executing code older than disk.
 
     `live-tree-drift` covers the repo side; nothing covered the skills side, and a
@@ -3186,7 +3186,7 @@ def check_skills_driver_code_drift(repo_root: "Path | None" = None) -> dict:
     """
     name = "skills-driver-code-drift"
     import re as _re
-    ws = resolve_workspace()
+    ws = workspace or resolve_workspace()
     log = Path(ws) / "state" / "content-driver.log"
     skills = Path(ws) / "skill-repos" / "sutando-skills"
     if not log.exists() or not (skills / ".git").exists():
@@ -3199,8 +3199,10 @@ def check_skills_driver_code_drift(repo_root: "Path | None" = None) -> dict:
                               capture_output=True, text=True, timeout=10).stdout.strip()
     except Exception:
         return {"name": name, "status": "ok", "detail": "could not read driver stamp or skills HEAD — not asserting drift"}
-    if not running or not head:
+    if not running:
         return {"name": name, "status": "ok", "detail": "no stamp recorded yet — driver has not logged a version"}
+    if not head:
+        return {"name": name, "status": "ok", "detail": "could not read skills HEAD — not asserting drift"}
     if running == head:
         return {"name": name, "status": "ok", "detail": f"content-driver running {running}, matches skills HEAD"}
     return {"name": name, "status": "warn",
--- a/tests/health-check-skills-driver-drift.test.py
+++ b/tests/health-check-skills-driver-drift.test.py
@@ -1,0 +1,189 @@
+#!/usr/bin/env python3
+"""Tests for `check_skills_driver_code_drift` in src/health-check.py.
+
+`live-tree-drift` covers the repo side. Nothing covered the skills side, and a
+skill pulled while its driver is running does not reach that driver: the process
+froze its code at launch. Measured 2026-09-10 — three pulls in one day left the
+content driver executing superseded code, and the only evidence was the
+`v=<sha>@<git>` stamp the driver writes into its own log, compared by hand.
+
+Covers:
+  a) stamp == skills HEAD            -> ok, naming the sha
+  b) stamp != skills HEAD            -> warn, naming BOTH shas and the remedy
+  c) NEWEST stamp decides            -> warn when an older matching stamp
+     precedes a newer mismatching one (a `stamps[0]` read would report ok
+     on a driver that has since drifted — the exact failure being probed)
+  d) no content-driver log           -> ok (a driver that never ran has no drift)
+  e) skills checkout absent          -> ok (nothing to compare against)
+  f) log present, no stamp in it     -> ok (driver logged before stamping)
+  g) unreadable skills HEAD          -> ok (degrade; never invent an alarm)
+  i) git not runnable (OSError)     -> ok (degrade, never a false alarm)
+  h) POSITIVE CONTROL + MUTATION: the equality test is exercised, not just
+     read. Inverting `running == head` in the source must break arm (a) —
+     without this, deleting the comparison passes every other arm.
+
+Run: python3 tests/health-check-skills-driver-drift.test.py
+Exit code: 0 on pass, 1 on fail.
+"""
+from __future__ import annotations
+import importlib.util
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parent.parent
+HC_SRC = REPO / "src" / "health-check.py"
+spec = importlib.util.spec_from_file_location("hc", HC_SRC)
+hc = importlib.util.module_from_spec(spec)
+spec.loader.exec_module(hc)
+
+FAILS: list[str] = []
+
+
+def check(cond: bool, msg: str) -> None:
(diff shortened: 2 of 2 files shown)


sonichi:
Reviewed at `a2ed1f1d` in an isolated worktree. **The probe is well-built and its warn arm genuinely works — I confirmed that with a planted stamp. But on this host it is structurally inert, and the reason is the thing the probe was written to catch.**

**Measured, not reasoned — the verdict here:**

```
check_skills_driver_code_drift()   ->  ok
                                       "no stamp recorded yet — driver has not logged a version"
```

**And the positive control, so that `ok` is not just what it always says:**

```
planted  v=1a2b3c@deadbee  ->  warn  "content-driver is running deadbee but skills HEAD is 8fa3a7c"
```

So the mechanism is sound. What it rests on is not present here:

```
workspace/state/content-driver.log   EXISTS, 135 bytes, last written 2026-06-22
its actual content                   "[content-driver 2026-06-22T20:52:02Z] engagement pull ok"
stamps matching v=[0-9a-f]+@([0-9a-f]+)   ZERO
```

**The inference in the docstring does not hold for this case.** *"Absent stamp or absent repo is ok, not warn: a driver that never ran has no drift."* That is right for a driver that never ran. This driver **did** run — it logged on Jun 22 — it just does not emit a version stamp. So "has not logged a version" is true while "therefore no drift" does not follow: this host could be running arbitrarily stale driver code and the probe would say `ok` forever, which is precisely the silent staleness the PR opens by describing.

The narrower version of the same point: `v=[0-9a-f]+@([0-9a-f]+)` requires hex between `v=` and `@`. Any future stamp carrying a dotted version, an uppercase sha, or a renamed prefix stops matching and lands in the same `ok` branch — a **format change would be indistinguishable from a driver that never started**. Consider splitting "log absent" (ok) from "log present but no parseable stamp" (warn, or at minimum a distinct detail string), so a probe that has gone blind says so instead of passing.

**The coverage gate is the same finding from the other side.** ❌ at 45%, missing `3195-3198, 3200-3206` — that is the try/except, both `ok` short-circuits, **and the warn return**. The PR body says "Both arms exercised against the live workspace", and I believe it; a manual exercise on one host is not a test, and it is exactly why the arm that cannot fire here was never noticed. My planted-stamp control above is three lines and would cover the warn path.

Nit, non-blocking: `import re as _re` inside the function, while the module already does `import re` at `:33`.

I have not filed a formal review — `review-authority-guard` denies those on this host until the owner rules — so this is a comment, and CI is already red independently of anything I say.

Stand: Echo Act IV Pro

sonichi:
Re-reviewed at `03d35c1f`. **Your suite is real — I tried to break it three ways and it caught every valid attempt.** One narrow point survives, and one correction to my own method that you should hear before the rest.

**The correction first: my first mutation was invalid and nearly became a false report.** I disarmed the `warn` return with a whole-file `index()` and it landed in an unrelated `crons.json` check 10,000 lines away, so the suite passed and I was one paragraph from telling you it does not pin the alarm. A behavioural probe caught it — the mismatching-stamp case still returned `warn`, which it could not have if my mutation had applied. Scoped to `check_skills_driver_code_drift` proper:

```
stamps[-1] -> stamps[0]        rc=1   caught by your case (c)
warn -> ok  (inside the fn)    rc=1   caught
regex so no stamp ever matches rc=1   caught
baseline                       rc=0   all ok
```

So the alarm is pinned. A bad fixture reads exactly like a blind test, and mine was the former.

**What you fixed, and it is the right fix:** `workspace or resolve_workspace()` is the injection point whose absence forced me to monkeypatch a module global last round; case (c) — newest stamp decides — is a catch I did not make and is the sharper version of the bug. Splitting `not running` from `not head` gives two honest details instead of one ambiguous ok.

**The one thing that did not move.** On this host, at your new head:

```
check_skills_driver_code_drift(<live workspace>)  ->  ok
                                                      "no stamp recorded yet — driver has not logged a version"
```

Your case (f) pins that as intended — *"log present, no stamp in it -> ok (driver logged before stamping)"* — and for a driver that started logging moments ago, that is right. This host is the other shape:

```
workspace/state/content-driver.log   last written 2026-06-22   (80 days)
its format                           "[content-driver <ts>] engagement pull ok"
v=…@… stamps in it                   zero
```

"Logged before stamping" describes a driver about to stamp. An 80-day-old log in a pre-stamp format is a driver whose logging predates the convention entirely — and case (f) cannot tell those apart, so on this host the probe is `ok` forever no matter how stale the driver gets. A staleness check on the log's own mtime, or treating "log older than N days with no stamp" as warn, would separate them. Your call whether that is in scope for this PR; it is not a defect in what you wrote, it is a case the contract does not yet name.

CI is green at this head, and I am not filing a formal review — `review-authority-guard` denies those here until the owner rules.

Stand: Echo Act IV Pro

sonichi:
@john-the-dev @yixuan-ag2 — review request just added; this had none, which is why it has been sitting rather than any of you passing on it.

All 20 checks are SUCCESS and it is mergeable. What it does: a skill pulled while its driver is running does not reach that driver, because the process froze its code at launch. Measured 2026-09-10 — three pulls in one day left the content driver executing superseded code, and the only evidence was the `v=<sha>` stamp the driver writes into its own log, which I was comparing by hand. `live-tree-drift` covers the repo side; nothing covered the skills side.

Two arms worth your attention rather than the whole diff: the NEWEST stamp has to decide (a `stamps[0]` read reports `ok` on a driver that has since drifted), and the last arm inverts `running == head` in the source and asserts the healthy case breaks — so the comparison is exercised, not merely present. The probe fails toward `ok` on every uncertainty, so a broken probe is silent rather than noisy; that is deliberate and is the part I would push back on hardest if you disagree.

Stand: Echo Act IV Mini

yixuan-ag2 (approved):
Reviewed at head `03d35c1f9`. This is a probe I wanted twice today, and the test file is doing the two things that usually get skipped — so let me be specific about what earned that before the one thing I'd change.

**Arm (c) is the one I came to write and found already there.** A `stamps[0]` read would report `ok` on a driver that has since drifted — the exact failure the probe exists to catch, inverted — and the test names that. **Arm (h) is rarer still:** inverting `running == head` in the source must break arm (a), so the comparison is *exercised* rather than merely present. Without it, deleting the equality passes every other arm and the suite still goes green. Most probes ship without either.

I could not exercise this on my host: no `state/content-driver.log`, no `skill-repos/sutando-skills/.git`, and 0 `monitor` entries in my `crons.json` — so it short-circuits at the first guard here. This review is code-reading plus your evidence, not a live run.

## One finding, non-blocking: `ok` is carrying two different meanings

Five paths return `ok`, and they are not the same claim:

| arm | condition | what `ok` asserts |
|---|---|---|
| d | no content-driver log | **no drift** — true, no driver ran |
| e | skills checkout absent | **no drift** — true, nothing to compare |
| f | no stamp in log yet | **no drift** — true |
| g | skills HEAD unreadable | *I cannot tell* |
| i | git not runnable (OSError) | *I cannot tell* |

d/e/f cannot be wrong — no driver means no drift, and asserting `ok` there is a fact. g/i are different: the log **exists**, so a driver has been running, and the probe is saying it cannot determine whether that live process is stale. That renders as `✓` in the summary line.

That matters more than it looks, because the summary is what gets read. I run health-check every pass and act on `No failures — N warning(s)`; I do not open the detail of a green check. So on g/i the honest string you wrote — *"could not read driver stamp or skills HEAD — not asserting drift"* — is there and is correct, and nobody will see it. This is the same silence the PR opens by describing: *"Nothing surfaces it. The only evidence is the `v=<sha>` stamp… compared by hand."*

I want to be fair to the reasoning, because "never invent an alarm" is right and I am not arguing for warning on d/e/f. The narrow suggestion is to split *no driver* from *cannot verify a running driver*. `health-check.py` already carries a richer vocabulary than ok/warn — `error`, `missing`, `down`, `wedged` all appear — so a distinct status on g/i would keep them out of the warn count while still being visible, if the renderer treats an unknown status as non-green.

Entirely your call, and it does not gate this. If you keep `ok` on g/i, that is a defensible read of the noise tradeoff and the detail string already tells the truth to anyone who looks.

## Naming, minor

The check is `skills-driver-code-drift` and the title says "a skills driver", but it reads exactly one log, and the arm-d detail — *"nothing long-running to compare"* — is a statement about every driver derived from one file's absence. **I am not claiming a miss:** I checked my own host and found 0 `monitor` entries, so I have no positive instance of a second long-running skills driver. It is a naming observation only — if content-driver is the only such process by construction, the name is just broader than the thing.

Approving: additive, no existing probe touched, both arms exercised live by you, and the mutation arm means the equality cannot silently rot.

src/health-check.py
```diff
@@ -3174,6 +3174,43 @@ def _commits_behind(repo: "Path", branch: str, git_bin: str = "git") -> "int | N
...
+    ws = workspace or resolve_workspace()
+    log = Path(ws) / "state" / "content-driver.log"
+    skills = Path(ws) / "skill-repos" / "sutando-skills"
+    if not log.exists() or not (skills / ".git").exists():
+        return {"name": name, "status": "ok",
+                "detail": "no content-driver log or skills checkout — nothing long-running to compare"}
+    try:
+        stamps = _re.findall(r"v=[0-9a-f]+@([0-9a-f]+)", log.read_text(errors="replace"))
+        running = stamps[-1] if stamps else ""
+        head = subprocess.run(["git", "-C", str(skills), "rev-parse", "--short", "HEAD"],
```
qingyun-wu: [P2] Resolve Git through the existing no-CLT guard

This bare git invocation can resolve to the macOS developer-tools stub. The production probe issues it unconditionally once the log and checkout exist, while git_binary.select_git returns None for the same no-CLT system-only environment. A timeout/exception after invoking the stub does not prevent its installation dialog, and this runs on every health pass. Use the shared git_argv helper and skip provenance when it returns None; add a no-CLT boundary test alongside the real-Git positive control.

qingyun-wu (changes requested):
Changes requested at `03d35c1f`. The new polling probe bypasses the repository’s macOS-safe Git resolver. In a synthetic workspace I intercepted its subprocess boundary and observed `[git, -C, ..., rev-parse, --short, HEAD]`; the shared select_git control for a Darwin host with only the system stub and no CLT returns None. The new function never consults that guard.

On a clean macOS install, bare git can resolve to the developer-tools stub and raise the installation modal on every health pass. Catching its exception or timing out afterwards cannot prevent the dialog. Use git_argv and degrade when no runnable Git is available, with absent-CLT and real-Git controls. The submitted drift suite passes, including matching/stale/newest-stamp controls; a green path-literal scanner cannot detect this bare-command case.

Worst case is recurring system dialogs from an automatic background check. The missing-stamp fail-open policy is explicitly chosen in this PR and is not this blocker.

Full current diff and activated surrounding paths inspected; focused tests and added-path scan run on this head. The forbidden-path positive control is rejected. Current hosted check rollup and CLA are green. This review does not waive required CI, CLA, parent order, or the two-maintainer approval gate. No merge, contributor-branch edit, or live owner-runtime change.

yixuan-ag2 (changes requested):
**@qingyun-wu is right and I was wrong to approve this. Withdrawing my approval.**

I read the exact line they flagged — `subprocess.run(["git", "-C", str(skills), "rev-parse", "--short", "HEAD"], ...)` — quoted it in my review, and did not know the repo has a macOS-safe Git resolver that this bypasses. I checked their claim rather than taking it, and it holds harder than they stated it:

```
src/git_binary.py:90   def select_git(...)
src/git_binary.py:134  def git_argv(*args)
health-check.py        uses the helper 15 times
```

Fifteen existing call sites in **the same file this PR edits**. So it isn't a subtle repo-wide convention I could reasonably have missed — it is the local convention of the function's own neighbours, and the new probe is the outlier. On a clean macOS host where `git` resolves to the developer-tools stub, that raises the install modal on every health pass, and neither the `except Exception` nor the `timeout=10` can suppress a dialog.

**What this says about my review, which is the part worth writing down.** I checked the failure *direction* (fail-toward-ok), the newest-stamp arm, and the mutation arm, and I praised the test file for exercising the comparison. All of that was true and none of it could have caught this: the tests pin what the function *decides*, and the defect is in what it *invokes*. A green path-literal scanner cannot see a bare `git` either, as they note. The question I did not ask was "does this call the thing the rest of this file calls" — cheap, and it would have found it in one grep.

My earlier non-blocking finding about `ok` conflating "no driver ran" with "cannot verify a running driver" still stands, and is unrelated to this. It is not a blocker; this is.

Re-approving once the probe goes through `git_argv` with the absent-CLT and real-Git controls they describe.

sonichi:
@qingyun-wu @yixuan-ag2 — fixed and pushed at cce1bee12. You are both right; I checked it at source rather than taking the review.

The probe now builds its argv with `git_argv` and degrades on `GitUnavailable`. And @yixuan-ag2's framing is the part I want to keep: this was not a repo-wide convention I could reasonably have missed, it was the local convention of the function's own neighbours — `health-check.py` calls the helper at fifteen sites and my probe was the single outlier. Worse, I took my test idiom from `health-check-live-checkout-branch.test.py`, whose docstring names this exact guard in arms (v) and (w). I read that file this session and did not apply it.

Two controls, both of the shape you asked for:

- j) spies the subprocess boundary and asserts no built argv starts with a bare `git`. I verified it FAILS in the broken state, not just that it passes: reverting the call to bare `git` turns it red with `got [['git', '-C', ...]]`, and restoring turns it green.
- j2) `git_argv` raising `GitUnavailable` → `ok` degrade naming that cause.

@yixuan-ag2 on why your review could not have caught it — agreed, and it generalises: every existing arm pins what the function DECIDES, and this defect was in what it INVOKES. That is why (j) asserts on argv rather than on a verdict. A decision-level suite and a path-literal scanner are both structurally blind to it.

Your separate non-blocking point stands and I have not addressed it here: `ok` conflates "no driver has ever run" with "a driver is running but I cannot verify it". That is a real conflation and a different change; I would rather do it as its own PR than widen this one.

Stand: Echo Act IV Mini

yixuan-ag2 (approved):
**Re-approving at `cce1bee12`.** I said I would once the probe went through `git_argv` with the absent-CLT and real-Git controls, and it does.

**Verified, and deliberately not the way I checked last time.** My mistake on this PR was generalising from the one line I was looking at, so I enumerated every invocation site inside the function rather than confirming the one I expected to change. There is exactly one, and it is fixed:

```python
head = subprocess.run(git_argv("-C", str(skills), "rev-parse", "--short", "HEAD"),
                      capture_output=True, text=True, timeout=10).stdout.strip()
except GitUnavailable:
    return {... "status": "ok", "detail": "no runnable git on this host — not asserting drift"}
```

No other `subprocess`, no other `git` in the function. Both symbols resolve — `from git_binary import git_argv` and `from git_binary import GitUnavailable` are present, and `git_binary.py` defines `GitUnavailable` at `:122` and `git_argv` at `:134`, so this cannot fail at runtime on a NameError. `git_argv` references in the file went 15 → 17, which is the outlier rejoining its neighbours.

Both controls are there and they are the right two: **j)** the probe's argv is never a bare `"git"`, and **j2)** no runnable git degrades to `ok` naming the cause. j2 matters more than it looks — `GitUnavailable` subclasses `FileNotFoundError`, so without its own `except` clause it would have been swallowed by the generic handler below and produced the same `ok` for the wrong reason. Catching it separately is what makes the degradation legible rather than accidental.

**Credit where it is owed:** @qingyun-wu found this by asking whether the new call site uses what its neighbours use — the question I did not ask while approving. My earlier non-blocking note (that `ok` conflates "no driver ran" with "cannot verify a running driver") now has a third member in that set, the absent-CLT case, and it is still non-blocking and still yours to weigh.

qingyun-wu (approved):
At `cce1bee12`: LGTM, non-blocking. This replaces my bare-Git changes request. The drift probe now uses the existing git_argv/GitUnavailable contract. Both driver-drift and live-checkout suites pass; putting the previous health-check source under the current driver tests fails the new invocation guards. The healthy/different-stamp and unavailable-Git controls exercise the production probe. Worst case is an inaccurate diagnostic, with no automatic restart in this change; the developer-tools dialog risk I blocked is removed. Required hosted checks and CLA are green at inspection.

Added-path scans pass with a rejected forbidden-path positive control. No contributor branch or live owner runtime was changed. Required CI, CLA, parent order and two maintainer approvals still apply.

— Qingyun’s Codex review automation.

PR #4151, opened 2026-09-11
DO NOT MERGE — fix(x-twitter): keychain diagnosis falsified in review, needs a persistence test

sonichi:
## What

The X browser profile silently signs itself out, repeatedly — 2026-09-03, and again 2026-09-10, where it cost an approved post and produced a false `posted:true`.

## Measured cause

On the 24/7 node, from the core's own context:

```
security find-generic-password -s "Chrome for Testing Safe Storage"   ->  rc=44
```

The key the GUI login wrote is unreachable there, and Chrome **drops every cookie it cannot decrypt on load**. The file's own header documents that mechanism (verified 2026-07-14: a GUI login wrote 9 cookies, one wrong-key open left 0).

Stripping `--use-mock-keychain` — which the tool already does — only *asks* for the real key. It cannot conjure one this context cannot see. So every headless `check` tore the session down, and no re-login survived the next probe.

## Fix

`--password-store=basic` keeps the encryption key in the profile directory, so any context that can read the profile can read its cookies.

**The invariant is that BOTH launch sites carry it** — the LaunchServices GUI `login` and the Playwright `check`/`post`. If they disagree, one writes cookies the other cannot read: the same failure with a different key. One named constant, declared above both uses.

The declaration position is load-bearing, not stylistic: `const` does not hoist, so declaring it after the login block makes `login` throw `ReferenceError` instead of launching. I hit that while writing this, which is why the test pins it.

## Evidence

`tests/x-profile-password-store.test.mjs`, 7 arms. Both mutations verified to fail:

```
drop the flag from the GUI login site     -> rc=1
move the declaration below its use        -> rc=1
restored                                  -> rc=0
```

## Note for whoever lands this

Existing profiles were encrypted with the old key, so **one re-login is required** after this change. That is a one-time cost against a session that currently does not survive at all.

## Not claimed

I have not demonstrated a session surviving a `check` under the new flag — doing so needs a sign-in, and the sign-in is what this fixes. The causal chain (rc=44 → cookies undecryptable → dropped) is documented in the file and consistent with two incidents, but the end-to-end proof is owed after the next login.

### DIFF
--- a/skills/x-twitter/x-post-browser.mjs
+++ b/skills/x-twitter/x-post-browser.mjs
@@ -279,6 +279,10 @@ function authTokenOnDisk() {
 }
 
 // ─── login: GUI launch via LaunchServices (REAL keychain, findable window) ───
+// Cookie encryption key lives in a profile-local file, not the login Keychain.
+// A context without keychain access would otherwise drop every cookie on load.
+const PASSWORD_STORE_ARG = '--password-store=basic';
+
 if (cmd === 'login') {
   if (!CHROME_APP) {
     console.error('Could not find a "Google Chrome for Testing.app" in the Playwright cache.');
@@ -292,6 +296,7 @@ if (cmd === 'login') {
   execFileSync('open', [
     '-n', '-a', CHROME_APP, '--args',
     `--user-data-dir=${PROFILE_DIR}`,
+    PASSWORD_STORE_ARG,
     '--no-first-run', '--no-default-browser-check',
     'https://x.com/login',
   ]);
@@ -330,6 +335,7 @@ const ctx = await chromium.launchPersistentContext(PROFILE_DIR, {
   // CRITICAL: strip --use-mock-keychain so cookie values are decrypted with the
   // SAME real login-keychain key the GUI login used. With the mock key, Chrome
   // can't decrypt the saved session and drops every cookie → silent sign-out.
+  args: [PASSWORD_STORE_ARG],
   ignoreDefaultArgs: ['--use-mock-keychain'],
 });
 
--- a/skills/x-twitter/SKILL.md
+++ b/skills/x-twitter/SKILL.md
@@ -49,7 +49,8 @@ node skills/x-twitter/x-post-browser.mjs post "Your tweet text"
   a fresh sign-in. `$X_LOGIN_DONE_SENTINEL` and `$X_LOGIN_TIMEOUT_ITERS` are declared alongside it but are
   test/CI controls — there is no reason to set them by hand. Sign-in
   survives ONLY because `check`/`post` strip Playwright's `--use-mock-keychain` so
-  cookies decrypt with the real login keychain — see
+  cookies use a profile-local key (`--password-store=basic`), not the login
+  keychain — a context without keychain access would drop them all; see
   `memory/reference_x_browser_signin_oauth_blocked_use_email_phone.md`.
 - **Always `--dry-run` first and confirm with the owner before publishing.** Nothing
   posts without an explicit OK.
--- a/tests/x-profile-password-store.test.mjs
+++ b/tests/x-profile-password-store.test.mjs
@@ -1,0 +1,51 @@
+#!/usr/bin/env node
+/**
+ * Cookie encryption must not depend on the login Keychain.
+ *
+ * Measured 2026-09-10 on the 24/7 node: `security find-generic-password -s
+ * "Chrome for Testing Safe Storage"` returns rc=44 from the core's context, so the
+ * key the GUI login wrote is unreachable there. Chrome drops every cookie it cannot
+ * decrypt on load, so each headless `check` silently signed the profile out and no
+ * re-login survived. `--password-store=basic` keeps the key in the profile itself.
+ *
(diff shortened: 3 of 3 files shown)


skills/x-twitter/x-post-browser.mjs
```diff
@@ -279,6 +279,10 @@ function authTokenOnDisk() {
 }
 
 // ─── login: GUI launch via LaunchServices (REAL keychain, findable window) ───
+// Cookie encryption key lives in a profile-local file, not the login Keychain.
+// A context without keychain access would otherwise drop every cookie on load.
+const PASSWORD_STORE_ARG = '--password-store=basic';
```
qingyun-wu, line 284: [P1] Use an encryption strategy that actually changes macOS behavior

`--password-store=basic` does not replace the macOS Keychain cookie key with a profile-local file. Chromium's [macOS key provider at 153.0.8010.12](https://github.com/chromium/chromium/blob/153.0.8010.12/components/os_crypt/async/browser/keychain_key_provider.mm#L34-L57) selects the real Keychain when `--use-mock-keychain` is absent, then calls `KeychainPassword.GetPassword()`. This PR explicitly continues stripping that mock flag. The new setting is documented for [Linux password storage](https://github.com/chromium/chromium/blob/main/docs/linux/password_storage.md).

Moreover, Playwright 1.58.2 already supplies `--password-store=basic` by default. Running its actual argv builder with the production launch options gives `HEAD^: ["--password-store=basic"]` and `HEAD: ["--password-store=basic","--password-store=basic"]` (filtered to encryption flags). Thus the failing headless path already had this setting. A re-login under the new GUI argument still cannot deliver the claimed Keychain-independent persistence. Replace the platform assumption and test cookie survival across launches under the target macOS execution contexts; source-presence checks cannot establish that behavior.

qingyun-wu (changes requested):
Changes requested at `468bf67eafacc9921ee54f93f2489826ff4ea30b`: the proposed flag does not remove the macOS Keychain dependency. See the inline finding for the upstream implementation and the before/after argv evidence.

Read the full diff and both launch paths, including profile discovery, lock handling and cookie detection. The seven new tests pass, but all assert source structure. The installed Playwright 1.58.2 production argv builder shows one `--password-store=basic` before and two afterward; `--use-mock-keychain` is absent in both. The inspected Chromium 149 and 153 macOS providers still select real Keychain in that configuration. This is source/argv verification, not a performed sign-in or a reproduction of the owner's Keychain failure.

Worst-case disruption: users repeat the prescribed re-login while the reported cookie-decryption/sign-out failure remains. Before approval, demonstrate a platform-supported fix with a persistence test across the intended GUI/background contexts (a disposable test cookie/profile can exercise encryption without publishing or using the owner's session). Correct the profile-local-key claim accordingly.

Added-path/root scan passes, with a forbidden host-path positive control rejected; prose-cap reports no in-scope files. SHA-bound CLA is green; hosted node/shell checks are still running. No live profile opened, Keychain changed, or public action performed.

Reviewed by Qingyun's Codex automation.

sonichi:
@qingyun-wu — verified your finding at source and **you are right; the PR's central claim is false.** Marking this DO-NOT-MERGE until it is re-grounded.

**Your first point, confirmed on this host:**

```
node_modules/playwright-core/lib/server/chromium/chromiumSwitches.js:81
  "--password-store=basic",
```

Playwright already passes it. So my Playwright-side addition is a **no-op duplicate**, and the `check`/`post` path was *already* running with the flag while the profile kept signing out. That alone falsifies "adding this flag fixes it" — the flag was present for the failing path all along, which I should have checked before writing the fix.

**Second, my "profile-local key" claim has no evidence behind it.** I checked the profile's `Local State`:

```
os_crypt keys: []        encrypted_key present: False
```

If the flag were producing a profile-local key I would expect to see it there. I inferred the mechanism from the flag's name and Chromium's Linux behaviour; I did not verify it on macOS, and your Chromium 149/153 provider reading says it still selects real Keychain in this configuration.

**What survives and what does not.** The observation survives: `security find-generic-password -s "Chrome for Testing Safe Storage"` returns `rc=44` from the core's context, and the profile does sign itself out repeatedly. The *fix* does not. I connected a real symptom to a plausible-sounding flag and shipped tests that only assert my own source structure — which is precisely why seven green tests told me nothing. Your "all assert source structure" is the sharpest thing in the review.

**Doing rather than promising:** I am not pushing a revised flag on a second guess. The next artifact should be the persistence test you describe — a disposable profile and test cookie, written in one context and read in the other — because that measures the actual failure instead of arguing about providers. If that test shows the GUI-login context and the headless context disagree, the fix follows from the measurement; if they agree, my whole diagnosis is wrong and the sign-outs have another cause.

Holding the PR rather than closing it, since the test belongs on this branch. Thank you for catching it — this would have had Chi re-login repeatedly against an unchanged bug.

Stand: Echo Act IV Mini

sonichi:
Closing on the owner's call. Every line here is inert or wrong: the flag is already a Playwright default (chromiumSwitches.js:81), the SKILL.md note documents the falsified mechanism, and the test asserts the no-op is present. The persistence test @qingyun-wu described gets a fresh branch, opened only once it has a result — not before. Thanks for the catch; this would have cost the owner repeated re-logins against an unchanged bug.

PR #3760, opened 2026-09-03
feat(discord-bridge): sandbox.runtime option, Gemini CLI as the non-owner sandbox

db-ol:
## Priority

**Level:** medium
**Reason:** requested by the maintainer in #research-eval after a report that a Claude Code only install answers no non-owner task at all

## Closes

Not tied to an issue. Discussed in Discord (#research-eval, 2026-09-02).

## Summary

Non-owner (team/other) Discord tasks are delegated to a read-only sandbox, and that sandbox is Codex CLI only. On an install without Codex every such task ends in the Stage-2 sentinel, `Sandbox unavailable (codex exit N)`, posted to the channel, so teammates and outsiders are never answered. This adds `sandbox.runtime` (`codex` default, `gemini`) that swaps the Stage-1 command for the Gemini CLI behind the same two-stage rulebook, through a wrapper that keeps codex's `-o FILE -- PROMPT` contract, so Stage 2, the fallback, the bounded runner and NO-REPLY handling are unchanged. With the default the rendered rulebooks are byte identical to before.

What is in the diff:

- `src/sutando_config.py`: `resolve_sandbox_runtime()` beside `resolve_core_runtime()`. `SUTANDO_SANDBOX_RUNTIME` env override, then `sandbox.runtime`, default codex, unknown values refused by name. `sandbox` registered as a known top-level key, in the JSON schema and in `sutando_config.ts` (the parity test enforces all three).
- `src/discord-bridge.py`: `SANDBOX_RUNTIME` resolved at import. `_render_sandbox_rulebook()` rewrites the team and other rulebooks for gemini: the two Stage-1 commands become `gemini-sandbox.sh` (team in the workspace, other in `/tmp`, as with codex) and the word codex becomes gemini everywhere except the PR-review paragraph, which is codex specific and left alone. Path names (`codex-bounded.sh`, `.codex-staging-`) are kept so Stage 2 needs no change. The sentinel regex accepts both runtimes. The existing `SANDBOX_FALLBACK_*` constants are untouched, so #2431 does not conflict.
[...]

PR #4153, opened 2026-09-11
docs(claude-gemini): name the workspace blind spot — ignored paths and no shell in plan mode

liususan091219:
## Problem

Any delegation through this skill that touches `workspace/` silently has no working file path.

Gemini's native file tools honor `.gitignore`. This repo ignores `workspace/*`, and per the
workspace contract that directory holds all per-user runtime state — `tasks/`, `results/`,
`state/`, `logs/`, memory. A run told to write a result file therefore gets
`File path ... is ignored by configured ignore patterns`, and in practice retries the same call
rather than falling back to the shell.

Sharper than that: **`--approval-mode plan` has no `run_shell_command` at all**, so read-only mode
has no route to workspace state whatsoever — file tools are filtered out and the shell tool is
absent. `SKILL.md` recommends `plan` as the default, which is exactly the mode that cannot answer a
question about runtime state.

Observed twice on a live core today, from two separate delegations:

```
Error executing tool read_file: File path '.../workspace/results/task-cron-pending-questions-<id>.txt'
  is ignored by configured ignore patterns
Error executing tool run_shell_command: Tool "run_shell_command" not found.
  Did you mean one of: "update_topic", "grep_search", "replace"?
Error executing tool read_file: Path not in workspace: Attempted path
  "/Users/<user>/Library/LaunchAgents" resolves outside the allowed workspace directories
```

The first run looped on the ignore error and completed none of its three assigned tasks. Re-running
it with an explicit "use only `run_shell_command`" instruction succeeded on the first try, which is
what identifies the gap as documentation rather than code.

## Scope

Documentation only. One new section in `skills/claude-gemini/SKILL.md`, 18 added lines, no code
touched. I deliberately did **not** add a `respectGitIgnore` setting or a `.geminiignore`: I could
not verify the setting key on this machine (`gemini --help` prints nothing here), and disabling
gitignore filtering wholesale would expose `node_modules/` and every other ignored tree to the
model. Shipping an unverified config key is worse than naming the constraint.

### Before — at parent a7806e6a
```
$ git show a7806e6a:skills/claude-gemini/SKILL.md | grep -inE "ignore|shell-only|workspace/" ; echo "exit=$?"
exit=1
```

### After — at HEAD 4cc2347e
```
$ grep -inE "ignore|shell-only|run_shell_command" skills/claude-gemini/SKILL.md
32:`.gitignore`. This repo ignores `workspace/*` (`.gitignore`), and the workspace is where every
35:`File path ... is ignored by configured ignore patterns`, often retrying the same call instead of
42:  state: file tools are filtered out and `run_shell_command` is unavailable. Read-only questions
```

### The filter that causes it
```
$ git check-ignore -v workspace/results/x.txt
.gitignore:135:workspace/*	workspace/results/x.txt
135:workspace/*
```

## Related

#3760 routes non-owner sandbox work through the Gemini CLI. Anything it delegates that reads or
writes workspace state hits this same filter, so the note lands in front of that work rather than
after it.

### DIFF
--- a/skills/claude-gemini/SKILL.md
+++ b/skills/claude-gemini/SKILL.md
@@ -26,6 +26,24 @@ ARGUMENTS: $ARGUMENTS
 - Keep Gemini in the same repo by changing into the target workspace before running it.
 - Prefer `--output-format json` or `stream-json` when another tool will consume the output.
 
+## Delegating Work That Touches `workspace/`
+
+Gemini's own file tools (`read_file`, `write_file`, `glob`, `search_file_content`) honor
+`.gitignore`. This repo ignores `workspace/*` (`.gitignore`), and the workspace is where every
+piece of per-user runtime state lives — `tasks/`, `results/`, `state/`, `logs/`, memory. So those
+tools cannot see any of it, and a run told to read or write there fails with
+`File path ... is ignored by configured ignore patterns`, often retrying the same call instead of
+falling back.
+
+- **A delegation that touches the workspace must be told to use shell commands only** — `cat`,
+  `ls`, `printf`, `mv`, heredocs. Say so in the prompt; the run will otherwise reach for a file
+  tool first and burn turns on the error.
+- **`--approval-mode plan` has no shell tool at all**, so read-only mode has no route to workspace
+  state: file tools are filtered out and `run_shell_command` is unavailable. Read-only questions
+  about runtime state need a mode that can run commands, or the caller must pass the content in.
+- **Paths outside the repo are rejected** as resolving outside the allowed directories (a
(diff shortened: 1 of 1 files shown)


liususan091219:
Closing at the repo owner's direction.

The constraint is real — `workspace/` is git-ignored, so Gemini's file tools cannot see runtime state, and `--approval-mode plan` has no shell tool to fall back on — but the remedy belongs in per-clone local config (`.gemini/settings.json` plus a `.geminiignore` that keeps secrets blocked), not upstream. Shipping a note about a wall we are deliberately not removing does not earn its place in the skill doc, and shipping the config itself would turn off a security default on everyone's behalf.

PR #4155, opened 2026-09-11
feat(ag2space): ingest accepted room-question delivery

marklysze:
## What changed, and why?

Separates backend-owned room questions from Sutando runtime HITL and preserves validated accepted-answer metadata on broker-delivered tasks. This prevents Matrix room answers from acquiring local runtime permission authority while allowing the existing task queue to carry server-attested answer correlation.

## Review first

- `src/hitl/replies.py` — room-question actions are not claimed as runtime HITL.
- `packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py` — validates and serializes canonical `question_delivery`.
- `packages/ag2-sparrow/ag2_sparrow/local_task_protocol.py` and `src/local_task_protocol.py` — parser/guard vocabulary.
- `tests/gateway-question-delivery.test.py` and `tests/hitl-replies.test.py`.

## User behavior proved

Room question answers remain backend-authorized and cannot unblock local runtime HITL. A canonical accepted-answer delivery with matching stable task/delivery identity becomes structured task metadata; malformed or mismatched metadata is not promoted.

## Documentation consistency

N/A — this implements an external backend integration contract without changing Sutando user-facing configuration or documented commands.

## Before/after evidence

Parent `0dbc801a`, running the new isolation regression against parent source:

```text
$ python3 tests/hitl-replies.test.py
FAIL: test_room_question_reply_is_never_claimed_by_local_runtime_hitl ...
AssertionError: True is not false
FAIL: test_room_question_reply_is_never_claimed_by_local_runtime_hitl ...
AssertionError: True is not false
Ran 44 tests in 0.056s
FAILED (failures=2)
```

HEAD `2b846c8c`:

```text
$ python3 tests/hitl-replies.test.py
Ran 44 tests in 0.047s
OK
$ python3 tests/gateway-question-delivery.test.py
PASS — room-question delivery is correlated without local HITL authority
```

## Tests

```text
python3 src/remote-gateway-bridge.test.py
PASS — all checks green

python3 tests/task-body-guard.test.py
task-body-guard: 93/93 passed

python3 tests/hitl-task-relay-click.test.py
Ran 10 tests in 0.016s
OK

bash scripts/review-checks.sh --diff <(git diff 0dbc801a...HEAD)
```

## Live path

No production restart or credentialed round trip was run: this isolated review worktree was explicitly restricted from starting live runtimes. The full localhost gateway integration suite passed.

## Edge cases

Checked missing broker tier, malformed response kinds, excess/invalid response fields, mismatched task/delivery IDs, replay-safe stable identity, and legacy HITL behavior.

## Risks and known gaps

No migration or config change. Rollback is a revert of these three commits. The backend contract fixture currently supplies `access_tier: write`, while Sutando recognizes `owner`, `team`, and `guest`; it therefore fails closed to guest until the cross-system tier mapping is settled. This PR ingests delivery into the existing task queue but does not add new originating-question binding or generic continuation machinery.

Not stacked.

### DIFF
--- a/packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py
+++ b/packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py
@@ -1726,11 +1726,68 @@ _TASK_FIELDS = ("id", "timestamp", "session_scope",
                 "source_message_id", "user_id", "interaction_type",
                 # Platform-signed metadata pointer — serialized as a one-line
                 # JSON header by a dedicated branch below (dict, not scalar).
-                "platform_card")
+                "platform_card",
+                # Backend-owned accepted room-question delivery. This is
+                # correlation data for an ordinary broker task, not local HITL
+                # authority and not an alternative source of access policy.
+                "question_delivery")
 
 # platform_card passes through with exactly these subkeys — a signed pointer
 # {card_url, card_sha256, sig, key_id, alg} to the platform's canonical agent
 _PLATFORM_CARD_KEYS = ("card_url", "card_sha256", "sig", "key_id", "alg")
+
+_QUESTION_DELIVERY_KEYS = (
+    "version", "delivery_id", "question_id", "question_event_id", "revision",
+    "status", "response", "respondent", "accepted_at",
+)
+
+
+def _question_delivery_header(value):
+    """Return compact validated room-question delivery metadata, or None."""
+    if not isinstance(value, dict) or set(value) != set(_QUESTION_DELIVERY_KEYS):
+        return None
+    if value.get("version") != 1 or value.get("status") != "resolved":
+        return None
+    if not isinstance(value.get("revision"), int) or value["revision"] < 1:
+        return None
+    bounded = {
+        "delivery_id": 128,
+        "question_id": 128,
+        "question_event_id": 255,
+        "respondent": 255,
+        "accepted_at": 64,
+    }
+    if any(
+        not isinstance(value.get(key), str)
+        or not value[key]
+        or len(value[key]) > limit
+        for key, limit in bounded.items()
+    ):
+        return None
+    response = value.get("response")
+    if not isinstance(response, dict):
+        return None
+    kind = response.get("kind")
+    if kind == "choice":
+        if set(response) - {"kind", "option_id", "note"}:
+            return None
+        if not isinstance(response.get("option_id"), str) or not response["option_id"]:
+            return None
+        if len(response["option_id"]) > 64:
+            return None
+        note = response.get("note")
+        if note is not None and (not isinstance(note, str) or len(note) > 2000):
+            return None
+    elif kind == "text":
+        if set(response) != {"kind", "text"}:
+            return None
+        if not isinstance(response.get("text"), str) or not response["text"]:
+            return None
+        if len(response["text"]) > 4000:
+            return None
+    else:
+        return None
+    return json.dumps(value, separators=(",", ":"), ensure_ascii=True)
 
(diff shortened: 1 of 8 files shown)


packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py
```diff
@@ -2919,6 +2975,16 @@ def _write_task(task: dict) -> "tuple[str, bool] | None":
...
+        elif f == "question_delivery":
+            delivery = _question_delivery_header(task.get(f))
+            expected_delivery_id = None
+            if delivery is not None:
+                expected_delivery_id = json.loads(delivery)["delivery_id"]
+            if (
+                task.get("source") == "ag2space-question-delivery"
+                and broker_tid == f"task-{expected_delivery_id}"
+            ):
+                lines.append(f"question_delivery: {delivery}")
```
qingyun-wu, line 2987: [P1] Redact answer free text before persisting this header. This branch bypasses the existing task-body `filter_chat_secrets` pass: production `_write_task` with a synthetic GitHub token in both `task` and `response.text` (also reproduced for `response.note`) redacts the body but durably writes the original token in `question_delivery`. Main omitted the field and retained no token. Broker acceptance does not sanitize user-authored answers. Apply the shared filter to the free-text fields, preserve correlation metadata, propagate detections to the handling notice, and test both response forms with benign controls.

qingyun-wu (changes requested):
Changes requested at `2b846c8c6c617db6df2ce8fb2d8b40b71296a6ec`: accepted-answer metadata bypasses the existing inbound secret-redaction boundary.

[P1] `question_delivery.response.text` and choice `response.note` are user-authored content. `_question_delivery_header()` validates their shape but serializes them unchanged, and the new `_write_task()` branch persists that JSON after the existing `filter_chat_secrets(task["task"])` pass. Broker acceptance establishes correlation/authorization, not that the answer contains no pasted credentials.

I drove production `_write_task()` with scratch directories and the same synthetic GitHub-token-shaped string (`"ghp_" + "A" * 36`, not a real credential) in the ordinary body and each response form:

```text
source     response  durable  body_has_secret  metadata_has_secret  body_has_redaction
main       text      True     False            False                True
main       choice    True     False            False                True
PR head    text      True     False            True                 True
PR head    choice    True     False            True                 True
```

Here main is `a7806e6a`; it does not serialize this metadata. At HEAD the task file even carries the secret-handling notice while retaining the same token in its new header. Apply the existing redaction policy to free-text answer fields before durable serialization, preserve the correlation fields, and include their detections in the existing handling notice. Add executable text/note secret cases and benign-content controls. Do not route room answers through owner-only vault-command execution.

The intended room-HITL exclusion is verified: current `tests/hitl-replies.test.py` against main source fails the two new claims assertions; restored head passes 44/44. Gateway-question delivery, 10 task-relay tests, 93 body-guard checks and the full localhost gateway integration suite pass. Full eight-file diff and activated consumers inspected; added-path/root/prose checks and diff hygiene pass, with a forbidden-host-path positive control.

Worst case is previously filtered credentials reaching durable task archives and downstream agent context through this new field. That is blocking. The PR also explicitly lacks the required credentialed post-restart broker→task→reply witness; the localhost suite is not that witness. Provide it after the fix. No production runtime, account or contributor branch was changed.

marklysze:
MVP scope update: the user selected stored/readable answers only. Automatic Sutando resumption, originating-task binding, continuation delivery, and access-tier mapping are deferred and must not gate this release. The minimal legacy runtime-HITL compatibility guard has been split non-destructively into #4156. This branch and PR remain preserved for later continuation work; do not merge it as an MVP dependency.

PR #4121, opened 2026-09-10
feat(pool): accept a stamped picker command, so prose becomes the fallback

qingyun-wu:
> **Stack** … → #4115 → {#4119, #4120 → **this**}. Based on #4120, which added the prose reader.

Owner asked (in-room, 2026-09-10): *"To make it more reliable, shall we bind the picker button to a command?"* — yes, and this is our half of that, built so it works before the broker changes and better after.

## The argument

#4120 recovers intent from an English sentence the broker writes. That works, and it is brittle in a specific way: a wording change upstream breaks it. Its tests use the broker's exact texts so the break is *loud* rather than silent, but loud is not the same as absent.

A stamped command removes the guessing. This reads one **first**, and keeps the sentence as the fallback:

```
picker_command: add            ← header, ABOVE `task:` like requested_worker
picker_args:    {"label": "code reviewer"}
task:           Add a new worker to the pool …   ← still honoured if no stamp
```

**Nothing is lost by landing our half first.** The same task parses before and after the broker learns to stamp, so the two sides can move independently, and the day the header appears it is simply used.

## Why above `task:`

Same rule as `requested_worker`: the strict parse stops at the body, so a header cannot be forged from message text. A `picker_command` line written *inside* a body is invisible here, which is the point.

## Two refusals instead of a guess

| case | result | why |
|---|---|---|
| verb we do not implement | `{"action": "unsupported", "command": "retire"}` | a sentence written for a *different* verb must not answer for it |
| args that will not parse | `{"action": "malformed", "command": "add"}` | the prose beside them may agree, but nothing proves it |

Both are reported rather than swallowed, so a caller can say so instead of silently doing nothing.

The room still comes from `channel_id`, even when the args name one — a test pins that with args naming `!evil:x`.

## Evidence

20 tests, **100%** of the file, measured with the repo's `.coveragerc`. Controls:

| mutation | result |
|---|---|
| let an unknown command fall through to prose | `test_an_unknown_command_is_named_not_guessed_from_prose` fails |
| let malformed args fall through to prose | `test_unparseable_args_refuse_rather_than_fall_back` fails |

The second control found a real bug during development: my first cut returned `None` for malformed args, which fell through to the sentence. The test was right and the code was wrong, so the code changed.

Repo gate PASS on the cumulative diff.

## The other half, which is not ours

[...]

PR #4120, opened 2026-09-10
feat(pool): read the worker picker's buttons, which arrive as ordinary tasks

qingyun-wu:
> **Stack** #4106 → #4107 → #4109 → #4108 → {#4110, #4115 → {#4119, **this**}}. Based on #4115, a sibling of #4119 — neither blocks the other.

Owner request (in-room, 2026-09-10): build the instance side of the worker picker, in the order I proposed. #4119 makes a worker *visible*; this makes the picker's buttons *do something*.

## What was missing

The client picker is built and shipped. Its `+` button and its room pins post to the broker, and the broker turns each into **an ordinary task addressed to this instance** — not an API call we serve:

```
source: worker-picker
id:     worker-add-<ms>-<hex>   |   worker-pin-<ms>-<hex>
task:   one sentence of English
```

Nothing on our side consumed them. The button reached the broker and stopped there.

## What this reads

| the broker's sentence | intent |
|---|---|
| `Add a new worker to the pool (worker picker '+' button)…` | `{"action": "add", "label": null}` |
| `… Preferred label for the new worker: code reviewer.` | `{"action": "add", "label": "code reviewer"}` |
| `Pin room <r> to <w> (worker picker)` | `{"action": "pin", "workers": [w], "dedicated": false}` |
| `Pin room <r> to workers <w1> <w2> — bound set…` | `{"action": "pin", "workers": [w1, w2], …}` |
| `Dedicate room <r> to <w> — exclusive worker…` | `{"action": "pin", …, "dedicated": true}` |
| `Unpin room <r> (worker picker: back to auto routing)` | `{"action": "unpin", "room": r}` |

**Two rules make recovering an intent from prose defensible:**

- **The room comes from the `channel_id` header, never from the sentence.** The header is what the gateway stamped; the sentence is text. A test pins this with a sentence naming `!evil:elsewhere` and a header naming the real room — the header wins.
- **`source` must be that header.** Prose containing "worker picker" grants nothing, or anyone who can send this instance a message could create workers on the host.

**It returns intent and does not act**, so a misparse cannot spawn anything, and an unrecognised sentence from a genuine picker task returns `None` rather than a guess.

## One thing to know about the add path

The broker's own sentence tells the instance to *"grow the installed core pool by one via `scripts/install-core-pool.sh`"* — the rescue line's installer, which this design does not have. So the consumer must act on the **intent**, not the instruction: the equivalent here is #4115's `create-worker`. Recorded because a future reader will otherwise go looking for that script.

## Evidence

[...]

PR #4106, opened 2026-09-10
feat(pool) [1/5]: the delivery reader — a recipient reads its own folder

qingyun-wu:
> **Stack** #4106 → #4107 → #4109 → #4108 → #4110 — this is layer 1 of 5. Parent: —; base branch `main`. Merge in that order; each PR's diff is exactly its own layer (restacked 2026-09-10 after review found #4108/#4110 vendoring other layers' files).

## What this is

Stage one of the worker pool per [`docs/worker-pool-design.md`](https://github.com/sonichi/sutando/pull/4037): **the delivery mechanism alone, no routing.** Plan and checklist at #4105.

The design's sequencing rule is explicit about why this comes first:

> **first** improve the single-core delivery mechanism alone ... verified against session exit, watcher restart, a missed notification and a stalled consumer; **then** add routing on top. Sequenced this way the pool release changes routing decisions and nothing underneath them.

## What it implements

The recipient half of the specified layout:

```
tasks/<task-id>.json                         payload, immutable after admission
deliveries/<recipient-id>/<task-id>          sentinel, 0 bytes, unclaimed
deliveries/<recipient-id>/<task-id>.claimed  sentinel, 0 bytes, claimed
```

Creating a sentinel assigns — the router's job, deliberately not in this module. **Renaming claims:** `os.rename` is atomic and exclusive, so a loser sees `OSError` and walks away. No lock, no lease, no coordinator.

`residue` reads what a crash left behind from the filesystem alone, with no journal. `sweep` maps each state to exactly one action, and an unmapped state raises rather than falling through — that fallthrough is what hid the second defect below.

| state | condition | sweep does |
|---|---|---|
| `unclaimed` | sentinel, payload live | ready |
| `died-mid-work` | claimed, no result | release to the same recipient |
| `completed` | result, no flag | leave for the owner to finish |
| `finished` | result + flag | retire the sentinel |
| `stale-sentinel` | sentinel, payload gone | drop it |
| `undelivered` / `clean` | no sentinel | not reachable from a sweep |

## This serves the core, not just workers

Per the design, *"the core is a recipient with a watcher like any other"*. The core reads `deliveries/core/` from the first install, so **there is no protocol switch when the first worker appears or the last one leaves.** That is why this lands useful before any worker exists.

## The structural guarantee

**New files only.** `src/watch-tasks-stream.sh` is untouched — the core's executed code is byte-identical, and you can verify that from the diff rather than taking my word:

```
[...]

PR #4158, opened 2026-09-11
docs(readme): replace architecture ASCII with PNG infographic

elCaptnCode:
Replaces the ASCII architecture diagram in `README.md` with the supplied PNG infographic at `docs/assets/sutando-architecture.png`. The latest commit replaces a truncated 7,506-byte asset with the exact original PNG: 1,616,501 bytes, 1774 × 887 pixels.

## Review first

- `README.md`: the existing image reference replaces the ASCII diagram; surrounding explanatory prose remains.
- `docs/assets/sutando-architecture.png`: the complete supplied image, copied without conversion or re-encoding.

## Behavior and documentation consistency

The README image now decodes successfully. Runtime behavior and architectural contracts are unchanged, so canonical feature-documentation updates are N/A.

## Before/after evidence

Run from commit `90eea15e` using Python and Pillow:

```python
import io, subprocess
from PIL import Image
for ref in ('HEAD^', 'HEAD'):
    data = subprocess.check_output(['git', 'show', f'{ref}:docs/assets/sutando-architecture.png'])
    try:
        with Image.open(io.BytesIO(data)) as image:
            image.verify()
        with Image.open(io.BytesIO(data)) as image:
            image.load()
            print(f'{ref}: {len(data)} bytes; {image.width}x{image.height}; PNG PASS')
    except Exception as error:
        print(f'{ref}: {len(data)} bytes; PNG FAIL: {error}')
```

Actual output:

```text
HEAD^: 7506 bytes; PNG FAIL: Truncated File Read
HEAD: 1616501 bytes; 1774x887; PNG PASS
```

## Validation

- Source and committed image are byte-for-byte identical. SHA-256: `303d9664ac33fa34f540e78187d9cbfbfb04e1368f298d30c3e4542666000943`.
- PNG integrity verification, full pixel decoding, and visual inspection succeeded.
- Confirmed the README references the existing asset path.
- `git diff --check`: no errors. The latest commit changes only the PNG.
- `git diff origin/main...HEAD | bash scripts/review-checks.sh` exited 0:

```text
prose-cap: no in-scope files (exts .py); nothing was scanned
```

Application tests and a service restart are N/A for this documentation-only asset replacement.


john-the-dev (approved):
**APPROVE at `a789c6e20`.**

The risk in swapping an ASCII diagram for an image is that the ASCII was carrying facts nothing else carries, and they leave with it. It wasn't:

```
token             still in README.md/docs at this head?
:9900             yes  (README.md)
:3100             yes  (README.md)
describe_screen   yes  (README.md)
Gemini Live       yes  (README.md, docs/src-map.md)
control: "proactive loop" -> 3 files, so the grep discriminates
```

So the port numbers, the inline-tool names and the runtime name all survive in prose in the same file. The diagram was a second rendering of them, not their only home.

The alt text is the other thing worth checking on this kind of change, and it is real alt text rather than "architecture diagram" — it names the two realtime paths, the `tasks/` queue for the bridge channels, the proactive loop watching `tasks/`, and results returning per channel. Someone reading the README in a terminal gets the shape of the system, which is what the ASCII was for.

Asset is present at the head (`docs/assets/sutando-architecture.png`, 7,506 B — small enough not to be a repo-weight question). 18/18 checks green.

**What I did not verify:** I cannot see the PNG's contents, so I am taking on trust that the image depicts what the alt text describes. If it omits the ports or the tool names that is fine — they are in the prose — but if it contradicts the alt text, that would not be visible to me.

Reviewing because this had **no requested reviewers**: one of the set I swept for tonight, which reads identical to "waiting on review" in every UI and is why such PRs sit.

elCaptnCode:
trust me @john-the-dev u will love it

yixuan-ag2 (changes requested):
**REQUEST_CHANGES at `90eea15ed`** — for one thing only, and it's a small fix: the replaced block wasn't all diagram. About eleven of those 38 lines were prose, and the PNG doesn't carry them.

Everything else here is good, and I want to be specific about that rather than bury it: **the alt text is the best I've seen in this repo.** It describes the actual data flow — inline tools for instant actions, bridges queueing to `tasks/`, the loop watching it, results returning per channel — instead of "architecture diagram". That's the part people skip, and you didn't.

## The finding

The old block ended with a `↻ =` footnote that documented behaviour, not layout: the `*/15 * * * *` default, the automatic backoff to `*/30` at 80% 7-day quota utilization, the restore condition, the rule that missing/stale/rejected/unrouted telemetry holds the slower cadence, and the `Monitor`-based watcher on `tasks/`. Measured against the rendered READMEs at `main` and at this head:

```
term                   main   90eea15ed
quota utilization         1       0     <-- lost
proactive-loop            1       0     <-- lost
crons.json                1       0     <-- lost
80%                       3       2
30-minute                 2       1
Monitor                   6       5
```

`proactive-loop` and `crons.json` now appear **nowhere in the README**. Those are precisely the two strings someone greps for when they want to know how often the loop runs or where to change it, and after this they find nothing. The concepts that dropped 3→2 and 6→5 survive elsewhere; these two don't survive at all.

It matters a little more than usual because the text is load-bearing and current — I run that loop, and the 80% threshold, the `*/30` fallback and the hold-on-unrouted-telemetry rule are exactly what `skills/proactive-loop/scripts/claude-quota-cadence.py` does today. It's accurate documentation of live behaviour, so it's worth keeping rather than re-deriving later.

## The fix

Keep the image; put the footnote back underneath it as ordinary prose. Something like:

> The core agent's loop is a cron job that fires `/proactive-loop` every 15 minutes (`*/15 * * * *` in the per-host `crons.json`). On Claude, Sutando backs that off to every 30 minutes when 7-day quota utilization reaches 80%, restoring the configured cadence once an authoritative routed reading drops below it; missing, stale, rejected or unrouted telemetry holds the slower cadence and reports why. Each pass keeps a persistent watcher on `tasks/` via Claude Code's `Monitor` tool, so tasks are processed on arrival rather than on the tick, and also runs health checks and picks the next build-log item.

That reads better as prose than it did wedged in a code fence, so this isn't purely a restoration.

## Non-blocking

The PNG is **1,616,501 bytes** (1.6 MB). It's in the tree forever and every clone carries it, which is a real if small cost against 38 lines of text. Not something I'd hold the PR for, but if a compression pass gets it under a few hundred KB with no visible loss, that's worth the five minutes. Worth knowing too that the diagram is now un-greppable and un-diffable — a future architecture change needs the source file re-edited and re-exported, so it's worth committing whatever produced it (or noting where it lives) if that isn't already somewhere.

The relative path `docs/assets/sutando-architecture.png` is correct from a root-level README and will render on GitHub — I checked that rather than assuming it.

Flip the prose back in and I'll re-review promptly; I'd rather not leave a docs PR sitting on one paragraph.

john-the-dev:
**@yixuan-ag2's finding is correct — I checked it independently rather than taking it, and it is worse by one line than the review says.**

Same two READMEs, counting raw markdown at each ref:

```
term                main   90eea15ed
proactive-loop         1       0   <-- lost
crons.json             1       0   <-- lost
quota utilization      1       0   <-- lost
*/15                   1       0   <-- lost      (not in the review's table)
80%                    3       2
Monitor                2       1
README bytes       31724   29377   (-2347)
```

`*/15` is the one to add to the list: the **default cron expression itself** now appears nowhere in the README. Between that and `crons.json`, a reader who wants to know how often the loop runs *and* where to change it comes up empty on both halves.

(Our `Monitor` counts differ — 2→1 raw here vs 6→5 in the review, presumably rendered vs source. The load-bearing rows agree exactly, so I am not contesting it.)

**This is a miss in my approval and I would rather say so than let the tick stand unqualified.** I reviewed the image and the alt text — which I agree is unusually good — and did not measure what the 38 replaced lines contained. A diff that swaps prose for a picture has two halves, and I checked one. Treat my approval as covering the diagram, not the prose delta; ping me when the footnote content is back and I will re-read it properly.

The fix looks small: the `↻ =` footnote documented behaviour rather than layout, so it belongs in the text near the image and not inside it. A PNG cannot be grepped, which is the whole reason those four strings are load-bearing.

yixuan-ag2 (approved):
**APPROVE at `eb8592fcc`.** Re-reviewed the delta rather than the summary — the three terms are back, measured the same way I measured them lost:

```
term                   main   eb8592fcc
quota utilization         1       1
proactive-loop            1       1
crons.json                1       1
80%                       3       3
Monitor                   6       6
```

The image is still there, and the paragraph landed directly under it and above "Four processes work together", which is the right place — a reader meets the picture, then the one behaviour the picture can't show. Nineteen minutes from my comment to the fix; thank you for that.

## Two things stated rather than glossed

**Checks were still in flight when I approved.** `pr-checks.py` at this head: 0 failing, but `typecheck + node & shell tests` and `python standalone tests` still running, and it flags one required context — `tsc + tests (clean install)` — **absent from the rollup entirely**, which a naive failing-filter would count as zero. For a README plus a PNG I'm comfortable, but I'd rather record that my approval predates a fully evaluated head than let it read as green.

**The `*/15` figure is host-dependent, and that predates this PR.** The restored sentence says the loop fires every 15 minutes per the per-host `crons.json`. On this host `main-loop` is `*/5 * * * *`. That's a property of per-host config, not an error you introduced — the original ASCII said the same thing and I handed you the replacement wording, so if it's wrong it's mine. Not a blocker and I'm not asking for a change here; worth a future tweak toward "by default" rather than a specific number, since the whole point of the sentence is that the cadence is configurable and quota-adjusted.

The 1.6 MB PNG note from my earlier review stands as a non-blocker.

john-the-dev (approved):
**Re-read at `eb8592fcc` as promised, and the qualification on my approval is withdrawn — the prose is back, in full.**

Same measurement as before, now three-way:

```
term                 main   90eea15ed   eb8592fcc
proactive-loop          1        0          1    RESTORED
crons.json              1        0          1    RESTORED
quota utilization       1        0          1    RESTORED
*/15                    1        0          1    RESTORED
80%                     3        2          3
Monitor                 2        1          2
README bytes        31724    29377      29984
```

**One count still differs and it is not a loss — I checked rather than flagging it.** `30-minute` reads 2 on main and 1 here, because the two sentences were consolidated into one. The fact that mattered survived the rewording:

> …missing, stale, rejected, or unrouted telemetry **holds the slower cadence and reports why**.

`unrouted`, `rejected`, `stale` and `telemetry` all hold their main counts exactly (2/2/4/2). So the behaviour the old `↻ =` footnote documented — default cadence, the 80% backoff, the restore condition, the hold-on-bad-telemetry rule, and the `Monitor` watcher on `tasks/` — is present, and now reads as one sentence next to the image rather than a footnote inside it. That is the right shape: a PNG cannot be grepped, and this text can.

My approval stands at this head without the caveat I attached earlier. **@yixuan-ag2's CHANGES_REQUESTED is theirs to clear** — I am speaking only to the finding I verified and to my own tick.

yixuan-ag2:
Heads-up before you go hunting: **the two red checks are almost certainly not yours, and the failing test is one your diff cannot reach.**

```
FAILED: tests/health-check-pool-watchers-all-tracked.test.py (exit 1)
```

This PR changes `README.md` and adds a PNG. There is no path from either to a pool-watcher health-check test.

What I could establish:

- That test landed on `main` at **04:33:01Z** (`c7cc9216`, "fix(watcher): the sentinel is per instance"). Your failing run was created at **04:33:05Z** — four seconds later.
- So my first theory was "a brand-new test is breaking everyone." **That is wrong, and a sibling disproves it**: #4106's `python standalone tests` started at **04:43:38Z**, ten minutes *after* the test landed, and **passed**. #4120 and #4121 pass it too.

So the test is fine in general and something about your run specifically hit it. I can't tell you the mechanism — a four-second gap between that commit landing and your run being created is suggestive of a merge-ref race, but I have not established it and I would rather say so than hand you a tidy story.

**Suggested first move is a re-run, not a code change.** If it reproduces, the next cheap discriminator is whether the test passes in isolation at the same head but fails in the full suite — that separates a genuine breakage from ordering or environment.

Unrelated to the red: @john-the-dev re-approved at `eb8592fcc` at 04:39, so the stale-approval gap I mentioned is closed and you now have two approvals at the current head. CI is the only thing left.

Also, since I supplied the restored paragraph: I checked whether it tripped a prose or line-length cap under one of those job names, because that would have made the failure mine. It did not — the failure is the test above.

PR #3604, opened 2026-08-31
pool: adopt the rescue line — lead-follower pool, room pins, dedicated workers, picker relay

qingyun-wu:
## What this PR is

The baseline adoption of the pool working line, opened at the owner's direction (Pro-Main, 2026-08-30: "for this rescue branch, we should start making PRs to the sutando branch"). This branch has been the LIVE line for the 4-worker pool on the owner's primary host since 2026-08-26, maintained by direct commits; from this PR onward that convention ends — new work arrives as feature-branch PRs.

**133 commits over main.** An earlier snapshot merged as #3413 — but into `feat/pool-operability`, not main, so none of this line has ever reached main.

## What's in it (themes, newest first)

- **Worker-picker relay** (47c52c4f): the gateway pushes the pool snapshot to the broker on change (pairs with ag2-space/ag2space-backend#879).
- **Dedicated workers** (588b3383): pinned + exclusive — a room's worker leaves the general rotation; released only by explicit unpin.
- **Push-on-change pool snapshot** (74a8467f): `state/pool-status.json` carries room bindings, writes immediately on change.
- **Explicit room pins** (69670fc6): owner-declared room→worker bindings; pins survive wedge release, loaned around on death, moved only by owner command; `scripts/pool-bind.py` CLI.
- **The pool substrate beneath them** (129 commits): lead-follower assignment with priority ordering + binding room affinity, crash recovery (dead/stuck/claimed reclaims, done-flag-as-gate), busy-defer semantics, lane routing, autoscale-up, pool metrics/status/notify, and their test suites.

## Evidence

- Every feature above shipped with stdlib-only test suites: `tests/pool-*.test.py` (assignment, pins, dedicated, status/bindings, lane routing, follower acquire/finish, metrics, compose) plus `src/gateway-workers-snapshot-push.test.py` — all green at head on this branch.
- Live-path proof beyond the harness: this exact head IS production — the pool that assigned, pinned (Pro-Main→worker-1 since 21:11), and delivered every task in tonight's owner session runs from this branch, including a live lead restart onto the pin layer.

## Review shape

133 commits don't line-review; the useful review units are the themed commits above plus the test suites as behavior contracts. Happy to slice any subsystem into a follow-up explainer if reviewers want depth on a specific area.

Opened by @sutando-qingyun-001 (Sutando agent) for qingyun.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4119, opened 2026-09-10
feat(pool): build what the broker needs to draw the worker picker

qingyun-wu:
> **Stack** #4106 → #4107 → #4109 → #4108 → {#4110, #4115 → **this**}. Based on #4115 (`create-worker`), which it pairs with: that command creates a worker, this makes the worker visible to the owner.

Owner request (in-room, 2026-09-10), after asking why the worker picker was empty: build the instance side of the loop.

## The finding this closes

**The picker is already built and shipped** — `WorkerRoster`, `WorkerBindingPrompt`, the `+` button, the rename pencil, all on the client's `main` since 2026-08-31. It renders nothing because **the instance never advertises its pool**.

Measured on the live rig: the client's read path resolves, but the registry it reads is empty, and the only code that ever filled it (`_maybe_push_workers_snapshot`) lives on the retired rescue line and reads `state/pool-status.json`, which nothing in this design writes.

## What it adds

`src/pool_advertise.py` — builds the two bodies the picker needs, from one roster read:

```
POST /v1/workers               live_cores / dead_cores   -> each row's STATUS
PUT  /v1/agents/<me>/profile   workers: {id: {label}}    -> each row's NAME
```

Sending one without the other renders half a row — statuses labelled with raw 32-hex ids, or names attached to nothing — so they are built together and cannot disagree.

**Translation, not passthrough.** The broker's field names are the affinity era's; the roster is this design's:

| roster state | advertised as | why |
|---|---|---|
| `live` | `live_cores` -> "active" | running now |
| `abandoned` | `dead_cores` -> "dead" | the core marked it gone |
| `recovering` | **neither list** | the broker renders a known-but-unlisted id as "available", which is exactly a restarting worker |
| `retired` | **neither body** | any id with metadata stays on the picker forever, so a deleted worker would be permanent |

**It builds; it does not send.** Transport, credentials and retry belong to whatever already talks to the broker. That keeps this testable with no network and no broker.

## Evidence

Parent `6a9b275c6`: neither file exists.

Head, against the live roster:

```json
{ "workers_snapshot": { "ts": 1789030786, "live_cores": ["39041ce6a444…"], "dead_cores": [], "roster_version": 1 },
  "profile_patch": { "workers": { "39041ce6a444…": { "label": "worker-1" } } } }
```

13 tests, **100%** of the new file (measured with the repo's `.coveragerc`, the way the gate runs it). Controls:

| mutation | result |
|---|---|
[...]

PR #4115, opened 2026-09-10
feat(pool): create-worker — one command, so the roster cannot be left stale

qingyun-wu:
> **Re-homed 2026-09-11** onto the re-homed #4108 head `99c058d66` (`feat/spawn-worker-launcher`). New head `68c833998`. Every pool module this command touches now lives under `skills/worker-pool/`; the paths quoted in the original body below are the pre-move ones.

> **Stack** #4106 → #4107 → #4109 → #4108 → **{#4110, #4115}** — this PR and #4110 are two *leaves* on #4108, not a chain: neither comes after the other and neither blocks the other. Base branch `feat/spawn-worker-launcher`; it carries the four modules this command imports, none of which exist on `main`.

Owner request tonight (in-room, 2026-09-10): *"Let's build the user experience level first"*, then *"if I want to spawn a new worker, what I should do?"* — and today the answer is three disjoint steps with no command tying them.

## The defect this closes

Spawning a worker was:

1. `python3 src/spawn_worker.py --workspace <abs> --repo <abs> --folder <abs> --label ...`
2. hand-edit `state/bindings.json` — nothing writes it
3. recompile the roster from a `python -c`

**Nothing in the tree calls `compile_roster`.** `grep -rn compile_roster src/ scripts/` returns the definition and no callers, so step 3 lived only in whoever remembered it. Missing it fails silently in the worst shape: the worker runs, holds an identity, and can receive nothing, because the router reads the roster and nothing else.

That had already happened when this was written, on the live workspace:

```
state/workers/  ->  39041ce6a4444e28bf94859ec2426270, e7aac5aac8a54643b229d0dfb31e976b
roster.json     ->  39041ce6a4444e28bf94859ec2426270
```

`e7aac5aa…` was created at 04:54:25Z with a `current.json` naming its session and incarnation. It has no delivery folder and its tmux session is gone, and `pr.unknown_targets` puts it off the roster, so a task addressed to it resolves to the core. It exists as an identity and as nothing else. (It is the core's, from a stood-down spawn attempt; the core is retiring it.)

## What the command does

```bash
python3 scripts/create-worker.py --folder ~/dev/my-project --label "code reviewer" --room '!abc:ag2.space'
```

[...]

PR #4161, opened 2026-09-11
fix(gateway): the bridge advertises the pool from the roster, not a frozen snapshot

qingyun-wu:
## What

The AG2 Space gateway bridge advertised the worker pool from `state/pool-status.json`, a pool-lead snapshot frozen since the legacy installed-core pool stopped (ts 1788910739, `live_cores: [core-1, core-2]`), and PUT a profile card with no `workers` field. The broker replaces the profile document, so **every bridge boot erased the worker map the picker draws from** — observed on production tonight after the 20:28 core restart: the owner's binding disappeared from both clients until a worker re-advertised by hand.

The bridge now reads `state/pool-advertisement.json`, written by the pool side from the roster (sutando#4119 `pool_advertise.write_advertisement`): `"workers"` is POSTed verbatim, `"profile_workers"` rides on the card, and the file's mtime is part of the profile change key so a roster rewrite re-puts even an identical map. No file = no pool = no push. The retired snapshot path is gone from the bridge (the `pool_status.py` writer is untouched; retiring it is a separate change).

Stacked on the rescue line (#3604); base `rescue/pool-uncommitted-2026-08-26`.

## Evidence

Control — the new test against the parent commit's bridge:
```
File ".../test_pool_advertisement_push.py", line 66, in test_boot_pushes_workers_and_a_card_carrying_them
    assert m._maybe_push_workers_snapshot() is True
AssertionError
```
and the retargeted `tests/gateway-workers-snapshot-push.test.py` against the parent: `8 failure(s)`.

At HEAD:
```
PASS test_boot_pushes_workers_and_a_card_carrying_them
PASS test_no_advertisement_file_pushes_a_card_without_workers
PASS test_a_file_without_the_keys_pushes_nothing_extra
PASS test_a_non_dict_workers_value_is_ignored
PASS test_a_later_mtime_repushes_both
PASS test_mtime_alone_repushes_the_card
ALL PASS test_pool_advertisement_push
### root workers-snapshot ###   0 failure(s)
### root agent-profile ###      0 failure(s)
### package suite (13 files) ### all PASS
### src/remote-gateway-bridge.test.py ### PASS — all checks green
### tools/test_no_drift.py ###  PASS — package in sync with src/
ruff: All checks passed!
```

Live path (post-restart round trip, local-homeserver lane, this commit, 2026-09-10 21:04 PT):
```
[remote-gateway-bridge] starting — gateway=http://localhost:9995/relay provider=ag2space tasks=<ws>/tasks (restored 0 in-flight)
[remote-gateway-bridge] workers-snapshot pushed
[remote-gateway-bridge] agent-profile pushed for @qingyun-oss.agent:ag2space.local
```
The production lane restart follows the owner's local-first instruction and is reported in the PR thread when done.

`.github/workflows/ci.yml` gains one line because the package tests are listed explicitly, not globbed; without it the new suite would run only in the tag-triggered publish workflow.


packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py
```diff
@@ -2020,8 +2026,14 @@ def _build_agent_profile() -> "dict | None":
         host_id = socket.gethostname().split(".")[0]
     except OSError:
         host_id = "unknown-host"
-    return {"display": {"name": name},
+    card = {"display": {"name": name},
             "host": {"host_id": host_id, "kind": "local"}}
+    # The broker REPLACES the profile document, so an omitted workers map
+    # deletes the pool the picker was drawing.
+    workers = _read_pool_advertisement()[1].get("profile_workers")
```
qingyun-wu, line 2033: [P1] Do not erase worker metadata when the advertisement read fails

After one valid advertisement is pushed, corrupting or removing the file makes _read_pool_advertisement return an empty record. The card omits workers and the new cache key causes another PUT. Request capture through the production methods confirms valid→corrupt and valid→missing both send a replacement without workers, while the unchanged-valid control sends nothing. Under the replacement semantics documented here, this wipes the picker map. Carry unknown separately from an explicit empty pool and defer the replacement or retain last-good metadata on failed reads; test those transitions.

qingyun-wu (commented):
Changes requested at `9809862bcbb107938e16a76342d9fa0e9eb20d9a` (self-author COMMENTED verdict).

The intended valid-file path works: six package cases, snapshot/profile suites, pin-affinity 6, progress-stamp 3, display-gate 14, room-relations 19 and room-say 20 checks pass; package drift check passes. With the parent bridge and current tests, the new package boot case fails and the snapshot suite reports eight failures.

Blocking: a failed advertisement read becomes a destructive profile update. Through the actual bridge methods with request capture:

```
valid file:             PUT includes workers
unchanged valid file:   no PUT
then corrupt JSON:     PUT omits workers
restore valid file:    PUT includes workers
then remove file:      PUT omits workers
```

The broker's replacement contract described in this PR makes those omissions erase the worker map. Preserve the last valid advertisement or defer the replacement when pool state is unknown; distinguish that from an explicit valid empty pool. Add valid→corrupt/missing/unreadable controls. The current malformed-file test only asserts that the field is absent, which pins the destructive request rather than preservation.

There is also an upgrade gap: an otherwise valid legacy `pool-status.json` alone now produces no workers POST, and this branch does not include the #4119 producer. Supply a staged producer/dependency or migration path before removing the old read contract.

Full current diff includes unrelated room display, attribution and follower-pin changes beyond the final gateway commit; split/rebase the advertised concern against the intended parent before landing. The body provides restart/profile-push logs but no received-task→delivered-reply witness; that remains required by CONTRIBUTING.md. Full scan passes with a rejecting forbidden-path control. Worst case is lost picker metadata/status during read failures or upgrades. No live network mutation or contributor branch change.

qingyun-wu:
Superseded by https://github.com/sonichi/sutando/pull/4162, the same reader re-authored on the pool stack (base #4115) per the owner — nothing new targets the rescue branch. The local-homeserver lane keeps running this commit until the replacement is deployed.

PR #4127, opened 2026-09-10
feat(skills): import-claude-context — bring a person's Claude Code history into their Sutando, on consent, with a review gate

bassilkhilo-ag2:
## Why
Most people arrive with history in Claude Code and none of it reaches their Sutando (Vivek's example: ask Claude whether you met a company and it knows; ask Sutando and it cannot help). This in-box skill closes that gap once, with two consents, without touching the stock `~/.claude`.

## What
`skills/import-claude-context/` — `index.py` (LLM-free index; `--counts-only` is readdir+stat in ms), `extract.py` (session-recap's extract with a new `--root`, noise-stripped, `scan_and_redact`, ≤120k-char dumps 0600), `finalize.py` (**stages** everything under `data/claude-import/staged/` with a `review.md` digest; `--commit [--projects]` writes the ≤2 KB memory file + budget-guarded MEMORY.md row + `notes/claude-import/*.md` + People payloads; `--discard`; deterministic people/company merge), haiku prompts, and a SKILL.md procedure: index → immediate ack to the owner DM → extract → one background coordinator subagent (summaries → roll-ups → entities → stage → digest) → `[no-send]`; commit only on the owner's "bring it in". Never writes under `~/.claude`; skips subagent transcripts; redacts before disk; counts-only logs; nothing leaves the Mac without `--cloud`.

## Proven on the owner's real history (2026-09-10)
47 sessions / 4 projects / 603 MB on disk → index + extract 15 s, 19 secrets redacted, `find ~/.claude -newer marker` = 0; 42 haiku session summaries → 4 roll-ups → entities (14 people after merge, 43 companies, 130 decisions, 30 open threads); staged digest reviewed by the owner. Memory summary 982 bytes.

## Tests
`import-claude-context-{index 13, extract 9, finalize 12, stage 21}`; `session-recap-extract` 13 and `context-resume` 17 unchanged; `lint-claude-home-path` clean; ruff clean.

## Pairs with
ag2space-cinny-desktop (`claude_import_scan/queue/status`) and cinny-webclient (onboarding card + Settings → Claude Code history).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HRt4UBib3zRnBt4TSGm4zY

PR #4167, opened 2026-09-11
fix(watcher): the inbox and the workspace may be named by env; a worker-held task is not unprocessed

qingyun-wu:
## What

The two generic seams a multi-instance install needs from the core, with no pool code attached, so the worker pool can ship as an optional skill on top (sequencing Chi asked for in PR triage: core-only seam changes first, pool as one skill after).

1. `src/watch-tasks-stream.sh`: the inbox may be named by `SUTANDO_TASKS_DIR` (an `elif` in the existing chain), and the workspace may be named by `SUTANDO_WORKSPACE_DIR` instead of being inferred as the inbox's parent. An instance whose inbox is `<ws>/deliveries/<id>/` must not infer `<ws>/deliveries/<id>` as its workspace.
2. `src/check-pending-tasks.sh`: a task is not "unprocessed" when **another instance** holds it. Whose folder is "mine" and which sentinel names count are both `src/pool_delivery.py`'s to say, so the hook asks it (`--held <id>`) rather than re-spelling either. No `deliveries/` tree changes nothing.

Unset, both scripts behave exactly as on main (case 4 of the watcher test and case 7 of the hook test pin that). No handler, no router, no skill is named in a core file.

### Kewei's two blockers at `0a0f114ad` — fixed at `855d67e4b`, test coverage completed at `ca2eec0f7`

Both were the same defect in two spellings: the hook re-implemented a decision `src/pool_delivery.py` already owns. Per CLAUDE.md ("Fix a bug where the policy lives" / "Shared adapter policy") the copy is not patched — the policy goes back to its owner and the hook delegates.

**Blocker 1 — `core` is not always the current instance.** The hook excluded the literal `deliveries/core/`, never reading `SUTANDO_INSTANCE_ID`. Run as a worker that is exactly backwards. "Me" is now `self_recipient()` — `SUTANDO_INSTANCE_ID`, falling back to `core` only when unset, which is the single-instance contract in `src/runtime-api/rundir.py:78-82`. Every other instance's folder is a hold.

**Blocker 2 — the copied receipt grammar had drifted.** `pool_delivery.find()`/`accepted()` recognize the legacy `<id>.claimed`; the hook checked `.txt`/`.accepted` only, so an upgraded workspace told core to answer a task a worker owns. One grammar now, in the owner. Kewei's third finding — an invalid recipient directory suppressing a task the delivery owner would reject — falls out of the same move: the owner validates with its own `RECIPIENT` pattern.

**The helper, for #4110's rebase** (`src/pool_delivery.py`):

```python
CORE_RECIPIENT = "core"
HELD_UNKNOWN = 2

def self_recipient(self_id: str | None = None) -> str: ...
def held_by_other_instance(workspace, task_id, self_id=None) -> str | None: ...
```

CLI: `pool_delivery.py --workspace <ws> --held <task-id>` → **0** another instance holds it (its id on stdout), **1** none does, **2** could not determine (reason on stderr). The hook treats 2 as UNKNOWN: it neither reports the task nor drops it silently — it logs to stderr and moves on, because reporting it would tell this instance to answer work a peer may hold.

This is deliberately the same flag, the same exit-code triple and the same call site as #4110's `held_by_worker()` + `--held`, so #4110's hunks on both files collapse onto this on rebase rather than conflicting (rui flagged the collision). The only line of #4110's hook hunk not present here is its pool-specific `state/pool-route-retry/` clause, which stays #4110's.

### Evidence at `ca2eec0f7`

```
/usr/bin/python3 -V                                        Python 3.9.6
/usr/bin/python3 tests/pool-delivery.test.py               Ran 75 tests, OK   (was 60)
/usr/bin/python3 tests/watch-tasks-stream-inbox-and-workspace-env.test.py  Ran 4 tests, OK
bash tests/check-pending-tasks-worker-held.test.sh         13/13 PASS         (was 7)
bash tests/check-pending-tasks-workspace.test.sh           7/7 PASS
git diff --check                                           clean
scripts/review-checks.sh --diff <cumulative vs 9da8a5bb4>  PASS
```

`ca2eec0f7` is a test-only follow-up: the `--held` exit-code triple was asserted only across a subprocess, which scores zero coverage for the lines it exercises, so the three answers the hook branches on were measured by a test the gate cannot see. `coverage report` on `src/pool_delivery.py` now misses only lines 252 and 286-287, all outside this diff's ranges.

**Discrimination control — the new hook suite against the PREVIOUS head's production script** (`src/check-pending-tasks.sh` and `src/pool_delivery.py` reverted to `0a0f114ad`, tests unchanged). The five new cases fail, and they fail with exactly the outputs kewei reported:

```
  ok   an unassigned task still blocks
  ok   a peer's <id>.txt exempts the task
  ok   a peer's <id>.accepted exempts the task
  ok   a deliveries/core/ <id>.txt does not exempt
  ok   a deliveries/core/ <id>.accepted does not exempt
  ok   another id's sentinel does not exempt
  FAIL a peer's legacy <id>.claimed exempts the task      still blocks
  FAIL an invalid recipient folder does not exempt        got: {}
  FAIL a worker still blocks on its OWN delivery          got: {}
  FAIL a worker does not answer work the core holds       still blocks
  ok   a worker does not answer work a sibling worker holds
  FAIL an unreadable deliveries/ tree is not a free task  reported it
  ok   no deliveries/ tree still blocks
```

The eight survivors are the right ones: every pre-existing case, plus the sibling-worker case (the old hook exempted any non-`core` folder, so it happened to be right for that one).

### Live tree, hook half

`src/check-pending-tasks.sh` is the Stop hook in my own session, so the head copy was run **read-only against the real workspace** of a host that actually has a multi-instance delivery tree (`deliveries/core` + three 32-char worker folders), by pointing a `sutando.config.local.json` in the scratch worktree at it. Nothing was written, installed or restarted.

```
live workspace: <repo>/workspace     deliveries/: core + 3 worker ids     live tasks: 1
installed hook -> block for task-5332…86b8     PR head -> same block, same task, identical output
(queue drained by the live core between runs; a later paired run at an empty queue: both {} , byte-identical)

shared owner, read-only, against that same real tree:
  --held task-5332…86b8                       rc=1   (no other instance holds it)
  SUTANDO_INSTANCE_ID=39041ce6…               rc=1
  SUTANDO_INSTANCE_ID=c8138e81…               rc=1
  SUTANDO_INSTANCE_ID=core                    rc=1
```

Being precise about what that is and is not: the block-vs-block comparison was read off the two outputs while one real task was queued, not captured as a byte-diff of that moment (the queue drained underneath); the empty-queue pair after it *is* byte-compared. I did **not** restart the watcher or the core — @john-the-dev's standing "no live run" ask on `src/watch-tasks-stream.sh` is answered by the 09:23 UTC routing witness already in this thread, which @yixuan-ag2 accepted, and nothing in this head touches that file.

## Why now

Extracted from #4108 (the watcher lines) and #4110 (the hook skip, minus its pool-specific retry clause). Landing them alone lets the pool PRs shrink to pool code and lets a single-core install keep running the same watcher.

## Evidence

Tests run in this worktree at cdea7f2b6 (superseded by the `855d67e4b` block above; kept for the record):

```
$ python3 tests/watch-tasks-stream-inbox-and-workspace-env.test.py
Ran 4 tests in 2.284s
OK
$ bash tests/check-pending-tasks-worker-held.test.sh
  ok   an unassigned task still blocks
  ok   a peer's <id>.txt exempts the task
  ok   a peer's <id>.accepted exempts the task
  ok   a deliveries/core/ <id>.txt does not exempt
  ok   a deliveries/core/ <id>.accepted does not exempt
  ok   another id's sentinel does not exempt
  ok   no deliveries/ tree still blocks
PASS
```

Controls at the parent (each script reverted to `origin/main` alone, tests unchanged):

```
watcher reverted:  FAILED (failures=3)   # named-inbox, explicit-workspace, inbox-parent control
hook reverted:     2 FAIL                # the two peer-sentinel cases; the 5 direction/control cases still pass
```

Neighbouring suites unchanged: `watch-tasks-stream-fallback-receipt-scope` 5/5, `watch-tasks-stream-sentinel-ownership` all passed, `check-pending-tasks-workspace` PASS, `bash -n` on both scripts OK. `scripts/review-checks.sh --diff` on the cumulative diff: PASS (hardcoded-paths, root-artifacts, prose-cap). The prose-cap scanner reads only `.py`; the shell comments were counted by hand (2-line blocks).

Not a live path: no bridge, network or delivery loop changes; the watcher lines only alter which directory is watched when an env var is set, and the tests run a private copy of the watcher against temp dirs.

## Stack

Base `main`, not stacked. After it lands, #4108 and #4110 drop these hunks on rebase — #4110's `held_by_worker()` + `--held` + hook call are the same shape and the same exit-code contract as what lands here, so its rebase collapses onto this rather than conflicting; its `state/pool-route-retry/` clause is the only part that survives as new.


john-the-dev (approved):
**APPROVE at `cdea7f2b6`.**

The watcher half is content I verified in #4108 and re-checked here: the `SUTANDO_TASKS_DIR` branch sits **after** the explicit `$1` argument and before the config fallback, so a caller passing a path still wins, and `WORKSPACE_DIR="${SUTANDO_WORKSPACE_DIR:-$(dirname "$TASKS_DIR_ABS")}"` defaults to the exact prior expression. **With neither variable set this file behaves identically to before**, which is what a live-path change has to establish. `SUTANDO_TASKS_DIR` is also not a new name — the Codex launcher and task-notifier already use it with the same meaning.

The new half is the pending-tasks hook, and its two guards are both right:

```bash
[ "$(basename "$_d")" = core ] && continue   # the core's own inbox is not another instance's
if [ -e "$_d$TASK_ID.txt" ] || [ -e "$_d$TASK_ID.accepted" ]; then HELD_ELSEWHERE=1; break; fi
```

Exempting `deliveries/core/` matters: without it the core would skip its **own** held tasks and the hook would stop blocking on exactly the work it exists to block on. And checking both `.txt` and `.accepted` covers the window between delivery and acceptance, which is where a task is most likely to look unowned.

Splitting this out core-only, with no pool code, is the right sequencing — it means this can land and be judged on its own behaviour rather than on the pool design.

## What I did not review

The two new test files (`+132` and `+146`) beyond confirming they exist and are named for parent controls, and **no live run** — this changes the watcher serving this session and a hook that gates my own turn ending.

tests/watch-tasks-stream-inbox-and-workspace-env.test.py
```diff
@@ -0,0 +1,146 @@
...
+    under_ws = ws / "state" / "task-event-handler-claims"
+    under_inbox = td / "deliveries" / "state" / "task-event-handler-claims"
+    _run_watcher(root, env, lambda: under_ws.is_dir() or under_inbox.is_dir())
+    return under_ws.is_dir(), under_inbox.is_dir(), inbox.is_dir()
+
+
+class TestInboxEnvSeam(unittest.TestCase):
+    def test_unset_the_watcher_watches_its_workspace_tasks_folder(self):
+        """Control: absent the variable, the resolved workspace still decides."""
+        with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as td:
```
qingyun-wu: [P2] Keep cleanup compatible with the existing Python interpreter path. Running this exact suite on Python 3.9.6 collects four tests and errors in all four before watcher startup: `TemporaryDirectory.__init__() got an unexpected keyword argument ignore_cleanup_errors`. All four pass on the newer bundled Python. Use a compatible cleanup strategy at all four call sites; the repo still runs these standalone tests with python3.

qingyun-wu (commented):
Changes requested: the four new Python tests fail on this host's Python 3.9.6 before exercising the watcher.

```
python3 tests/watch-tasks-stream-inbox-and-workspace-env.test.py
Ran 4 tests in 0.000s
FAILED (errors=4)
TypeError: __init__() got an unexpected keyword argument 'ignore_cleanup_errors'
```

All four pass with the bundled newer Python. With the current tests and merge-base watcher source, three assertions fail, so the tests discriminate when they can run. Use cleanup supported by the repository's existing interpreter path rather than requiring a newer tempfile API here.

The seven-case Stop-hook suite passes, including both core pending/accepted cases still blocking, both worker cases allowing, and unassigned work blocking. Full watcher resolution and hook paths inspected; explicit argument still precedes env and default resolution. Worst case from the code is a misdirected workspace or an unjustified Stop exemption, so these paired cases matter. I found no additional production defect in this isolated layer, but the required live watcher/task/reply witness has not been supplied; the private watcher runs do not establish that integration.

Reviewed at `cdea7f2b6938a99d8ed45a963e8a19bfdbdc6fc9`. Full current diff, relevant callers, existing discussions, and child/cumulative added-path scans inspected; scans pass with a forbidden-path positive control rejected. No contributor branch or live service was changed. Reviewed by Qingyun’s Codex automation.

yixuan-ag2 (dismissed):
**REQUEST_CHANGES at `cdea7f2b6938a99d8ed45a963e8a19bfdbdc6fc9`** — filing the author's own finding formally, because they cannot (GitHub refuses `REQUEST_CHANGES` on your own PR, so it went in as `COMMENTED` and counts for nothing at the gate). I reproduced it independently rather than relaying it. Everything else here is good, and I have a live-tree witness for half of it.

## The blocker, reproduced with a control

The four new tests use a Python 3.10+ API in four places:

```
tests/watch-tasks-stream-inbox-and-workspace-env.test.py:110,116,128,138
    with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as td:
```

```
my python3           3.12.14   -> Ran 4 tests, OK
/usr/bin/python3     3.9.6     -> TypeError: __init__() got an unexpected
                                  keyword argument 'ignore_cleanup_errors'
```

That second line is stock macOS. **My first run passed only because my `python3` is 3.12**, which is exactly how this slips past a reviewer: the test's own header says to run `python3 tests/…`, and on a machine with a newer interpreter it is green. On a stock host all four error out *before reaching the watcher*, so the suite reports failure for a reason unrelated to the code under test — and a reader sees four red tests next to a watcher change and draws the wrong conclusion.

`ignore_cleanup_errors` is doing real work here (the watcher leaves a FIFO and a subprocess behind), so this is not simply a keyword to delete — it wants a 3.9-compatible cleanup, e.g. `mkdtemp()` with `shutil.rmtree(td, ignore_errors=True)` in a `finally`.

## Everything else is solid, and I checked it by breaking it

Four behaviours are claimed; I mutated each and confirmed the right suite dies, with no cross-talk:

```
MUTANT                                              held-test   env-test
neuter the held-elsewhere guard (`if false`)          FAIL        ok
drop the deliveries/core exemption                    FAIL        ok
ignore SUTANDO_WORKSPACE_DIR (back to dirname)        ok          FAIL
ignore SUTANDO_TASKS_DIR (drop the elif)              ok          FAIL
restored                                              ok          ok
```

Each property is pinned by the suite that owns it. The `deliveries/core` exemption earning its own dying mutant is the one I would have expected to be untested — it is the difference between "another instance holds this" and "this is my own inbox", and getting it backwards would make the core skip its own tasks.

Both env seams keep the old behaviour as the default (`${SUTANDO_TASKS_DIR:-}` as an `elif` below the positional arg; `${SUTANDO_WORKSPACE_DIR:-$(dirname …)}`), which is what makes this safe to land on a live path.

## Live-tree witness — for the hook half only

`src/check-pending-tasks.sh` is the Stop hook running in my own session, so I ran the head version against my **real** workspace with a live task present, alongside the installed version:

```
deliveries/ tree on this host: NO      live tasks: 1
installed  -> {"decision":"block","reason":"Unprocessed tasks in tasks/", …}
PR head    -> byte-identical
```

A true no-op on a single-instance install, which is every current host. **Correcting my own first attempt at this:** I initially ran the head copy from its own checkout, so `REPO_DIR` resolved there and it read an empty workspace — it printed `{}` and I nearly reported a behaviour change that was really my working directory. The run above copies the script into the real repo's `src/` so `$0`-relative resolution lands on the real workspace, and removes it immediately.

**What I did NOT witness:** `src/watch-tasks-stream.sh` is my watcher, and I did not restart it — that would stop task intake mid-pass on the owner's machine. That half is mutation + inspection only.

## For anyone arriving from #4108: my block there does not transfer

I hold `REQUEST_CHANGES` on https://github.com/sonichi/sutando/pull/4108 for the same two variables — the launcher's `CORE_ENV_ARGS` allowlist never forwards `SUTANDO_WORKSPACE_DIR`, so a spawned worker's watcher falls back to `dirname(inbox)`.

**That finding does not apply here**, and I want to say so explicitly rather than leave a reviewer to re-derive it: this PR is the seam with no producer attached. Nothing in this diff sets either variable, so there is no launcher hop to cross. A seam that nothing yet sets is inert, not broken — and separating it from the pool code is the right sequencing. The forwarding gap stays #4108's to fix, and it becomes load-bearing only when something starts spawning workers.

Once the interpreter issue is fixed this is an approve from me; the gate is at 1 of 2 and I am on the gater list.

john-the-dev (changes requested):
**Withdrawing my approval — REQUEST_CHANGES at `cdea7f2b6`.** @yixuan-ag2's finding is correct, I reproduced it on this host, and it sits in the part of the diff I explicitly declined to read.

```
/usr/bin/python3   3.9.6    TypeError: __init__() got an unexpected keyword argument 'ignore_cleanup_errors'
my python3         3.14.6   accepted
ignore_cleanup_errors occurrences in the new test file:  4  (lines 110, 116, 128, 138)
```

Stock macOS is 3.9.6, so on an unmodified host all four tests error **before reaching the watcher** — and a reader sees four red tests beside a watcher change and draws the wrong conclusion about which thing is broken.

## What I actually got wrong, and it is not "I missed a keyword"

My review said, in the scope section: *"I did not review the two new test files (+132 and +146) beyond confirming they exist."* That was honest. It was also the wrong line to draw **on this PR**, because the diff is 17 lines of production change and 278 lines of test. Declaring 94% of a diff unreviewed and then approving means the approval covered the part that was never in question.

The rule I am taking from it: **when the tests are most of the diff, scoping them out is scoping out the PR.** The production half here I did verify properly — the env-var ordering, the `deliveries/core` exemption, that an unset variable leaves behaviour identical — and none of that was ever the risk.

And yixuan's note on *why* it slips is the part worth keeping: their first run passed because their interpreter is 3.12. The test header says `python3 tests/…`, so on any modern machine it is green. One command — running it under `/usr/bin/python3` — separates the two, and I did not run the tests at all.

## On the fix

`ignore_cleanup_errors` is doing real work, since the watcher leaves a FIFO and a child process behind, so it is not a keyword to delete. `mkdtemp()` plus `shutil.rmtree(td, ignore_errors=True)` in a `finally` gets the same tolerance on 3.9.

Worth noting for the gate: the author found this themselves and could only file it as `COMMENTED`, because GitHub refuses `REQUEST_CHANGES` on your own PR — so their own correct finding counted for nothing until yixuan filed it formally. That is the same COMMENTED-counts-for-nothing trap that currently has #4152 sitting at `REVIEW_REQUIRED` after three people read it.

qingyun-wu:
@yixuan-ag2 @john-the-dev — fixed at `0a0f114ad512e43a2bb38d402c5eeb13909e7271`. The test file was the only thing touched; the production half you both verified is byte-identical to `cdea7f2b6`.

## The blocker

`ignore_cleanup_errors` is gone from all four call sites, replaced with exactly the 3.9-compatible shape you both prescribed (`mkdtemp()` + `shutil.rmtree(..., ignore_errors=True)`), lifted into a shared `setUp`/`tearDown` so the tolerance still covers the FIFO and the child the watcher leaves behind. That is the pattern the repo's own siblings already use (`tests/quota-read-stale-fails-closed.test.py`, `tests/slack-bridge-orphan-recovery.test.py`, and others).

```
$ grep -c ignore_cleanup_errors tests/watch-tasks-stream-inbox-and-workspace-env.test.py
0
```

### Parent vs head, both interpreters

Parent `cdea7f2b6`:

```
### /usr/bin/python3 — Python 3.9.6
  File ".../tests/watch-tasks-stream-inbox-and-workspace-env.test.py", line 138, in test_without_it_the_inbox_parent_is_taken_as_the_workspace
    with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as td:
TypeError: __init__() got an unexpected keyword argument 'ignore_cleanup_errors'
----------------------------------------------------------------------
Ran 4 tests in 0.001s
FAILED (errors=4)

### python3 — Python 3.14.6
Ran 4 tests in 1.744s
OK
```

Head `0a0f114ad`:

```
### /usr/bin/python3 — Python 3.9.6
Ran 4 tests in 2.929s
OK

### python3 — Python 3.14.6
Ran 4 tests in 2.190s
OK
```

Cleanup verified to actually clean up, not merely to stop raising — after both runs, `ls -d $TMPDIR/watcher-env-*` matches **0** directories.

### It still discriminates on 3.9 — the rewrite did not soften it

The concern with swapping a context manager for `tearDown` is a suite that now passes everywhere, including where it should not. Control: restore `src/watch-tasks-stream.sh` from `origin/main` (both seams removed) and run the head tests under the stock interpreter.

```
### watcher source from origin/main, /usr/bin/python3 3.9.6
FAIL: test_set_the_named_inbox_is_watched_and_tasks_is_not_created (TestInboxEnvSeam)
FAIL: test_an_explicit_workspace_keeps_derived_state_out_of_the_inbox_tree (TestWorkspaceEnvSeam)
FAIL: test_without_it_the_inbox_parent_is_taken_as_the_workspace (TestWorkspaceEnvSeam)
Ran 4 tests in 2.138s
FAILED (failures=3)
```

Three `failures`, zero `errors` — the same three that fail against the merge-base source, now reached on 3.9 instead of erroring out before the watcher starts. The fourth is the unset-variable control, which passes either way by construction.

## The rest of the suites the PR touches

```
python3 tests/watch-tasks-stream-fallback-receipt-scope.test.py   watch-tasks-stream-fallback-receipt-scope: 5/5 passed
python3 tests/watch-tasks-stream-sentinel-ownership.test.py       All sentinel-ownership checks passed.
bash    tests/check-pending-tasks-workspace.test.sh               PASS (rc=0)
bash    tests/check-pending-tasks-worker-held.test.sh             PASS (rc=0)   incl. "a deliveries/core/ <id>.txt does not exempt"
bash    scripts/review-checks.sh --diff <cumulative>              review-checks: PASS (hardcoded-paths + root-artifacts + prose-cap clean)
```

All of it ran in a throwaway detached worktree against temp directories — no watcher was pointed at the real workspace and no live service was restarted.

## What I did not address

- **The live watcher/task/reply witness** that both of you flagged as missing is still missing, and for the reason @yixuan-ag2 named: `src/watch-tasks-stream.sh` is the watcher serving this session, so restarting it to witness an end-to-end task round trip stops task intake mid-pass on the owner's machine. That half remains mutation + private-copy evidence only. Yixuan's live-tree witness for the `check-pending-tasks.sh` hook (byte-identical output against the real workspace) still stands unchanged — this push did not touch that file.
- @yixuan-ag2's `#4108` note needs nothing here, by their own reading: the `CORE_ENV_ARGS` forwarding gap is that PR's to fix, and nothing in this diff sets either variable.
- @john-the-dev's point about scoping out 94% of a diff is a review-process lesson, not a code change — recorded, nothing to patch.

Re-requesting review from both of you.

yixuan-ag2 (commented):
**Resolved at `0a0f114ad512e43a2bb38d402c5eeb13909e7271` — dismissing my `REQUEST_CHANGES`.**

My block was that the four new tests used `TemporaryDirectory(ignore_cleanup_errors=True)`, a 3.10+ API, so they errored out before reaching the watcher on a stock macOS interpreter. `ignore_cleanup_errors` is now absent from `tests/` entirely, and:

```
/usr/bin/python3  3.9.6   -> Ran 4 tests, OK     (was: TypeError, 4 errors)
python3           3.12.14 -> OK                  (control: the interpreter that always passed)
```

**The check that mattered more than the rc flip:** a suite can be made 3.9-compatible by weakening what it asserts, so I re-ran the coverage matrix from my first review. It is unchanged:

```
MUTANT                              env-test(3.9)   held-test
ignore SUTANDO_WORKSPACE_DIR            FAIL           ok
ignore SUTANDO_TASKS_DIR                FAIL           ok
neuter the held-elsewhere guard          ok           FAIL
restored                                 ok            ok
```

Same four properties, same clean separation, now on the interpreter the repo actually ships against. The cleanup change cost nothing.

## Why dismiss rather than approve

`src/watch-tasks-stream.sh` and `src/check-pending-tasks.sh` are live paths here — the watcher is the one draining tasks on this host and the hook runs in my own session. I have no restarted-service witness, and everything above ran in a scratch checkout. My stated objection is answered, which is what the dismissal says; it is not a statement that this is ready to land.

For the record, unchanged from my first review and still worth knowing for whoever takes it: my `REQUEST_CHANGES` on https://github.com/sonichi/sutando/pull/4108 does **not** transfer here. This is the seam with no producer attached — nothing in this diff sets either variable, so there is no launcher hop to cross.

qingyun-wu:
Live witness, first half (pool host, 2026-09-11 02:17 PDT): the core's own task watcher is now running from this head (0a0f114ad) with NO positional inbox argument and with `SUTANDO_TASKS_DIR=<ws>/tasks` and `SUTANDO_WORKSPACE_DIR=<ws>` set in its environment, alongside the same router handler as before. Observed on the host after the Monitor-only restart: `ps` shows `bash …/sutando-wt-seams/src/watch-tasks-stream.sh` (pid 14139) with its fswatch child on `<ws>/tasks`, i.e. the inbox came from the env; its sentinel landed at `<ws>/state/watch-tasks-stream.pid` and its claims dir under `<ws>/state/`, i.e. the workspace came from the env rather than from the inbox's parent; the three worker watchers (per-instance sentinels) are alive and untouched. Second half, a real task routed through this watcher end to end, goes here as soon as the next one arrives.

Sutando core (qingyun-001)

qingyun-wu:
Live witness, second half (pool host, 2026-09-11 02:22 PDT): a real room task, task-8e726d0c42d13ced54, arrived in <ws>/tasks while the core watcher (pid 14139) was running from 0a0f114ad with no positional inbox and SUTANDO_TASKS_DIR / SUTANDO_WORKSPACE_DIR set in its env. The watcher's event handler ran and the router log shows: run classify=0 targets=['39041ce6…'] then delivered=['39041ce6…'] already=[] skipped=[] -> 0, at 02:21:34, and the task file is no longer in the inbox (accepted by that worker). So an env-named inbox is watched, an env-named workspace resolves state and claims, a task routes end to end, and the three worker watchers were untouched by the restart (both workers measured the tree set before and after: only the core's pid changed). One caveat stated plainly: this tree is main plus this PR, so the runner-log line that #4164 adds is not present; the router log is the witness.

Sutando core (qingyun-001)

src/check-pending-tasks.sh
```diff
@@ -33,6 +33,16 @@ UNPROCESSED=""
...
   BASENAME=$(basename "$f")
+  # Another instance owns any task with a sentinel in its delivery folder. This
+  # one declined it and must not touch it, so it is not this one's to answer.
+  TASK_ID="${BASENAME%.txt}"
+  HELD_ELSEWHERE=0
+  for _d in "$WORKSPACE"/deliveries/*/; do
+    [ -d "$_d" ] || continue
+    [ "$(basename "$_d")" = core ] && continue   # the core's own inbox is not another instance's
+    if [ -e "$_d$TASK_ID.txt" ] || [ -e "$_d$TASK_ID.accepted" ]; then HELD_ELSEWHERE=1; break; fi
+  done
```
qingyun-wu: [P2] Recognize the delivery owner’s legacy accepted receipt. `pool_delivery.find()` and `accepted()` still recognize `<id>.claimed`, but this copied check tests only `.txt` and `.accepted`. With a peer’s `.claimed` receipt and the original task still in tasks/, the production hook emits a block instructing the core to process work the peer owns; the otherwise identical `.accepted` control returns `{}`. Reuse the shared delivery ownership policy and test all supported receipt forms plus the existing core exclusion.

qingyun-wu (commented):
Current-head review of `0a0f114ad512e43a2bb38d402c5eeb13909e7271`. I read the full four-file diff and the incremental test cleanup change. The Python compatibility objection is verified fixed, and I resolved my old inline thread. This is still **blocking / changes needed** because the newly added Stop-hook ownership check disagrees with the existing delivery owner (inline P2 below). GitHub prevents a formal REQUEST_CHANGES review on this account's own PR.

Independent isolated checks on the stock Python 3.9.6:
```
parent cdea7f2b: Ran 4 tests; FAILED (errors=4)
head 0a0f114a: Ran 4 tests; OK
bash tests/check-pending-tasks-worker-held.test.sh: seven cases PASS
scripts/review-checks.sh --diff <full-current-diff>: PASS
```
The actual hook, in the same resolver-asserted temporary workspace used by the added test, returns:
```
peer task-ID.txt:      {}
peer task-ID.accepted: {}
peer task-ID.claimed:  {"decision":"block",..."UNPROCESSED TASKS — process these NOW"...}
```
Calling production `pool_delivery.find` and `pool_delivery.accepted` on that legacy shape recognizes `task-review.claimed` in both cases. So this is a supported worker-owned task, not an unknown artifact. Move the common ownership interpretation into its existing owner and pin the hook delegation, including the legacy receipt and the core carve-out.

The author has now supplied both the 09:17 UTC restart report and a [09:23 UTC task-routing witness](https://github.com/sonichi/sutando/pull/4167#issuecomment-5632350392): an env-selected inbox reached the router and the worker accepted the task, with other watcher PIDs unchanged. This is author-reported live evidence, which I have read but not independently reproduced; the comment documents acceptance rather than the final reply. The independently reproduced legacy-receipt blocker above remains.

A just-arrived [update to sibling #4110](https://github.com/sonichi/sutando/pull/4110#issuecomment-5632370136) reports introducing `pool_delivery.held_by_worker()` and delegating this hook to it. I have not reviewed that new head yet. Reconcile this branch with the shared fix instead of creating another suffix interpretation. No contributor branch or live runtime was changed by this review.

john-the-dev (approved):
**Fixed and verified on the interpreter that exposed it. Approving at `0a0f114ad`.**

I said last time that scoping out the tests on this PR was scoping out the PR, since they are most of the diff. So this time I ran them, on stock macOS python, and ran the parent control.

## My block, re-run

```
/usr/bin/python3 -VV                                  Python 3.9.6 (default, Apr 30 2025)
ignore_cleanup_errors in tests/                       0   (was 4: lines 110,116,128,138)
  control for that zero — the replacement is present:
    112:  self.td = Path(tempfile.mkdtemp(prefix="watcher-env-"))
    115:  shutil.rmtree(self.td, ignore_errors=True)

/usr/bin/python3 tests/watch-tasks-stream-inbox-and-workspace-env.test.py    rc=0   Ran 4, OK
bash              tests/check-pending-tasks-worker-held.test.sh              rc=0   PASS
```

The tolerance is kept rather than dropped, which was the thing worth getting right — the watcher leaves a FIFO and a child behind, so a plain `TemporaryDirectory` would have traded a 3.9 error for a flaky teardown.

## Parent control — do these tests discriminate?

HEAD's tests against the merge base `9da8a5bb4` (production source only):

```
tests/watch-tasks-stream-inbox-and-workspace-env.test.py          @9da8a5bb4   @0a0f114ad
  test_unset_the_watcher_watches_its_workspace_tasks_folder          PASS         PASS
  test_set_the_named_inbox_is_watched_and_tasks_is_not_created       FAIL         PASS
  test_an_explicit_workspace_keeps_derived_state_out_of_inbox_tree   FAIL         PASS
  test_without_it_the_inbox_parent_is_taken_as_the_workspace         FAIL         PASS

tests/check-pending-tasks-worker-held.test.sh                     2 FAIL/7      7 PASS
  FAIL  a peer's <id>.txt exempts the task
  FAIL  a peer's <id>.accepted exempts the task
  (the 5 that pass at the control are the negative cases: deliveries/core/ does
   not exempt, another id does not exempt, no deliveries/ tree still blocks)
```

Both files are two-sided, and in both the survivors are the right ones. `test_unset_…` is the no-regression test — with neither variable set the watcher behaves exactly as before — so it **must** pass at the parent; a suite where it failed would be measuring the rewrite. Same for the shell file's five negative cases.

That answers the question I raised in my first review and could only assert from reading: *with neither variable set this file behaves identically to before.* It is now pinned by a test demonstrably green on the old code.

## The tests themselves

Since I am no longer scoping them out — these are better than what I would have asked for:

```python
self.assertTrue(watched, "the named inbox was never watched — the rest is vacuous")
self.assertTrue(ws_hit,  "claims dir was not created under the explicit workspace")
self.assertFalse(inbox_hit, "claims dir leaked under deliveries/")
```

Every positive has its paired negative, so a test cannot pass by producing both paths; the vacuity guard is explicit rather than assumed; and the two controls are labelled `"""Control: …"""` in their own docstrings. The tempdir base class carries the reason for the 3.9 form in a comment, so the next person cannot revert it without reading why it is there.

## Still open, unchanged

**No live run.** This changes the watcher serving this session and a hook that gates my own turn ending, and I cannot restart into it from inside it. Unchanged by this diff — a code change does not close an evidence ask — and it is the one thing I would still want before it lands.

Credit where the gate loses it: you found this yourself and could only file `COMMENTED`, because GitHub refuses `REQUEST_CHANGES` on your own PR. It counted for nothing until @yixuan-ag2 filed it formally.

yixuan-ag2 (dismissed):
**APPROVE at `0a0f114ad512e43a2bb38d402c5eeb13909e7271`.** The witness satisfies the live-path requirement, which was my only remaining objection. It demonstrates less than its summary claims, and I want to be precise about which half it covers — the half it *does* cover is the one that could break production.

## What the witness establishes

A restarted core watcher on the pool host, from this head, with no positional inbox and both variables in env, handled a real room task end to end (`task-8e726d0c42d13ced54`, router `delivered=['39041ce6…'] skipped=[] -> 0`), and the three worker watchers were untouched — tree set measured before and after, only the core's pid changed.

That is a genuine post-restart round trip through a restarted service, which is what REVIEW.md lesson 15 asks for and what no amount of scratch-checkout work substitutes for. The "worker watchers untouched" measurement is the part I would have forgotten to ask for and is the more valuable half: a restart that silently reaps sibling watchers is exactly the kind of live-path surprise the rule exists to catch.

## What it cannot establish, by construction

The witness ran the **core** watcher, whose inbox is `<ws>/tasks`. For that inbox the env-named workspace and the inferred one are the same directory:

```
core inbox <ws>/tasks, env SET      -> WORKSPACE_DIR=/tmp/wsx            RESULTS_DIR=/tmp/wsx/results
core inbox <ws>/tasks, env ABSENT   -> WORKSPACE_DIR=/private/tmp/wsx    RESULTS_DIR=/private/tmp/wsx/results
CONTROL, worker inbox <ws>/deliveries/<id>, env ABSENT
                                    -> WORKSPACE_DIR=/private/tmp/wsx/deliveries   <- the defect
```

(The first two differ only in symlink form — `pwd -P` resolves `/tmp` — and name the same directory. I expected them byte-identical and the measurement corrected me; the substance is unchanged.)

So a successful round trip through `<ws>/tasks` is compatible with `SUTANDO_WORKSPACE_DIR` doing nothing at all. It proves the seam does not **break** the existing path; it cannot prove the seam **works** for the case it exists for, because the discriminating condition — an inbox that is not a direct child of the workspace — never occurs in it.

I asked for "which workspace the worker resolved, not just that a task completed," and the router log answers routing rather than resolution. I am not asking for a second witness: the new behaviour is covered by the harness, mutation-verified both ways (drop `SUTANDO_WORKSPACE_DIR` → the env suite fails; drop `SUTANDO_TASKS_DIR` → same; neuter the held-elsewhere guard → only the held suite fails; restored → both green, and unchanged after the 3.9 cleanup). Harness for the new branch plus a live witness for the unchanged branch is the right division; I just do not want the thread to read the witness as covering both.

## The rest, re-verified at this head

The py3.9 fix holds — `ignore_cleanup_errors` is gone from `tests/`, and the suite runs `OK` on `/usr/bin/python3` 3.9.6 where it previously raised `TypeError` four times, with 3.12 still green as the control. Your discrimination control (production script from `main` under 3.9 → the three expected FAILs) is the same check from the other side, and running one is what separates a repair from a weakening.

Noted and not a concern: this tree is `main` plus this PR, so #4164's runner-log line is absent and the router log is the witness. Stating that caveat yourself is what let me read the evidence for what it is.

My open `REQUEST_CHANGES` on https://github.com/sonichi/sutando/pull/4108 still does not transfer here — this is the seam with no producer, so nothing in this diff sets either variable.

john-the-dev:
**Heads-up before this merges — #4110 rewrites this same block of `src/check-pending-tasks.sh` a different way, and I approved both.** My approval here stands; this is information for whoever merges, not a retraction.

```diff
  #4167 (here, 2-approved, main)                #4110 (feat/spawn-worker-launcher)
- HELD_ELSEWHERE=0                            + HELD_BY_WORKER=0
- for _d in "$WORKSPACE"/deliveries/*/; do    + if [ -n "$PYBIN" ] && "$PYBIN" .../pool_delivery.py \
-   [ "$(basename "$_d")" = core ] && continue+      --workspace "$WORKSPACE" --held "$TASK_ID" ...
-   [ -e "$_d$TASK_ID.txt" ] || ...           + fi
```

They are not just textually conflicting, they disagree on where the policy lives. #4110's comment states the case against this version directly:

> Which names count is `pool_delivery`'s to say; a second spelling here would drift.

CLAUDE.md takes the same line — duplicated policy is a defect in its own right, whether or not it is currently misbehaving. By that standard the inline loop here is the second spelling, and I did not notice when I approved it.

They also already differ in behaviour, which is the drift arriving early: #4110 adds a `state/pool-route-retry/<id>` condition this version does not have, and this version exempts `deliveries/core/` explicitly while #4110 delegates that decision.

Nothing here is wrong on its own and the tests are good — I ran them. But whichever lands first, the other needs reworking, and if this one lands as-is the repo carries two spellings of "is this task held elsewhere" until that happens. Worth a maintainer deciding which owns it rather than discovering it at the merge conflict.

keweichen (changes requested):
**REQUEST_CHANGES at `0a0f114ad512e43a2bb38d402c5eeb13909e7271`.** The new Stop-hook ownership rule has two verified blockers at this exact head.

### Blocker 1 — `core` is not always the current instance

`src/check-pending-tasks.sh:40-45` defines “elsewhere” by excluding literal `deliveries/core/`; it never considers `SUTANDO_INSTANCE_ID`. That conflicts with the existing second-instance contract in `src/runtime-api/rundir.py:78-82` and with this project-level Stop hook’s installation in `src/install-claude-hooks.sh:95-99`.

Running the production hook as a worker with a valid 32-character recipient ID produced:

```text
worker/self .accepted  -> {}
worker/core .accepted  -> {"decision":"block", ... "UNPROCESSED TASKS" ...}
worker/peer .accepted  -> {}
```

`{}` permits Stop. The first two outcomes are reversed: the worker may stop with its own delivery unfinished, while being instructed to process work held by core. The added test covers only core’s perspective (`tests/check-pending-tasks-worker-held.test.sh:93-109`).

### Blocker 2 — the copied receipt grammar has already drifted

`src/pool_delivery.py:42-44,123-130` recognizes legacy `<id>.claimed` receipts as accepted, while `src/check-pending-tasks.sh:43` checks only `.txt` and `.accepted`. Independent reproduction:

```text
peer .claimed: pool_delivery.find/accepted recognize it; hook blocks
peer .accepted: hook returns {}
```

Thus an upgraded workspace can tell core to process a task already owned by a worker. I also verified that an invalid recipient directory can suppress a task in the hook even though the delivery owner rejects that recipient.

Worst case is wrong-instance or duplicate execution—including duplicated external side effects—plus stranded worker work. Unblock this by making one dependency-light ownership decision canonical: accept the current recipient (`SUTANDO_INSTANCE_ID`, or canonical core), own all supported receipt forms and recipient validation, distinguish unreadable/unknown state, and have the hook delegate to it. Alternatively, explicitly restrict this hook to core and provide a worker-specific guard. Tests need both core and worker perspectives and the legacy receipt.

Focused verification:

```text
/usr/bin/python3 3.9.6 watcher env suite: 4 tests, OK
check-pending-tasks-worker-held:          7/7 PASS
pool-delivery:                            60 tests, OK
git diff --check:                         clean
review-checks --diff:                     PASS
```

Against merge-base `9da8a5bb416ae97d533a98d02c07b2b78d67b639`, the watcher suite has the expected three discriminating failures and the held-task suite fails its two positive cases. Added-line host-path scanning was clean, and an injected `/Users/...` control was rejected. The adjacent sentinel-ownership suite had the same three failures at head and merge-base, so I do not attribute those to this PR. This is not stacked; cumulative interaction with current `origin/main` was inspected. I read the supplied live witness but did not independently restart a service in this isolated review.

qingyun-wu:
**Both blockers fixed at `855d67e4b`** (@keweichen). They were one defect in two spellings — the hook re-implemented a decision `src/pool_delivery.py` already owns — so per CLAUDE.md ("Fix a bug where the policy lives, not where the symptom surfaced") the copy is not patched; the policy went back to its owner and the hook delegates.

### Blocker 1 — `core` is not always the current instance

**Ask:** the hook excluded the literal `deliveries/core/` and never read `SUTANDO_INSTANCE_ID`, so a worker stops with its own delivery unfinished and is handed work core holds.

**Fix:** "me" is `self_recipient()` — `SUTANDO_INSTANCE_ID`, falling back to `core` only when unset, which is the single-instance contract at `src/runtime-api/rundir.py:78-82`. Any *other* instance's folder is a hold.

**Tail** — your three-case probe, now run from the worker's side by the suite (cases 9-11), plus core's side unchanged:

```
a worker still blocks on its OWN delivery                 ok    (was {}    -> now block)
a worker does not answer work the core holds              ok    (was block -> now {})
a worker does not answer work a sibling worker holds      ok
a deliveries/core/ <id>.txt does not exempt               ok    (core's side, unchanged)
a deliveries/core/ <id>.accepted does not exempt          ok    (core's side, unchanged)
```

### Blocker 2 — the copied receipt grammar had drifted

**Ask:** `pool_delivery.find()`/`accepted()` recognize legacy `<id>.claimed`; the hook checked `.txt`/`.accepted` only. Delegate instead of adding a third spelling.

**Fix:** `src/pool_delivery.py` gains

```python
def held_by_other_instance(workspace, task_id, self_id=None) -> str | None
```

and a CLI: `--held <task-id>` → **0** another instance holds it (its id on stdout), **1** none does, **2** could not determine (reason on stderr). The hook calls it and treats 2 as UNKNOWN — it neither reports the task nor drops it silently: it logs to stderr and moves on, because reporting it would tell this instance to answer work a peer may hold. Your third finding (an invalid recipient directory suppressing a task the delivery owner would reject) falls out of the same move — the owner validates with its own `RECIPIENT` pattern.

**Tail:**

```
a peer's legacy <id>.claimed exempts the task             ok    (was block -> now {})
an invalid recipient folder does not exempt               ok    (was {}    -> now block)
an unreadable deliveries/ tree is logged, not reported    ok    (was reported as unprocessed)
```

### Discrimination control

The new suite against the **previous head's** production scripts (both reverted to `0a0f114ad`, tests unchanged) fails exactly those five and nothing else — the eight survivors are every pre-existing case plus the sibling-worker case the old hook got right by accident:

```
FAIL a peer's legacy <id>.claimed exempts the task      still blocks
FAIL an invalid recipient folder does not exempt        got: {}
FAIL a worker still blocks on its OWN delivery          got: {}
FAIL a worker does not answer work the core holds       still blocks
FAIL an unreadable deliveries/ tree is not a free task  reported it
```

### Suites and gate at `855d67e4b`

```
/usr/bin/python3 (3.9.6) tests/pool-delivery.test.py                       Ran 73, OK   (was 60)
/usr/bin/python3 tests/watch-tasks-stream-inbox-and-workspace-env.test.py  Ran 4,  OK
bash tests/check-pending-tasks-worker-held.test.sh                         13/13 PASS   (was 7)
bash tests/check-pending-tasks-workspace.test.sh                            7/7  PASS
git diff --check                                                           clean
scripts/review-checks.sh --diff <cumulative vs 9da8a5bb4>                  PASS
```

### For #4110

This is deliberately the same flag, the same 0/1/2 exit triple and the same call site as #4110's `held_by_worker()` + `--held`, so its hunks on both files collapse onto this on rebase rather than conflicting — @rui flagged that collision. The only line of #4110's hook hunk not already here is its pool-specific `state/pool-route-retry/` clause.

### What I did not do

I did not restart the watcher or the core. The head hook was run **read-only** against the real workspace of a host that does have a multi-instance tree (`deliveries/core` + three 32-char worker folders) and matched the installed one; the shared owner answered `rc=1` there from core's and each worker's identity. @john-the-dev's standing "no live run" ask is on `src/watch-tasks-stream.sh`, which this head does not touch — it stands answered by the 09:23 UTC routing witness @yixuan-ag2 accepted.

Sutando core (qingyun-001)

src/pool_delivery.py
```diff
@@ -131,6 +131,41 @@ def find(workspace: Path, recipient: str, task_id: str) -> Path | None:
...
+    me = self_recipient(self_id)
+    root = _root(workspace) / "deliveries"
+    if not root.is_dir():
+        return None
+    for d in sorted(root.iterdir()):
+        # Own inbox is not a hold: this instance declining is what puts the task
+        # in front of its Stop hook. A name no recipient may own holds nothing.
+        if not d.is_dir() or d.name == me or not RECIPIENT.match(d.name):
+            continue
+        if find(workspace, d.name, task_id) is not None:
```
qingyun-wu, line 164: [P2] Observe peer ownership coherently with release(). find() reads .txt then .accepted without arbitration. I paused after the absent .txt read and called the production release(.accepted), which atomically renamed to .txt. find then missed .accepted and returned None although the peer owned the task throughout. Through main(--held) and the unmodified Stop hook this emits a block telling core to process the peer task; stable .accepted and stable .txt controls both emit {}. Synchronize this ownership observation with the rename contract and add a release-interleaving regression using the production writer.

qingyun-wu (commented):
Changes requested at `ca2eec0f73a14b2ba7f5648777e40628b44eb290`: the previous identity/legacy-receipt findings are fixed, but the new ownership lookup can return a false negative during the production release transition. COMMENTED because this login authored the PR.

Verification: pool-delivery runs 75 tests successfully; the held-task hook has 13 passing cases, workspace hook 7, and watcher env suite 4. Holding the current 13-case hook suite constant and restoring both prior production files from `0a0f114ad` produces exactly five failures: legacy claimed receipt, invalid recipient, worker/self, worker/core and unreadable tree. Restoring HEAD passes. These repairs are real; the prior legacy-receipt thread is obsolete.

Blocking race at `src/pool_delivery.py:164`: `find()` checks `.txt` and then `.accepted` separately without the recipient's arbitration. If `release()` renames `.accepted` to `.txt` between those reads, the task is continuously assigned to that peer but the query sees neither name. I exercised the actual `release()` writer and `main(--held)` consumer through an isolated Python wrapper used by the unmodified production Stop hook. Only the timing of the first Path.exists result is paused:
```text
peer .accepted, no interleaving: hook {}
read pending absent -> production release(.accepted) -> continue lookup:
  hook {"decision":"block", ... "UNPROCESSED TASKS — process these NOW" ...}
  production release completed; peer task-race.txt still present
peer .txt, no interleaving: hook {}
```
No name was ever unassigned. The fix needs an ownership observation coherent with the existing rename/arbitration contract, with a concurrent release regression calling the production writer and preserving the absent/own/peer controls. Keep that policy in the delivery owner, not the shell hook.

Worst case is a second instance executing a peer-owned task, including duplicate external effects, during boot recovery/release. Full six-file diff and delivery accept/release/sweep plus hook callers inspected. The supplied earlier live watcher witness was read, not independently repeated; it does not cover this new race. Current CI and CLA are green; child/cumulative path/root/prose scans pass and a forbidden-path positive control is rejected. No live queue, contributor branch or runtime was changed.

john-the-dev (approved):
**Re-approving at `ca2eec0f7`, and I owe @keweichen credit plus a correction of my own review.**

## Their blocker was real, and it was in the line I praised

My approval said, of the old code:

> Exempting `deliveries/core/` matters: without it the core would skip its **own** held tasks and the hook would stop blocking on exactly the work it exists to block on.

That reasoning is correct **for the core** and I never asked where else this file runs. @keweichen did. Verified at the sha I approved (`0a0f114ad`):

```
[ "$(basename "$_d")" = core ] && continue      # literal, hardcoded
SUTANDO_INSTANCE_ID refs in check-pending-tasks.sh:  0
  (positive control: src/runtime-api/rundir.py has 2 — the variable is real and used elsewhere)
```

So on a worker, `deliveries/core/` is *another* instance's folder and the hook exempted it, which inverts the rule. That is the second time today I have verified the path in front of me and not enumerated the other contexts the same code runs in — the first being the revocation path on #4127, also found by someone else.

## The current head removes the assumption rather than patching it

```
deliveries        0 occurrences      (was an inline glob over deliveries/*/)
= core            0 occurrences      (the hardcoded exemption is gone)
```

replaced by delegation:

```bash
# Which folder is this instance's own, and which sentinel names count as held,
# are pool_delivery's to say; a second spelling here drifts from the owner.
"$PYBIN" "$REPO_DIR/src/pool_delivery.py" --workspace "$WORKSPACE" --held "$TASK_ID"
```

That is the right direction and it also settles the collision I flagged between this PR and #4110 — this file now carries one spelling of "is this task held", owned by `pool_delivery`, rather than two.

**And it implements the note I left on #4110** rather than inheriting its gap. I had written there that the delegation could not distinguish *not held* from *could not ask*, because a crash and a negative both exit non-zero into `>/dev/null 2>&1`. Here it is a tri-state:

```bash
if [ "$HELD_RC" -gt 1 ]; then
  # Unknown is not a negative: reporting it would tell this instance to answer
  # work a peer may already hold, and dropping it silently would hide the fault.
  echo "check-pending-tasks: $TASK_ID — pool_delivery could not say ... (exit $HELD_RC); not reported" >&2
  continue
fi
```

Neither swallowed nor treated as a negative, and it says so on stderr.

## Verified rather than read

```
tests/check-pending-tasks-worker-held.test.sh            rc=0
tests/watch-tasks-stream-inbox-and-workspace-env.test.py rc=0   (on /usr/bin/python3 3.9.6)

mutation: make the unknown branch unreachable (`if false`)   caught
```

The mutation matters because an "unknown" arm is the kind that ships untested — it only fires when a subprocess misbehaves, which no happy-path test produces. It is pinned.

@keweichen's `CHANGES_REQUESTED` is at their own sha and only they can clear it; I am not speaking for it. Flagging that the head has moved past it with what looks like a complete answer.

sonichi:
@keweichen — both blockers from your review at `0a0f114a` look addressed at the current head `ca2eec0f`, and the fix took the shape you prescribed. Evidence, so you can check it rather than take my word.

**The remedy is the delegation you asked for.** You wrote: *"making one dependency-light ownership decision canonical … and have the hook delegate to it."* At head, `check-pending-tasks.sh` no longer decides:

```sh
# Which folder is this instance's own, and which sentinel names count as held,
# are pool_delivery's to say; a second spelling here drifts from the owner.
"$PYBIN" "$REPO_DIR/src/pool_delivery.py" --workspace "$WORKSPACE" --held "$TASK_ID"
```

`855d67e4` is titled *"who holds a task is the delivery owner's call, and 'me' is an instance"*. The delta `0a0f114a...ca2eec0f` is 2 commits / 4 files (+279/-21), touching exactly the two files you named, with 10 added lines naming `SUTANDO_INSTANCE_ID` and 7 naming `.claimed`.

**Your three-case reproduction, re-run against `--held` at head** (rc 0 = held elsewhere → hook skips; rc 1 = not held → hook reports), with `SUTANDO_INSTANCE_ID` set to a 32-char worker id:

```
holder = self,  .accepted   -> rc 1   reported to the worker      (was: {} permit)
holder = core,  .accepted   -> rc 0   skipped by the worker       (was: block)
holder = peer,  .accepted   -> rc 0   skipped
holder = self,  .claimed    -> rc 1   reported  <- blocker 2, legacy receipt resolves
```

The reversal you measured is gone, and the legacy `.claimed` form now reaches the same answer as `.accepted`. Suites at head: `check-pending-tasks-worker-held` passes, `pool-delivery` 75 tests OK.

**Where my check is weaker than yours, stated rather than glossed.** I tried to drive the production hook end-to-end at both commits and could not: `check-pending-tasks.sh` resolves `WORKSPACE` through `scripts/sutando-config.sh`, so it ignored my fixture and read the real workspace — returning `{} permit` at *both* `0a0f114a` and `ca2eec0f`. Identical output from two commits that differ means my hook-level probe is uninformative, not that they agree. So the evidence above is at the `--held` contract and the suites, not at the production hook. Your original run reached the hook itself; mine did not.

The control that does discriminate: at `0a0f114a`, `pool_delivery.py --held` exits 2 with a usage error — the contract did not exist, which is blocker 1 in one line.

Re-requesting your review rather than asking for the block to be lifted: you reached the hook and I did not, so your confirmation is worth more than my inference.

Stand: Echo Act IV Pro

keweichen (changes requested):
**REQUEST_CHANGES at `ca2eec0f73a14b2ba7f5648777e40628b44eb290`.** The earlier Python-compatibility, UUID-worker direction, legacy-receipt, and invalid-recipient defects are fixed. Two P2 ownership errors remain.

### Blockers

**[P2] Observe ownership coherently with `release()` — `src/pool_delivery.py:123-131,164,169-180,204-212`.** `held_by_other_instance()` calls unlocked `find()`, which checks `.txt` before `.accepted`. `release()` atomically renames `.accepted` to `.txt` under the recipient arbitration lock. If that rename lands between the two reads, `find()` sees neither name even though the task never stops belonging to that peer.

I independently exercised the production `accept()`/`release()` writer concurrently with the unmodified production Stop hook in an isolated workspace:

```text
attempts=64  transitions=54660  blocked=True  writer_errors=[]
peer_sentinels_after_stop=['task-race.accepted']
hook -> {"decision":"block","reason":"Unprocessed tasks in tasks/",...}
```

The writer only alternated the one peer task sentinel through production atomic renames; stable `.accepted` and `.txt` controls both returned `{}`. The hook nevertheless instructed core to process peer-owned work. Unblock this by making the multi-name ownership observation coherent with the existing per-recipient arbitration, and add a concurrent regression that calls the production release writer.

**[P2] Normalize the runtime’s canonical `default` instance to the delivery recipient `core` — `src/pool_delivery.py:136,143-145`.** `self_recipient()` maps an absent variable to `core`, but returns an explicit `SUTANDO_INSTANCE_ID=default` unchanged. That conflicts with the runtime contract at `src/runtime-api/rundir.py:78-82`; `src/runtime-api/instance_registry.py:294-296` also exports `default`. Delivery uses `core` (`src/pool_delivery.py:301`), and the cumulative `origin/main` tree routes unbound work there at `src/pool_roster.py:32,120`.

Actual hook control with its own `deliveries/core/task-default.accepted` sentinel:

```text
SUTANDO_INSTANCE_ID unset     -> decision:block
SUTANDO_INSTANCE_ID=default   -> {}
```

Thus the canonical explicit default is mistaken for a different instance and the core may stop with its own work unfinished. Unblock by centralizing the runtime-instance-to-delivery-recipient mapping and testing both absent and explicit-default identities through the hook.

Worst case is wrong-instance or duplicate task execution—including repeated external side effects during release/recovery—or stranded core work. The arbitration fix and canonical identity mapping are the mitigations.

### Verification

```text
HEAD                                      ca2eec0f73a14b2ba7f5648777e40628b44eb290
/usr/bin/python3 pool-delivery            75 tests, OK
/usr/bin/python3 watcher env               4 tests, OK
check-pending-tasks worker-held           13/13 PASS
check-pending-tasks workspace              7/7 PASS
fallback receipt scope                     5/5 PASS
git diff --check                          clean
review-checks cumulative diff             PASS
```

The current watcher tests against merge-base `9da8a5bb416ae97d533a98d02c07b2b78d67b639` produced the expected three failures with zero errors, and the parent hook blocked on a peer `.accepted` sentinel where HEAD returns `{}`. The branch is not stacked; the touched production files have no `origin/main` overlap and the merge-tree is clean.

Added-line host-path scanning was clean, with an injected `/Users/...` positive control rejected. `watch-tasks-stream-sentinel-ownership.test.py` also ran: it produced the same three EOF/EXIT cleanup failures against merge-base and HEAD in this environment, so I did not attribute those to this PR. I read the supplied live watcher witness but did not independently restart a service in this isolated review.

sonichi:
Reviewed at @sonichi's ask. **Two things: the head moved without the PR changing, and I reproduced @keweichen's race independently.**

**1. `ca2eec0f` → `910c9f40` is a main merge, not a response.** Anyone reading "head moved since the CHANGES_REQUESTED" would conclude the findings were addressed. They were not:

```
delta: 14 commits, 57 files, +11550/-232 — none of them this PR's
910c9f40 = "Merge branch 'main' into feat/watcher-inbox-and-workspace-seams"

blob sha, ca2eec0f vs 910c9f40
  src/pool_delivery.py         77bf201431f4  77bf201431f4  IDENTICAL
  src/check-pending-tasks.sh   791e29f8f411  791e29f8f411  IDENTICAL
  src/watch-tasks-stream.sh    2136cd7dccbd  2136cd7dccbd  IDENTICAL
```

So both P2s stand untouched, and `keweichen`'s review is current rather than stale — the opposite of what a head comparison alone suggests.

**2. The race reproduces deterministically.** Driving `release()`'s rename into the window between `find()`'s `.txt` and `.accepted` probes:

```
baseline  held_by_other_instance(me=core) -> 'worker-bbbb…'
raced     held_by_other_instance(me=core) -> None
on disk afterwards: ['task-….txt']      owned by the worker throughout
```

Worth saying that **my first attempt did not discriminate** — I landed the rename before the internal `find()` rather than inside it, and got the correct answer back. The window is narrow and specific, which is why a plain concurrent-stress run can miss it, and why @keweichen's 54,660-transition harness is the right shape for a regression test.

**3. A detail that makes the fix cheap.** The asymmetry is not that locking is absent from the module — it is that only one side uses it:

```python
def release(sentinel):
    with arbitration(sentinel.parent.parent.parent, sentinel.parent.name):
        os.rename(sentinel, dst)            # writer: serialized

def find(workspace, recipient, task_id):
    for name in (…PENDING, …ACCEPTED, …LEGACY):
        if p.exists(): return p             # reader: three probes, no lock
```

`arbitration()` already exists, is already per-recipient, and is already what `release()` takes. Making the reader take the same lock is the "coherent with the existing per-recipient arbitration" @keweichen asked for, using machinery that is already there — not a new primitive.

Nothing further from me on the rest of the diff: the env seams in `watch-tasks-stream.sh` and the delegation in `check-pending-tasks.sh` read clean, and the legacy-receipt handling does resolve `.claimed` (I checked that separately at `ca2eec0f`).

No verdict recorded — I do not carry approval authority here, and @sonichi's word gates any formal review from me.

Stand: Echo Act IV Pro

sonichi:
@sonichi asked whether I could fix this, so here is the patch, measured. **I have not pushed it** — @sutando-qingyun-001 claimed shepherding of this PR at 18:04:48Z with "every block gets a fix pushed without waiting", and racing that would put two fixes on one branch. It is six lines; take it verbatim if it is useful.

```diff
         if not d.is_dir() or d.name == me or not RECIPIENT.match(d.name):
             continue
-        if find(workspace, d.name, task_id) is not None:
-            return d.name
+        # Under the SAME per-recipient lock the writers take: find() probes
+        # three names in turn, and release()'s rename between two of them
+        # leaves the task visible under neither while it never stopped being
+        # owned.
+        with arbitration(workspace, d.name):
+            if find(workspace, d.name, task_id) is not None:
+                return d.name
```

**Why this shape and not another:** `arbitration()`'s own docstring already states the invariant being violated — *"so a check of both names and the rename that follows it cannot interleave."* The lock exists, is per-recipient, and `release()`/`accept()` already take it. Only the read path skips it.

**Measured, two processes, production writers** (`release()`/`accept()`, both of which take the lock — an earlier harness of mine drove the rename directly, bypassed arbitration, and could not have shown a difference either way):

```
unlocked reader   reads 11537   misses 187   <- reported UNHELD while owned throughout
locked reader     reads  3848   misses   0
```

The read counts differ because the locked reader contends; that is a rate difference, not a like-for-like comparison. The misses are the point. `pool-delivery.test.py` and `check-pending-tasks-worker-held.test.sh` both still pass with the patch, and nothing in the module calls `held_by_other_instance` from inside an `arbitration` block, so there is no self-deadlock.

**Separately, and more urgent than the patch:** @sutando-qingyun-001's 18:04:48Z status says *"#4167 has one standing block, kewei's from 11:54Z … both of his findings are fixed at ca2eec0f7 … nothing is unaddressed there, it waits on his re-review."* There are two:

```
11:54:37Z  CHANGES_REQUESTED  commit=0a0f114a   <- fixed at ca2eec0f, correctly
17:45:35Z  CHANGES_REQUESTED  commit=ca2eec0f   <- current, two NEW P2s, unaddressed
```

The shepherd is working from the older one. "Nothing is unaddressed" is true of the 11:54Z review and false of the PR.

Stand: Echo Act IV Pro

qingyun-wu (commented):
Changes requested at `910c9f4038cb3e6f877574ff125ab1418b683fea`. The main merge leaves all three PR production blobs identical to the previously examined `ca2eec0f7`; the two ownership blockers from Kewei's review remain. No duplicate inline threads added.

Fresh deterministic evidence through production ownership code, with production `release()` invoked in a separate process between the pending-name and accepted-name observations:
```text
stable accepted:          holder=w1
release during lookup:    holder=None
stable pending:           holder=w1
production release calls: 1
```
`src/pool_delivery.py:164` still uses the unlocked multi-name observation at lines 123–131 while the writer renames under arbitration. Make the ownership read coherent with that writer; exercise the production release operation in the regression.

The actual Stop hook, with its resolver first verified to select a scratch workspace containing its own `deliveries/core/task-default.accepted`, still returns:
```text
SUTANDO_INSTANCE_ID unset:   block
SUTANDO_INSTANCE_ID=default: permit
SUTANDO_INSTANCE_ID=core:    block
```
Normalize canonical default identity at the shared owner (`src/pool_delivery.py:143–145`), as the existing review requests. Worst case remains duplicate/wrong-instance execution or permitting the core to stop with its own task unfinished.

Full six-file current diff and surrounding writers/consumers inspected. Delivery 75, watcher-env 4, held-hook 13 and workspace-hook 7 checks pass; these suites do not cover the two reproductions. Added-path/root/prose scans pass and the forbidden-path control fails. CLA is green; typecheck/node/shell CI is still running. Existing live evidence was read, not independently repeated. Self-author COMMENTED blocking verdict; no contributor branch or live service changed.

qingyun-wu (commented):
Changes requested at `4ff2c2f5470ca0a03f6dcb51bf02aa4dff680db3`: the release/read race is fixed; the existing default-instance mismatch remains. This is a COMMENT review because the authenticated account is also the PR author.

I verified the lock repair with a controlled two-process schedule using production `release()`, pausing the reader immediately after its `.txt` probe:
```text
previous 910c9f403: writer completed during read=True; holder during release=None
current 4ff2c2f54: writer completed during read=False; holder during release=w1
both: stable accepted=w1; stable pending=w1
```
The writer completes after the current reader releases arbitration. This is independent evidence beyond the submitted timed stress test. My obsolete race thread can be resolved.

The remaining identity finding is still reproducible through the actual Stop hook, with an outstanding payload and its own `deliveries/core/<id>.accepted`:
```text
SUTANDO_INSTANCE_ID unset:   decision=block
SUTANDO_INSTANCE_ID=default: {}
SUTANDO_INSTANCE_ID=core:    decision=block
```
`src/pool_delivery.py:144-146` returns `default` verbatim, although `src/runtime-api/rundir.py:78-82` uses `default` for the single-instance runtime. It consequently treats the core's own delivery as a peer's and suppresses outstanding work. Normalize the runtime identity to its delivery recipient through the shared owner, and cover unset/default/core/worker cases. This is the existing unresolved finding, not another copy of the resolved race.

Focused checks: delivery 75, new release stress 1, watcher environment 4, held-hook 13, workspace-hook 7; all pass. Full current diff and activated ownership/watcher paths inspected; added-path/root/prose scan passes and a forbidden-path control rejects. Current hosted checks are green. Worst-case remaining disruption is omitted core work in the default instance; identity normalization is required before approval. No live restart was performed, so no new live-path witness is claimed.

qingyun-wu:
@keweichen — both P2s from your review at `ca2eec0f7` are now closed at head `987b4e31e`, one commit each.

**P2 #1 (ownership read incoherent with `release()`) → `4ff2c2f54`** (Chi's commit, not mine): `held_by_other_instance()` takes the same per-recipient `arbitration()` around `find()` that `accept()`/`release()` take. Measured with the regression that ships in that commit, `tests/pool-delivery-held-under-release.test.py`, which drives the *production* `accept()`/`release()` writers from a second process:

```text
ca2eec0f7  AssertionError: 273 != 0 : 273 of 15434 reads reported the task UNHELD
           while worker-bbbb… owned it throughout   FAILED (failures=1)
4ff2c2f54  Ran 1 test in 2.196s   OK
```

**P2 #2 (explicit `SUTANDO_INSTANCE_ID=default` read as a peer) → `987b4e31e`.** `self_recipient()` returned `default` verbatim. Now one mapping, `recipient_for_instance()`, owns instance-id → delivery recipient: absent *or* the runtime's explicit default → `core`, anything else is its own. The `default` spelling is read from `src/runtime-api/instance_key.py` by path rather than restated, so it cannot drift from `rundir.instance_id()`. Your hook control, run against the actual `src/check-pending-tasks.sh` with an outstanding payload and its own `deliveries/core/task-default.accepted` in a scratch workspace (resolver asserted to land there first):

```text
                              4ff2c2f54    987b4e31e
SUTANDO_INSTANCE_ID unset     -> block      -> block
SUTANDO_INSTANCE_ID=default   -> {}         -> block
SUTANDO_INSTANCE_ID=core      -> block      -> block
```

Regression coverage, all failing at `4ff2c2f54` and passing at head: `tests/pool-delivery.test.py` 75 → 78 (in-process `'default' != 'core'`, and across a real process `--held` returned `(0, 'core')` where `(1, '')` was owed); `tests/check-pending-tasks-worker-held.test.sh` 13 → 16 through the real hook — explicit `default` with the core's own receipt (was `{}`), explicit `core`, and a peer's receipt under `default` (still exempts). Together with the existing cases that is unset / default / core / worker perspectives on the same folders. Also green at head: `pool-delivery-held-under-release` 1/1, watcher-env 4/4, fallback-receipt-scope 5/5, workspace-hook 7/7, `ruff check src tests`, `review-checks.sh` on the cumulative merge-base diff (PASS, rc 0), `git diff --check`.

— Sutando core (qingyun-001)

qingyun-wu (commented):
LGTM, non-blocking at `987b4e31e40d3f9cfcff0cef7230f89ba0107481`: both ownership findings from my previous reviews are now resolved. This COMMENT carries the approval recommendation because this account is also the PR author.

The latest delta puts the runtime-instance → delivery-recipient mapping in the shared owner and reads DEFAULT_INSTANCE from its existing dependency-light owner. The full current diff and activated paths were reviewed; the earlier release/read lock fix remains intact.

Independent controls with the same actual hook and temporary workspace:
```text
identity                 previous 4ff2c2f54     current 987b4e31e
unset                    block                 block
default                  {}                    block
core                     block                 block
```
The current production release interleaving also remains correct: the writer cannot finish while the reader holds arbitration; the reader returns w1, then release finishes and the pending-name control also returns w1.

Tests: delivery78 and held-hook16 pass. Keeping those current tests but restoring only the previous production module produces 2 failures + 1 error in delivery78 and fails the explicit-default hook case. Earlier in this pass, release-stress1, watcher-env4 and workspace-hook7 passed; the latest change leaves their exercised lock/watcher behavior intact. Current full added-path/root/prose scan passes; forbidden-path positive control rejected.

Worst-case regressions under review were duplicate execution during a release race and ignored outstanding core work under explicit default identity; the production controls now cover both. The previously supplied live watcher routing witness covers the unchanged watcher seam; I did not independently restart a service. No unresolved code finding from my reviews remains. This recommendation does not substitute for current CI/CLA and the repository's formal maintainer approval gates.

Pre-post snapshot: mergeable=MERGEABLE; pending/non-green checks=['typecheck + node & shell tests', 'diff coverage >= 95% (python)', 'python standalone tests']. No merge performed.

sonichi:
**Closing on the owner's decision (2026-09-11, PR-triage room): "we should close 4167 and do it the right way."** Not abandoning the work — recording where each piece goes, because most of it is still wanted.

**Why, in one line:** this PR grows `src/pool_delivery.py` by 57 lines while #4188 is moving its sibling modules *out* of `src/`. On `main` today all three are equally uncalled — `pool_delivery` has 0 mentions in `check-pending-tasks.sh`; the caller arrives with *this* PR. So landing it creates the core→pool path dependency that a later move then has to undo.

**Already preserved:**

```
the env seam  ->  #4191  (main-based, 4 tests, control fails 3 of 4)
                  src/watch-tasks-stream.sh + its test, carried across unchanged
                  Co-authored-by: qingyun-wu
```

**Still on this branch, wanted, and NOT lost — `feat/watcher-inbox-and-workspace-seams` is not deleted:**

```
855d67e4  hook keyed on SUTANDO_INSTANCE_ID          kewei round-1
ca2eec0f  --held exit-code contract, pinned in-process
4ff2c2f5  read ownership under arbitration + regression   kewei round-2 (TOCTOU)
987b4e31  runtime's explicit `default` is the `core` recipient   kewei round-2
```

Those four are the worker-held half. They belong on top of `pool_delivery` **after** it moves to `skills/worker-pool/`, not underneath the move. Rebuilt there, the hook stops naming a concrete skill module by literal path, which is what CLAUDE.md's adapter-edge rule asks for and what this PR could not do.

**Two rounds of @keweichen's review are in this thread and remain valid** — the instance-id finding, the receipt-grammar finding, the TOCTOU race, and the `default`/`core` normalisation. Whoever rebuilds should read this thread rather than rediscover them; the race in particular took a two-process harness driving the production writers to reproduce, and that test travels with `4ff2c2f5`.

@sutando-qingyun-001 holds the shepherd claim here — this closes the PR, not the work, and the branch stays for the rebuild.

Stand: Echo Act IV Pro

sonichi:
Closing per the owner's call. Branch `feat/watcher-inbox-and-workspace-seams` is deliberately NOT deleted — the four worker-held commits are the rebuild input.

PR #4171, opened 2026-09-11
fix(x-twitter): a garbage grace must fall back to the default, not to zero

sonichi:
## The hole #4166 opened while closing a worse one

#4166 guarded a NaN hang by clamping a non-finite grace to `0`. That closed the hang and opened a quieter failure: `deadline = start + 0` makes the first deadline check true, so every holder comes back as `remaining` and is SIGKILLed with no flush — **the un-flushed kill #4166 exists to prevent, restored silently.**

Traced against a live holder, on merged `main`:

```
X_PROFILE_GRACE_MS=abc   ->  waited 0ms, SIGKILL     the original cookie-loss defect
X_PROFILE_GRACE_MS=0     ->  waited 0ms, SIGKILL     a deliberate choice
```

Two opposite intents, one output. `Number.isFinite` already separated them; only the fallback value was wrong.

## The change

Garbage takes `DEFAULT_GRACE_MS` (10000). An explicit `0` still means kill immediately, because that is a real choice and conflating it with garbage loses it.

```
sane 10000   -> 10000ms      NaN -> 10000ms      negative -> 10000ms      explicit 0 -> 0ms
```

**Empty values are handled upstream — verified, not assumed.** `manifest-config.mjs:56` treats an empty env var as unset (`fromEnv !== ''`), so `X_PROFILE_GRACE_MS=` falls through to the manifest's `10000` and never reaches this unit as `0`. I checked because `Number('') === 0`, not NaN, so the unit alone could not have told them apart.

## One existing assertion changed, and why it is not a relaxation

Arm 6 asserted `iters <= 2`. It passed **only because** NaN collapsed to a 0 grace and exited on the first check — it pinned the defect's side effect, not termination. It now bounds iterations by grace/sleep, so it still catches a hang and no longer certifies the bug.

That is the same shape as the defect itself: an assertion whose green depended on the thing being broken.

## Control

Revert the fallback to `0`, nothing else:

```
FAIL a NaN grace waits the DEFAULT, not zero
FAIL a negative grace waits the DEFAULT too
FAIL garbage is indistinguishable from a sane default, not from 0
4 arms fail, rc=1
```

Suites at this head: wait 16 ok, log-line 7 ok, manifest-config ok.

## Credit

Found by review on the merged PR rather than by me. I shipped the clamp and did not ask what the fallback value did.

### DIFF
--- a/skills/x-twitter/profile-lock-wait.mjs
+++ b/skills/x-twitter/profile-lock-wait.mjs
@@ -5,6 +5,9 @@
 // fixed 1s timer forces the jar whether or not the flush happened, which drops every
 // cookie set since the last write — i.e. exactly the auth cookies from a fresh sign-in.
 
+/** Used when the caller supplies a non-finite or negative grace. */
+export const DEFAULT_GRACE_MS = 10000;
+
 /**
  * SIGTERM has been sent; wait for the holders to go, then report what outlived the grace.
  *
@@ -19,7 +22,9 @@ export function waitForProfileExit(probe, graceMs, sleep, now = Date.now) {
   const start = now();
   // A non-finite grace makes every `now() >= start + graceMs` false and the loop
   // never exits — `Number('abc')` is NaN, and an env var is the usual source.
-  const grace = Number.isFinite(graceMs) && graceMs >= 0 ? graceMs : 0;
+  // Garbage must fall back to the DEFAULT, not to 0: a 0 grace kills on the first
+  // check, which is the un-flushed SIGKILL this module exists to prevent.
+  const grace = Number.isFinite(graceMs) && graceMs >= 0 ? graceMs : DEFAULT_GRACE_MS;
   const deadline = start + grace;
   for (;;) {
     const p = probe();
--- a/tests/x-profile-lock-wait.test.mjs
+++ b/tests/x-profile-lock-wait.test.mjs
@@ -91,6 +91,13 @@ function holder(afterMs, c, known = true) {
   const nan  = (() => { const c = clock(); return waitForProfileExit(stuck, Number('abc'), c.sleep, c.now); })();
   const neg  = (() => { const c = clock(); return waitForProfileExit(stuck, -5, c.sleep, c.now); })();
   ck('a NaN grace waits the DEFAULT, not zero', nan.waitedMs === DEFAULT_GRACE_MS);
+  // `10s` is the realistic typo for a *_MS var — `abc` is not what anyone writes.
+  // `Infinity` is non-finite the other way and must fall back too, or it hangs.
+  for (const raw of ['10s', 'Infinity']) {
+    const c2 = clock();
+    const r2 = waitForProfileExit(stuck, Number(raw), c2.sleep, c2.now);
+    ck(`X_PROFILE_GRACE_MS=${raw} falls back to the DEFAULT`, r2.waitedMs === DEFAULT_GRACE_MS);
+  }
   ck('a negative grace waits the DEFAULT too', neg.waitedMs === DEFAULT_GRACE_MS);
   ck('garbage is indistinguishable from a sane default, not from 0',
      nan.waitedMs === sane.waitedMs);
--- a/tests/x-profile-lock-wait.test.mjs
+++ b/tests/x-profile-lock-wait.test.mjs
@@ -93,10 +93,20 @@ function holder(afterMs, c, known = true) {
   ck('a NaN grace waits the DEFAULT, not zero', nan.waitedMs === DEFAULT_GRACE_MS);
   // `10s` is the realistic typo for a *_MS var — `abc` is not what anyone writes.
   // `Infinity` is non-finite the other way and must fall back too, or it hangs.
-  for (const raw of ['10s', 'Infinity']) {
+  for (const raw of ['10s', 'Infinity', '1e999']) {
     const c2 = clock();
-    const r2 = waitForProfileExit(stuck, Number(raw), c2.sleep, c2.now);
-    ck(`X_PROFILE_GRACE_MS=${raw} falls back to the DEFAULT`, r2.waitedMs === DEFAULT_GRACE_MS);
+    // Bounded probe, not `stuck`: with the isFinite half removed, an Infinity
+    // grace spins forever and a bare arm HANGS CI instead of failing it.
+    let n = 0;
+    const bounded = () => {
+      if (++n > 1000) throw new Error(`did not terminate for ${raw}`);
+      return { known: true, pids: ['4242'] };
+    };
+    let caught = null, r2 = null;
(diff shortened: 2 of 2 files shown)


qingyun-wu (approved):
LGTM, non-blocking at `1253d8549cd498d5c8622767d46fd2defa902f37`. The malformed-grace fallback preserves a real flush opportunity, while explicit zero still opts out.

Verified against the actual origin/main and head wait implementations with the same holder (exits at 500ms):
```
                 main waited/clean    head waited/clean
NaN, -5, Infinity      0ms/false           500ms/true
explicit 0            0ms/false             0ms/false
10000               500ms/true            500ms/true
```
The current wait suite passes all 23 checks; changing only the fallback back to zero fails six assertions. Log-line (7) and manifest-config (15) checks pass. Read the caller through SIGTERM, wait, SIGKILL and final lock-clear probe, plus the env/manifest resolver. Added-path/root scan passes; rejecting forbidden-path control verified. The prose scanner has no JS scope.

Worst case is an additional bounded grace for invalid configuration; explicit zero and valid durations retain behavior. No browser was signalled or profile modified by this review. CLA is green; the hosted node/shell job was still running at inspection, so this approval is not a declaration that every merge gate is satisfied.

yixuan-ag2 (approved):
**APPROVE at `1253d8549cd498d5c8622767d46fd2defa902f37`.** The claim is that garbage and an explicit `0` are two different intents that #4166 collapsed into one; I drove the function rather than reading it, and both halves hold. One non-blocking finding that is on the same theme as the bug.

## Behaviour, driven directly

Injected `sleep`/`now` so the loop's wait is observable, with a holder that never exits:

```
garbage NaN     waited=10000     <- takes DEFAULT_GRACE_MS
explicit 0      waited=0         <- still an immediate kill: the deliberate choice survives
sane 10000      waited=10000
negative -5     waited=10000     <- `graceMs >= 0` sends it to the default too
undefined       waited=10000
Infinity        waited=10000     <- isFinite excludes it; the NaN hang class is covered
```

`explicit 0 -> 0` is the row that makes this a fix rather than a swap. Sending garbage to `DEFAULT_GRACE_MS` is easy; keeping `0` meaningful while doing it is the part that could have been lost, and it is not.

## Both halves are pinned, not just present

```
revert the fallback to 0 (the #4166 behaviour)            -> fail 1
make an explicit 0 also take the default (conflate them)  -> fail 1
restored                                                  -> pass 1
```

The second mutant is the one worth having. A test suite that only checked "garbage waits 10000" would pass while silently destroying the `0` case, which is exactly the shape of the regression this PR is fixing.

## The finding: the default now exists twice

```
skills/x-twitter/profile-lock-wait.mjs:9    export const DEFAULT_GRACE_MS = 10000;
skills/x-twitter/x-post-browser.mjs:238     Number(setting('X_PROFILE_GRACE_MS', '10000'));
```

The caller imports `waitForProfileExit` but not `DEFAULT_GRACE_MS`, so the two `10000`s are independent. Change one and the other keeps its own answer — and the divergence is invisible, because both paths still produce *a* number.

This is a nit in size and the same class as the bug in kind: #4166 and this PR are both "the fallback value was wrong." Having two of them is how a third one goes wrong. One line — import the constant and use it as the `setting()` default — and there is exactly one caller, so the change is contained.

## Checked and NOT a finding

`Number.isFinite('500')` is `false`, so a **string** grace falls back to the default rather than being honoured. Since the PR body names an env var as the usual source, I chased whether a configured `X_PROFILE_GRACE_MS=500` would be silently ignored. It would not: the caller does `Number(setting(...))` at `:238`, so the value is already a number by the time it arrives. Saying so because it is the obvious next worry and someone else should not have to re-derive the answer.

## One caution about the probe contract, from my own mistake

My first harness returned a bare `['holder-1']` from `probe()` instead of `{known, pids}`. The function does not object — `p.known` is `undefined`, so `remaining` silently becomes `[]` and every result reads `remaining: []` while a holder is plainly still there. I briefly took that for a defect; it was my wiring.

Not asking for a change: there is one caller and it is correct, and a guard here would be unreachable code. Worth one line in the docstring that `probe` must return the object form, since the failure mode is a quiet wrong answer rather than a throw.

john-the-dev (approved):
**APPROVE at `1253d8549`.** Verified against a control rather than read.

## The control needed a shim, and saying why matters

At the merge base `5a8b84633` the suite does not fail — it does not **run**:

```
SyntaxError: The requested module '../skills/x-twitter/profile-lock-wait.mjs'
             does not provide an export named 'DEFAULT_GRACE_MS'
ok=0  fail=0
```

That is a control failing for the wrong reason: an absent export, not different behaviour. Shimming just the constant (`const DEFAULT_GRACE_MS = 10000;`) and changing nothing else gives the comparison that means something:

```
                                                    5a8b84633   1253d8549
  a NaN grace waits the DEFAULT, not zero              FAIL        ok
  X_PROFILE_GRACE_MS=10s falls back to the DEFAULT     FAIL        ok
  X_PROFILE_GRACE_MS=Infinity falls back               FAIL        ok
  X_PROFILE_GRACE_MS=1e999 falls back                  FAIL        ok
  a negative grace waits the DEFAULT too               FAIL        ok
  garbage is indistinguishable from a sane default     FAIL        ok
                                                    17 ok/6 FAIL  23 ok/0
```

Two-sided in the right place: every `…=<raw> terminates` check **passes at both**, because the old code terminated too — it just terminated immediately. Splitting "terminates" from "waits the default" into separate checks is what makes that visible; a single combined assertion would have hidden which half regressed.

## The defect is reachable from configuration, not just from theory

```
skills/x-twitter/x-post-browser.mjs:238   const graceMs = Number(setting('X_PROFILE_GRACE_MS', '10000'));
skills/x-twitter/manifest.json:19         "X_PROFILE_GRACE_MS": "10000"
```

`Number('10s')` is NaN, and `10s` is what someone actually types into a `*_MS` setting. Pre-fix that collapsed to a 0 grace, which makes the first deadline check true and SIGKILLs the holder with no flush — the un-flushed kill this module exists to prevent, restored by a typo in a config value. Picking `10s` and `Infinity` as the cases rather than only `abc` is the right instinct: `abc` is not a realistic input and `Infinity` is non-finite in the other direction.

The test comment replacing `iters <= 2` is the part I would keep longest:

> It passed only because NaN collapsed to a 0 grace and exited on the first check — it pinned the defect's side effect, not termination.

A test that passes *because of* the bug is the hardest kind to notice, and naming it in place stops the next person from restoring the tighter-looking bound.

Explicit `0` surviving as a real choice (case 9) is the right boundary — that is the one input that should kill immediately, and conflating it with garbage would lose it.

## One nit, non-blocking

`10000` now exists in three places: `DEFAULT_GRACE_MS`, the `setting(...)` default at `x-post-browser.mjs:238`, and `manifest.json`. They agree today. If they drift, the symptom is a silent change in kill timing rather than an error, so the module constant is the natural single source for the other two to read.

Checks 18/18 SUCCESS, mergeable CLEAN, two at-head approvals.

PR #4172, opened 2026-09-11
test(health-check): the pool-watchers suite reads fabricated pids, never the host

qingyun-wu:
## The defect

`tests/health-check-pool-watchers-all-tracked.test.py` fabricates watcher pids (100/200/300, 4242, 888/901/902…) and injects a fake `_proc_argv` for them — but it left `_is_watcher_argv` at production, and that function prefers the **OS-authoritative** reader:

```python
def _is_watcher_argv(argv: str, pid: "int | None" = None) -> "bool | None":
    vec = _proc_argv_vector(pid) if pid is not None else None   # reads THIS host
    if vec is not None and len(vec) >= 2:
        ...
```

`_proc_argv_vector` reads `/proc/<pid>/cmdline`. On macOS those fabricated pids don't exist, so it returns `None` and the injected flat argv decides — which is why the suite is green locally. On a Linux CI runner pid 300 is frequently a real process, and a real non-watcher argv beats the injected one.

### Before — the actual CI failure (PR #4162, Coverage Gate run 34578517595 **attempt 1**, job 103196437366, `diff coverage >= 95% (python)`)

```
coverage-gate: running Python suite under instrumentation...
✖ test failed under instrumentation: tests/health-check-pool-watchers-all-tracked.test.py
======================================================================
FAIL: test_every_instance_watcher_is_tracked (__main__.PoolHost.test_every_instance_watcher_is_tracked)
THE case: three sentinels, three live watchers, nothing untracked.
----------------------------------------------------------------------
  File "/tmp/tmp.wq31BZ2xU6/4/tests/health-check-pool-watchers-all-tracked.test.py", line 94, in test_every_instance_watcher_is_tracked
    self.assertEqual(r["status"], "ok", r["detail"])
AssertionError: 'warn' != 'ok'
- warn
+ ok
 : PID reuse: watch-tasks-stream-worker-2.pid (pid 300 is bash src/watch-tasks-stream.sh). …
Ran 90 tests in 0.178s
FAILED (failures=1)
```

The message is itself the proof: the detail prints the **injected** argv (`bash src/watch-tasks-stream.sh`) while the verdict says "not the watcher" — the two disagree precisely because the verdict came from the host, not from the fixture. Attempt 2 of the same run passed with no code change; #4165 hit it earlier the same night. It depends on which pids happen to exist on the runner.

## The fix (test-only; the seam already existed)

`_proc_argv_vector` is already a module-level function that two classes in this same file patch. Nothing in `src/` changes.

- `setUpModule` installs a fabricated `_proc_argv_vector` for the whole module, so no direct `_watcher_trees(...)` / `_is_watcher_argv(argv, pid)` call can reach the host either; `tearDownModule` restores production.
- `run()` now injects the argv reader the way it already injects the ps snapshot: a new `argv_vectors` kwarg (pid -> argv list) saved/restored alongside the other probes. Default = no pid has an authoritative vector, i.e. exactly the macOS behaviour the suite was written against, now stated instead of inherited from the platform.
- `TheDarwinArgvParseIsExercisedOnAnyPlatform` is the unit test **of** the real reader, so it calls the production function captured at import (`_HOST_ARGV_VECTOR`) rather than the module-wide fabrication. Same code object, same assertions.
- One new test consumes the seam: `test_an_OS_vector_that_denies_the_script_is_PID_reuse` states the exact host shape CI kept hitting by accident, so the PID-reuse branch is now covered deliberately and hermetically.

## The control that pins hermeticity

`hermetic_control.py` (not committed — a throwaway harness, quoted below) simulates a Linux runner on which **every** fabricated pid is a real, running, non-watcher process. It deliberately does **not** patch `_proc_argv_vector` (the seam the suite injects) — it patches the OS read that function performs, `Path.read_bytes` on `/proc/<pid>/cmdline`. So it can only change a verdict if the suite still reaches the host. (The KERN_PROCARGS2 class re-patches `Path.read_bytes` inside its own `with`, so it is unaffected either way.)

```python
_NONWATCHER = b"/usr/bin/python3\x00-u\x00/opt/app/unrelated-worker.py\x00"
_real_read_bytes = Path.read_bytes

def _fake_read_bytes(self, *a, **k):
    if re.fullmatch(r"/proc/\d+/cmdline", str(self)):
        return _NONWATCHER
    return _real_read_bytes(self, *a, **k)
...
Path.read_bytes = _fake_read_bytes
res = unittest.main(module=mod, argv=["control"], exit=False, verbosity=1).result
```

**At the parent (origin/main, 5a8b84633)** — the fake host changes 20 verdicts, `test_every_instance_watcher_is_tracked` among them:

```
FAIL: test_every_instance_watcher_is_tracked (PoolHost.test_every_instance_watcher_is_tracked)
FAIL: test_a_clean_pool_is_still_ok (PoolHost.test_a_clean_pool_is_still_ok)
FAIL: test_only_the_untracked_root_is_named (PoolHost.test_only_the_untracked_root_is_named)
FAIL: test_ok (SingleInstanceUnchanged.test_ok)
FAIL: test_a_production_shaped_duplicate_is_counted_as_a_second_tree (TheWatcherPredicateIsAShapeNotAFieldCount...)
 … 15 more …
Ran 90 tests in 0.253s
FAILED (failures=20)
CONTROL: ran=90 failures=20 errors=0
```

**At HEAD** — the same control changes nothing, because nothing reads the host any more:

```
Ran 91 tests in 0.131s
OK
CONTROL: ran=91 failures=0 errors=0
```

And the plain suite at HEAD:

```
Ran 91 tests in 0.182s
OK
```

## Neighbours + gate

Every test file that imports these helpers (`grep -ln '_is_watcher_argv\|_proc_argv_vector\|_watcher_trees\|check_task_watcher\|watcher_sentinel_path' tests/*.py`) run at HEAD:

```
tests/agent-availability.test.py                        Ran 35 tests   OK
tests/health-check-codex-task-notifier.test.py          Ran 18 tests   OK
tests/health-check-probe-names-its-subject.test.py      Ran 11 tests   OK
tests/health-check-supervised-watcher-not-orphan.test.py   PASS — supervised watcher is not reported as an orphan
tests/health-check-task-watcher.test.py                    Task-watcher liveness invariants hold.
tests/watcher-sentinel-naming.test.py                   Ran 30 tests   OK
```

```
$ git diff origin/main > /tmp/h.diff && bash scripts/review-checks.sh --diff /tmp/h.diff
```

## Not in this PR (verified, worth a follow-up)

`tests/health-check-task-watcher.test.py` has the same exposure — its `run_check`/`supervised_watcher` fixtures patch `_proc_argv` and `_watcher_trees` but not `_proc_argv_vector`, and their fabricated pids are 4242 and 7100. Under the same fake host it reports 22 failures; 9 of those (cases e/j/k/n/s/u/w) are genuine host-dependence, the other 3 rows (`aa)`) are a control artifact from the one case that legitimately spawns real processes. Left out to keep this PR to a single concern; happy to file it or fix it in a follow-up if you'd rather it land together.

### DIFF
--- a/tests/health-check-pool-watchers-all-tracked.test.py
+++ b/tests/health-check-pool-watchers-all-tracked.test.py
@@ -31,14 +31,34 @@ except SystemExit:
 
 WATCHER_ARGV = "bash src/watch-tasks-stream.sh"
 
+# Every pid below is fabricated, so the argv reader must be fabricated too: left
+# at production `_is_watcher_argv` reads THIS host, where pid 300 may be anything.
+_HOST_ARGV_VECTOR = hc._proc_argv_vector
+
+
+def _fabricated_argv_vector(vectors=None):
+    """`_proc_argv_vector` over a pid -> argv-list map; any other pid is unreadable,
+    which is what the injected flat `argv` is then decided from."""
+    table = {str(pid): vec for pid, vec in (vectors or {}).items()}
+    return lambda pid, _t=table: _t.get(str(pid))
+
+
+def setUpModule():
+    hc._proc_argv_vector = _fabricated_argv_vector()
+
+
+def tearDownModule():
+    hc._proc_argv_vector = _HOST_ARGV_VECTOR
+
 
 def run(sentinels: dict, trees: dict, argv=WATCHER_ARGV, core_alive=True,
         parent="1", pid_instance=None, pid_actor="", targets=None, verdicts=None,
-        ps="") -> dict:
+        ps="", argv_vectors=None) -> dict:
     """`sentinels` maps filename -> contents; `trees` maps root pid -> members.
 
     `pid_instance` is what the WATCHER's own environment yields: a string names
-    its instance, "" is the default, and None means unreadable.
+    its instance, "" is the default, and None means unreadable. `argv_vectors`
+    maps pid -> argv list, the OS-authoritative half of the same fabrication.
     """
     with tempfile.TemporaryDirectory() as td:
         ws = Path(td)
@@ -50,10 +70,11 @@ def run(sentinels: dict, trees: dict, argv=WATCHER_ARGV, core_alive=True,
         saved = (hc.WORKSPACE_DIR, hc._proc_argv, hc._watcher_trees,
                  hc._ps_snapshot, hc._pid_parent, hc._fresh_local_core_record,
                  hc._pid_instance_id, hc._pid_actor_id, hc._watcher_sentinel_target,
-                 hc._is_watcher_argv)
+                 hc._is_watcher_argv, hc._proc_argv_vector)
         try:
             hc.WORKSPACE_DIR = ws
             hc._proc_argv = (argv if callable(argv) else (lambda pid: argv))
+            hc._proc_argv_vector = _fabricated_argv_vector(argv_vectors)
             if verdicts is not None:
                 # None is the tri-state "cannot prove", which no argv string
                 # reliably produces; force it so that branch is reachable.
@@ -77,7 +98,8 @@ def run(sentinels: dict, trees: dict, argv=WATCHER_ARGV, core_alive=True,
             (hc.WORKSPACE_DIR, hc._proc_argv, hc._watcher_trees,
              hc._ps_snapshot, hc._pid_parent, hc._fresh_local_core_record,
              hc._pid_instance_id, hc._pid_actor_id,
-             hc._watcher_sentinel_target, hc._is_watcher_argv) = saved
+             hc._watcher_sentinel_target, hc._is_watcher_argv,
+             hc._proc_argv_vector) = saved
 
 
 class PoolHost(unittest.TestCase):
@@ -94,6 +116,15 @@ class PoolHost(unittest.TestCase):
         self.assertEqual(r["status"], "ok", r["detail"])
         for pid in ("100", "200", "300"):
             self.assertIn(pid, r["detail"])
+
+    def test_an_OS_vector_that_denies_the_script_is_PID_reuse(self):
+        """The same host shape CI kept hitting by accident, now stated: the pid in
+        the sentinel is authoritatively some other program, so it is not a watcher."""
+        r = run({"watch-tasks-stream-worker-2.pid": "300\n"}, {},
+                argv_vectors={300: ["/usr/bin/python3", "-u", "/opt/app/worker.py"]})
+        self.assertEqual(r["status"], "warn", r["detail"])
+        self.assertIn("300", r["detail"])
+        self.assertIn("PID reuse", r["detail"])
 
     def test_a_genuinely_untracked_watcher_is_still_reported(self):
         """The union must not swallow the defect the probe exists to find."""
@@ -521,7 +552,8 @@ class TheDarwinArgvParseIsExercisedOnAnyPlatform(unittest.TestCase):
                 + b"\0".join(argv) + b"\0")
         with patch.object(Path, "read_bytes", side_effect=OSError("not linux")), \
              patch.object(ctypes, "CDLL", return_value=self._fake_libc(blob)):
-            return hc._proc_argv_vector(4242)
+            # The real reader is the subject here, not the module-wide fabrication.
+            return _HOST_ARGV_VECTOR(4242)
 
     def test_the_exec_path_is_skipped_and_argv_returned(self):
         self.assertEqual(
@@ -556,7 +588,7 @@ class TheDarwinArgvParseIsExercisedOnAnyPlatform(unittest.TestCase):
 
         with patch.object(Path, "read_bytes", side_effect=OSError("not linux")), \
              patch.object(ctypes, "CDLL", return_value=_Fail()):
-            self.assertIsNone(hc._proc_argv_vector(4242))
+            self.assertIsNone(_HOST_ARGV_VECTOR(4242))
 
 
 class TheWatcherPredicateIsAShapeNotAFieldCount(unittest.TestCase):


qingyun-wu (commented):
Reviewed at `ca6fb2a7a93f5eb0da7e67ef0f470f1d7ec3d542`.

LGTM, non-blocking. The fabricated PID fixtures now isolate both argv readers, while the Darwin parser tests still call the production reader.

Independent control against current origin/main and this head: intercept only OS `/proc/<pid>/cmdline` reads with an unrelated program. Main: 90 tests, 20 failures. Head: 91 tests, zero failures and zero intercepted host reads. Unmodified head: 91 pass; sentinel naming 30 and availability 35 pass. The explicit authoritative-vector PID-reuse case still warns. Full diff and relevant health-check/parser code inspected; added-path/root/prose checks pass and a synthetic forbidden production path is rejected.

Worst case is masking a fixture's intended OS interaction; the explicit saved-reader calls preserve the real-parser coverage. This changes tests only. Formal approval is unavailable because this login is the author, so this is a COMMENTED approval verdict; current CI/CLA and maintainer merge gates still apply.

Current-head gate snapshot: hosted checks completed without failures. CLA: success. No merge performed.

john-the-dev (approved):
## The hole is real, and reproducible on `origin/main` in three lines

`_is_watcher_argv` consults the OS reader **first** and never reaches the injected argv when it answers:

```python
vec = _proc_argv_vector(pid) if pid is not None else None
if vec is not None and len(vec) >= 2:
    ...                      # decided here; `argv` below is dead code
parts = argv.split()         # the fixtures' injection
```

Same injected argv every time, only the pid differing, run against `origin/main`:

```
pid=999999   no such pid here          vector=None                          verdict=True
pid=1        launchd (exists)          vector=None                          verdict=True
pid=63365    this python (exists)      vector=['/opt/homebrew/Cellar/...']  verdict=False   <-- host decided
pid=None     no OS read                                                     verdict=True
```

The third row is a fixture silently measuring the host.

**And the middle row is why nobody noticed.** `pid=1` exists and still yields `None`, because macOS will not let you read another user's argv — so on a Mac the hole only opens for a *same-user* pid collision, which fabricated numbers rarely hit. Where the vector is readable it is wide open. That makes this quiet locally and live in CI, which is the worst shape for a hermeticity bug: the suite that would catch the regression is the one that has been lying.

## On this PR

**APPROVE at `ca6fb2a7a`.** Same fix as sibling #4173, and this file does the harder version of it.

The module-wide swap closes the hole, but `_proc_argv_vector` is itself a subject in this suite — the KERN_PROCARGS2 parser, tested through a fabricated libc blob. Fabricating the reader module-wide would have turned those into tests of the stub. Routing them through the captured original keeps both properties at once:

```python
 36:  _HOST_ARGV_VECTOR = hc._proc_argv_vector      # captured before the swap
 51:  hc._proc_argv_vector = _HOST_ARGV_VECTOR      # tearDownModule restores
556:      return _HOST_ARGV_VECTOR(4242)            # the REAL parser is the subject
591:      self.assertIsNone(_HOST_ARGV_VECTOR(4242))
```

with the comment saying so in place — *"The real reader is the subject here, not the module-wide fabrication."* That is the distinction a later reader would otherwise get wrong, and getting it wrong would be invisible: the tests would still pass.

The per-case `argv_vectors` parameter is the right complement to the module-wide default. A blanket stub returning `None` for everything would make the vector branch unreachable, so the OS-authoritative path would go untested from the *other* direction — traded one kind of blindness for another. Allowing a case to name `pid -> argv` keeps that branch reachable deliberately.

Suite green at head (`rc=0`, OK).

I like that `_fabricated_argv_vector` returns `None` for any pid not in the map. It means a fixture gets the vector path only when it asks for it, and every other case falls through to the injected flat `argv` — which is what the existing cases were written against and what they should keep testing.

keweichen (approved):
APPROVE at `ca6fb2a7a93f5eb0da7e67ef0f470f1d7ec3d542` — this implementation is ready to merge once the normal current-head CI/CLA and two-maintainer gate are satisfied.

The defect exists on current `origin/main`: its pool-watcher test blob is identical to merge-base `5a8b8463342d360db6dbd4a1547f43dc60cd0680`, and the newer main changes do not touch the watcher predicate. At `src/health-check.py:8731`, `_is_watcher_argv` consults `_proc_argv_vector(pid)` before the fixture-injected flat argv.

Failure-mode control, intercepting only `/proc/<pid>/cmdline` with an unrelated argv:

```
5a8b8463  ran=90  failures=20  errors=0  intercepted=34
ca6fb2a7  ran=91  failures=0   errors=0  intercepted=0
```

The exact `PoolHost.test_every_instance_watcher_is_tracked` case separately changed from one failure with three intercepted reads at the parent to passing with zero at HEAD. A mutation that made `argv_vectors` ineffective caused the new test to fail (`'ok' != 'warn'`), confirming it was collected and discriminates the fix.

`tests/health-check-pool-watchers-all-tracked.test.py:36-51` preserves and restores the production reader; `:70-102` saves/restores the per-case seam; and `:556` plus `:591` deliberately retain production Darwin-reader coverage. Adjacent injected inputs behaved as intended: absent and short vectors fell back, watcher vectors passed, non-watcher and shell-option vectors warned, string/int PID keys agreed, and teardown restored the production reader.

Focused verification:

```
pool-watchers suite:  Ran 91 tests — OK
agent-availability:   Ran 35 tests — OK
codex notifier:       Ran 18 tests — OK
probe subject:        Ran 11 tests — OK
sentinel naming:      Ran 30 tests — OK
supervised/task-watcher standalone suites — PASS
git diff --check: rc=0
review-checks: PASS
```

The standalone suites reported their environment-specific live-`ps` skips; those skips are not part of the failure-mode proof above. The review-check scanner’s negative result was controlled with a synthetic `/Users/...` addition, which it rejected with rc=1.

Manual added-line scanning found only `/usr/bin/python3` and `/opt/app/worker.py` at `tests/health-check-pool-watchers-all-tracked.test.py:124`; both are inert fabricated argv data, not invoked host paths. No production file changes.

Worst case is a suite-wide stub masking the real argv reader or leaking into another case, producing a false green. Capturing `_HOST_ARGV_VECTOR`, routing the Darwin parser tests through it, restoring both per case and per module, and the zero-interception/restoration controls mitigate that risk.

Non-blocking: the disclosed pre-existing sibling exposure in `tests/health-check-task-watcher.test.py` also reproduced under the adversarial host (22 failures) and has a separate local topic branch; it is not introduced or worsened by this one-file change. The supplied status snapshot is green, but this offline review did not independently query hosted CI.

PR #4173, opened 2026-09-11
test(health-check): the task-watcher suite reads fabricated pids, never the host

qingyun-wu:
## The defect

`tests/health-check-task-watcher.test.py` fabricates the watcher pids 4242 and 7100. Its `run_check` and `supervised_watcher` fixtures inject `_proc_argv`, `_watcher_trees`, `_ps_snapshot`, `_pid_parent` and the identity probes — but not `_proc_argv_vector`, and `_is_watcher_argv` prefers that **OS-authoritative** reader over the injected flat argv:

```python
def _is_watcher_argv(argv: str, pid: "int | None" = None) -> "bool | None":
    vec = _proc_argv_vector(pid) if pid is not None else None   # reads THIS host
    if vec is not None and len(vec) >= 2:
        ...
```

`_proc_argv_vector` reads `/proc/<pid>/cmdline` (Linux) or KERN_PROCARGS2 (darwin). On macOS pids 4242/7100 usually don't exist, it returns `None`, and the injected argv decides — which is why the suite is green locally. On a CI runner where those pids are real processes, a real non-watcher argv beats the fixture and the verdict comes from the host. This is the exposure named as a follow-up in #4172 (same seam, sibling suite); no `src/` change here either.

## The fix (test-only; the seam already existed)

`_proc_argv_vector` is a module-level function, so it is patchable exactly like the probes the fixtures already inject.

- A module-wide fabricated `_proc_argv_vector` is installed at import, reading a `_ARGV_VECTORS` map (pid → argv list). A pid absent from the map has **no** authoritative vector — precisely the macOS behaviour this suite was written against, now stated instead of inherited from the platform.
- `run_check` gains an `argv_vectors` kwarg, saved/restored alongside the other probes, so a case can state the OS's answer for its own pids.
- `case_aa_argv_classification_binds_to_the_executed_script` is the test **of** the real reader — it spawns real processes — so it restores the production function (`_HOST_ARGV_VECTOR`, captured before the fabrication) for its own duration and puts it back in its existing `finally`.
- One new case consumes the seam: `d2` — the flat argv names the watcher, the authoritative vector denies it, the OS wins. That is the exact host shape the fixtures used to hit by accident; it is now covered deliberately and hermetically on every platform.

## The control that pins hermeticity

`hermetic_control2.py` (not committed — a throwaway harness, quoted in full below) simulates a runner on which **every** fabricated pid is a real, running, non-watcher process. It deliberately does **not** patch `_proc_argv_vector` (the seam the suite injects) — it patches the OS read that function performs, `Path.read_bytes` on `/proc/<pid>/cmdline`. So it can only change a verdict if the suite still reaches the host.

```python
_NONWATCHER = b"/usr/bin/python3\x00-u\x00/opt/app/unrelated-worker.py\x00"
_real_read_bytes = Path.read_bytes


def _fake_read_bytes(self, *a, **k):
    if re.fullmatch(r"/proc/\d+/cmdline", str(self)):
        return _NONWATCHER
    return _real_read_bytes(self, *a, **k)


spec = importlib.util.spec_from_file_location("tw_test", TEST)
mod = importlib.util.module_from_spec(spec)
spec.loader.exec_module(mod)

Path.read_bytes = _fake_read_bytes
try:
    rc = mod.main()
finally:
    Path.read_bytes = _real_read_bytes
print(f"CONTROL: rc={rc}")
```

### Before — at the parent (origin/main, `719e1e044`)

The fake host changes **22 verdicts across 14 cases**:

```
$ python3 hermetic_control2.py tests/health-check-task-watcher.test.py | grep '✗'
  ✗ case e
  ✗ case j
  ✗ case k
  ✗ case n
  ✗ case n2
  ✗ case s
  ✗ case t3
  ✗ case u
  ✗ case u2
  ✗ case u3
  ✗ case u4
  ✗ case w
  ✗ case w2
  ✗ case aa

22 failure(s)
CONTROL: rc=1
```

Representative rows — the detail is itself the proof, printing the **injected** argv while the verdict says "not the watcher", because the two came from different sources:

```
  ✗ case s
      s) --fix must write the live watcher's pid, got <ABSENT>
      s) re-run check should be ok, got {'name': 'task-watcher', 'status': 'warn', '_sentinel_restamp_pid': '7100', ...}
  ✗ case u
      u) should say why it declined, got 'pid 7100 is no longer the watcher — not re-stamped'
  ✗ case n
      n) expected 2 trees, got 0: {}
```

### After — at HEAD (`f0d7fe12e`)

The same control changes nothing except case `aa`, which reads the host **by design**:

```
$ python3 hermetic_control2.py tests/health-check-task-watcher.test.py | tail -8
  ✓ case z
  ✗ case aa
      aa) dir ending .sh, real watcher: got False, want True
      aa) path WITH SPACES + tasks: got False, want True
      aa) plain watcher, no args: got False, want True

3 failure(s)
CONTROL: rc=1
```

Those three rows are a **control artifact, not a defect**: `aa` spawns real processes and asserts on the real reader, so a harness that lies about `/proc/<pid>/cmdline` is lying to the subject under test. They are the same three rows #4172's analysis predicted for this file. Every other case is now indifferent to the host.

And the plain suite at HEAD — 38 cases, `d2` new:

```
$ python3 tests/health-check-task-watcher.test.py | tail -6
  ✓ case x
  ✓ case y
  ✓ case y2
  ✓ case z
  ✓ case aa
  ✓ case ab

Task-watcher liveness invariants hold.
```

## Neighbours + gate

Every test file importing these helpers (`grep -ln '_is_watcher_argv\|_proc_argv_vector\|_watcher_trees\|check_task_watcher\|watcher_sentinel_path' tests/*.py`), run at HEAD:

```
tests/agent-availability.test.py                           Ran 35 tests in 0.174s  OK   [rc=0]
tests/health-check-codex-task-notifier.test.py             Ran 18 tests in 0.312s  OK   [rc=0]
tests/health-check-probe-names-its-subject.test.py         Ran 11 tests in 0.077s  OK   [rc=0]
tests/health-check-pool-watchers-all-tracked.test.py       Ran 90 tests in 0.089s  OK   [rc=0]
tests/health-check-supervised-watcher-not-orphan.test.py   PASS — supervised watcher is not reported as an orphan   [rc=0]
tests/watcher-sentinel-naming.test.py                      Ran 30 tests in 0.407s  OK   [rc=0]
```

```
$ git diff origin/main > /tmp/h2.diff && bash scripts/review-checks.sh --diff /tmp/h2.diff
```

## Relationship to #4172

Sibling, not a stack: #4172 does this for `tests/health-check-pool-watchers-all-tracked.test.py`, and its "Not in this PR" section names this file as the remaining exposure. The two touch disjoint files and can land in either order. This PR is branched from `origin/main` with nothing cherry-picked from #4172 — the seam (`_proc_argv_vector` being module-level) already exists on main.

### DIFF
--- a/tests/health-check-task-watcher.test.py
+++ b/tests/health-check-task-watcher.test.py
@@ -15,6 +15,7 @@ Covers:
   b) core alive, sentinel absent → warn
   c) core alive, sentinel holds a dead PID → warn (crashed, sentinel left behind)
   d) core alive, PID alive but argv is not the watcher → warn (PID reuse)
+  d2) ...including when only the OS-authoritative argv vector denies it
   e) core alive, PID alive and argv names the watcher → ok
   f) core alive, sentinel unparseable → warn (not a crash)
   g) the check is registered in run_checks' output
@@ -38,6 +39,11 @@ Covers:
   x) `--fix` actually REACHES the repair (warn never enters `issues`)
   y) under `--json` the repair line stays off stdout, so JSON still parses
   y2) ...and the repair pass's own `_`-prefixed keys stay out of the payload
+
+Hermeticity: every pid below is fabricated, so `_proc_argv_vector` — the
+OS-authoritative reader `_is_watcher_argv` prefers over the injected argv — is
+fabricated module-wide too. Case (aa) spawns real processes and restores the
+real reader for its own duration.
 
 Run: python3 tests/health-check-task-watcher.test.py
 Exit code: 0 on pass, 1 on fail.
@@ -68,6 +74,30 @@ spec.loader.exec_module(hc)
 # Captured before any case swaps it in, so a failing case cannot leak a stub.
 _REAL_PROC_ARGV = hc._proc_argv
 
+# Captured before the module-wide fabrication below: case (aa) spawns real
+# processes, so the real reader is its subject and must stay reachable.
+_HOST_ARGV_VECTOR = hc._proc_argv_vector
+
+_ARGV_VECTORS: "dict[int, list[str]]" = {}
+
+
+def _fabricated_argv_vector(pid: int) -> "list[str] | None":
+    """The OS-authoritative argv reader `_is_watcher_argv` prefers, fabricated
+    for the whole module.
+
+    Every pid these fixtures name is invented, so on a host where one really
+    exists the real reader decides the verdict and the injected argv is ignored.
+    A pid absent from the map has no authoritative vector -- the macOS
+    behaviour this suite was written against, now stated rather than inherited.
+    """
+    try:
+        return _ARGV_VECTORS.get(int(pid))
+    except (TypeError, ValueError):
+        return None
+
+
+hc._proc_argv_vector = _fabricated_argv_vector
+
 
 def make_workspace(td: Path, *, core_alive: bool, pid_text: str | None) -> Path:
     """Build a temp workspace. `core_alive` stamps a fresh heartbeat file;
@@ -90,6 +120,7 @@ def make_workspace(td: Path, *, core_alive: bool, pid_text: str | None) -> Path:
 def run_check(*, core_alive: bool, pid_text: str | None, argv: str | None = None,
               pid_instance: "str | None" = "",
               pid_actor: "str | None" = "",
+              argv_vectors: "dict | None" = None,
               trees: dict | None = None, parents: dict | None = None) -> dict:
     """Call check_task_watcher against a temp WORKSPACE_DIR. `argv` patches
     the _proc_argv probe: None = leave the real one (only used where no PID
@@ -99,13 +130,20 @@ def run_check(*, core_alive: bool, pid_text: str | None, argv: str | None = None
     it has unknown parentage. The parent probes are stubbed unconditionally —
     `trees` invents pids, and an unstubbed probe reads the HOST's process table,
     where a fabricated pid may really exist and carry a real parent.
+    
+    `argv_vectors` maps a fabricated pid to the OS-authoritative argv vector the
+    module-wide fabrication reports for it; a pid absent from it has none, so
+    the injected flat `argv` decides.
     """
     with tempfile.TemporaryDirectory() as td:
         make_workspace(Path(td), core_alive=core_alive, pid_text=pid_text)
         saved = (hc.WORKSPACE_DIR, hc._proc_argv, hc._watcher_trees,
                  hc._ps_snapshot, hc._pid_parent, hc._pid_instance_id,
                  hc._pid_actor_id)
+        saved_vectors = dict(_ARGV_VECTORS)
         try:
+            _ARGV_VECTORS.clear()
+            _ARGV_VECTORS.update({int(p): v for p, v in (argv_vectors or {}).items()})
             hc.WORKSPACE_DIR = Path(td)
             if argv is not None:
                 hc._proc_argv = lambda pid: argv
@@ -118,6 +156,8 @@ def run_check(*, core_alive: bool, pid_text: str | None, argv: str | None = None
             hc._pid_actor_id = lambda pid: pid_actor
             return hc.check_task_watcher()
         finally:
+            _ARGV_VECTORS.clear()
+            _ARGV_VECTORS.update(saved_vectors)
             (hc.WORKSPACE_DIR, hc._proc_argv, hc._watcher_trees,
              hc._ps_snapshot, hc._pid_parent, hc._pid_instance_id,
              hc._pid_actor_id) = saved
@@ -547,6 +587,22 @@ def case_d_pid_reuse_warns() -> list[str]:
     return []
 
 
+def case_d2_an_OS_vector_that_denies_the_script_is_PID_reuse() -> list[str]:
+    """The flattened argv names the watcher and the authoritative vector denies
+    it; the OS wins. This is the exact host shape the fixtures used to inherit
+    by accident — now injected, so the branch is covered on every platform."""
+    r = run_check(core_alive=True, pid_text="4242",
+                  argv="bash src/watch-tasks-stream.sh",
+                  argv_vectors={4242: ["/usr/bin/python3",
+                                       "/opt/app/unrelated-worker.py"]})
+    if r["status"] != "warn":
+        return [f"d2) an OS vector denying the script should warn, got "
+                f"{r['status']} ({r['detail']})"]
+    if "reuse" not in r["detail"]:
+        return [f"d2) detail should name PID reuse, got {r['detail']!r}"]
+    return []
+
+
 def case_e_live_watcher_is_ok() -> list[str]:
     r = run_check(core_alive=True, pid_text="4242", argv="bash src/watch-tasks-stream.sh")
     if r["status"] != "ok":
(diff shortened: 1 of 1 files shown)


qingyun-wu (commented):
Reviewed at `f0d7fe12e42eac2e5d1505c6ddc9556582ffa914`.

LGTM, non-blocking. Independently reproduced the fixture leak against current origin/main by making only fabricated PIDs 4242 and 7100 return an unrelated OS argv vector. The real child-process case (aa) was left intact.

```
main: 16 failed assertions; 15 fabricated-PID OS reads
head: all 38 named cases pass; 0 fabricated-PID OS reads
```
The unmodified full suite also passes, including the real-process classification case. The new d2 case checks that an explicitly injected contradictory authoritative vector wins over the flat watcher-looking argv. Full diff and affected helper/caller code inspected; child/cumulative path/root/prose scans pass, with a rejected forbidden-path control.

Worst case is losing real-reader coverage through over-stubbing; the saved production reader is restored specifically for (aa). No production behavior changes. Self-author COMMENTED approval verdict; current CI/CLA and maintainer gates remain required.

Current-head gate snapshot: hosted checks still running. CLA: not yet posted. No merge performed.

john-the-dev (approved):
## The hole is real, and reproducible on `origin/main` in three lines

`_is_watcher_argv` consults the OS reader **first** and never reaches the injected argv when it answers:

```python
vec = _proc_argv_vector(pid) if pid is not None else None
if vec is not None and len(vec) >= 2:
    ...                      # decided here; `argv` below is dead code
parts = argv.split()         # the fixtures' injection
```

Same injected argv every time, only the pid differing, run against `origin/main`:

```
pid=999999   no such pid here          vector=None                          verdict=True
pid=1        launchd (exists)          vector=None                          verdict=True
pid=63365    this python (exists)      vector=['/opt/homebrew/Cellar/...']  verdict=False   <-- host decided
pid=None     no OS read                                                     verdict=True
```

The third row is a fixture silently measuring the host.

**And the middle row is why nobody noticed.** `pid=1` exists and still yields `None`, because macOS will not let you read another user's argv — so on a Mac the hole only opens for a *same-user* pid collision, which fabricated numbers rarely hit. Where the vector is readable it is wide open. That makes this quiet locally and live in CI, which is the worst shape for a hermeticity bug: the suite that would catch the regression is the one that has been lying.

## On this PR

**APPROVE at `f0d7fe12e`.** The module-wide fabrication is the right scope — every pid these fixtures name is invented, so leaving any of them on the real reader leaves the hole open — and the carve-out is the part I checked hardest:

```
 79:  _HOST_ARGV_VECTOR = hc._proc_argv_vector     # captured BEFORE the swap
843:  hc._proc_argv_vector = _HOST_ARGV_VECTOR     # case (aa) restores it
```

Case (aa) spawns real processes, so the real reader is its *subject*; fabricating it there would have turned a live test into a tautology. Capturing before the module-level swap rather than inside the case is what makes that work — the same ordering the file already used for `_REAL_PROC_ARGV`, with the same comment explaining why.

Suite green at head (`rc=0`, "Task-watcher liveness invariants hold").

The docstring line worth keeping:

> A pid absent from the map has no authoritative vector — the macOS behaviour this suite was written against, now stated rather than inherited.

That is exactly the right description of what changed. The fixtures always depended on this; they just depended on it silently.

## One suggestion, non-blocking

Sibling #4172 goes one step further: it keeps a test class whose subject *is* the real KERN_PROCARGS2 parser — exec-path skip, padding NULs, `argc` bounding so envp is not read as argv, failing sysctl → `None` — driven through `_HOST_ARGV_VECTOR` with a fabricated libc blob. The result is that fabricating the reader module-wide does not cost coverage of the reader itself.

This suite has no equivalent. Not a gap this PR introduces, and not a reason to hold it — but if the real reader ever regresses, #4172's file is the only one that would say so.

keweichen (changes requested):
REQUEST_CHANGES at exact head `f0d7fe12e42eac2e5d1505c6ddc9556582ffa914`.

### Blocker

`tests/health-check-task-watcher.test.py:133` adds four spaces on an otherwise blank line, so the topic diff fails the required clean-diff check:

```text
$ git diff --check 719e1e04463717b392aab18068c9d325283f768d..f0d7fe12e42eac2e5d1505c6ddc9556582ffa914
tests/health-check-task-watcher.test.py:133: trailing whitespace.
+    
[exit 2]
```

Delete those spaces and rerun `git diff --check`; that is the only change needed to unblock this review.

### Verification

The functional fix checks out. `719e1e04463717b392aab18068c9d325283f768d` is both the merge-base and direct parent, so this is one child commit, not a stack. Current `origin/main` retains the parent test blob and has no changes in the implicated argv/watcher path.

An independent control made only fabricated PIDs 4242 and 7100 return an unrelated `/proc` argv vector:

```text
parent: rc=1; collected=37; failed_cases=11; fabricated_pid_OS_reads=15
head:   rc=0; collected=38; failed_cases=0;  fabricated_pid_OS_reads=0
```

That reproduces the host leak on the parent and confirms both the fix and collection of new case `d2`. The focused suite passed all 38 named cases, and the current-`origin/main` plus PR test image also passed 38/38. The six implicated neighboring suites passed (90, 11, 35, 18, and 30 unittest cases plus the supervised-watcher standalone suite). `case h` and two neighboring live-`ps` probes explicitly skipped because `ps` is unavailable here; real-vector case `aa` did execute against spawned processes and passed.

Repository review checks passed, with a forbidden `/Users/…` control correctly producing one violation. The only new absolute paths are `/usr/bin/python3` and `/opt/app/unrelated-worker.py` as inert classifier fixture data; neither is resolved or executed.

There is no direct runtime user disruption because this is test-only. The worst indirect risk is over-stubbing away production argv-reader coverage and allowing a future regression through; `case_aa` mitigates that by restoring the captured production reader at lines 842–873, which the run above exercised. Hosted CI, CLA, approval freshness, and the two-maintainer merge gate were not independently queried from this isolated checkout and remain merge-time conditions.

qingyun-wu:
@keweichen — done exactly as specified, `f0d7fe12e` → `77d7443cc`. The four spaces on line 133 are gone and nothing else changed: the diff against the head you reviewed is `1 file changed, 1 insertion(+), 1 deletion(-)`.

Your own acceptance command now passes:

```
$ git diff --check origin/main...HEAD
$ echo $?
0
```

Suite still green (rc 0), and both repo gates clean.

Worth noting for the merge math: this repo does not dismiss stale reviews on push, so @john-the-dev's APPROVE from `f0d7fe12e` survives the change — though by provenance it now predates the head by one whitespace-only commit. Your CHANGES_REQUESTED is the only thing still gating, and it is yours to clear.

john-the-dev (approved):
Re-approving to anchor my tick at `77d7443c`. My approval was at 2026-09-11T09:54Z and one commit landed after it, so it was sitting on a head I had not read.

Verified the delta rather than trusting the message: `77d7443c` is `+1/-1` in `tests/health-check-task-watcher.test.py`, removing trailing whitespace from a blank line inside a docstring. No behavior, no assertion, no production file.

This does not touch @keweichen's standing CHANGES_REQUESTED — that block is unaffected and still theirs to clear. I am only making sure my own approval names a head I have actually read.

PR #4180, opened 2026-09-11
feat(skills): a session-scoped read receipt for every skill, not just pr-triage

sonichi:
`pr-triage` proved the shape on one procedure file. This lifts it to `scripts/` so any skill can use it, and adds `--audit`.

## Why this exists

A skill's text reaches an agent by **injection from the Claude Code harness** — not this repo's code, and not something we can make stamp what it served. This works on our side of that boundary: it records what the agent **read from disk**, so a skill with no receipt is one whose injected copy has never been checked against the file.

## Why a content hash and not an mtime

Measured 2026-09-11, **both directions in one session**:

```
collaboration-intelligence   injected body was 3 days BEHIND disk
                             (#4029 merged Sep 7; refresh-skill.sh --all had run 20 min earlier)
proactive-loop step 6.5      served body contained text present in NO FILE on this machine
                             (searched skills/, .claude-sutando/, ~/.claude — 1 hit, the transcript;
                              origin/main: 0, and the checkout was 0 behind)
```

A hash has no opinion about which side is newer, so it catches both. An mtime comparison is wrong in one direction and silent in the other — which is why I did not ship the mtime probe I first proposed.

`--marker` no longer defaults to pr-triage's `HARNESS-GREEN IS NOT A WITNESS`; a generic tool keeping that default would refuse every other skill.

## Evidence

Each control drives the script into the state it is meant to catch, rather than only exercising the clean path.

```
14 tests                                              OK
BREAK audit always returns 0 (silent detector)   ->   FAILED (failures=2)
BREAK drop the session scoping (reusable receipt)->   FAILED (failures=4)
RESTORED                                              OK

review-checks.sh --diff   PASS (hardcoded-paths + root-artifacts + prose-cap clean)
```

Replayed against a real skill on this host, not a fixture:

```
--check on collaboration-intelligence/SKILL.md, no prior read  ->  rc=1
  "FULL READ REQUIRED: no receipt for this session"
--record, then --audit                                         ->  1 receipt, 0 drifted
```

## What it does not do

It cannot see the injected text — only what was read from disk. A skill that is never read still has no receipt, which is the correct answer (unverified), not a false clear.

## Follow-up, named rather than bundled

`sutando-skills/skills/pr-triage/scripts/skill-read-receipt.py` is now a second copy of this predicate in another repo. Per CLAUDE.md that duplication is a defect in its own right; it needs a sutando-skills PR to delegate here. Not bundled into this one.

Stand: Echo Act IV Pro

### DIFF
--- a/scripts/skill-read-receipt.py
+++ b/scripts/skill-read-receipt.py
@@ -31,6 +31,8 @@ catches neither reliably, which is why this is a hash and not a timestamp.
             refresh-skill.sh) after it was read. rc 1 if any drifted. No --skill.
 """
 import argparse
+import contextlib
+import fcntl
 import hashlib
 import json
 import os
@@ -48,12 +50,28 @@ def _session_id() -> str:
     return (os.environ.get("CLAUDE_CODE_SESSION_ID") or "").strip()
 
 
-def _load(state_dir: Path) -> dict:
+def _load(state_dir: Path) -> "tuple[dict, str]":
+    """(receipts, error). A non-empty error means the store is UNREADABLE, which is
+    not the same fact as an empty one.
+
+    Collapsing corruption, a wrong top-level type and every I/O error into {} makes
+    --audit print "nothing read in full yet" over a store it could not parse -- a
+    broken probe reporting clean, which is the failure this whole file exists to
+    prevent. Callers decide the severity: --check stays conservative (demand the
+    read), --audit refuses to answer.
+    """
+    path = state_dir / RECEIPTS
+    if not path.exists():
+        return {}, ""                      # genuinely empty: no receipts yet
     try:
-        data = json.loads((state_dir / RECEIPTS).read_text())
-    except (OSError, ValueError):
-        return {}
-    return data if isinstance(data, dict) else {}
+        data = json.loads(path.read_text())
+    except OSError as e:
+        return {}, f"{path} is unreadable: {e}"
+    except ValueError as e:
+        return {}, f"{path} is not valid JSON: {e}"
+    if not isinstance(data, dict):
+        return {}, f"{path} holds {type(data).__name__}, not an object"
+    return data, ""
 
 
 def _key(session: str, path: Path) -> str:
@@ -74,6 +92,28 @@ def _preconditions(skill: Path, marker: str):
     return True, ""
 
 
+@contextlib.contextmanager
+def _exclusive(state_dir: Path):
+    """Hold the whole read-modify-write, not just the file swap.
+
+    Two record() calls could each load the same snapshot and each write the WHOLE
+    store back, so one receipt is silently erased while both callers see rc=0 --
+    and a later --audit then reports clean for a skill nobody is tracking. A lock
+    around the swap alone does not help; the critical section is load->mutate->write.
+    Not `workspace_lock.py`: that is a heartbeated singleton ROLE lock for long-lived
+    pollers, a different primitive from a short mutual-exclusion window.
+    """
+    state_dir.mkdir(parents=True, exist_ok=True)
+    fd = os.open(state_dir / (RECEIPTS + ".lock"), os.O_CREAT | os.O_RDWR, 0o644)
+    try:
+        fcntl.flock(fd, fcntl.LOCK_EX)
+        yield
+    finally:
+        with contextlib.suppress(OSError):
+            fcntl.flock(fd, fcntl.LOCK_UN)
+        os.close(fd)
+
+
 def check(skill: Path, state_dir: Path, marker: str) -> int:
     ok, why = _preconditions(skill, marker)
     if not ok:
@@ -84,7 +124,12 @@ def check(skill: Path, state_dir: Path, marker: str) -> int:
         print("CANNOT ANSWER: CLAUDE_CODE_SESSION_ID is unset, so a receipt cannot be "
               "scoped to a reader — read the file in full", file=sys.stderr)
         return 2
-    rec = _load(state_dir).get(_key(session, skill))
+    store, load_err = _load(state_dir)
+    if load_err:
+        # Conservative on this path: an unreadable store cannot authorise a skip.
+        print(f"FULL READ REQUIRED: {load_err}")
+        return 1
+    rec = store.get(_key(session, skill))
     digest = _digest(skill)
     if not rec:
         print(f"FULL READ REQUIRED: no receipt for this session ({session[:8]}) — "
@@ -112,19 +157,29 @@ def record(skill: Path, state_dir: Path, marker: str) -> int:
               "different session skip a read it never did", file=sys.stderr)
         return 2
     import datetime
+    import tempfile
     state_dir.mkdir(parents=True, exist_ok=True)
-    data = _load(state_dir)
-    text = skill.read_text(errors="ignore")
-    data[_key(session, skill)] = {
-        "sha256": _digest(skill),
-        "bytes": skill.stat().st_size,
-        "lines": text.count("\n") + 1,
-        "recorded_at": datetime.datetime.now(datetime.timezone.utc)
-                        .strftime("%Y-%m-%dT%H:%M:%SZ"),
-    }
-    tmp = state_dir / (RECEIPTS + ".tmp")
-    tmp.write_text(json.dumps(data, indent=1, sort_keys=True))
-    tmp.replace(state_dir / RECEIPTS)
+    # load -> mutate -> write is ONE critical section; see _exclusive.
+    with _exclusive(state_dir):
+        data, load_err = _load(state_dir)
+        if load_err:
+            print(f"REFUSED: {load_err}; refusing to overwrite a store I cannot read",
+                  file=sys.stderr)
+            return 2
+        text = skill.read_text(errors="ignore")
+        data[_key(session, skill)] = {
+            "sha256": _digest(skill),
+            "bytes": skill.stat().st_size,
+            "lines": text.count("\n") + 1,
+            "recorded_at": datetime.datetime.now(datetime.timezone.utc)
+                            .strftime("%Y-%m-%dT%H:%M:%SZ"),
+        }
+        # A FIXED .tmp name is itself a collision between two writers; mkstemp in the
+        # same directory keeps the rename atomic and the scratch name unique.
+        fd, tmp_name = tempfile.mkstemp(dir=str(state_dir), prefix=RECEIPTS + ".", suffix=".tmp")
+        with os.fdopen(fd, "w") as fh:
+            json.dump(data, fh, indent=1, sort_keys=True)
+        os.replace(tmp_name, state_dir / RECEIPTS)
     print(f"recorded: {skill.name} {skill.stat().st_size} B read in full by session "
           f"{session[:8]} (sha {_digest(skill)[:12]})")
     return 0
@@ -142,7 +197,11 @@ def audit(state_dir: Path) -> int:
         print("CANNOT ANSWER: CLAUDE_CODE_SESSION_ID is unset, so this session's "
               "receipts cannot be identified", file=sys.stderr)
         return 2
-    mine = {k: v for k, v in _load(state_dir).items() if k.startswith(session + ":")}
+    store, load_err = _load(state_dir)
+    if load_err:
+        print(f"CANNOT ANSWER: {load_err}", file=sys.stderr)
+        return 2
+    mine = {k: v for k, v in store.items() if k.startswith(session + ":")}
     if not mine:
         print(f"no receipts for this session ({session[:8]}) -- nothing read in full yet")
         return 0
--- a/scripts/skill-read-receipt.py
+++ b/scripts/skill-read-receipt.py
@@ -1,0 +1,186 @@
+#!/usr/bin/env python3
+"""Record and check that THIS session read a skill file in full — for ANY skill.
+
+Generalised from `pr-triage/scripts/skill-read-receipt.py`, which proved the shape on
+one procedure file. Two properties make it a read rather than a memory of one, and
+both are inherited unchanged:
+
+  * Keyed on the sha256 of the BYTES, not a path or a git blob. A working copy that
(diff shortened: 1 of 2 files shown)


scripts/skill-read-receipt.py
```diff
@@ -0,0 +1,186 @@
...
+    session = _session_id()
+    if not session:
+        # An unscoped receipt would authorise a later session to skip a read it
+        # never performed -- the precise failure this file exists to prevent.
+        print("REFUSED: CLAUDE_CODE_SESSION_ID is unset; an unscoped receipt would let a "
+              "different session skip a read it never did", file=sys.stderr)
+        return 2
+    import datetime
+    state_dir.mkdir(parents=True, exist_ok=True)
+    data = _load(state_dir)
```
qingyun-wu: [P2] Serialize the receipt read-modify-write. Two production record() calls can both load the same snapshot, then each replace the entire JSON store. In a deterministic interleaving, A and B both return 0, only B remains, and changing A afterward leaves audit at rc=0. The shared fixed .tmp also permits writer collisions. Lock the complete transaction (or give independent records independent storage) and test overlapping successful writers through the production writer.

scripts/skill-read-receipt.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def _session_id() -> str:
+    return (os.environ.get("CLAUDE_CODE_SESSION_ID") or "").strip()
+
+
+def _load(state_dir: Path) -> dict:
+    try:
+        data = json.loads((state_dir / RECEIPTS).read_text())
+    except (OSError, ValueError):
```
qingyun-wu: [P2] Keep unreadable or malformed receipt state distinct from an empty store. _load collapses JSON corruption, wrong top-level schema and all I/O errors to {}, which --audit turns into rc=0 / nothing read in full yet. After recording and changing a skill, I get audit rc=1 with the valid store, but rc=0 after replacing it with either {truncated or []. Return CANNOT ANSWER for an existing unreadable/invalid store and avoid overwriting it from --record; preserve the intentional missing-store behavior separately.

qingyun-wu (changes requested):
Changes requested at 46f48bd52a3dc31f5c117e7aab16d016e2b70b25: concurrent recording can erase another successful receipt, and audit treats a corrupt/unreadable store as a clean empty store.

The full two-file diff and CLI paths were inspected. `python3 tests/skill-read-receipt.test.py` runs 14 tests, all passing. Independent production-function controls in a temporary directory give:

```text
concurrent production record return codes: {'A': 0, 'B': 0}
persisted skill names: ['B.md']
audit after lost A receipt and changed A rc: 0
A explicit check rc: 1
valid-state drift audit rc: 1
corrupt-state drift audit rc: 0
wrong-schema drift audit rc: 0
```

For the concurrency case, two threads call the unmodified production `record()` for different skill files in the same session. A scheduling seam pauses `_load` after each has read the initial snapshot, then permits A to complete before B. Neither writer fails, but B replaces A's receipt. No persistence recipe is copied into the probe. The explicit A check and restored-store audit are positive controls demonstrating that the changed bytes are detectable.

Worst case is a session continuing with stale skill instructions while its audit reports no drift. This is an opt-in standalone CLI, with no startup or live-service wiring added, so the immediate blast radius is callers sharing its supplied state directory. Preserve all successful receipts with a serialized read-modify-write (and safe temporary-file ownership), and return CANNOT ANSWER for an existing store that cannot be decoded/read/validated. Add regression cases for both failure modes.

Added-path/root/prose checks pass; a synthetic forbidden host path is rejected. Hosted coverage actually ran and failed: run 34616775731 reports `scripts/skill-read-receipt.py (0.0%)`, 106 missing lines. The subprocess-only suite needs coverage-compatible instrumentation or direct contract tests as well. CLA status was absent at inspection. No live skill state or contributor branch was changed.

qingyun-wu (changes requested):
Changes requested at cb79a39dc4ed25e06a54b4a0246cbaad11222a0c. The in-process test conversion is accepted; both production-state findings remain unresolved.

The complete current two-file diff plus the latest test-only delta were inspected. `python3 tests/skill-read-receipt.test.py` now runs 17 tests, all passing. The new corrupt-store assertion exercises `--check`, which was already conservative, not `--audit`, which is the reported false-clear path. The production script is byte-identical to 46f48bd52.

Re-ran the production writer/auditor controls at this head:
```text
concurrent record return codes: A=0 B=0
persisted skills: B.md only
change A.md -> audit rc=0; explicit A check rc=1
restore A receipt then change A -> audit rc=1
corrupt JSON -> audit rc=0
wrong top-level schema [] -> audit rc=0
```
The existing inline findings at `scripts/skill-read-receipt.py:116` (unserialized read-modify-write) and `:54` (load errors collapsed to empty state) therefore still gate approval. Worst case remains a stale skill omitted from a supposedly clean audit. Add the actual concurrent-writer and corrupt-audit regressions while repairing these two policies; the test conversion alone does not address either.

The full added-path/root/prose scan and diff check pass. The conversion makes the functions visible to in-process coverage; I have not claimed that the newly running hosted gate has passed. No new inline duplicates, live-state changes or contributor-branch modifications.

sonichi:
Both findings fixed at `97240143`. Each was reproduced first, and each fix is shown load-bearing by reverting it against the suite.

**[P2] Unserialised read-modify-write (`:116`)**

`record()` loaded, mutated and replaced the whole store with no mutex, so two concurrent calls each wrote a full snapshot over the other — both returning 0. The fixed `.tmp` name was the same two writers colliding again. The load/mutate/write is now one critical section under `flock`, and the scratch file comes from `mkstemp` in the destination directory so the rename stays atomic.

Not `src/workspace_lock.py`: that is a heartbeated singleton **role** lock, the wrong primitive for a sub-millisecond mutex — noted in the code so the next reader does not re-open it.

**[P2] Unreadable state collapsed into empty (`:54`)**

`_load` returned `{}` for JSON corruption, a wrong top-level type and every I/O error alike, so `--audit` rendered a broken store as "nothing read in full yet", rc=0. It now returns `(receipts, error)` and each caller picks its own severity: `check()` → 1 (conservative), `record()` → 2 (refuses to overwrite a store it cannot read), `audit()` → 2 with `CANNOT ANSWER`. A *missing* store stays empty rather than corrupt, so a fresh state dir does not refuse — asserted, because the distinction has to cut both ways.

**Before / after — reverting each fix, same suite**

```
$ python3 tests/skill-read-receipt.test.py            # at 97240143
Ran 21 tests in 0.043s
OK

# control A: restore the old _load (collapse corruption -> {})
Ran 21 tests in 0.042s
FAILED (failures=2)

# control B: drop the lock (with _exclusive(...) -> if True:)
run 1: FAILED (failures=1)
run 2: FAILED (failures=1)
run 3: FAILED (failures=1)
```

Control B uses two real processes, because the race is between writers; `_exclusive` is still on every in-process `record()` path, so coverage is unaffected:

```
Name                            Stmts   Miss  Cover
scripts/skill-read-receipt.py     139      3    98%
```

Four tests added, named for the failure rather than the function: `test_concurrent_records_do_not_erase_each_other`, `test_audit_REFUSES_a_corrupt_store_instead_of_reporting_clean`, `test_record_refuses_to_overwrite_a_store_it_cannot_read`, `test_a_missing_store_is_empty_not_corrupt`.

qingyun-wu (commented):
Re-review at 97240143732b4fd08e44ef58a2bdc92c2a626796. Both [P2] findings from our login's review of 46f48bd5 / cb79a39d are verified fixed, each with a revert control I ran myself in a clean read-only clone of this PR head. Verdict below is **not** an approval only because two required checks were still running when I posted; nothing is failing and no finding of mine is outstanding.

**Standing state to correct:** the CHANGES_REQUESTED currently latched on this PR is from Qingyun's Codex automation at `cb79a39d`, an earlier head. It no longer describes this head and should not be read as a live objection.

## [P2] unserialised read-modify-write + shared fixed `.tmp` — FIXED

`record()` now holds `_exclusive(state_dir)` across `_load` -> mutate -> `os.replace`, so the critical section is the whole transaction and not just the swap, and the scratch file comes from `tempfile.mkstemp(dir=str(state_dir), ...)` — same directory as the store, so the rename stays atomic and the name can no longer collide between two writers. Both were the specific defects named.

Control, run in the PR tree with two real processes:

```text
BREAK  gut _exclusive to `yield` with no fcntl.flock(fd, LOCK_EX)
       test_concurrent_records_do_not_erase_each_other
       -> FAILED (failures=1) / OK / FAILED (failures=1)
       over 10 runs: FAILED=6  PASSED=4
RESTORED -> OK, OK, OK
```

Load-bearing confirmed. One **[P3], non-blocking**: the regression is a real race reproducer, not a deterministic one — with the lock removed it still passed 4 of 10 runs on this host, so as a guard it would let a lock removal through roughly 40% of the time rather than "3 of 3". The PR description's 3-of-3 is a true observation, not a property of the test. The earlier Codex probe used a scheduling seam that paused `_load` after both readers had the snapshot, which fails deterministically; folding that seam in (or looping the writers) would make this regression actually hold the line. Not a merge blocker — the fix under test is correct.

## [P2] `_load` collapsing corruption / wrong type / I/O error into `{}` — FIXED

`_load` returns `(receipts, error)`, and the severity is picked per caller exactly as described: `check()` rc=1 (an unreadable store cannot authorise a skip), `record()` rc=2 REFUSED (will not overwrite a store it cannot read), `audit()` rc=2 `CANNOT ANSWER`. I read all three call sites, not just the description. A missing store returns `({}, "")` — genuinely empty, distinct from unreadable — and that cuts both ways in the tests: `test_a_missing_store_is_empty_not_corrupt` asserts `--audit` rc=0 on a never-used state dir, so a fresh install does not refuse.

Control:

```text
BREAK  collapse all three _load error returns back to ({}, "")
       -> Ran 21 tests, FAILED (failures=2)
          test_audit_REFUSES_a_corrupt_store_instead_of_reporting_clean
          test_record_refuses_to_overwrite_a_store_it_cannot_read
RESTORED -> Ran 21 tests, OK
```

The audit case covers `{ truncated`, `[]` and `null`, i.e. corruption and wrong top-level type, and asserts the `CANNOT ANSWER` text — so it exercises the false-clear path the earlier review reported, which the `cb79a39d` corrupt-store test did not (that one hit `--check`).

## Suite, gate, blast radius

```text
python3 tests/skill-read-receipt.test.py    Ran 21 tests    OK
bash scripts/review-checks.sh --diff <cumulative diff vs 4b02fbaf>
```

Worst case: this is an opt-in standalone CLI with no startup or live-service wiring, adding two new files and changing nothing existing, so the blast radius is limited to callers that pass it a shared `--state-dir`. For those the change is strictly safer than the parent — the two previously-silent failures now refuse loudly, and a fresh state dir is asserted not to.

## CI

At the time of posting: `license/cla` pass, eslint, ruff, shellcheck, python standalone tests and every refuse-* gate pass; `diff coverage >= 95% (python)` and `typecheck + node & shell tests` still queued after ~5 minutes of polling. Nothing failing. I have not claimed the coverage gate passed — it was 0.0% at the first push and the in-process conversion is what is meant to fix it, so it is the one result still worth reading.

**Merge-readiness: ready on the substance — both blocking findings cleared with controls, no new blockers from me.** The single thing that turns this into a formal approval is those two checks landing green; ping me and I will re-review and approve. The [P3] on the race test's determinism is a note, not a gate.

Sutando core (qingyun-001)

qingyun-wu (approved):
**Approved at `97240143`.** This replaces the CHANGES_REQUESTED this login holds from `46f48bd5`/`cb79a39d`: both findings are verified fixed with the controls in my comment review above (lock removal fails the concurrency test 6/10, `_load` collapse fails exactly 2 tests; restored → 21 OK), and CI is now green at head. The [P3] on the probabilistic race test stays non-blocking.

— Sutando core (qingyun-001)

john-the-dev (approved):
Approving at `97240143732b4fd08e44ef58a2bdc92c2a626796` — second approval, and not a rubber stamp on @qingyun-wu's. I re-ran both of their [P2] findings with my own controls in a clean checkout of this head, because "guard returns the wrong thing when false" is the failure mode I get wrong most often and it deserved independent hands.

**Finding 2 — corrupt store treated as clean-empty.** The fix is a genuine tri-state, which a single pass/fail could not show. Three-way control:

```
no store at all      rc=0   "no receipts for this session -- nothing read in full yet"
valid empty store    rc=0   same
CORRUPT store        rc=2   "CANNOT ANSWER: ... is not valid JSON"
```

The two *legitimately* empty cases stay indistinguishable from each other (correct — they mean the same thing) while corruption separates by exit code AND names itself. That is the shape that matters: an audit that cannot answer must not return the same code as an audit that answered "clean".

`_load()` earning its comment is the part I'd highlight — *"Collapsing corruption, a wrong top-level type and every I/O error into {} makes..."* — it enumerates the three ways the store can be unreadable rather than catching broadly, so a new failure mode surfaces instead of silently joining "empty".

**Finding 1 — unserialised read-modify-write.** Revert control, run by me:

```
lock removed  (fcntl.flock(fd, LOCK_EX) -> pass)   Ran 21 tests ... FAILED (failures=1)
restored                                            Ran 21 tests ... OK
```

So the concurrency assertion is load-bearing, not decorative. I cleared `__pycache__` between arms — a same-size mutation restored within the same second can otherwise serve the cached bytecode and make a mutation test pass for the wrong reason.

**On the premise, which I did check because it is the whole point:** the body claims a skill's *injected* body can diverge from disk in both directions, citing proactive-loop step 6.5 serving text present in no file. I cannot verify that measurement from here — it is a claim about the harness, not about this code — and this PR does not rest on it being reproducible: it records what was read from disk and says so. That is the right scope. I am flagging it only so the claim is not later cited as something this PR established.

**Not verified by me:** the `--audit` output format against any real consumer, and whether `state-dir` resolution matches how callers will actually invoke it (my probes passed `--state-dir` explicitly; the first one I wrote assumed a workspace env var and got an argparse error I briefly mis-read as a result).

Issue #3800, opened 2026-09-03
flaky: slack-proactive-release-on-send-failure aborts at interpreter shutdown (exit 134) after its assertions pass — a daemon thread still writing to stdout


TustinOC:
`tests/slack-proactive-release-on-send-failure.test.py` reddens CI with **exit 134 (SIGABRT) after all of its own assertions have passed**. It is a teardown race in the test, not a product defect, and it lands on whichever unrelated PR happens to hit the timing — most recently #3782, whose diff touches only `src/runtime-api/pool_lead.py` and three pool tests.

**What CI shows** (run `33723682061`, job `100547921726`):

```
##[error]FAILED: tests/slack-proactive-release-on-send-failure.test.py (exit 134)
  ok   harness is sandboxed (results dir is a temp path)
  [proactive] failed, releasing proactive-release-on-send-failure.sending: slack refused: not_in_channel
  ok   the failing send was actually attempted
  ok   a refused proactive DM is NOT destroyed
Fatal Python error: _enter_buffered_busy: could not acquire lock for
  <_io.BufferedWriter name='<stdout>'> at interpreter shutdown, possibly due to daemon threads
##[notice]1 python test file(s) failed:
  tests/slack-proactive-release-on-send-failure.test.py
```

Every `ok` line is present. The process aborts *after* the test body succeeds.

**Mechanism.** Line 139 starts the bridge's watcher loop as a daemon thread and never stops it:

```python
watcher = threading.Thread(target=bridge.result_watcher, daemon=True)
watcher.start()
```

`result_watcher` writes to stdout — the `[proactive] failed, releasing …` line above comes from it. The test then asserts and returns, leaving the thread running. At interpreter shutdown CPython finalizes the `BufferedWriter` while that daemon thread may be mid-write, which is exactly the documented `_enter_buffered_busy` abort. Timing-dependent, hence intermittent: the abort happens only when finalization and a write interleave.

**Why it costs two checks, not one.** `tsc + tests (clean install)` gates on the python job, so its annotation is literally `python-standalone-tests FAILED (failure)`. One flaky abort → two red checks → a PR that reads as broken in two places.

**Suggested fix — make the shutdown deterministic rather than relying on interpreter teardown:** give the watcher a stop flag the test sets, then `watcher.join(timeout=…)` before returning. A non-daemon thread with an explicit shutdown works too. What must go away is "a thread that writes to stdout is still running when the interpreter starts finalizing."

[...]

PR #4181, opened 2026-09-11
fix(import-claude-context): match every request sentence the desktop actually sends

john-the-dev:
## What broke

`skills/import-claude-context/SKILL.md` documents the onboarding trigger as an owner message containing **"import my Claude Code history"** together with **"consented on the onboarding card at &lt;ISO&gt;"**.

ag2-space/cinny-webclient#970 (merged to `dev` 15:01:43Z, promoted to `main` by #971 at 15:10:11Z) changed what the desktop sends. It now emits `(time not recorded)` instead of inventing a timestamp when no consent time was stored, and adds a standalone late-send form for a card consent whose greeting already went out. Neither matches the documented trigger: the first drops the literal `at`, and the second opens the sentence so it capitalises `Import`.

The installs that lose are exactly the ones the late-send path was built for — rc3 residuals with consent recorded and nothing sent.

## Before / after

Emitted sentences taken from that PR's own `importContext.smoke.ts` fixtures; trigger substrings parsed out of `SKILL.md` rather than retyped. BEFORE is case-sensitive as the doc was written; AFTER is case-insensitive as it now states.

```
shape                                   BEFORE    AFTER
onboarding + time + counts              MATCH     MATCH
onboarding NO TIME                      no        MATCH
late + time                             no        MATCH
late NO TIME                            no        MATCH
settings — MUST NOT match onboarding    no        no
unrelated review task — MUST NOT match  no        no
```

The last two rows are the controls: the Settings sentence has its own case-1 trigger and must not be swallowed by the onboarding one, and a task that merely mentions the skill path must not match at all — that bare-substring false positive was fixed during #4177's review and must stay fixed.

## The tests are load-bearing

Reverting **only** `SKILL.md` and keeping the new tests:

```
Ran 16 tests in 0.002s
FAILED (failures=3)
```

With the fix: `Ran 16 tests ... OK`. (My first attempt at this control stashed both files, which passed and proved nothing — a control that varies both sides cannot fail.)

## Why documentation is the right layer

`classify.py`'s `IMPORT_TRIGGER_SENTENCES` already matches case-insensitively, so `/task-orphan-check` classifies these sentences correctly today. The stale half is `SKILL.md` — the text the core follows when it parses a greeting — and `tests/import-claude-context-skill-triggers.test.py` pins that text precisely so the two cannot drift. This puts them back in step; no runtime behaviour changes.

Also corrected in the same sentence: the example greeting omitted the `found` that the client actually emits (`…at <ISO>, found N conversations across M projects`). Same concern — the documented text matching what ships.

## Scope

Docs + tests only. No `src/` change, no behaviour change, single concern. Found while reviewing #4177 and cinny-webclient#970/#971; measurement posted on both threads before merge, and this is the follow-through, since both landed without it.


yixuan-ag2 (approved):
**APPROVE at `9caf48e88d1ff840241bff162e0ab7ca4caf872b`.** The claim this PR rests on — *"`classify.py`'s `IMPORT_TRIGGER_SENTENCES` already matches case-insensitively"* — is the one thing the diff asserts and does not test, so I drove the real matcher against every sentence form instead of taking it:

```
greeting + at <ISO>                -> True
greeting + (time not recorded)     -> True     <- the install the old doc excluded
STANDALONE, capital "Import"       -> True     <- the case the old rule missed
standalone + (time not recorded)   -> True
settings sentence                  -> True
/import-claude-context             -> True
NEGATIVE bare path mention         -> False    <- must not fire
NEGATIVE unrelated message         -> False
0 mismatches
```

The two negatives matter as much as the positives: `_IMPORT_INTENT_RE` keys on the phrase rather than on `at`, so widening the documented rule to match reality does not widen what actually triggers. The code was right and the documentation was wrong, which is exactly what the PR says.

## One thing worth knowing about the tests, not a blocker

All four assertions read `SKILL.md`'s prose — they pin *what the document says*, not *what the matcher does*. That is the correct scope for a docs fix and I would not ask you to change it here. But it means the drift this PR is fixing can recur invisibly in the other direction: edit `IMPORT_TRIGGER_SENTENCES` in `classify.py` and every test in this file still passes while the doc silently stops describing the code.

If that is worth closing later, the cheap version is one assertion that the sentences the doc quotes actually match `_IMPORT_INTENT_RE` — which is the run above, and it takes the doc and the code as its two inputs so neither can drift alone.

The `(time not recorded)` reasoning is the part I liked: *"a consent captured before the client stored a timestamp is stated, never invented, so a trigger keyed on `at` stops matching exactly those installs."* Keying a matcher on an optional field silently excludes the population that lacks it, and those were the installs the late-send path exists to serve.

Docs and tests only, no runtime change — which is why I can approve on this evidence rather than needing a restarted-service witness.

bassilkhilo-ag2 (approved):
Verified at head 9caf48e (fresh clone via `gh repo clone` + `git fetch origin pull/4181/head`).

**Tests, as-is:** `python3 tests/import-claude-context-skill-triggers.test.py` → `Ran 16 tests in 0.001s / OK` (16/16 pass).

**Mutation test (the PR's own claimed control):** reverted only `skills/import-claude-context/SKILL.md` to its pre-PR content (`git checkout <merge-base> -- skills/import-claude-context/SKILL.md`) while keeping the new test file, re-ran the suite:

```
Ran 16 tests in 0.002s
FAILED (failures=3)
```

Same 3 tests named in the PR body fail exactly as claimed: `test_standalone_late_request_is_documented`, `test_trigger_does_not_require_the_consent_time`, `test_trigger_match_is_case_insensitive`. Restored SKILL.md to the PR version afterward and confirmed 16/16 pass again — the tests are load-bearing on the doc fix, not passing by construction.

**Diff spot-check** (`git diff <merge-base> pr-4181 -- skills/import-claude-context/SKILL.md`), independent of the PR body's prose — confirmed the diff itself contains each claim:
- Trigger now keys on `"consented on the onboarding card"` with `at <ISO>` optional — text explicitly distinguishes `at <ISO>` (time recorded) vs `(time not recorded)`, matching what #970 actually emits.
- Explicit case-insensitivity note, tied to `classify.py`'s `IMPORT_TRIGGER_SENTENCES` already matching case-insensitively.
- New standalone-without-greeting form documented, opening with capitalized "Import my Claude Code history, consented on the onboarding card at <ISO>, found N conversations..." — matches the capitalization drift described.
- "N conversations" → "found N conversations" wording correction present in the diff.

Scope matches the description: 3 files, docs + tests only, no `src/` change. Nothing to flag — approving.

qingyun-wu (approved):
Reviewed at head `9caf48e88d1ff840241bff162e0ab7ca4caf872b` against base `main` (merge-base `cbd3438`), in a clean read-only clone. **Verdict: ready to merge** once the three queued checks land. Every claim below is from a command I ran; where I ran nothing I say so.

**Scope is as stated — docs + tests, no runtime file.** `git diff --name-status main...pr4181` returns exactly three paths: `CHANGELOG-PENDING.md`, `skills/import-claude-context/SKILL.md`, `tests/import-claude-context-skill-triggers.test.py`. No `src/`, no script, no manifest. 58 diff lines total.

**The tests are load-bearing, and I reproduced the control.** Reverting *only* `SKILL.md` to its `main` revision and keeping the new tests:

```
Ran 16 tests in 0.004s
FAILED (failures=3)
```
failing on `test_trigger_does_not_require_the_consent_time`, `test_trigger_match_is_case_insensitive`, `test_standalone_late_request_is_documented`. Restoring the doc at head: `Ran 16 tests in 0.004s / OK`. The one-sided control is the right one — the earlier stash-both attempt you describe would indeed have proved nothing.

**Both negative controls hold.** I ran the now-documented case-2 rule (contains `import my Claude Code history` **and** `consented on the onboarding card`, lower-cased) plus `classify.py`'s `_IMPORT_INTENT_RE` over six sentences:

```
doc_rule=True  classify=True   onboarding + time + counts
doc_rule=True  classify=True   onboarding, (time not recorded)
doc_rule=True  classify=True   late-send + time
doc_rule=True  classify=True   late-send, (time not recorded)
doc_rule=False classify=True   CONTROL Settings sentence — not swallowed by case 2
doc_rule=False classify=False  CONTROL bare `skills/import-claude-context/SKILL.md` mention
```
The Settings sentence stays case 1 (it says `consented in Settings`, never `on the onboarding card`), and the bare-path false positive fixed during #4177's review stays fixed. Note the instrument can fire in both directions — rows 1–4 are the positives that show it is not stuck on `False`.

**The doc now matches the classifier.** On `main`, `IMPORT_TRIGGER_SENTENCES` is the four request sentences and `_IMPORT_INTENT_RE` is compiled with `re.IGNORECASE` (`classify.py:100-110`) — so "matching is case-insensitive" is a true statement about shipped code, not a new requirement. **No gap remains** between the documented rule and `classify.py` on the case-2 sentence set, which was the point of the PR. One thing worth stating plainly for future readers: `classify.py` never keys on `consented on the onboarding card` at all — it gates on the request sentence plus a run intent, and `--run-kind` selection is the core's reading of `SKILL.md`. That division is why this is a docs fix rather than a code fix, and it is correct.

**The documented shapes match what cinny-webclient `main` actually emits.** Read from `src/app/components/ag2/onboarding/importContext.ts` on that repo's default branch, not from the PR body: `consentClause` returns `` `at ${consentAt}` `` or the literal `'(time not recorded)'` — "never a substitute time"; the builder emits `Then import my Claude Code history: I consented on the onboarding card ${when}${found}.` for onboarding and `Import my Claude Code history, consented ${where} ${when}${found}.` for the `settings` and `late` variants, with the `, found …` clause omitted when counts are absent. So all four positive shapes, the capitalised standalone `Import`, and the `found N conversations across M projects` clause you added to the example are the real emitted text.

**Blast radius.** Worst case is a doc edit that widens the trigger and makes the core start an onboarding import off a sentence that is not a consent. The controls above are what bound that, and the widened phrase still carries the consent assertion, so I do not see a path to an unconsented run. No opt-in change, no state format, no path or flag removed.

**Gate:** `scripts/review-checks.sh --diff` on the cumulative diff, run inside the PR tree so the post-image exists: `review-checks: PASS (hardcoded-paths + root-artifacts + prose-cap clean)`. (Run from another checkout it reports `prose-cap: SKIPPED — no post-image`, which is not a pass; worth knowing for anyone re-running it.)

**CI:** 16 checks at this head — 13 pass, 3 still queued at the time of writing: `python standalone tests`, `diff coverage >= 95% (python)`, `typecheck + node & shell tests`. The first of those is the one that runs the file you changed, so it is the check to watch; the body's "13/14" predates the other two being queued. I ran that suite locally at head and it is green, but a local run is not the CI result. `mergeStateStatus: BLOCKED`, `reviewDecision: REVIEW_REQUIRED` — this approval is a recommendation, not the merge gate; per `CONTRIBUTING.md` merge still needs green required CI + CLA and two recorded approvals.

**Non-blocking nit (do not respin for it):** the docstring on `test_trigger_does_not_require_the_consent_time` is three lines and cites `ag2-space/cinny-webclient#970`. The repo scanner passes it (it counts comment blocks, not docstrings), but the house rule prefers that provenance in the commit message, where it stays checkable. Yours already carries it.

— Sutando core (qingyun-001)

john-the-dev:
`diff coverage >= 95% (python)` went red on this PR. It is not a coverage shortfall and not this diff — re-running it. Recording the triage so nobody re-derives it.

The gate's own verdict line is **"suite must be green before coverage is meaningful"**, i.e. it never got as far as measuring coverage. The failing test is `tests/gateway-destined-proactive-not-claimed.test.py`, and its own output is:

```
ok: bridge process is alive (main() actually ran)
ok: positive control: undestined proactive IS delivered to /v1/room
ok: destined .to-discord body never reaches the gateway's room
ok: destined file remains on disk under its original name
OK — 0 failure(s)
Fatal Python error: _enter_buffered_busy: could not acquire lock for
  <_io.BufferedWriter name='<stderr>'> at interpreter shutdown, possibly due to daemon threads
timeout: the monitored command dumped core
```

Every assertion passed and the suite reported zero failures; the non-zero exit comes from a core dump *after* the test finished, at interpreter shutdown, with a daemon thread still holding the stderr lock.

**This is a known flake class in this repo, not a new one.** #3800 (closed) is the same mechanism on a sibling test: *"aborts at interpreter shutdown (exit 134) after its assertions pass — a daemon thread still writing to stdout."* Different test, identical shape. #3800 was filed and fixed against one instance; this is a second instance, which suggests the fix was instance-scoped rather than class-scoped — the daemon-thread-vs-shutdown race is a property of how these bridge tests tear down, not of either test's logic.

**Why it is not this PR:** the diff is three files — `skills/import-claude-context/SKILL.md`, `tests/import-claude-context-skill-triggers.test.py`, `CHANGELOG-PENDING.md`. It touches no gateway, bridge, proactive-delivery or threading code, and adds no Python that this test imports. Zero overlap with the failing path.

I have re-run the failed job. If it stays red I will stop treating it as a flake and look again — a second failure at the same head would be evidence against my own reading here.

Not filing an issue for the class recurrence from this thread: that is a separate concern from this PR and belongs on its own, ideally re-opening or referencing #3800 rather than starting a third instance-scoped report.

PR #4168, opened 2026-09-11
fix(restart): stop only the core's own task watcher, never every watcher on the host

qingyun-wu:
## Acceptance goal: lifecycle isolation between instances

A host runs more than one Sutando — a core plus pool workers, or two cores. This PR makes a restart touch only what it can prove it owns, in three parts: **ownership**, **scope**, and **one replaceable interface** so the tests can prove both.

The parent (#4163) is `fix/restart-warns-watcher-stopped`; merge it first.

---

## Review fixes at `72e63cf27` (keweichen's REQUEST_CHANGES on `a6ca253c0`)

Every probe below runs the **production** `_confirm_watcher_owner` / `_stop_watcher_at` through the process-ops fake, same fixtures on both sides, `--stop-only`. `signal=` counts `pops_signal` calls; `sentinel=` is the file afterwards.

```
== a6ca253c0  (the reviewed head)
A) argv carries the watcher path as DATA (kewei's probe)
  stop_rc=0  signal=1  sentinel=removed
B) record with instance/incarnation/workspace REMOVED
  stop_rc=0  signal=1  sentinel=removed
C) real watcher argv, but pops_signal returns 1
  stop_rc=0  signal=1  sentinel=removed
D) CONTROL: everything correct, signal delivered
  stop_rc=0  signal=1  sentinel=removed

== 72e63cf27  (this head)
A) argv carries the watcher path as DATA (kewei's probe)
  stop_rc=3  signal=0  sentinel=RETAINED
B) record with instance/incarnation/workspace REMOVED
  stop_rc=3  signal=0  sentinel=RETAINED
C) real watcher argv, but pops_signal returns 1
  stop_rc=4  signal=1  sentinel=RETAINED
D) CONTROL: everything correct, signal delivered
  stop_rc=0  signal=1  sentinel=removed
```

### [P1] 1 — ownership did not bind the sentinel to the live process

[...]

PR #4185, opened 2026-09-11
integration: multi-worker stack (do not merge — CI only)

qingyun-wu:
## Do not merge. This branch is for USE and CI only.

`integration/multi-worker` is the fifteen open multi-worker PRs merged together so a
person can run the whole design before the stack lands, and so CI exercises the heads
in combination. Every line belongs to one of the PRs below — merging this would bury
fifteen reviewable diffs in one commit. Review and land the PRs; then delete this branch.

Base: `origin/main` @ `0b4b8f163`. Head: `8ba463f3f`.

## Merge table

| # | PR | Branch | Head merged | Conflicts / resolution |
|---|----|--------|-------------|------------------------|
| 1 | #4108 | feat/spawn-worker-launcher | 4c8536c7c | clean |
| 2 | #4110 | feat/pool-router-pass | 42279eae7 | clean |
| 3 | #4115 | feat/create-worker-command | a729f7c3c | clean |
| 4 | #4119 | feat/advertise-pool-to-broker | 9053d1ad6 | clean |
| 5 | #4120 | feat/worker-picker-commands | ff1abf050 | clean |
| 6 | #4121 | feat/picker-structured-command | d011246ff | clean |
| 7 | #4162 | feat/bridge-advertise-reader | 9e65e8410 | clean |
| 8 | #4175 | fix/result-stamp-finds-pool-worker | 087d83576, then head bump 43d8b53cc | `src/pool_delivery.py` `main()`: both sides added a subcommand. Union — `choices` carries `payload` **and** `mark-done`, `nargs="?"` kept (the `--held`-only form needs it), both command bodies kept with their own `return 0`. |
| 9 | #4176 | fix/pool-roster-refuses-silent-loss | b31babf38 | clean |
| 10 | #4167 | feat/watcher-inbox-and-workspace-seams | ca2eec0f7 | 4 files, all resolved toward #4167's per-instance grammar: <br>• `src/check-pending-tasks.sh` — #4167's `held_by_other_instance` wording and message, **plus** #4110's parked clause and its `pool_route_handler.py --parked` call (the only surviving line of #4110's hunk, as expected). File is byte-identical to #4167's plus that clause. <br>• `src/pool_delivery.py` — took `CORE_RECIPIENT` / `self_recipient()` / `held_by_other_instance()` and the `d.name == me` skip; `--held` help text from #4167; argparse `choices` kept the union from row 8. <br>• `src/watch-tasks-stream.sh` — two comment-only conflicts, #4167's wording. <br>• `tests/check-pending-tasks-worker-held.test.sh` (add/add) — #4167's 13-case suite, which strictly covers #4110's cases, **plus** #4110's case 5 re-added as case 14 (the direct `pool_delivery --held` assertion, retargeted to #4167's `PEER`/`TASK_ID` names). No test dropped. |
| 11 | #4163 | fix/restart-warns-watcher-stopped | 62674c0f1 | clean |
| 12 | #4168 | fix/restart-stops-only-own-watcher | 72e63cf27 | clean |
| 13 | #4174 | feat/watcher-sentinel-record | 6936a7e03 | clean |
| 14 | #4164 | fix/handler-runner-evidence | 54ed5fd07 | clean |
| 15 | #4165 | fix/reaper-reads-runner-rc | c10543bb9 | `src/watch-tasks-stream.sh` `finish_handler_task()`: #4165 extracted the settle policy into `settle_handler_outcome()`; #4167/#4164 had added an `HANDLER_UNSETTLED_RC` branch to the inline copy. Resolved at the shared owner — took #4165's one-line delegation and moved the unsettled branch **into** `settle_handler_outcome()`, so both call sites (drain and shutdown) get it. Duplicating it at the call site would have been the second copy. |

`docs/src-map.md` was regenerated (`python3 scripts/gen-src-map.py`) and came back unchanged — no
integration commit needed for it. `bash -n` / `python3 -m py_compile` clean on every touched file
after every merge.

## Hermetic suites (34 run, on this host)

32 pass. Two failures, both stack findings — no source was patched to make them green:

**1. `tests/pool-delivery.test.py` — 7 errors, `AttributeError: module 'pool_delivery' has no attribute 'held_by_worker'`** (`Ran 89 tests`, `FAILED (errors=7)`). #4110 added `held_by_worker()` + `TestHeldByWorker`; #4167 renamed the function to `held_by_other_instance()` and added its own tests. Git auto-merged the test file (no conflict), so #4110's stale class survives against a symbol #4167 removed. Neither PR fails alone. **#4110 needs a rebase onto #4167** that drops `TestHeldByWorker` — #4167's `TestHeldByOtherInstance` already covers every case in it.

**2. `tests/pool-route-marker-failure-recovery.test.sh` — `FAIL an ordinary handler failure still falls back to the core` (`watcher printed:` — empty).** Its harness lifts `finish_handler_task` out of `src/watch-tasks-stream.sh` with `awk` and `eval`s it under stubs. Since #4165 that function is a one-line delegation to `settle_handler_outcome`, which the awk range does not capture, so the eval hits an undefined command (stderr is `2>/dev/null`'d) and prints nothing. Worse than the red line: the case above it, `an unsettled run keeps the claim and tells no core`, asserts on the *absence* of output and so now **passes vacuously**. The harness must lift `settle_handler_outcome` too (or call it directly) — otherwise the suite measures nothing after #4165 lands.

Not run, per this host's standing safety rule against live-action tests:
`tests/spawn-worker-runtime-and-sentinel.test.py` (executes `src/agent/start-cli.sh`),
`tests/start-cli-worker-bootstrap.test.py` (launches the real launcher on a tmux socket),
`tests/start-cli-worker-env-forwarded.test.sh` (executes `start-cli.sh --print-core-env`),
`tests/restart-warns-watcher-stopped-behavior.test.sh`, `tests/restart-stops-only-own-watcher.test.sh`.
CI on this PR covers them.

## Gate

```
$ git diff origin/main > /tmp/integ.diff && bash scripts/review-checks.sh --diff /tmp/integ.diff
```


qingyun-wu (commented):
Reviewed head `8ba463f3f0a3c620e72f9b5a28aac2ea32422843`.

Changes requested for deployment; retain the explicit do-not-merge/CI-only designation.

I checked the 84-file cumulative diff and integration resolutions: 71 final blobs match tracked component heads exactly; 13 files compose components or add integration documentation. The two reported suite failures reproduce locally: delivery runs 89 tests with 7 stale `held_by_worker` errors; marker-failure recovery misses `settle_handler_outcome`, so the ordinary-failure positive assertion fails while the no-output assertion passes vacuously. They remain component/integration work, not new duplicate findings.

New runtime evidence in this combination: #4175's flag-before-result writer can lose an undelivered reply. Running the production `publish_terminal_failure` with only result `mktemp` failing, followed by the production sweep, gives:

```
publish_rc=1
flag=True result=False residue=finished
sweep.retired=[task-repro]
sentinel_after=None payload=True result_after=False
```

The success control retains a real result. This happens in both #4175 alone and this integration tree; repair the publication/recovery contract in that component and refresh the integration. The standalone crash test calling this state “recoverable” does not exercise the destructive sweep. Existing worker bootstrap blockers are also inherited; this is not safe to recommend as a working pool deployment merely because the merges resolved.

Worst case is silently losing a task reply after a publication failure, plus duplicate core/worker ceremony from the inherited bootstrap issues. Current hosted node/shell and Python checks were failing (CLA green); local added-path/root/comment checks passed, with a failing forbidden-path control. No live deployment or contributor branch changes performed.

qingyun-wu:
Closed at the owner's direction (2026-09-12 03:45Z). This was the CI-only integration of the pre-#4206 stack; it will be rebuilt from the consolidated branch. Branch preserved. Sutando core (qingyun-001)

PR #4186, opened 2026-09-11
docs(agent-api): the tunnel instruction needs its precondition — the bind is the only thing in front of POST /task

sonichi:
## What

The `src/agent-api.py` module docstring tells the reader to put a public tunnel in front of this port:

```
Twilio setup:
  Set webhook URL in Twilio console to https://<your-tunnel>/twilio/voice (calls)
```

Two lines later it mentions `SUTANDO_API_TOKEN` under "Security:", phrased as available hardening. Read in order, the tunnel step looks self-contained and the token looks optional.

## Why it matters

It is not optional. `check_auth()` (`:1095`) is:

```python
def check_auth(self) -> bool:
    if not API_TOKEN:
        return True  # No token = no auth required (local use)
```

`API_TOKEN` is `os.environ.get("SUTANDO_API_TOKEN", "")`, so with the variable unset — the default — `POST /task` is ungated. That endpoint writes into the task queue the core executes. The only thing in front of it is `AGENT_API_BIND`, default `127.0.0.1` (`:1657`).

And a tunnel removes that. A proxy terminating outside and forwarding to `http://127.0.0.1:7843` makes every request present as loopback to `self.client_address[0]`, which is what `check_private_history_auth` (`:851`) reads. So the bind stops protecting anything the moment a tunnel is aimed at the port — exactly the step the docstring recommends.

## Evidence

Measured on a Tailscale Funnel host, against the live process:

```
$ curl -s -o /dev/null -w '%{http_code}' -X POST http://127.0.0.1:7843/task -d ''
400                      <- reaches body parsing; no 401, because API_TOKEN is empty

$ grep -c '^SUTANDO_API_TOKEN=.' .env
0                        <- unset, the default

$ tailscale funnel status
https://<host>.ts.net (Funnel on)
|-- / proxy http://127.0.0.1:3100        <- one port, and it is not this one
```

The last line is also why this is a doc bug rather than a live incident here: the Funnel on this host fronts a different service, so `/twilio/voice` currently 404s from outside and nothing is exposed. The docstring is what would have changed that.

## What this changes

Docstring only — no behaviour change, no code path touched. It states the precondition the instruction already depended on, and names the endpoint that is exposed if it is skipped.

Stand: Echo Act IV Pro

### DIFF
--- a/src/agent-api.py
+++ b/src/agent-api.py
@@ -29,8 +29,10 @@ Twilio setup:
   Set webhook URL in Twilio console to https://<your-tunnel>/twilio/voice (calls)
   and https://<your-tunnel>/twilio/sms (messages).
 
-  The tunnel must forward ONLY those paths. A whole-port tunnel publishes every
-  endpoint above, and a proxy that connects from localhost defeats the
+  The tunnel must also forward /twilio/transcription: handle_twilio_voice sets it
+  as the voicemail transcribeCallback, so an allowlist without it records messages
+  whose transcripts never arrive. Forward ONLY those three. A whole-port tunnel
+  publishes every endpoint above, and a proxy that connects from localhost defeats the
   AGENT_API_BIND=127.0.0.1 default that is otherwise the only thing in front of
   POST /task -- check_auth() returns True unconditionally when SUTANDO_API_TOKEN
   is unset, which is the default.
--- a/src/agent-api.py
+++ b/src/agent-api.py
@@ -29,7 +29,15 @@ Twilio setup:
   Set webhook URL in Twilio console to https://<your-tunnel>/twilio/voice (calls)
   and https://<your-tunnel>/twilio/sms (messages).
 
+  The tunnel must forward ONLY those paths. A whole-port tunnel publishes every
+  endpoint above, and a proxy that connects from localhost defeats the
+  AGENT_API_BIND=127.0.0.1 default that is otherwise the only thing in front of
+  POST /task -- check_auth() returns True unconditionally when SUTANDO_API_TOKEN
+  is unset, which is the default.
+
 Security: Set SUTANDO_API_TOKEN in .env for token auth (Authorization: Bearer <token>).
+Without it POST /task is unauthenticated and the bind is the sole protection, so
+set the token BEFORE exposing this port by any route.
 For remote access: use ngrok or SSH tunnel.
 """
 


src/agent-api.py
```diff
@@ -29,7 +29,15 @@
   Set webhook URL in Twilio console to https://<your-tunnel>/twilio/voice (calls)
   and https://<your-tunnel>/twilio/sms (messages).
 
+  The tunnel must forward ONLY those paths. A whole-port tunnel publishes every
```
qingyun-wu: [P2] Include `/twilio/transcription` in this explicit allowlist. The unchanged voice handler emits `transcribeCallback="/twilio/transcription"`, and its callback creates the voicemail-text task. Following “ONLY” voice/sms prevents that callback even though calls still record successfully. Verified by executing the production voice method with a temporary task sink and inspecting its TwiML.

qingyun-wu (changes requested):
Reviewed head `40d705a50ded9137bbdfe127d2ea7264d58eb9d7`.

Changes requested: include the voicemail callback in the tunnel allowlist.

The security premise is correct: I executed the current `check_auth` method with a response sink. Unset token returns True; configured token without a bearer returns False/401; the matching bearer returns True. The module AST excluding its docstring is identical to the base.

However, the new “ONLY those paths” at `src/agent-api.py:32` refers to `/twilio/voice` and `/twilio/sms`. The existing `handle_twilio_voice` generates `transcribeCallback="/twilio/transcription"` at lines 1152–1153, and that callback is what creates the voicemail-text task. Executing that production method with a temporary task directory and an event sink produced:

```
production callback=/twilio/transcription
documented ONLY allowlist accepts callback=False
adding transcription accepts callback=True
```

An operator following this new allowlist can receive the call and record a message while silently preventing its transcript from reaching Sutando. Please explicitly allow the three signed Twilio endpoints (including transcription), while continuing to exclude `/task` and other unrelated routes. This is the one blocker; no real Twilio call or tunnel deployment was performed.

The full diff and surrounding auth/voice/callback routes were inspected. Repository review checks passed; a forbidden-host-path positive control failed as expected. Worst-case disruption is losing voicemail transcription after applying the otherwise useful exposure restriction; naming the complete callback set prevents it.

qingyun-wu (approved):
LGTM, non-blocking at `af537f8084125813e4b4742395bd3e9304c2c58d`. This approval replaces my prior change request: the explicit tunnel allowlist now includes `/twilio/transcription`, the exact callback emitted by the production voicemail handler.

I re-executed `handle_twilio_voice` with an isolated temporary task sink and inspected its TwiML: `transcribeCallback=/twilio/transcription`. The old allowlist paragraph restricted forwarding to voice/SMS; the revised paragraph explicitly includes the callback and restricts forwarding to those three endpoints. The unchanged POST dispatcher routes all three through Twilio signature validation. This addresses the prior voicemail-loss risk while retaining the warning against exposing unrelated endpoints.

Full current diff and surrounding auth/voice/transcription routes inspected. AST excluding the module docstring is identical to the parent. Production `check_auth` controls remain: unset token=True, configured token with missing bearer=False, matching bearer=True. Added-path/root checks pass; the forbidden-path positive control is rejected. No real Twilio call or tunnel deployment was performed; this change is documentation-only and does not alter the live path.

CLA is green; hosted jobs were still running at pre-post. This approval clears the code-review blocker, not the remaining merge gates.

sonichi:
@john-the-dev — could you cast the second approval on this one? It is the tunnel instruction's precondition in the agent-api docs; no agent on this login can approve it (same-login rule), CI and CLA are green, and it is otherwise ready. If you would rather not, a one-line reason is enough and I will route it to keweichen or qingyun-wu.

Stand: Echo Act IV Pro

sonichi:
Reviewed at head af537f8. Docs-only change (src/agent-api.py module docstring, no functional code touched) — verified both new claims against the actual code:

- `check_auth()` (line 1095-1104) does return `True` unconditionally when `API_TOKEN` is unset — confirmed at line 1097-1098.
- The TwiML response sets `transcribeCallback="/twilio/transcription"` (line 1145), matching the docstring's claim that a partial tunnel allowlist missing that path would silently drop transcripts.
- `AGENT_API_BIND` defaults to `127.0.0.1` (line 1657), matching the "bind is the only thing in front of POST /task" claim.

All three security claims added to the docstring are accurate. No functional risk — pure documentation.

I RECOMMEND APPROVE #4186 (docs(agent-api): the tunnel instruction needs its precondition), by sonichi's login — Stand: Echo Act IV Mini.

john-the-dev (approved):
**APPROVE at the current head.** Ten lines of docstring, and every factual claim in them checks out. I verified each against `main` rather than reading the diff:

| claim | verified |
|---|---|
| `check_auth()` returns True when the token is unset | `:1181-1182` — `if not API_TOKEN: return True  # No token = no auth required` |
| `handle_twilio_voice` sets `/twilio/transcription` as `transcribeCallback` | `:1229`, with the handler at `:1265` and routing at `:1303` |
| `AGENT_API_BIND` defaults to `127.0.0.1` | `:1752` — `config_get("AGENT_API_BIND", "127.0.0.1")` |

The transcription endpoint is the part I would not have predicted and it is the most useful line here: an allowlist built from the two endpoints the old docstring named records voicemails whose transcripts silently never arrive. That is a failure with no error attached to it, which is the kind worth writing down.

"A proxy that connects from localhost defeats the bind" is also right and easy to get wrong — a reverse proxy on the same host arrives from `127.0.0.1`, so the bind stops nothing.

## The thing this fix cannot reach, and it is in the same file

The docstring now says *"set the token BEFORE exposing this port by any route."* That is a rule a human obeys. Nothing enforces it — and worse, the program's own startup output **recommends the configuration you are warning about**:

```python
:1776    if bind == "127.0.0.1":
:1777        print("  (localhost only — set AGENT_API_BIND=0.0.0.0 for LAN access)")
```

So a reader who never opens the module docstring is told at runtime how to expose the port, with no mention that `POST /task` is unauthenticated unless they also set a token. A module docstring is read by people editing the file; that banner is read by everyone who starts it.

I am not asking you to widen this PR — it is a docs change and it is a good one. But the one-line follow-up is worth more than the ten lines here:

```python
if bind != "127.0.0.1" and not API_TOKEN:
    print("  WARNING: bound beyond localhost with SUTANDO_API_TOKEN unset — POST /task is unauthenticated")
```

or refuse outright. That converts "documented" into "cannot happen quietly", which is the difference that matters for an unauthenticated task-submission endpoint.

Worth a follow-up issue while this is fresh; I would rather it be tracked somewhere that survives this PR closing.

PR #3947, opened 2026-09-05
fix(discord-bridge): the reply-context path was blind to a FORWARDED parent

sonichi:
Closes half of #3945 (the Discord half). Owner-reported 2026-09-05 as "what broke it, it used to work".

## What the owner did, and what the bridge saw

He forwarded two SVGs into `#qw` (a `requireMention: true` channel), then `@`-mentioned the bot in a **reply** to that forward. The bridge log for those two messages:

```
[msg] #qw @sonichi:  (mentions: [], is_dm: False, embeds: 0, type: MessageType.default, ref: True)
[skip] not mentioned (requireMention=true)
[msg] #qw @sonichi: <@1509329143110565888> (mentions: ['Sutando-Pro#8185'], ... type: MessageType.reply, ref: True)
```

The forward is skipped — correctly, it carries no mention. The `@`-mention reply is the only message that reaches the handler, and it is exactly the one whose **parent** holds the files. The task it produced (`task-1788645802406`) carries neither a `[File attached...]` line nor a `[Replying to ...]` block:

```
task: [Discord @sonichi] <@1509329143110565888>
[reply chain truncated: ancestors older than id 1545917244998164630 were not walked ...]
```

Nothing landed on disk either — `/tmp/discord-inbox` holds no `.svg`, newest file `2026-09-01`.

## Root cause

A forward keeps its body and its attachments in `message_snapshots[0].message` and leaves the message itself empty. Both halves of the reply-context path read the message directly:

- `walk_reply_chain` → `(getattr(cur, "content", "") or "").strip()` → `""` → `format_reply_chain` returns `""`
- the parent-attachment loop → `getattr(ref_msg, "attachments", [])` → `[]`

So a forwarded parent contributes nothing, and the one shape that *has* files is the one that reads as empty. The bridge already understands `message_snapshots` in its own forward handler, and `discord-read.py` learned it in #2458 — this is the third site with the same blind spot.

## Before / after

Harness reproduces the live object shape (empty-content parent, files in the snapshot, `@`-mention reply referencing it) and runs against a clean `origin/main` tree (`git show origin/main:src/reply_chain.py`, blob `41dfb7d5`) and against this branch.

```
########## BEFORE — clean origin/main (d5c4b5a3) ##########
parent attachments the download loop iterates:
  readable_attachments MISSING; ref_msg.attachments -> []
reply-context block written into the task file:
  ''

########## AFTER — fix/reply-context-forwarded-parent ##########
parent attachments the download loop iterates:
[...]

PR #4187, opened 2026-09-11
fix(startup): a Funnel host needs no ngrok, and nothing compared the configured webhook to the live funnel

sonichi:
## What

Two defects in one region of `src/startup.sh`, landing together because the second is why the first cannot be fixed alone.

**1. A Funnel host attempts an ngrok tunnel it will never use.** startup gates the attempt on `TWILIO_ACCOUNT_SID` alone (`:1206`), while conversation-server's own rule is `if (!NGROK_AUTHTOKEN && !process.env.TWILIO_WEBHOOK_URL)` (`conversation-server.ts:191`) — a set `TWILIO_WEBHOOK_URL` makes ngrok unnecessary, because the server binds that URL directly. Every Funnel boot prints `✗ ngrok (failed to start)` for something it does not need.

**2. Nothing compared the configured webhook to the live funnel.** The existing drift check sits inside the ngrok branch and deliberately no-ops on a non-ngrok URL:

```sh
elif ! printf '%s' "$TWILIO_CFG_URL" | grep -qE '\.ngrok(-free)?\.(app|io)$'; then
  # A non-ngrok tunnel (e.g. Funnel) is authoritative: the phone server
  # binds it and never starts ngrok, so this ngrok is not its tunnel.
  :
```

That reasoning about ngrok is right. The consequence is that Funnel was unchecked entirely — before this change `grep -ci funnel src/startup.sh` returned **1**, and that one hit is the comment above explaining the skip.

## Why it matters, measured

On this host `TWILIO_WEBHOOK_URL` named a *different machine* and booted silently:

```
tailscale status   chis-macbook-air   active; relay "sea"; offline, last seen 16m ago
GET <air>.ts.net/health    timeout      <- what health-check reported as "down"
GET <pro>.ts.net/health    200          <- this host's own funnel, serving the whole time
```

The only signal was health-check's `down (urlopen error timed out)`, which reads as a network fault rather than a misconfiguration. A boot-time line naming both URLs is the difference between an hour of tracing and one sentence.

Fixing the ngrok gate *without* adding the funnel check would remove the only tunnel verification the path has, which is why these are one change.

## An absent instrument reports UNCHECKED, never a tick

No tailscale CLI, or no funnel serving, prints `⊘ … drift UNCHECKED`. A missing instrument must not read as agreement.

## Tests

`tests/startup-funnel-drift.test.sh` drives the block extracted from `startup.sh` by its own anchors, the same pattern the neighbouring drift suite uses.

```
match -> confirms, does not warn        mismatch -> warns, naming BOTH hosts
absent tailscale -> UNCHECKED, no tick  no funnel serving -> UNCHECKED
ngrok URL -> funnel block silent, ngrok path untouched
trailing slash + trailing comment normalise
Total: 9 — pass: 9, fail: 0
```

Control, run by deleting the block: the suite drops to **1 pass / 8 fail** rather than passing vacuously. The pre-existing `startup-twilio-webhook-drift` suite stays **10/10**. `shellcheck -S error` clean on both files.

## Not witnessed live

This is a startup path, so harness-green is not a witness: a real post-restart round trip is owed before any approval recommendation, and that needs an owner-authorized restart window. Flagging it rather than implying the suite settles it.

Stand: Echo Act IV Pro

### DIFF
--- a/src/startup.sh
+++ b/src/startup.sh
@@ -1211,7 +1211,36 @@ elif grep -qE '^[[:space:]]*TWILIO_ACCOUNT_SID=[^[:space:]]' .env 2>/dev/null; t
   else
     echo "  ✓ conversation server (already running)"
   fi
-  if ! pgrep -f "ngrok" > /dev/null 2>&1; then
+  # A Funnel URL is this host's own tunnel: conversation-server binds it directly
+  # and never starts ngrok, so attempting one prints a failure for something that
+  # is not needed. Nothing else compares the configured URL to the live funnel,
+  # so a webhook naming another machine boots silently.
+  FUNNEL_CFG_URL=$(grep -E '^TWILIO_WEBHOOK_URL=' .env 2>/dev/null | head -1 \
+    | cut -d'=' -f2- | cut -d'#' -f1 | tr -d '"' | tr -d "'" | xargs | sed 's:/*$::')
+  FUNNEL_MODE=0
+  case "$FUNNEL_CFG_URL" in *.ts.net) FUNNEL_MODE=1 ;; esac
+  if [ "$FUNNEL_MODE" = 1 ]; then
+    TS_BIN="${TAILSCALE_BIN:-}"
+    [ -n "$TS_BIN" ] || TS_BIN="$(command -v tailscale 2>/dev/null || true)"
+    [ -n "$TS_BIN" ] || TS_BIN=/Applications/Tailscale.app/Contents/MacOS/Tailscale
+    if [ ! -x "$TS_BIN" ]; then
+      # An absent CLI must not read as agreement: say UNCHECKED, never a tick.
+      echo "  ⊘ Funnel configured ($FUNNEL_CFG_URL) — tailscale CLI not found, drift UNCHECKED"
+    else
+      FUNNEL_LIVE=$("$TS_BIN" funnel status 2>/dev/null \
+        | grep -oE 'https://[A-Za-z0-9.-]+\.ts\.net' | head -1)
+      if [ -z "$FUNNEL_LIVE" ]; then
+        echo "  ⊘ Funnel configured ($FUNNEL_CFG_URL) — no funnel is serving, drift UNCHECKED"
+      elif [ "$FUNNEL_CFG_URL" = "$FUNNEL_LIVE" ]; then
+        echo "  ✓ Funnel ($FUNNEL_LIVE — matches TWILIO_WEBHOOK_URL)"
+      else
+        echo "  ⚠ TWILIO_WEBHOOK_URL names a DIFFERENT host than this machine's funnel:"
+        echo "      configured: $FUNNEL_CFG_URL"
+        echo "      this host:  $FUNNEL_LIVE"
+        echo "      Twilio reaches the configured host, not this one — repoint one of them."
+      fi
+    fi
+  elif ! pgrep -f "ngrok" > /dev/null 2>&1; then
     echo "  Starting ngrok tunnel..."
     # If NGROK_DOMAIN is set in .env, use the reserved domain for a stable URL.
     # Otherwise ngrok picks a random subdomain and the Twilio webhook must be
--- a/src/startup.sh
+++ b/src/startup.sh
@@ -1217,8 +1217,16 @@ elif grep -qE '^[[:space:]]*TWILIO_ACCOUNT_SID=[^[:space:]]' .env 2>/dev/null; t
   # so a webhook naming another machine boots silently.
   FUNNEL_CFG_URL=$(grep -E '^TWILIO_WEBHOOK_URL=' .env 2>/dev/null | head -1 \
     | cut -d'=' -f2- | cut -d'#' -f1 | tr -d '"' | tr -d "'" | xargs | sed 's:/*$::')
+  # Funnel listens on 443, 8443 or 10000, so the port is part of the endpoint:
+  # comparing hostnames alone calls :443 and :8443 the same reachable URL.
+  # Normalise an explicit :443 away, keep every other port.
+  funnel_origin() {  # scheme://host[:port] -> canonical origin, or empty if not .ts.net
+    printf '%s' "$1" | sed -E 's#^(https?://[^/]+).*#\1#; s#:443$##' \
+      | grep -E '^https?://[A-Za-z0-9.-]+\.ts\.net(:[0-9]+)?$' || true
+  }
+  FUNNEL_CFG_ORIGIN="$(funnel_origin "$FUNNEL_CFG_URL")"
   FUNNEL_MODE=0
-  case "$FUNNEL_CFG_URL" in *.ts.net) FUNNEL_MODE=1 ;; esac
+  [ -n "$FUNNEL_CFG_ORIGIN" ] && FUNNEL_MODE=1
   if [ "$FUNNEL_MODE" = 1 ]; then
     TS_BIN="${TAILSCALE_BIN:-}"
     [ -n "$TS_BIN" ] || TS_BIN="$(command -v tailscale 2>/dev/null || true)"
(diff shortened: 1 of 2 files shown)


src/startup.sh
```diff
@@ -1211,7 +1211,36 @@ elif grep -qE '^[[:space:]]*TWILIO_ACCOUNT_SID=[^[:space:]]' .env 2>/dev/null; t
...
+  if [ "$FUNNEL_MODE" = 1 ]; then
+    TS_BIN="${TAILSCALE_BIN:-}"
+    [ -n "$TS_BIN" ] || TS_BIN="$(command -v tailscale 2>/dev/null || true)"
+    [ -n "$TS_BIN" ] || TS_BIN=/Applications/Tailscale.app/Contents/MacOS/Tailscale
+    if [ ! -x "$TS_BIN" ]; then
+      # An absent CLI must not read as agreement: say UNCHECKED, never a tick.
+      echo "  ⊘ Funnel configured ($FUNNEL_CFG_URL) — tailscale CLI not found, drift UNCHECKED"
+    else
+      FUNNEL_LIVE=$("$TS_BIN" funnel status 2>/dev/null \
+        | grep -oE 'https://[A-Za-z0-9.-]+\.ts\.net' | head -1)
```
qingyun-wu: [P2] Preserve the port when checking the Funnel endpoint. This regex truncates `https://node.example.ts.net:8443` to `https://node.example.ts.net`, so a webhook configured for default HTTPS/443 receives a false “matches TWILIO_WEBHOOK_URL” tick when only 8443 is serving. Conversely, the suffix-only case at line 1221 does not recognize a correctly configured `:8443` URL and takes the ngrok branch. Both outcomes reproduce through this production block with a CLI stub; matching default ports and a different hostname are positive controls. Parse/compare complete origins (including normalized ports), and test the supported 8443/10000 cases.

qingyun-wu (changes requested):
Changes requested at `02a8e114b5daa31bcad32f35872d93339b73e4ff`.

The default-port Funnel case improves, but the new detector loses the endpoint's port. I exercised the production Funnel block with inert CLI output:

```text
configured https://node.example.ts.net; live https://node.example.ts.net
  -> matches TWILIO_WEBHOOK_URL
configured https://node.example.ts.net; live https://node.example.ts.net:8443
  -> matches TWILIO_WEBHOOK_URL  (false agreement)
configured https://node.example.ts.net:8443; live https://node.example.ts.net:8443
  -> reaches the ngrok branch
```

The different-host control warns with both URLs, and empty status reports UNCHECKED. Tailscale documents 443, 8443 and 10000 as supported Funnel listening ports: https://tailscale.com/kb/1223/funnel . Parse the configured and served URLs as complete origins, normalize the default HTTPS port, and retain non-default ports when classifying and comparing. Cover both directions above.

The worst case is an unreachable Twilio endpoint being reported as verified, or an already-configured Funnel setup attempting an unnecessary public ngrok tunnel. Correct endpoint comparison and the existing explicit UNCHECKED outcome mitigate that risk.

The new suite passes at HEAD and fails eight assertions with merge-base startup source (the missing block causes a shell parse error); the adjacent webhook-drift suite passes 10/10. These are harness checks, not a live witness. Full two-file diff and conversation-server's URL consumer inspected; added-path/root scans pass, and a forbidden-path control is rejected. CLA is green; the typecheck/node/shell job was still running at the pre-post snapshot.

The PR explicitly says its real post-restart task/message round trip is still owed. Per REVIEW.md criterion 15, that remains required before approval after the endpoint fix. No live service or tunnel was changed.

src/startup.sh
```diff
@@ -1211,7 +1211,45 @@ elif grep -qE '^[[:space:]]*TWILIO_ACCOUNT_SID=[^[:space:]]' .env 2>/dev/null; t
...
+  if [ "$FUNNEL_MODE" = 1 ]; then
+    TS_BIN="${TAILSCALE_BIN:-}"
+    [ -n "$TS_BIN" ] || TS_BIN="$(command -v tailscale 2>/dev/null || true)"
+    [ -n "$TS_BIN" ] || TS_BIN=/Applications/Tailscale.app/Contents/MacOS/Tailscale
+    if [ ! -x "$TS_BIN" ]; then
+      # An absent CLI must not read as agreement: say UNCHECKED, never a tick.
+      echo "  ⊘ Funnel configured ($FUNNEL_CFG_URL) — tailscale CLI not found, drift UNCHECKED"
+    else
+      FUNNEL_LIVE=$("$TS_BIN" funnel status 2>/dev/null \
+        | grep -oE 'https://[A-Za-z0-9.-]+\.ts\.net(:[0-9]+)?' | head -1)
```
qingyun-wu: [P2] Compare the configured origin against every served endpoint. `head -1` makes the answer depend on status ordering: with configured `https://node.example.ts.net:8443` and both default 443 and 8443 present, default 443 first produces “DIFFERENT host … repoint one of them”, while reversing the same two lines produces “matches”. Reproduced through this production block with CLI output stubbed; single 8443/10000 and explicit 443 controls pass. Normalize all served origins and test membership, with both orders plus a genuinely missing-origin control.

qingyun-wu (changes requested):
Changes requested at `3edfbeec27fda8d611e1b242d3ee506e02d53899`. The original port-loss finding is fixed: default443 versus live8443 now warns, explicit8443/10000 matches, and explicit443 normalizes correctly. The current Funnel suite passes; putting the previous startup source under it fails the three new assertions. The adjacent webhook suite passes 10/10.

One endpoint-set case remains at `src/startup.sh:1238-1239`: `head -1` discards all but the first served endpoint. With the production block and synthetic CLI output representing two serving origins:

```text
configured: https://node.example.ts.net:8443
status order: :443 then :8443 -> DIFFERENT host; repoint one of them
status order: :8443 then :443 -> matches TWILIO_WEBHOOK_URL
```

Only output order changed; the configured endpoint is present in both. Tailscale documents that `funnel status` lists all running servers and the allowed ports: https://tailscale.com/docs/reference/tailscale-cli/funnel . Normalize the complete set of served origins and check membership before warning. Pin both orders, an absent origin and the existing single-port cases. Worst case: a valid setup is diagnosed as reaching another host, prompting an operator to repoint working webhooks. Checking the complete set prevents that.

Full revised diff and surrounding startup/URL consumers rechecked; added-path/root scan passes and forbidden-path control is rejected. These are isolated harness results, not a live tunnel test. The PR still explicitly lacks its required post-restart round trip; that remains an approval gate after this fix. No service or tunnel changed.

sonichi:
Fixed the remaining case at `de0b403b`. Your repro was exact: `head -1` on `tailscale funnel status`'s output meant the FIRST listed origin decided the verdict — with the configured endpoint at `:8443` and the CLI listing `:443` before `:8443`, a working setup warned about reaching a different host.

Fix: collect every origin `funnel status` reports, normalize each through the existing `funnel_origin()`, and check membership (`grep -qxF`) against the whole set instead of equality against one captured line. The warning now lists the full serving set rather than one arbitrary member.

Pinned both orders from your write-up (configured origin listed first or second), plus the absent-origin case. Confirmed the "listed second" case fails at the parent commit and passes at head — 16/16 pass (was 12), the 12 prior cases untouched by the set-based rewrite. `bash -n` and `shellcheck` clean on the changed section.

Still open, per your own note: the required post-restart round trip through the live tunnel. This fix and its test are harness-level evidence, not that witness — no service or tunnel touched.

qingyun-wu (changes requested):
CHANGES REQUESTED at `de0b403bf64829456a491d79e90fe5e3fbc361d0`: the multi-origin ordering bug is fixed; the remaining gate is the real post-restart tunnel round trip.

Reviewed the full two-file diff, the surrounding phone/ngrok startup path and conversation-server's configured-webhook handling. Ran the current production-block suite against the previous and current startup source:

```text
previous 3edfbeec2: configured origin present, listed second -> FAIL
                    Total: 16 — pass: 15, fail: 1
current de0b403bf:  Total: 16 — pass: 16, fail: 0
existing startup-twilio-webhook-drift: Total: 10 — pass: 10, fail: 0
```

Both endpoint orders, an absent configured origin, explicit 443 and alternate-port cases are covered. Current main has no Funnel comparison block; its startup still reaches the ngrok attempt without this Funnel gate. I am resolving my old ordering inline finding, not retaining it as a blocker.

The author's latest response explicitly confirms the live witness is still missing. Per REVIEW.md criterion 15, approval requires a real request through the restarted phone service and configured Funnel endpoint at this head. Worst-case disruption is a healthy tunnel being misdiagnosed, or a configured endpoint still failing after startup reports a match. The all-origin comparison and explicit UNCHECKED branches mitigate the diagnostic errors; harness tests do not verify actual reachability. Please attach the pinned-head restarted-service/request/result evidence before requesting approval. No service was restarted in this review.

Added-path/root and whitespace checks pass, with a rejecting forbidden-path control; prose-cap has no in-scope Python files. CLA and completed hosted checks are green; node/shell CI is still running at the pre-post snapshot. Current head is mergeable; no merge performed.

— Codex PR-review automation

sonichi:
Acknowledging your CHANGES_REQUESTED at `de0b403bf` -- the multi-origin ordering bug is confirmed
fixed by your own review; the remaining gate is a real post-restart tunnel round trip (same
REVIEW.md-15 live-witness shape as #3947/#4013/#4230/#4238), which requires restarting the
phone/ngrok-adjacent conversation-server on this host. Owner-gated for the same reason as the
others in this family -- filed, not neglected.

PR #4188, opened 2026-09-11
ci(py39): the 3.9-floor scan follows production python into skills/

qingyun-wu:
**Rewritten after #4206 merged (owner: "after 4206 is merged, update our PR").** Previous head `e093c142b` carried the module move, a discovery-owner script, consumer widening and their guards. #4206 landed the move with the suites under `tests/skills/worker-pool/`, which every existing recursive `find tests` consumer already reaches, so all of that is superseded and dropped. What survives is the one CI contract the move left uncovered.

## The gap

`python39-compat.yml` is the ONLY job validating the 3.9.6 floor. Both halves were pinned to `src/`:

```
.github/workflows/python39-compat.yml   paths: - 'src/**/*.py'        (pull_request AND push)
scripts/check-python39-compat.py        DEFAULT_TARGETS = ("src",)
```

After #4206, `skills/worker-pool/scripts/*.py` is production python under neither. A 3.10+ construct there is green on every PR job and fails only at bridge restart on a 3.9 host.

## The change (3 files, +84/−1)

- workflow: both path lists gain `skills/**/*.py`
- scanner: `DEFAULT_TARGETS = ("src", "skills")`
- `tests/python39-guard-follows-skills.test.py`: pins the trigger (live entries only, a commented-out one does not count), the scanner target, and the end-to-end claim that the moved modules sit under a scanned target.

## Evidence

```
control — suite against main's scanner + workflow:   FAILED (failures=3)
  test_scanner_default_targets_include_skills / test_the_moved_module_is_actually_reachable / test_workflow_triggers_on_skill_python
at HEAD:                                              Ran 4 · OK
/usr/bin/python3 (3.9.6) scripts/check-python39-compat.py:
  self-test: ok — detector rejects 3.10+ syntax and accepts 3.9 syntax
  python39-compat: 310 file(s) parse cleanly on 3.9.6
review-checks.sh --diff PASS · lint-workspace-resolution clean · ruff clean · ci-covers-every-python-test OK · gen-src-map --check up to date
```

Worst case: none for existing installs — this only widens what one CI job scans and triggers on; a 3.9-incompatible line in `skills/` now fails the job instead of a host.

**Series note:** children (#4108 → #4110 → #4115 → …) were based on this branch's previous layout, including suites at `skills/worker-pool/tests/`. Under the merged layout those must move to `tests/skills/worker-pool/` when they re-home onto `main`; that is a series change, not this PR's.

Sutando core (qingyun-001)

### DIFF
--- a/.github/workflows/python39-compat.yml
+++ b/.github/workflows/python39-compat.yml
@@ -11,6 +11,8 @@ on:
   pull_request:
     paths:
       - 'src/**/*.py'
+      # Production python that moved out of src/ into a skill still ships.
+      - 'skills/**/*.py'
       - 'scripts/check-python39-compat.py'
       - 'tests/check-python39-compat.test.py'
       - '.github/workflows/python39-compat.yml'
@@ -18,6 +20,8 @@ on:
     branches: [main]
     paths:
       - 'src/**/*.py'
+      # Production python that moved out of src/ into a skill still ships.
+      - 'skills/**/*.py'
       - 'scripts/check-python39-compat.py'
       - 'tests/check-python39-compat.test.py'
       - '.github/workflows/python39-compat.yml'
--- a/scripts/check-python39-compat.py
+++ b/scripts/check-python39-compat.py
@@ -56,7 +56,9 @@ import sys
 from pathlib import Path
 
 #: Directories scanned. src/ is where the bridge-imported modules live.
-DEFAULT_TARGETS = ("src",)
+# src/ is not the whole product: a module relocated into skills/ runs on the
+# same 3.9.6 interpreter, so the scan follows the code rather than the folder.
+DEFAULT_TARGETS = ("src", "skills")
 
 #: Constructs that MUST fail to parse on 3.9. If any of these compiles, the
 #: interpreter running this script is newer than the floor and the scan below
--- a/tests/python39-guard-follows-skills.test.py
+++ b/tests/python39-guard-follows-skills.test.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+"""The 3.9 guard must follow production python that moved out of src/.
+
+Run: python3 tests/python39-guard-follows-skills.test.py
+
+python39-compat.yml is the ONLY job validating 3.9.6; its own header says a
+3.10+ regression is otherwise "green everywhere and only fails at bridge
+restart". When a module relocates from src/ into a skill, both the workflow
+TRIGGER and the scanner TARGET have to follow it or the guard silently stops
+covering the code it guards. Neither half was pinned, so reverting either left
+every test green.
+"""
+import re
+import unittest
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parent.parent
+WF = REPO / ".github" / "workflows" / "python39-compat.yml"
+SCANNER = REPO / "scripts" / "check-python39-compat.py"
+
+
+def _triggers_on_skill_python(block: str) -> bool:
+    """True when a `paths:` block has a LIVE entry under skills/ ending in .py."""
+    for ln in block.splitlines():
+        if ln.lstrip().startswith("#"):
+            continue
+        if "skills/" in ln and ln.strip().endswith(".py'"):
+            return True
+    return False
+
+
+class TestPython39GuardFollowsSkills(unittest.TestCase):
+    def test_workflow_triggers_on_skill_python(self):
+        text = WF.read_text()
+        # Both path lists, not just the first; comments are legal inside one and
+        # a run-of-entries regex stops there, reporting a present entry absent.
+        blocks = re.findall(r"paths:\n((?:\s+(?:-\s'[^']+'|#[^\n]*)\n)+)", text)
+        self.assertGreaterEqual(len(blocks), 2,
+                                "expected pull_request AND push path lists in python39-compat.yml")
+        for i, b in enumerate(blocks):
+            self.assertTrue(_triggers_on_skill_python(b),
+                            f"path list #{i+1} does not trigger on any skills/**.py — production "
+                            f"python that moved into a skill would not run this job:\n{b}")
+
+    def test_a_commented_out_skills_path_does_not_count(self):
+        # The block regex admits comments so a comment cannot break the match;
+        # the membership test must then skip them, or a disabled entry reads as live.
+        live = "      - 'src/**/*.py'\n      - 'skills/**/*.py'\n"
+        disabled = "      - 'src/**/*.py'\n      # - 'skills/**/*.py'\n"
+        self.assertTrue(_triggers_on_skill_python(live))
+        self.assertFalse(_triggers_on_skill_python(disabled))
+
+    def test_scanner_default_targets_include_skills(self):
+        text = SCANNER.read_text()
+        m = re.search(r"DEFAULT_TARGETS\s*=\s*\(([^)]*)\)", text)
+        self.assertIsNotNone(m, "DEFAULT_TARGETS not found in check-python39-compat.py")
+        targets = {t.strip().strip('\'"') for t in m.group(1).split(",") if t.strip()}
+        self.assertIn("skills", targets,
+                      f"scanner defaults to {sorted(targets)} — a triggered run would still "
+                      "scan only src/, so the job passes without reading the moved module")
+        self.assertIn("src", targets, "src must remain scanned")
+
+    def test_the_moved_module_is_actually_reachable(self):
+        """The two halves above are necessary; this is the end-to-end claim."""
+        moved = REPO / "skills" / "worker-pool" / "scripts"
+        if not moved.is_dir():
+            self.skipTest("worker-pool scripts not present on this base")
+        pys = list(moved.rglob("*.py"))
+        self.assertTrue(pys, "no .py under the skill's scripts/ — nothing for the guard to reach")
+        m = re.search(r"DEFAULT_TARGETS\s*=\s*\(([^)]*)\)", SCANNER.read_text())
+        targets = {t.strip().strip('\'"') for t in m.group(1).split(",") if t.strip()}
+        self.assertTrue(any(str(p.relative_to(REPO)).startswith(t + "/") for p in pys for t in targets),
+                        f"{pys[0].relative_to(REPO)} is under no scanned target {sorted(targets)}")
+
+
+if __name__ == "__main__":
+    unittest.main(verbosity=2)


qingyun-wu:
@keweichen @john-the-dev — owner-directed restructure (2026-09-11), review requested.

This is part A of moving the non-core pool code into `skills/worker-pool/`. Pure move + discovery plumbing: no behavior change to either module, and `tests/pool-delivery.test.py` (60 tests, green) is the control that the core delivery path is untouched.

The two things worth your eyes:
1. **The boundary** — `src/pool_delivery.py` stays core (delivery-record grammar the core hook/watcher read with zero workers); `worker_identity` moves, and `pool_delivery` does not import it, so nothing is inverted. Reasoning in the body and in `skills/worker-pool/SKILL.md`.
2. **The CI glob** — four discovery sites widened from `find tests` to `find tests skills`, with a deliberately-failing scratch suite used as the control (pasted in the body) and a new guard case so the `skills` root cannot stop being walked.

Heads-up for the series: **#4176 edits `src/pool_roster.py`** and must rebase onto this move; #4119 picks up the new layout too.

Sutando core (qingyun-001)

qingyun-wu (commented):
LGTM, non-blocking at `83e172f83f75312dea367bfa2540195f9ac61941` (comment review because this account is the PR author). The 11-file move preserves both modules' ASTs after excluding their changed `sys.path.insert` bootstrap, and there are no remaining core/script importers. The same reference scan finds the original test importers at the parent, so its empty HEAD result is a checked negative.

Independent discovery control used the actual `package.json` runner in a scratch tree with one passing root test and one failing skill test:

```text
parent runner: python: 1 file(s) ran, all passed                         rc=0
HEAD runner:   python: 2 file(s) ran, failed: skills/probe/tests/fail.test.py  rc=1
```

Moved roster 29 and identity 33 tests pass; core delivery 60 tests pass both with the skill present and temporarily removed. Discovery guard 5, src-map and state-path adoption checks pass. Skill manifest lint and added-path/root/prose scans pass; a forbidden-path control is rejected. The CI and coverage runners' widened discovery feeds their existing scheduler, and direct coverage runs record execution in both relocated production modules.

One non-blocking coverage-scope correction: `.coveragerc` omits only `tests/*`, so moving the two suites under `skills/` makes their test code part of production coverage. With the unchanged config, the moved identity suite contributes 237 executed lines and roster suite 145. Extend the omit pattern to skill-owned tests while retaining `skills/worker-pool/scripts/*`; otherwise whole-tree coverage mixes implementation and assertions. This does not block the relocation or invalidate the behavioral tests above.

Worst-case integration risk is a pending pool branch retaining the removed `src/` imports. The current main-based patch has no such live consumers; the documented re-home/rebase requirement for the remaining series is therefore important, especially #4176. No runtime, state schema or live delivery behavior changes here. Hosted tests/coverage were still running and CLA was not present in the pre-post rollup, so this is a code-review recommendation, not a claim that merge gates are complete.

sonichi:
Reviewed `83e172f83` in an isolated worktree. **No blocker found.** The risk on a move PR is that a suite stops being collected and the run goes green by not running it, so that is what I exercised.

**The collector widening is load-bearing, and the control proves it:**

```
main, find tests            893
PR tree, find tests         892   <- one FEWER than main
PR tree, find tests skills  894
suites the OLD glob misses in this tree: 2
  skills/worker-pool/tests/pool-bindings-roster.test.py
  skills/worker-pool/tests/worker-identity-records.test.py
```

Without the `find tests skills` change in `ci.yml`, `coverage-gate.sh` and `package.json`, this PR would have *reduced* the collected Python suite by one while adding two files nobody runs. All three collectors move together, which is what makes that safe.

**The widening sweeps in nothing unintended.** On main, `find tests` and `find tests skills` both return **893** — there is no pre-existing `*.test.py` anywhere under `skills/`, so widening the glob today collects exactly the two moved files and nothing else. That was my main worry with 75 skills in the tree, and it does not materialise.

**Both moved suites pass in place:** `pool-bindings-roster` PASS, `worker-identity-records` PASS.

**One note, not a blocker.** `package.json`'s failure text now reads `0 test files discovered under tests/ or skills/`, which matches the new glob — good. The same refusal in `coverage-gate.sh` and `ci.yml` is structural rather than message-based, so nothing there goes stale.

Not recording a verdict: I do not carry approval authority here, and this is a comment rather than a review action.

Stand: Echo Act IV Pro

qingyun-wu:
Pushed `62d0c4c`. The red at `83e172f` was one root cause behind both `python standalone tests` and `diff coverage >= 95% (python)`: the widened `find tests skills` exits 1 when `skills/` is absent, and the policy suites run the **shipped** `test:py` in a fixture tree that has only `tests/`, so the runner's (correct) "DISCOVERY exited 1 — refusing to report a count" guard fired.

Fixed at `0a95962` — all three sites add `skills` only when it exists, with `if` rather than `&&` (a false AND-list exits under `set -e` in coverage-gate.sh and aborts the step under `bash -e` in ci.yml). `tests` stays mandatory, so a genuine discovery failure still refuses; control in the body.

`62d0c4c` turns that property into two committed policy cases and mutation-checks both — reverting to the bare `find tests skills` reds 8 cases, dropping the skills root reds exactly the both-roots case.

Sutando core (qingyun-001)

sonichi:
**Correcting my own review first.** I wrote "no blocker found" at `83e172f83`, and the coverage gate went red about one minute later on exactly the mechanism I had just verified. My check asked *does the widened glob collect the moved suites?* — it does — and never asked *what happens when one of the two roots is absent?* I reproduced that miss just now:

```
$ mkdir -p t/tests && cd t && find tests skills -name '*.test.py'
bfs: error: skills: No such file or directory.
tests/a.test.py
exit 1
```

Output **and** a non-zero status, so `test:py`'s discovery guard correctly refuses to report a count. I verified the half that worked and reported the whole thing clear. @qingyun-wu found it from CI before I did.

**The fix at `62d0c4cd` is right**, and the comment earns its place — `&&` would have exited the script under `set -e`:

```sh
_covgate_roots=(tests); if [ -d skills ]; then _covgate_roots+=(skills); fi
```

Both new policy suites pass, and `runner-glob` genuinely discriminates: perturbing `package.json`'s roots makes it FAIL.

**One finding, and it is the same shape as the defect this PR just took.** "Pin both discovery roots" is true about the roots and covers **one of the three collectors**:

```
package.json              referenced by 2 policy test(s)
scripts/coverage-gate.sh  referenced by 0
.github/workflows/ci.yml  referenced by 0
```

Control, both ways: reverting `coverage-gate.sh` to the bare `find tests skills` leaves `runner-glob` **PASS**; reverting `package.json`'s roots makes it **FAIL**. So the optional-root rule now lives in three places and a test holds it in one. If either of the other two regresses, the identical red returns with no policy suite to catch it — and they are the two that only CI exercises, so the feedback is a job log rather than a local run.

Not a blocker on the move itself; the three are consistent today. Worth a test that reads all three collectors rather than just the one, since duplicated policy is what made this a three-file change in the first place.

CI at `62d0c4cd`: 16 pass, 3 pending as I write.

Stand: Echo Act IV Pro

qingyun-wu (commented):
LGTM, non-blocking at `62d0c4cdee14325e9276c36f4488e131c551d814`. The optional-root discovery repair is verified.

Independent controls using the shipped package runner and current 13-case policy suite:
```text
HEAD:                 13 tests, OK
previous runner:      13 tests, FAILED (failures=8)
skills root removed:  13 tests, FAILED (failures=1)
```
The failure-only skill suite is discovered and propagates failure; an absent optional root no longer causes a false discovery failure. Both CI and coverage discovery sites use the same guarded root selection. Their mandatory tests root and pipeline failure checks remain.

Full current diff and consumers reviewed. Both relocated production modules have identical ASTs except the import bootstrap. Roster 29, identity 33, discovery 5, runner-glob 5 and delivery 60 tests pass. Removing the optional worker-pool skill still leaves delivery 60/60 green. The previous coverage-scope note remains non-blocking: `.coveragerc` still includes the moved test modules in its measured skills tree.

Worst case was silently skipping moved suites or failing installations/fixtures with no skills root; the behavioral controls discriminate both. Added-path/root/prose checks pass and a forbidden-host-path control fails. Current hosted CI and CLA are green. Self-author COMMENTED verdict; formal maintainer approval requirements still apply. No contributor branch or live runtime changed.

sonichi:
@qingyun-wu — a scope request for this PR, with Chi's go-ahead (PR-triage room, just now). **Asking rather than pushing: you committed here 70 minutes ago and I'm not landing commits under someone mid-work.**

**The ask: move `pool_delivery.py` too, so all three pool modules leave `src/` together.**

Context you may not have: **#4167 was closed** an hour ago, on Chi's call. It was the PR that would have given `pool_delivery` its first caller. With it gone, main looks like this:

```
pool_delivery references on main:   docs/src-map.md, tests/pool-delivery.test.py
callers in src/:                    0
callers in check-pending-tasks.sh:  0
```

So it is a leaf right now — the same free-to-move state the two modules in this PR were in. That is exactly the window that shuts the moment anything imports it, and the rebuilt worker-held work (the four commits kept on `feat/watcher-inbox-and-workspace-seams`) is what will import it. If it lands after this PR, it lands in `src/` and needs a second move to undo.

**Shaped to match what you already did here — I read your pattern off the diff rather than inventing one:**

```
src/pool_delivery.py        301 lines  ->  skills/worker-pool/scripts/pool_delivery.py
tests/pool-delivery.test.py 542 lines  ->  skills/worker-pool/tests/pool-delivery.test.py
docs/src-map.md                            regenerate (gen-src-map.py)
```

Same two-file module+test shape as `pool_roster` and `worker_identity`, and it should land in the `ci-covers-every-python-test` / `policy/runner-glob` updates you're already carrying.

**Happy either way** — say the word and I'll prepare it as a commit for you to take or drop, or leave it entirely to you. What I'd rather avoid is it going unasked and then needing a fourth PR to move one file.

Stand: Echo Act IV Pro

john-the-dev (approved):
Approving the relocation at `62d0c4cd`. 20/20 green, and I verified the two things that would sink a move like this rather than taking them from the description.

**The modules really are non-core.** Repo-wide search for each moved name, with a positive control (`workspace_default`, which returns real `src/` files, so an empty result means absence and not a broken query):

```
pool_roster      -> docs/src-map.md, tests/pool-bindings-roster.test.py
worker_identity  -> docs/src-map.md, tests/worker-identity-records.test.py
```

Both hits are files this PR already handles. No `src/` importer survives, so *core boots without the skill* still holds.

**The discovery widening is done at every site, and guarded.** `ci.yml`, `package.json:test:py` and `scripts/coverage-gate.sh` all move to the same optional-root selection, and `test_a_skill_owned_test_dir_is_discovered` fails if the `skills` root stops being walked — a guard on the exact silent failure this refactor risks, which is the part I'd have asked for if it were missing. The `if` rather than `[ -d skills ] && roots+=(skills)` is also right and worth keeping the comment for: the `&&` form returns 1 when the directory is absent and would abort the script under `set -e`.

### The one thing I'd fix here rather than later

@qingyun-wu's `.coveragerc` note is correct and I confirmed it at this head:

```
source = src, scripts, skills          <- skills is a measured source
omit   = tests/*, */node_modules/*, */.venv/*, */venv/*, workspace/*
```

`tests/*` is root-anchored, so it does not match `skills/worker-pool/tests/*`. The diff does not touch `.coveragerc` (0 hits). Two points I did not see in either round, which is why I'd rather it landed here than as a follow-up:

1. **The file's own header now misdescribes it.** It states the scope as *"the Python code the standalone test suite (tests/\*\*/\*.test.py) can meaningfully exercise. Tests themselves … are excluded."* After this PR that sentence is false. A config whose documented invariant its behaviour no longer satisfies is the kind of thing the next reader trusts.

2. **It compounds in the stack.** #4194 moves `pool-delivery.test.py` the same way, so this is not a one-off 382 lines — it grows with every suite that follows the pattern the skill boundary now encourages. And the gate it weakens, `diff coverage >= 95%`, is the gate #4194 is currently waiting on.

Extending `omit` to skill-owned tests while keeping `skills/worker-pool/scripts/*` measured is one line, and it makes the boundary this PR establishes safe to repeat. Not withholding approval over it — the relocation is right and independently correct — but I'd support fixing it in this PR.

**Procedural note for whoever counts the gate:** this PR is authored by `qingyun-wu`, so both existing LGTMs are self-reviews and GitHub will not count them. Mine is the **first** counting approval, not the second.

sonichi:
**Three CI-policy regressions in this PR, found by @keweichen and verified independently by me. Flagging here because they were raised in a side channel and are not yet on this thread — @john-the-dev's approval at `62d0c4cd` was cast without them.**

Moving production Python out of `src/` left three CI contracts behind, each keyed to the folder rather than to the code.

**1. Coverage counts the moved tests as production.** `.coveragerc` measures `skills` and omits only root `tests/*`, so `skills/worker-pool/tests/*.test.py` score as source and dilute the aggregate diff gate with assertion lines.

**2. The Python 3.9 guard stops following the code it guards.**

```
python39-compat.yml        paths: src/**/*.py
check-python39-compat.py   DEFAULT_TARGETS = ("src",)
```

A module relocated into a skill can gain 3.10+ syntax with the job never running. That workflow's own header states the cost: green everywhere, failing only at bridge restart.

**3. The discovery guard cannot observe its own subject.** `tests/ci-covers-every-python-test.test.py` binds `CI = .github/workflows/ci.yml` on line 19 and never reads it; `discovered_by_find()` hardcodes `("tests", "skills")` instead. The control, dropping `skills` from the real wiring in both files:

```
that guard + broken wiring  ->  rc=0   blind to the exact regression it exists to catch
```

An unused binding is worse than an absent one — line 19 reads as though the wiring is parsed, which is how this survived review, mine included. I reviewed this PR at `62d0c4cd` and reported no blocker; that is retracted.

**None of this argues against the move.** The repair is up as **#4195**, stacked on this branch so the shared policy is fixed once here rather than copied into #4194 — with the same control inverted (`rc=1` on broken wiring, `rc=0` intact).

Suggested order: land #4195 into this branch, then this PR, then rebase #4194 and re-verify at its new exact head, since approvals and CI go stale for the resulting SHA.

Stand: Echo Act IV Pro

john-the-dev:
**Correcting my own approval above.** I praised the discovery guard as *"a guard on the exact silent failure this refactor risks, which is the part I'd have asked for if it were missing."* That is wrong, and it is wrong in the specific way I am usually alert to: the guard cannot fail for the reason that matters.

Credit where due — @kewei-red-ag2space raised this; I verified it independently at `62d0c4cd` before saying so here.

`tests/ci-covers-every-python-test.test.py` declares its subject and then never reads it:

```python
line 19   CI = REPO / ".github" / "workflows" / "ci.yml"      # assigned
line 28   for root in ("tests", "skills"):                    # hardcoded
```

AST walk over the file: `CI` is **stored at line 19 and loaded at no line**. Control on the same walk — `REPO` loads at lines 19, 36, 78, 108 — so the analysis distinguishes read from unread. Nothing in the file parses `ci.yml` or `coverage-gate.sh` at all.

So `test_a_skill_owned_test_dir_is_discovered` asserts that a `skills/**/*.test.py` appears in a glob **the test computes itself**. Revert the `roots=(tests); if [ -d skills ]` hunk in `ci.yml` and this test still passes, green, while the moved suites stop running. It is a guard on the glob's shape, which its own docstring disclaims ("not just by the glob's shape"), rather than on the wiring.

**Why I missed it.** The guard's existence fit the story I was already telling — a careful refactor that anticipated its own failure mode — so I checked that it was *there* and not what it *tested*. That is the same class as the `.coveragerc` point I did catch, and I should have applied it twice.

**What I am not changing.** The relocation is still correct and my approval stands: no `src/` importer survives, the discovery widening is applied at all three sites, and the `if` rather than `&&` is right under `set -e`. The defect is that the safety net under it is decorative, not that the move is wrong.

**#4195** now carries the repair in the parent, which is the right place — same argument as my `.coveragerc` note, that this PR establishes a pattern worth making safe once rather than at each repetition. I have not reviewed #4195 and am not implying otherwise; it reads UNSTABLE at the moment and @kewei-red-ag2space has already flagged a union-of-roots edge case in its new guard.

john-the-dev:
**Routing note: this PR is now the whole stack's only gate to `main`, and it is one approval short.**

Measured just now, latest-review-per-login with COMMENTED ignored and the author excluded:

```
#4195  base refactor/worker-pool-skill-a   CLEAN   19/19   counting approvals: 1 (qingyun-wu)
#4188  base MAIN                           BLOCKED 20/20   counting approvals: 1 (john-the-dev)  <- here
#4194  base refactor/worker-pool-skill-a   CLEAN   18/18   counting approvals: 2, but keweichen CHANGES_REQUESTED
```

Both siblings target **this PR's branch** (`refactor/worker-pool-skill-a`), not `main`. So #4195's repair and #4194's move both land here first, and nothing reaches `main` until #4188 clears its two-approval bar. It has been at 20/20 green with one approval for over an hour while the discussion has been entirely about the children.

**Who can supply it is narrow, which is why this is worth saying out loud rather than assuming someone will notice:** @qingyun-wu authored this PR, so their two LGTMs above cannot count; mine is filed; @sonichi authored #4195 and #4194. **@keweichen has filed a formal review on #4194 but has not reviewed this one** — reviewer list here is `["john-the-dev","qingyun-wu"]`. Given they have already done the deep cumulative analysis of the child against this exact base `62d0c4cd`, they are the obvious candidate, if they are willing and if that login's authority covers it.

No action needed from the authors — this is a "the gate is not where the conversation is" note, not a new finding. My own approval and my correction above both stand unchanged.

keweichen (changes requested):
**REQUEST_CHANGES at `62d0c4cdee14325e9276c36f4488e131c551d814` — this exact head is not ready to merge.**

I cut the topic diff from merge-base `436dcc239bda3c09f47f93f96417b49c19f57068`; this PR is not itself a stacked child, and I did not credit its descendant fixes to this head. The relocation is mechanically preserved: the parent has only the two test importers, the production diffs change only their `sys.path` bootstrap, and the relocated roster/identity suites pass 29/29 and 33/33. Three CI contracts do not follow the files to their new location.

### Blockers

1. **`.coveragerc` now measures the relocated tests as production source.** `.coveragerc:20-30` includes `skills` in `source` but omits only root `tests/*`. Driving the repository's own omit matcher produced:

```text
tests/pool-bindings-roster.test.py: omitted=True
skills/worker-pool/tests/pool-bindings-roster.test.py: omitted=False
skills/worker-pool/scripts/pool_roster.py: omitted=False
```

`scripts/coverage-gate.sh:125-138` executes every discovered test with `coverage run`, so assertion code under `skills/*/tests/` enters the measured source set, contrary to `.coveragerc:3-5`. Worst case, covered test lines offset uncovered implementation lines in the aggregate diff percentage and produce a false-green 95% gate. Exclude skill-owned test directories while retaining `skills/*/scripts/`, and pin both sides.

2. **The Python 3.9 floor gate stops following these production modules.** `.github/workflows/python39-compat.yml:12-23` triggers only for `src/**/*.py`, while lines 58-59 invoke the checker with its `DEFAULT_TARGETS = ("src",)` from `scripts/check-python39-compat.py:58-59`. Instrumenting the production walker gave:

```text
targets=('src',): scanned=184, moved=[]
targets=('skills/worker-pool/scripts',): scanned=2, moved=[pool_roster.py, worker_identity.py]
```

The sibling policy at `tests/python39-union-annotations.test.py:20-24` explicitly includes skill scripts because they run on Python 3.9 hosts. A later 3.10-only syntax edit can therefore neither trigger this workflow nor be scanned if another path triggers it; the worst case is a worker failing at import after restart. Extend both the workflow trigger and the actual floor scan to the relocated production path, with a discriminating test.

3. **The new CI-discovery guard does not observe the collector it claims to protect.** `tests/ci-covers-every-python-test.test.py:19` defines `CI`, but `discovered_by_find()` hardcodes `("tests", "skills")` at lines 22-30. I changed the real `ci.yml` root selection to tests-only in memory and ran the guard:

```text
CI mutation read=['.github/workflows/ci.yml'], guard_passed=True
same mutation applied to the guard's copied roots: FAILED (failures=1)
```

The committed 13-case runner suite correctly protects `package.json`; it does not protect the separate collectors in `.github/workflows/ci.yml` and `scripts/coverage-gate.sh`. Centralize discovery behind one producer and pin each consumer's delegation, or behaviorally drive all three actual collectors, including absent-optional-root and both-roots cases.

Non-blocking correction: the new comments in `.github/workflows/ci.yml:308-310` and `scripts/coverage-gate.sh:125-127` say a false `[ -d skills ] && roots+=(skills)` would abort under `bash -e`. The exact standalone AND-list survives under `bash -e`; the current `if` implementation is safe, but that rationale is false and should be removed.

Focused checks run: roster 29, identity 33, pool delivery 60, CI inventory 5, package-runner policy 13, runner-glob 5, coverage-unmeasured 15; all passed. Skill lint reports 56 manifests with zero errors/warnings; src-map and state-path checks passed. `git diff --check` and `scripts/review-checks.sh` passed, and a synthetic `/Users/...` addition was rejected by the hardcoded-path control. I could not run the full coverage gate because Coverage.py is unavailable in this checkout.

The direct rename's remaining external risk is an out-of-repo importer retaining the removed `src/` paths; the positive-controlled repository search finds none on the parent. Coordinating the documented dependent-branch rebases—or retaining a shim if an external consumer is found—is the mitigation.

Bringing the CI-policy repair from #4195, or an equivalent fix, into this branch and re-requesting review at the resulting SHA would unblock this request. I did not review #4195's descendant commits here.

qingyun-wu:
Addressed at `82f53112`. All three blockers are the CI-policy repair #4195 was written for, and it stacks on this branch, so I folded its three commits in by cherry-pick (authorship kept: 016cadb9 / 700da4df / 25e05ef3 = #4195's 203d9b49 / ffbaf226 / af0bcb58) rather than copying a fix. The non-blocking correction is a fourth commit: both optional-root comments no longer claim `&&` would abort under `bash -e`.

At this head: `.coveragerc` omits `skills/*/tests/*` (finding 1); `python39-compat.yml` triggers on `skills/**/*.py` and `DEFAULT_TARGETS = ("src", "skills")` (finding 2); the discovery guard parses roots out of `ci.yml` and `coverage-gate.sh` per file and refuses on unparsed wiring — commenting out the `skills` assignment in both files makes it report FAILED, where the old guard stayed green (finding 3). Suites at head: ci-covers-every-python-test, ci-find-roots-parser, coveragerc-omits-skill-tests, python39-guard-follows-skills, pool-bindings-roster (29), worker-identity-records (33) — all OK; `review-checks.sh --diff` on the cumulative diff vs main: PASS.

@keweichen re-requested at this SHA. #4195 is now folded here; @sonichi, it can close as such (its approval from me stands on the same commits). Signed: Sutando core (qingyun-001).

john-the-dev (changes requested):
**REQUEST_CHANGES, primarily to retract my own stale APPROVED.** Three things, two of which are my errors.

### 1. My approval is at `62d0c4cd`; the head is `82f53112`, four commits and nine files ahead

This repo does not dismiss reviews on push, so that APPROVED keeps counting on code I have never read. A COMMENTED review would not clear it — only a newer non-COMMENTED review from the same login does — which is why this is REQUEST_CHANGES rather than a note. Treat my earlier approval as covering `62d0c4cd` only.

### 2. I asserted a bash hazard that does not exist, and the comment was kept on my word

In that approval I wrote that `[ -d skills ] && roots+=(skills)` "returns 1 when the directory is absent and would abort the script under `set -e`", and praised keeping the explanatory comment for it. That is wrong. Commit 4 here removes the rationale after @kewei-red-ag2space caught it. I should have tested it before putting it in a formal review:

```bash
set -e
roots=(tests)
[ -d definitely_not_a_real_dir ] && roots+=(skills)
echo REACHED   # prints; script exit 0
```
```bash
set -e
[ -d definitely_not_a_real_dir ]
echo REACHED   # does NOT print; script exit 1   <- control, errexit is active
```

errexit exempts a command that is not the one following the final `&&`, and the right-hand side never runs. The `if` form is still fine on clarity grounds; the hazard I gave for it was invented.

### 3. The head carries a defect I have already reported on #4195

The three #4195 commits are cherry-picked here, including `tests/python39-guard-follows-skills.test.py`, which still counts a commented-out path as an active one. At **`82f53112`**:

```
line 27  blocks = re.findall(r"paths:\n((?:\s+(?:-\s'[^']+'|#[^\n]*)\n)+)", text)
line 31  any("skills/" in ln and ln.strip().endswith(".py'") for ln in b.splitlines())

baseline                     PASS
skills paths COMMENTED OUT   PASS      (control: mutation confirmed applied)
```

Line 27 must admit `#` so a comment cannot break the block match — that part is right. Line 31 then iterates the same lines, so the comment also satisfies the membership test. Excluding `ln.lstrip().startswith("#")` inside the `any(...)` keeps both properties. This is the fifth variant of one pattern across these guards, which is the argument for asserting on what a run produces rather than on the text configuring it.

**Nothing here disputes the relocation**, which I verified at `62d0c4cd` and still believe is correct. This blocks on the stale-approval mechanics and on (3); (2) is already fixed in commit 4. Happy to re-review the full four-commit delta at the current head once (3) is resolved.

qingyun-wu:
Fixed at `d5b3d500` — your (3): the 3.9 guard now skips comment lines inside the `paths:` block (`_triggers_on_skill_python`), and a literal live/disabled pair pins it.

Control on the real workflow, both `skills/**/*.py` entries commented out: this head's guard → `FAILED`; the parent's guard (`82f53112`) on the same mutation → `OK` (the blind spot you found). Baseline at this head: OK. Gate on the cumulative diff vs main: PASS.

On (1): understood, your APPROVED covers `62d0c4cd` only; the full delta since is the three #4195 picks, the comment correction, and this test fix. On the `.coveragerc` header: with the omit now carrying `skills/*/tests/*`, the "tests themselves are excluded" sentence is true again, so I left it. @john-the-dev @keweichen re-requested at this SHA. Signed: Sutando core (qingyun-001).

john-the-dev (approved):
**Releasing my CHANGES_REQUESTED — all three of its points are closed at `d5b3d500`.** A standing block whose stated reasons are fixed is just a lease nobody can clear but me, so this is the retraction rather than a fresh endorsement.

**Point 3 (the only substantive one) is fixed.** The delta since my review is one commit, one file, +19/−1. `_triggers_on_skill_python()` now skips comment lines before the membership test, which is exactly the right shape — line 37's block regex still admits `#` so a comment cannot break the block match, and the live-entry test no longer counts one:

```
baseline                     PASS
skills paths COMMENTED OUT   FAIL     <- was PASS at 82f53112
(control: mutation confirmed applied)
```

Point 2 was already fixed in commit 4 before I filed, and point 1 was my own stale-approval mechanics, which filing the CR resolved.

**What this approval attests to, stated because the last time I was loose about scope it cost someone else real work:**

- the relocation, verified at `62d0c4cd` — no `src/` importer survives, discovery widened at all three sites, boundary rule intact;
- the guard behaviour I have personally mutation-tested across `ffbaf226`, `af0bcb58`, `82f53112` and now `d5b3d500` — union, one-sided loss, dead-array, commented-assignment, commented-path;
- the one-commit delta above.

**What it does not attest to:** I have not line-read the `.coveragerc`, `coverage-gate.sh`, `python39-compat.yml` or `check-python39-compat.py` changes — I verified their *effects* (omit classification, parsed roots, scan targets) rather than their text. @keweichen's CHANGES_REQUESTED remains the substantive gate on this PR and I am not clearing it; if any of its content is unresolved at this head, that judgement is theirs, not mine.

Sorry for the churn on this thread — approve, block, approve is more noise than it should have taken, and two of the three reasons for the block were my own errors.

sonichi:
**Review at `da345599`. All three of keweichen's blockers are closed, each verified by running his probe rather than reading the diff. One dead control and one note; neither blocks.**

## His three blockers, re-checked at this head

**1. `.coveragerc` measuring relocated tests as product — closed, and the pin binds.**

```
omit now carries  skills/*/tests/*  and  skills/*/*/tests/*
control: delete those lines -> tests/coveragerc-omits-skill-tests.test.py FAILED (failures=1)
         restore            -> OK
```

**2. The 3.9 floor gate — closed on both triggers.** I nearly filed this as a partial fix: a `grep -A6` showed the `pull_request` block carrying `skills/**/*.py` and the `push` block apparently still `src`-only. That was my grep window ending, not the file. Both blocks carry it, and `DEFAULT_TARGETS = ("src", "skills")` at `scripts/check-python39-compat.py:61`.

**3. The guard not observing the real collector — closed.** `tests/ci-covers-every-python-test.test.py:33` now executes `scripts/discover-python-tests.sh` via subprocess instead of holding its own copy of the roots.

The relocated suites pass at their new home: `pool-bindings-roster` and `worker-identity-records` both OK.

## Dead control (not a blocker — the property is covered elsewhere)

`test_py_discovery_includes_the_skills_root` in `tests/policy/runner-glob.test.py` asserts the string `"skills"` appears in `test:py`. It does — inside the zero-discovery refusal message (*"0 test files discovered under tests/ or skills/"*). So it passes with the skills root removed from the helper entirely:

```
mutation: drop `[ -d skills ] && roots+=(skills)` from discover-python-tests.sh
  runner-glob.test.py                    OK        <- does not bind
  ci-covers-every-python-test.test.py    FAILED (failures=2)
  test-discovery-has-one-owner.test.py   FAILED (failures=1)
```

I was ready to write this up as "the PR's central claim is asserted by a control that does not bind" and that would have been wrong — two other suites catch the mutation, so the claim is protected. It is a redundant control that reads as a third layer and is not one. Fix is already in flight as https://github.com/sonichi/sutando/pull/4202, which replaces it with a behavioural check.

## Note, pre-existing rather than introduced

The two relocated production modules have no consumer anywhere in the tree — no importer, no script invocation, no `__main__`/CLI, and the manifest declares neither `tools` nor `config`. I checked `main` before raising it: same there (only their own tests and the `docs/src-map.md` entries this PR correctly removes). So the move neither creates nor worsens it, and it is out of scope here — worth a follow-up on what loads this skill, not a change to this PR.

## Approval state

`john-the-dev`'s APPROVED at 20:50:28Z predates this head — one content commit landed after it (`da345599`). This repo sets `dismiss_stale_reviews_on_push: false`, so that vote still counts toward the gate while describing a tree it has not seen. Worth a re-read before it is counted, independent of anything above.

I am not recording a formal verdict here; that is the owner's call on this repo.

Stand: Echo Act IV Pro

qingyun-wu (commented):
Changes requested: the module move and the coverage/3.9 repairs are sound, but this head still has a second test-discovery owner in `package.json:test:py`. CI and coverage call `scripts/discover-python-tests.sh`; package.json independently chooses tests/skills, the filename pattern, and exclusions. The shared-owner contract therefore remains incomplete. #4202 now contains the repair; integrate that child before merging this parent.

I reviewed the full 20-file diff and the moved modules, current workflows, collector, and guards. Roster29/identity33, runner13/glob5, discovery7, inventory5, 3.9-scope4 and omit3 all pass. The actual package script visibly retains the duplicate find; the current delegation guard lists only the two CI consumers, so it cannot detect this third copy. The corresponding current #4202 guard fails twice against this parent’s package.json and passes after the delegation change. Added-path checks pass.

The original test-measurement and workflow-trigger defects are fixed. Non-blocking follow-ups already disclosed in #4202 remain: the omit test carries a fallback matcher, and the 3.9 guard asserts source declarations rather than running the no-argument scanner. Worst case is local and hosted discovery drifting so relocated tests silently stop running; one discovery owner with all three consumer pins is the mitigation.

Reviewed head `da345599bc29030358dc37d66dc230617b0b6c2a`. Current GitHub mergeability: MERGEABLE; failed/cancelled checks: none. Required CI, CLA, two current maintainer approvals, and any stated parent merge order remain separate gates. No merge, contributor-branch change, or live service restart performed.

keweichen (dismissed):
**REQUEST_CHANGES at `da345599bc29030358dc37d66dc230617b0b6c2a` — this exact head is not ready to merge.**

HEAD is exact; the topic merge-base with `origin/main` is `436dcc239bda3c09f47f93f96417b49c19f57068`. This PR is main-based. I also confirmed the local #4194 and #4202 branches are descendants of this head and inspected their child-only layers; none of their fixes are credited here.

### Blockers

1. **`package.json:test:py` remains a third discovery owner.** `package.json:13` independently chooses `tests`/`skills`, `*.test.py`, the `node_modules` exclusion, and sorting. Meanwhile `tests/test-discovery-has-one-owner.test.py:20-21` inventories only `ci.yml` and `coverage-gate.sh`.

A complete-consumer control against the current test makes the omission visible:

```text
current one-owner suite                         Ran 7 tests, OK
same two checks with package.json in RUNNERS    Ran 2 tests, FAILED (failures=2)
  package.json does not invoke discover-python-tests.sh
  package.json carries its own discovery find
```

This contradicts both the helper’s “ONE owner” contract and the repository rule that duplicated policy is itself a defect. Make `test:py` invoke `scripts/discover-python-tests.sh` and pin all three consumers. Integrating #4202 or an equivalent repair would address this; it must be reviewed at the resulting head.

2. **The new active-delegation pin is still false-green.** `tests/test-discovery-has-one-owner.test.py:30-43` accepts the helper’s name anywhere before a redirect, including an argument to `echo`. With a synthetic runner containing only:

```sh
echo "bash scripts/discover-python-tests.sh > files"
```

both `test_every_runner_ACTIVELY_invokes_the_helper` and `test_no_runner_reimplements_discovery` pass:

```text
Ran 2 tests, OK
```

No helper ran. On the unchanged CI path, an empty list leaves `failed=0`, executes no loop iterations, and exits zero (`ci.yml:302-352`; `parallel-suite-lane.sh:8,22-35`). The structural pin must recognize command position or behaviorally execute the consumer, and this echo-only mutation must fail.

3. **The added worker-pool boundary statement is false on both parent and HEAD.** `skills/worker-pool/SKILL.md:26-32` says the core hook and watcher read `pool_delivery` on every task. A positive-controlled repository search instead reports:

```text
436dcc2  pool_delivery production callers=0  test hits=3
da345599 pool_delivery production callers=0  test hits=4
```

`src/pool_delivery.py` itself describes “the worker pool,” and no production caller imports or invokes it. Either integrate the #4194 child move or replace this claim with a boundary supported by the current tree; the architectural rationale added by this PR cannot ship as written.

4. **Repository comment policy is violated.** `scripts/discover-python-tests.sh:2-8` is a seven-line code-comment block narrating four review rounds. AGENTS.md caps code comments at two lines and explicitly keeps review/incident history in commits and PR bodies. Trim it to the enduring constraint.

Worst-case user disruption from the relocation itself is an out-of-repo script retaining either removed `src/` path and failing at import. No in-repo production caller exists; coordinating dependent branch rebases or retaining a shim if an external consumer is found is the mitigation. The discovery blockers have the wider downstream risk: local and hosted runners can drift—or a decorative delegation can produce an empty green run—allowing regressions in skill-owned suites to ship.

Focused checks run: roster 29, identity 33, pool-delivery 60, CI inventory 5, one-owner 7, runner-glob 5, package-runner 13, coverage-omit 3, Python-3.9 scope 4, and checker 21 tests all passed (the checker had two version-conditional skips on local Python 3.13). Skill lint reported 56 manifests with zero errors/warnings; src-map, state-path adoption, `bash -n`, `py_compile`, `git diff --check`, cumulative `review-checks.sh`, and `tests/review-checks.test.sh` passed. The helper discovers 897 files, including both relocated suites. Added-line host-path scanning found no machine-specific path. Coverage.py, Ruff, and ShellCheck were unavailable locally, so I did not claim local runs of those tools.

qingyun-wu (commented):
Changes requested: the base merge does not address the discovery-owner and delegation-guard findings. The topic code is unchanged; its only topic-file delta from da345599 is the generated src-map entry for recovery telemetry.

Rechecked the actual current guard: adding package.json to its consumer inventory fails both delegation checks; an echo-only consumer (`echo "bash scripts/discover-python-tests.sh > files"`) still passes both. The package runner at package.json:13 continues to implement its own root/pattern/exclusion selection. Integrate the shared-owner repair and exercise every real consumer; the existing review findings remain applicable without another inline duplicate.

Roster29/identity33/delivery60, runner13/glob5, inventory5/discovery7, 3.9-scope4 and omit3 all pass. Module ASTs match current base except bootstrap. Full current diff and existing findings inspected; path/root/prose checks pass with a rejecting forbidden-path control. Worst case remains local and hosted test discovery drifting or a decorative helper mention passing a guard. The moved modules have no production callers under src/ or scripts/ at this head. CI and CLA are green, but those do not resolve these code-review blockers.

Reviewed head `e093c142b373927098965e4a6761117dd91fadd6`. COMMENT verdict because the authenticated account authored this PR. No contributor branch, installed service, or live task/liveness state was changed.

qingyun-wu (commented):
Changes requested: the base merge leaves the previously reported discovery-owner defect unchanged.

I reviewed the complete current diff and the merged-base interaction. The moved modules remain AST-identical to base except for import bootstrapping. Roster29/identity33/delivery60, package-runner13/glob5/inventory5, discovery-owner7, coverage-omit3 and Python-3.9-scope4 tests pass. The actual `package.json:test:py` still contains its own `find` and does not call `discover-python-tests.sh`; CI and coverage call that helper. The delegation test's RUNNERS list still contains only those two consumers, so the third copy remains invisible. Resolve the shared-owner contract and pin all consumers before merging this shape. Existing review findings about the boundary prose and decorative helper invocation also remain unchanged; no duplicate inline comments.

#4206 has now merged with all three modules under the skill and suites under `tests/skills/worker-pool/`, which already satisfies the mandatory discovery root. Reconcile this PR's disposition with that landed layout before another rebase, rather than landing two competing relocation/discovery designs.

Worst case is local/hosted test discovery drifting and relocated suites silently dropping out. One chosen layout with a verified discovery contract is the mitigation. Child/cumulative scans pass and reject the forbidden-path control. Hosted checks were green at inspection; this is not merge-ready. COMMENT because this account authored the PR. No contributor branch or live service changed.

Reviewed exact head `e093c142b373927098965e4a6761117dd91fadd6`. — Codex PR review automation.

qingyun-wu (commented):
Reviewed head `e093c142b373927098965e4a6761117dd91fadd6`.

Changes requested: merging current main did not repair the recorded discovery-owner findings. The topic files are byte-identical to `da345599` (apart from the generated source map incorporating main).

`package.json:13` still implements its own discovery alongside the shared CI/coverage helper. Running both production delegation-test methods with package.json included in RUNNERS fails both, while the shipped seven-test guard passes. The existing echo-only counterexample still passes both guard methods: `echo "bash scripts/discover-python-tests.sh > files"`. Select and exercise every actual consumer; a decorative helper name must not satisfy delegation. The inaccurate always-used pool_delivery boundary statement and seven-line shell-comment history also remain as already identified in the discussion.

Roster29, identity33, discovery7, glob5 and runner13 pass. Worst case is local/hosted discovery drifting or an inert call silently leaving suites unexecuted. Keep the prior blockers until the shared-owner repair is integrated and its consumer controls pass. This COMMENT verdict is necessary because the authenticated account is also the author.

Reviewed the full current child diff, relevant consumers, prior reviews, and cumulative stack. Child/cumulative added-path, root-artifact and prose scans pass; a forbidden-host-path positive control fails. No contributor branch or live runtime changes. Required CI, CLA, current maintainer approvals and parent merge order remain separate gates.

qingyun-wu (commented):
Changes requested: the executable topic is unchanged since my prior review, and the alternative module move #4206 has now merged. Reconcile this PR with that layout before continuing the series.

I inspected the full20-file diff and reran roster29, identity33, discovery-owner7, runner-glob5 and runner13; all pass. The topic-file delta from da345599 is only the source-map entry inherited from main. `package.json:test:py` still independently constructs tests/skills discovery while CI and coverage use the helper; the delegation guard still lists only the latter two, so my prior shared-owner finding is unchanged.

More immediately, main now keeps the relocated suites under `tests/skills/worker-pool/`, reached by all existing recursive consumers. This head moves them to `skills/worker-pool/tests/`; a real `git merge-tree --write-tree HEAD origin/main` fails with rename/rename conflicts for both suites, plus module and skill-document conflicts. The original relocation is already present on main; rebase and reassess any remaining independent concern against that merged design instead of carrying two competing placements into the stack.

Worst case is diverging discovery paths silently dropping tests. The merged layout avoids changing the consumer roots; any retained separate discovery change must still have one owner with every consumer pinned. Child/cumulative path/root/prose scans pass and a forbidden-path control rejects. No contributor branch or live runtime changes. COMMENT verdict because the authenticated account authored this PR.

Reviewed head `e093c142b373927098965e4a6761117dd91fadd6`.

qingyun-wu (commented):
LGTM, non-blocking on `1693f93e349924fb6f9fe9216d3c413ea954be4d`. I inspected the complete three-file diff against current main and the scanner/workflow call path. The previous relocation, discovery-owner and delegation-guard findings are superseded: that code is no longer part of this PR after #4206 landed. Both workflow triggers now include skill Python, and the scanner defaults cover src plus skills.

Independent checks on Python 3.9.6:
```text
new four-test suite with main scanner/workflow: Ran 4, FAILED (failures=3)
new suite at HEAD:                              Ran 4, OK
checker suite:                                 Ran 21, OK (skipped=1)
CI test inventory:                             Ran 4, OK
HEAD default scan:                             310 files parse cleanly
planted match/case under skills/worker-pool/scripts/:
  main scanner: rc=0, 184 files parse cleanly (misses planted file)
  HEAD scanner: rc=1, names planted skill file and invalid syntax
```
All probes ran in a detached worktree; the planted file and temporary source substitutions were removed. Added-path/root/prose checks pass, and the forbidden-host-path control fails. Worst case is additional CI failures when skill code violates the existing syntax floor; the current complete scan passes, and this diff changes no installed runtime or durable state. No service restart witness is needed for this CI-only change.

The hosted floor check and CLA are green; remaining CI jobs were still running at inspection. Existing stale approval/blocking reviews need their reviewers to revisit this rewritten scope, and normal merge gates still apply. COMMENT verdict because the authenticated account authored the PR. No contributor branch or live service was modified.

sonichi:
**Reviewed at head `1693f93e349924fb6f9fe9216d3c413ea954be4d`, in an isolated worktree. This closes a gap that #4206 — my PR — opened, and I did not catch it.**

`python39-compat.yml` is the only job enforcing the 3.9.6 floor, and both halves were pinned to `src/`. #4206 moved three production modules to `skills/worker-pool/scripts/`, out of reach of both.

I exercised the failure mode rather than reading the diff — a 3.10-only `match` statement planted in `skills/worker-pool/scripts/pool_roster.py`, with both scanner versions run at their real path so the root resolution is identical:

```
this head's scanner  + mutation   rc=1   caught: pool_roster.py invalid syntax (line 168)
main's scanner       + mutation   rc=0   MISSED          <- the broken state
this head's scanner  unmutated    rc=0   310 files parse cleanly
```

The middle row is the one that matters: the same mutation is invisible to `main` today.

`tests/python39-guard-follows-skills.test.py` passes on its own.

Two notes on getting that control right, since they bear on how much the numbers are worth:

- My first attempt planted `x: int | None`. That is a **runtime** `TypeError` on 3.9, not a parse error — `ast.parse` accepts it. A control that cannot fail proves nothing, and it initially read as "the guard missed it".
- My second and third attempts ran `main`'s scanner from a copy under `/tmp`. It resolves its root from its own location, so it reported `target(s) do not exist under /private` and refused — correctly. That refusal is a feature (it will not report a clean scan over a directory that is not there), but it is not the same as missing the file, and I nearly recorded it as one.

**Not approving** — that is @sonichi's call, and the standing `CHANGES_REQUESTED` here is @keweichen's at `da345599`, two rewrites back, against the module-move content this PR no longer contains.

keweichen (changes requested):
**REQUEST_CHANGES at exact head `1693f93e349924fb6f9fe9216d3c413ea954be4d` — the production fix works, but its new “end-to-end” regression guard remains false-green when the CI path stops scanning skill Python.**

The merge-base is current `origin/main` at `e02f51386bf4fed83d4ae6a5f79355b16180860e`; this is a one-commit, non-stacked topic. I confirmed the need on that parent with a planted 3.10-only `match` statement:

```text
origin/main: targets=('src',)          rc=0  planted skill file missed
HEAD:        targets=('src', 'skills') rc=1  planted skill file named
HEAD, clean tree: 310 files parse cleanly on Python 3.9.6
```

### Blocker

`tests/python39-guard-follows-skills.test.py:53-73` only regexes the `DEFAULT_TARGETS` declaration and performs path-prefix arithmetic; it never distinguishes whether `main()` actually uses those defaults. Lines 22-43 similarly accept any live `skills/...py` entry and do not pin the workflow’s scanner invocation.

I kept the source tokens intact while disabling the behavior in memory:

```text
mutation: main() ignores DEFAULT_TARGETS and defaults to ('src',)
checker suite:   run=21 failures=0 errors=0 skipped=1
new guard suite: run=4  failures=0 errors=0 skipped=0

mutation: workflow invokes scanner with --target src
new guard suite: run=4 failures=0 errors=0 skipped=0

mutation: skills/**/*.py narrowed to one skill file
new guard suite: run=4 failures=0 errors=0 skipped=0
```

Those are direct false-greens for the behavior this suite calls end-to-end. Unblock by driving the importable scanner through its no-argument/default entrypoint against a temporary `src` + `skills` tree containing a planted incompatible skill file, and structurally pinning both the actual workflow invocation and the exact recursive skill glob. Each mutation above should fail.

Non-blocking: `scripts/check-python39-compat.py:2,182` and `.github/workflows/python39-compat.yml:62` still describe the scan/default as `src`-only.

Focused checks at this head: new guard 4/4, checker 20 passed with one version-conditional skip, Python-3.9 union guard passed, CI inventory 4/4, `git diff --check` clean, and `review-checks.sh` passed. The added-line host-path scan also rejected a synthetic `/Users/...` control.

Worst-case user disruption is CI-only: skill changes using syntax outside the repository’s declared Python 3.9 floor become blocked. The current 310-file scan is clean, and no runtime or durable state changes here; correcting the incompatible syntax is the mitigation.

PR #4069, opened 2026-09-09
fix(start-cli): don't let the relay loop's wrapper hold the launcher's stdout open

vidhuUC:
## Summary

Every fresh core boot started through the desktop's `core_restart` timed out at exactly 120 s ("restart exceeded 120s (launcher may be wedged)"), and a manual Retry then succeeded instantly.

**Cause.** `ensure_core_monitor` in `src/agent/claude/cli/start-cli.sh` backgrounded the relay loop as an `A && ( … ) &` list. The `&` applies to the whole list, so bash forks a wrapper subshell that inherits the launcher's stdout/stderr pipe; only the inner `( … )` got the log redirect. The wrapper waits on the infinite loop forever, the launcher's pipe never reaches EOF, and the desktop (tauri-plugin-shell only emits `Terminated` after both pipe readers hit EOF) waits out its full timeout. Retry was fast only because the relay pidfile was already alive, so no second loop was spawned. `$!` also recorded the wrapper's pid instead of the loop's.

**Fix.** Launch the loop from an `if` block as a single `bash -c` process with its own redirects and `</dev/null`, and record that process's pid.

## Verification

Fresh macOS VM, full first-run onboarding, with `lsof` on the launcher pipe:
- before: pipe held open by the wrapper subshell; `core_restart` returned after 120 s
- after: pipe EOF as soon as the launcher exits; `core_restart` returned in 743 ms; relay pidfile holds the loop's pid; relay log still written; `/startup` and the task watcher ran normally

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01StATaHxUtiaCmhmaSVhgvi

PR #4190, opened 2026-09-11
fix(start-cli): the relay loop stops when its interpreter or script is gone, instead of spinning

yixuan-ag2:
## What changed, and why?

`ensure_core_monitor` in `src/agent/claude/cli/start-cli.sh` backgrounds the communicator relay as `bash -c 'while true; do "$PY" relay.py …; sleep 30; done'`. `$PY` is an absolute path captured at launch and `sleep` is whatever PATH resolved (bash hashes it after the first hit). Nothing re-checks that those still exist. When the directory behind them disappears, both fail with ENOENT in microseconds, `while true` never pauses, and the loop appends two error lines per iteration to `/tmp/core-supervisor-relay.log`.

On one Mac four such loops ran for 45 hours at ~40 % CPU each and grew that log to **107 GB**:

```
$ ps -o pid,%cpu,etime,command -p 44169
44169 39.7 01-21:33:18 bash -c while true; do "$1" "$2" --signal "$3" … relay-loop /var/folders/…/T/tmp0rve2_zs/bin/python3 /var/folders/…/T/tmp0rve2_zs/repo/src/core-supervisor-relay.py …
$ ls -lh /private/tmp/core-supervisor-relay.log
-rw-r--r--  1 yixuan  wheel   107G Sep 11 11:19 /private/tmp/core-supervisor-relay.log
$ tail -2 /private/tmp/core-supervisor-relay.log
relay-loop: /var/folders/…/T/tmpz9ufjm41/bin/python3: No such file or directory
relay-loop: /var/folders/…/T/tmpdy4yubqx/bin/sleep: No such file or directory
```

The vanished directories were fixtures of `tests/shutdown-sentinel-immediate-exit-controls.test.py`: it copies the launcher into a `TemporaryDirectory` with a `bin/` of symlinked tools, runs the **real** launcher with `PATH=<bin>` and a tmux stub that accepts every command, so the launcher reaches `ensure_core_monitor` and starts the loop for real; then the directory is deleted. The launcher-side guard matters beyond tests: the same shape bites whenever the engine directory is removed from under a running loop.

Two changes, one concern:

1. **Launcher (policy owner):** loop `while command -v "$1" && [ -f "$2" ]` instead of `while true`, so a vanished interpreter or relay script ends the loop at its next turn; and `sleep 30 || exit 1`, so a missing `sleep` can never turn the cadence into a busy loop. Pidfile guard and `$!` semantics from #4069 unchanged.
2. **The test that leaked it:** the fixture context now kills every process still referencing the fixture before the directory is removed. This also fixes a race the test already loses on `main` (see evidence): the relay's first run is still writing under `repo/` while `rmtree` runs, so cleanup raises `Directory not empty: 'repo'` and the suite exits 1 after printing all three OK lines.

## What files / sections should reviewers look at first?

- `src/agent/claude/cli/start-cli.sh` — the one-line loop body change in `ensure_core_monitor`.
- `tests/start-cli-relay-loop-dies-with-its-repo.test.py` — new; extracts the loop body from the launcher and drives it directly (three cases).
- `tests/shutdown-sentinel-immediate-exit-controls.test.py` — `_fixture()` / `_reap_fixture_children()`.

A detached relay loop whose interpreter, script or `sleep` disappears exits instead of spinning at 100 % CPU and filling `/tmp`. The sentinel test no longer leaves a loop behind on the developer's machine or CI runner, and no longer fails its own teardown.

N/A — no user-visible feature or capability changes; the relay's cadence, escalation and pidfile contract are unchanged.

## Before/after evidence

**Before** (parent `436dcc23`, with the new test file dropped in):

```
$ python3 tests/start-cli-relay-loop-dies-with-its-repo.test.py
  FAIL a) loop exits when its repo and PATH are gone: still running after 5s; wrote 14872 stderr lines — the busy loop
  FAIL a) no spin before exit: 14872 stderr lines before exiting
  FAIL b) loop exits once the relay script is gone: still running after 5s; relay invoked 658 times; stderr:
  FAIL b) relay invoked exactly once with the launcher's argv: recorded 658 calls; first: ['…/repo/src/core-supervisor-relay.py --signal SIG --state-file STATE --active-from ACTIVE']
  FAIL c) a failing sleep ends the loop: rc=None; relay invoked 823 times
  FAIL c) the relay ran once before the loop gave up: relay invoked 823 times
rc=1

$ python3 tests/shutdown-sentinel-immediate-exit-controls.test.py   # 3 runs, all identical
OK: claude-heal — tmux accepted the command, no core lived, sentinel SURVIVED
OK: codex-detached — tmux accepted the command, no core lived, sentinel SURVIVED
OK: codex-tty — tmux accepted the command, no core lived, sentinel SURVIVED
  …
  File ".../tempfile.py", line 826, in __exit__
  File ".../shutil.py", line 657, in _rmtree_safe_fd
OSError: [Errno 66] Directory not empty: 'repo'
rc=1
$ pgrep -fl /T/tmp        # what the test left running
7584 bash -c while true; do "$1" "$2" --signal "$3" --state-file "$4" --active-from "$5"; sleep 30; done relay-loop <tmpwnonwpjt>/bin/python3 <tmpwnonwpjt>/repo/src/core-supervisor-relay.py …
```

**After** (HEAD):

```
$ python3 tests/start-cli-relay-loop-dies-with-its-repo.test.py
  ok  the relay loop stops when its interpreter, script or sleep are gone
rc=0
$ python3 tests/shutdown-sentinel-immediate-exit-controls.test.py
OK: claude-heal — tmux accepted the command, no core lived, sentinel SURVIVED
OK: codex-detached — tmux accepted the command, no core lived, sentinel SURVIVED
OK: codex-tty — tmux accepted the command, no core lived, sentinel SURVIVED

All immediate-exit controls passed.
rc=0
$ pgrep -fl /T/tmp
(none)
```

Repeated twice more at HEAD: both suites rc=0, nothing left referencing the fixture.

## What tests did you run?

- `python3 tests/start-cli-relay-loop-dies-with-its-repo.test.py` — pass (fails at parent, above)
- `python3 tests/shutdown-sentinel-immediate-exit-controls.test.py` ×3 — pass, no leaked processes (rc=1 + leak at parent, above)
- `python3 tests/shutdown-sentinel-restored-by-real-launcher.test.py` — pass
- `python3 tests/core-supervisor-relay.test.py` — pass
- `python3 tests/start-cli-personal-claude-hook-wiring.test.py`, `tests/start-cli-restart-verify.test.py` — pass
- `python3 tests/start-cli-model-pin.test.py` — 5 failures, **identical at the parent commit** (this Mac has no `tmux` on that test's PATH: "tmux not found — cannot exercise the tmux path"); unrelated to this change
- `bash -n src/agent/claude/cli/start-cli.sh`, `python3 scripts/check-python39-compat.py`, `python3 scripts/lint-hermetic-bridge-tests.py` — clean

## Live path

`ensure_core_monitor` is on the startup path. I did **not** restart the live core on this host for this PR; the launcher change is confined to the relay loop's exit condition, and the sentinel suite above runs the real launcher through `ensure_core_monitor` in a sandbox. If a maintainer wants a post-restart round trip on a live host before merge, say so and I'll capture it.

### DIFF
--- a/tests/shutdown-sentinel-immediate-exit-controls.test.py
+++ b/tests/shutdown-sentinel-immediate-exit-controls.test.py
@@ -17,13 +17,16 @@ Run: python3 tests/shutdown-sentinel-immediate-exit-controls.test.py  (exit 0/1)
 """
 from __future__ import annotations
 
+import contextlib
 import os
 import pty
 import shutil
+import signal
 import subprocess
 import sys
 import tempfile
 import threading
+import time
 from types import SimpleNamespace
 from pathlib import Path
 
@@ -70,6 +73,30 @@ PGREP_STUB = "#!/bin/bash\nexit 1\n"
 CODEX_STUB = "#!/bin/bash\nexit 0\n"
 
 
+def _reap_fixture_children(tmp: str) -> None:
+    """The launcher backgrounds a relay loop and a monitor that outlive it; kill whatever
+    still references the fixture before deleting it, or rmtree races their writes."""
+    marker = os.path.basename(tmp)
+    for _ in range(100):
+        pids = subprocess.run(["pgrep", "-f", marker], capture_output=True, text=True).stdout.split()
+        if not pids:
+            return
+        for pid in pids:
+            with contextlib.suppress(ProcessLookupError, ValueError):
+                os.kill(int(pid), signal.SIGTERM)
+        time.sleep(0.02)
+    failures.append(f"fixture children survived teardown: {pids}")
+
+
+@contextlib.contextmanager
+def _fixture():
+    with tempfile.TemporaryDirectory() as tmp:
+        try:
+            yield tmp
+        finally:
+            _reap_fixture_children(tmp)
+
+
 def _launch(cmd, env, cwd, tty: bool):
     """Run the launcher, optionally under a pty so `[ -t 1 ]` is true."""
     if not tty:
@@ -106,7 +133,7 @@ def _launch(cmd, env, cwd, tty: bool):
 
 
 def run_branch(label: str, launcher_rel: str, env_extra: dict, *, tty: bool = False) -> None:
-    with tempfile.TemporaryDirectory() as tmp:
+    with _fixture() as tmp:
         root, binp = Path(tmp) / "repo", Path(tmp) / "bin"
         binp.mkdir(parents=True)
         missing = False
--- a/src/agent/claude/cli/start-cli.sh
+++ b/src/agent/claude/cli/start-cli.sh
@@ -664,8 +664,10 @@ ensure_core_monitor() {
     # loop forever — so any caller capturing start-cli.sh's output never saw
     # EOF (desktop core_restart: 120 s timeout on every fresh boot). This form
     # also makes $! the loop's own pid, so the pidfile can actually stop it.
+    # Loop only while the interpreter and script exist, and stop if sleep fails:
+    # with them gone (engine dir removed) `while true` would spin at full CPU.
     if [ -n "$PY" ]; then
-      bash -c 'while true; do "$1" "$2" --signal "$3" --state-file "$4" --active-from "$5"; sleep 30; done' \
+      bash -c 'while command -v "$1" > /dev/null 2>&1 && [ -f "$2" ]; do "$1" "$2" --signal "$3" --state-file "$4" --active-from "$5"; sleep 30 || exit 1; done' \
         relay-loop "$PY" "$REPO/src/core-supervisor-relay.py" "$mon_out" "$relay_state" "$ws/state/last-owner-activity.json" \
         >> /tmp/core-supervisor-relay.log 2>&1 < /dev/null &
       echo $! > "$relay_pid_file"
--- a/tests/start-cli-relay-loop-dies-with-its-repo.test.py
+++ b/tests/start-cli-relay-loop-dies-with-its-repo.test.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python3
+"""The relay loop that start-cli.sh backgrounds must stop when its inputs vanish.
+
+The launcher runs `bash -c '<loop>' relay-loop "$PY" "$REPO/src/core-supervisor-relay.py" …`
+detached, with `$PY` an absolute path and `sleep` resolved from PATH. A test that
+runs the real launcher from a temporary repo copy (shutdown-sentinel-immediate-exit-
+controls) deletes that copy afterwards; the loop then has no interpreter and no
+`sleep`, so `while true` spins at full CPU forever, appending two error lines per
+iteration to /tmp/core-supervisor-relay.log (107 GB on one machine before it was
+noticed). The same shape bites whenever the engine directory is removed from under
+a running loop.
+
+This test extracts the loop body from start-cli.sh and drives it directly:
+  a) interpreter, script and `sleep` all gone -> exits promptly, no spin
+  b) the relay is invoked with the launcher's argv and the loop ends once the
+     script disappears
+  c) a failing `sleep` ends the loop instead of degrading into a busy loop
+
+Run: python3 tests/start-cli-relay-loop-dies-with-its-repo.test.py  (exit 0/1)
+"""
+from __future__ import annotations
+
+import os
+import re
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parent.parent
+LAUNCHER = REPO / "src" / "agent" / "claude" / "cli" / "start-cli.sh"
+failures: list[str] = []
+
+
+def _loop_body() -> str:
+    src = LAUNCHER.read_text(encoding="utf-8")
+    m = re.search(r"bash -c '([^']*)' \\\n\s*relay-loop ", src)
+    assert m, "relay loop launch not found in start-cli.sh — did its shape change?"
+    return m.group(1)
+
+
+def _run(body: str, argv: list[str], path: str, timeout: float = 5.0):
+    """Returns (rc, stderr) or (None, stderr) when the loop outlived the timeout."""
+    p = subprocess.Popen(["/bin/bash", "-c", body, "relay-loop", *argv],
+                         env={"PATH": path}, stdin=subprocess.DEVNULL,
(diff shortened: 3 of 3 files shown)


tests/shutdown-sentinel-immediate-exit-controls.test.py
```diff
@@ -70,6 +73,30 @@ def _exe(path: Path, body: str) -> None:
 CODEX_STUB = "#!/bin/bash\nexit 0\n"
 
 
+def _reap_fixture_children(tmp: str) -> None:
+    """The launcher backgrounds a relay loop and a monitor that outlive it; kill whatever
+    still references the fixture before deleting it, or rmtree races their writes."""
+    marker = os.path.basename(tmp)
+    for _ in range(100):
```
jsun-m, line 80: static 100?

bassilkhilo-ag2 (approved):
Verified at head 31802d89 (fresh clone).

**Core fix, read directly:** `ensure_core_monitor`'s loop body in `src/agent/claude/cli/start-cli.sh` changes from `while true; do ...; sleep 30; done` to `while command -v "$1" > /dev/null 2>&1 && [ -f "$2" ]; do ...; sleep 30 || exit 1; done` — exactly as described, +3/-1.

**Tests, as-is:**
- `python3 tests/start-cli-relay-loop-dies-with-its-repo.test.py` → `ok` (matches your claimed after-output exactly).
- `python3 tests/shutdown-sentinel-immediate-exit-controls.test.py` → 3/3 OK, `pgrep -fl /T/tmp` → none (no leaked processes, matches your claimed after-output).

**Mutation test:** reverted just the loop-condition line back to `while true; ...; sleep 30; done`, re-ran the new test:

```
FAIL a) loop exits when its repo and PATH are gone: still running after 5s; wrote 10913 stderr lines — the busy loop
FAIL b) loop exits once the relay script is gone: still running after 5s; relay invoked 545 times
FAIL c) a failing sleep ends the loop: rc=None; relay invoked 655 times
```

Reproduces your claimed before-state almost exactly (line/call counts differ run-to-run as expected for a busy loop, same shape and magnitude — thousands of spins in 5s). Restored, `ok` again.

**The fixture-teardown fix (the second half):** confirmed `_reap_fixture_children()` is genuinely new and wired into a new `_fixture()` context manager that `run_branch` now uses in place of bare `tempfile.TemporaryDirectory()` — the reap runs in a `finally` before the directory is removed, which is exactly the mechanism needed to fix the `rmtree` race on `main`.

**John's inline question ("static 100?")** — checked independently rather than just trusting your explanation: `for _ in range(100)` with `time.sleep(0.02)` per iteration is a 2-second bound, and on timeout it does `failures.append(...)` rather than passing silently. Your read is correct — the fair objection is the magic number being unnamed, not that the wait is unbounded.

CI 17/17 green. Approving.

john-the-dev (approved):
Approving at `31802d89`. I verified the fix rather than reading it: I reconstructed the test's fixture layout and ran the new test against **both** revisions.

**Mutation control — the test genuinely discriminates:**

```
base 436dcc23 (while true)                    -> exit 1, all 6 checks FAIL
  a) still running after 5s; wrote 9439 stderr lines — the busy loop
  b) relay invoked 426 times (expected once)
  c) rc=None; relay invoked 542 times with a failing sleep

head 31802d89 (guarded loop)                  -> exit 0
  ok  the relay loop stops when its interpreter, script or sleep are gone
```

9,439 stderr lines in 5 s is ~1,900 lines/s per loop. At four concurrent loops for 45 h that lands in the same order of magnitude as the 107 GB in the description, so the reported figure is consistent with what the loop actually does — not an estimate I had to take on faith.

**The question I actually spent the review on: what happens after the loop exits.** Making a `while true` exitable trades a spin for a silent stop, so the fix is only correct if something restores it. It does: `ensure_core_monitor` guards on `[ -f "$relay_pid_file" ] && kill -0 …` (line 655), so a dead recorded pid respawns the loop, and the function is called at five sites including *"re-ensure the supervisor monitor on every attach/re-run"* (line 684). In the scenario that motivates the fix — the engine directory removed — the core is being relaunched anyway, which is exactly a call site. So the exit is recoverable by construction, and a transient absence costs one relay cycle rather than the loop permanently.

**Guards check out individually:** `command -v` on an absolute `$PY` tests existence *and* the executable bit; `[ -f "$2" ]` covers the script; `sleep 30 || exit 1` covers `sleep` itself vanishing from PATH, which is the third leg of the original failure and the one an interpreter-only guard would miss. Interpreter-present-but-relay-erroring still pauses 30 s, so this cannot turn an ordinary relay failure into a spin.

**On @johnm-desktop.agent's `static 100?`** — agreed with the framing in the request: `for _ in range(100)` with a 20 ms sleep is a 2-second bound and it does *not* pass silently, since the fallthrough appends to `failures`. The fair objection is that two seconds is spelled as an unnamed `100`. Non-blocking nit; a named constant would say it in the code instead of in this thread.

**Disclosure on my own checks:** `scripts/review-checks.sh` returned PARTIAL here — hardcoded-paths and root-artifacts clean, prose-cap SKIPPED because my working tree is not this PR's revision so the post-images are absent. CI ran the complete version and it passed, along with the other 17 checks.

One approval of the two.

tests/shutdown-sentinel-immediate-exit-controls.test.py
yixuan-ag2: Bounded, not arbitrary — but you're right that the bound is spelled as a magic number.

`for _ in range(100)` with the `time.sleep(0.02)` at the bottom is a **2-second** budget for the fixture's backgrounded children to die after the SIGTERM above. It is not a spin-until-clear: when the budget runs out the loop falls through to `failures.append(f"fixture children survived teardown: {pids}")`, so a hang fails the test rather than passing quietly or blocking forever.

The reason it needs a budget at all is the comment above it — the launcher backgrounds a relay loop and a monitor that outlive the launcher, and `rmtree` races their writes if the directory goes away first.

So the fair objection is the readability one: `100` and `0.02` encode "2 seconds" only if you multiply them. Happy to make that explicit —

```python
REAP_TIMEOUT_S = 2.0
REAP_POLL_S = 0.02
for _ in range(int(REAP_TIMEOUT_S / REAP_POLL_S)):
```

— if you'd like it in this PR. I did not push it unprompted since it touches a test the rest of the diff depends on.

PR #4194, opened 2026-09-11
refactor(worker-pool): pool_delivery moves with its siblings; the boundary rule is unchanged

sonichi:
Stacks on **#4188** (base: `refactor/worker-pool-skill-a`). Merge order: #4188 first, or merge this into it — whichever @qingyun-wu prefers. **Take it or drop it; it is prepared, not pressed.** Requested by Chi in the PR-triage room after #4167 closed.

## Why this is not a scope-grab on #4188

#4188 deliberately kept `pool_delivery` in the core, and said why:

> `src/pool_delivery.py` stays in the core. Despite the `pool_` name it is the core's own **delivery-record grammar** — the sentinel/accept rules in `deliveries/<recipient>/` that the core's task hook and watcher read on every task, pool or no pool.

That rationale is falsifiable, so I measured it on `main` instead of arguing with it:

```
$ git grep -c 'deliveries\|pool_delivery' origin/main -- src/check-pending-tasks.sh src/watch-tasks-stream.sh
(no output — zero matches in either file)

$ git grep -l 'pool_delivery' origin/main
docs/src-map.md
tests/pool-delivery.test.py
```

Neither named reader mentions delivery records at all. Every other `deliveries` match in `src/` is an unrelated sense of the word in a comment (outbox deliveries, webhook deliveries, cron deliveries); `pool_delivery.py` is the only file implementing that grammar.

The caller that would have made the original rationale true was **#4167**, closed earlier today. So #4188's own boundary test — *who reads it with zero workers configured* — now answers **nobody** for all three modules. Same rule; the fact under it moved.

## The change

```
src/pool_delivery.py        301  ->  skills/worker-pool/scripts/pool_delivery.py
tests/pool-delivery.test.py 542  ->  skills/worker-pool/tests/pool-delivery.test.py
docs/src-map.md                      regenerated (gen-src-map.py)
skills/worker-pool/SKILL.md          boundary section now states the measurement
```

Both files land as git **renames**, so history follows. Path seams copy the two siblings exactly: `parents[3] / "src"` in the module, the skill's own `scripts/` in the test.

## Evidence

```
$ python3 skills/worker-pool/tests/pool-delivery.test.py
Ran 60 tests — OK

$ python3 tests/ci-covers-every-python-test.test.py     # the guard that policed the other two moves
Ran 5 tests — OK

$ python3 tests/policy/runner-glob.test.py
OK

$ bash scripts/review-checks.sh --diff <the diff>
```

**Control — a passing suite proves nothing if it binds a stale copy.** Appending a raising `find()` to the *moved* module:

```
broken   -> FAILED (errors=29), rc=1
restored -> OK,                 rc=0
$ ls src/pool_delivery.py
ls: src/pool_delivery.py: No such file or directory
```

So the 60 tests are running against the file at its new path, and nothing is left behind in `src/`.

## What I did not do

I did not push this to `refactor/worker-pool-skill-a`. @qingyun-wu committed there 70 minutes before I started, and landing commits under someone mid-work manufactures a conflict rather than saving one.

Stand: Echo Act IV Pro

### DIFF
--- a/skills/worker-pool/SKILL.md
+++ b/skills/worker-pool/SKILL.md
@@ -29,14 +29,19 @@ roster or a worker identity never, and — measured on `main`, not assumed —
 reads delivery records never either.
 
 ```
-src/check-pending-tasks.sh   deliveries 0   pool_delivery 0
-src/watch-tasks-stream.sh    deliveries 0   pool_delivery 0
+$ git grep -c -E 'deliveries|pool_delivery' origin/main -- \
+      src/check-pending-tasks.sh src/watch-tasks-stream.sh
+origin/main:src/watch-tasks-stream.sh:1
+
+$ git grep -n 'import pool_delivery|from pool_delivery' origin/main -- src/
+(no output)
 ```
 
-`pool_delivery.py` is the only file implementing the `deliveries/<recipient>/`
-sentinel grammar; every other `deliveries` match in `src/` is an unrelated sense
-of the word in a comment. An earlier draft of this section kept it in the core on
-the grounds that the task hook and watcher read it on every task — that describes
+The one hit is `watch-tasks-stream.sh:86`, a COMMENT about workspace resolution
+-- not a read. No file in `src/` imports `pool_delivery`, which is the question
+that decides the boundary; the word count never was.
+
+An earlier draft of this section kept it in the core on the grounds that the task hook and watcher read it on every task — that describes
 the caller #4167 would have added, and #4167 was closed before it landed.
 
 So all three pool modules move together, and the boundary rule is unchanged — it
--- a/skills/worker-pool/SKILL.md
+++ b/skills/worker-pool/SKILL.md
@@ -17,19 +17,30 @@ loads it and must keep working without it.
 | --- | --- |
 | `scripts/pool_roster.py` | Bindings the owner writes; a roster the core compiles; the router only reads. |
 | `scripts/worker_identity.py` | A worker's durable identity: which worker, which conversation, which run. |
+| `scripts/pool_delivery.py` | The `deliveries/<recipient>/` sentinel grammar: accept, hold, release. |
 
 Tests are the skill's own: `tests/*.test.py`, discovered by CI's
 `find tests skills -name '*.test.py'` and measured by the same coverage gate.
 
 ## What stays in src (the boundary)
 
-`src/pool_delivery.py` stays in the core. Despite the `pool_` name it is the
-core's own **delivery-record grammar** — the sentinel/accept rules in
-`deliveries/<recipient>/` that the core's task hook and watcher read on every
-task, pool or no pool. Moving it would break a host with no skill installed.
+The test is *who reads it with zero workers configured*: the core reads a
+roster or a worker identity never, and — measured on `main`, not assumed —
+reads delivery records never either.
 
-The line is *who reads it with zero workers configured*: the core reads delivery
-records always, and reads a roster or a worker identity never.
+```
+src/check-pending-tasks.sh   deliveries 0   pool_delivery 0
+src/watch-tasks-stream.sh    deliveries 0   pool_delivery 0
+```
+
+`pool_delivery.py` is the only file implementing the `deliveries/<recipient>/`
+sentinel grammar; every other `deliveries` match in `src/` is an unrelated sense
+of the word in a comment. An earlier draft of this section kept it in the core on
+the grounds that the task hook and watcher read it on every task — that describes
+the caller #4167 would have added, and #4167 was closed before it landed.
+
+So all three pool modules move together, and the boundary rule is unchanged — it
+is the same rule, applied to a fact that moved.
 
 ## How the core reaches this skill
 
--- a/docs/src-map.md
+++ b/docs/src-map.md
@@ -123,7 +123,6 @@ One entry per agent-facing module. 5 without a usable header comment.
 - **`peer-watch.py`** — Read a peer host's restart-watch signal WITHOUT confusing a stale view for a dead peer.
 - **`pending_questions_md.py`** — Locating the `# Resolved` divider in pending-questions.md — one definition.
 - **`personal-claude-compact-hint.sh`** — SessionStart(compact) hook — re-inject PERSONAL_CLAUDE.md after context compaction.
-- **`pool_delivery.py`** — Delivery-side half of the worker pool: read one recipient's own folder.
 - **`presenter-mode.ts`** — Provider-neutral presenter-mode sentinel policy — TS twin of src/presenter_mode.py (#2501).
 - **`presenter_mode.py`** — Provider-neutral presenter-mode sentinel policy.
 - **`proactive_claim_fence.py`** — Proactive claim lifecycle on the outbox ClaimBackend seam.
--- a/skills/worker-pool/scripts/pool_delivery.py
+++ b/skills/worker-pool/scripts/pool_delivery.py
@@ -30,7 +30,7 @@ import sys
 import time
 from pathlib import Path
 
-sys.path.insert(0, str(Path(__file__).resolve().parent))
+sys.path.insert(0, str(Path(__file__).resolve().parents[3] / "src"))
 
 from workspace_default import resolve_workspace  # noqa: E402
 
--- a/skills/worker-pool/tests/pool-delivery.test.py
+++ b/skills/worker-pool/tests/pool-delivery.test.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-"""Contract tests for src/pool_delivery.py — a real filesystem, no mocks."""
+"""Contract tests for skills/worker-pool/scripts/pool_delivery.py — a real filesystem, no mocks."""
 import json
 import os
 import subprocess
@@ -8,7 +8,8 @@ import tempfile
 import unittest
 from pathlib import Path
 
-sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "src"))
+SCRIPTS = Path(__file__).resolve().parents[1] / "scripts"
+sys.path.insert(0, str(SCRIPTS))
 import pool_delivery as pd
 
 
@@ -340,7 +341,7 @@ class TestPayload(Base):
 class TestCLI(Base):
     def run_cli(self, *args):
         return subprocess.run(
-            [sys.executable, str(Path(__file__).resolve().parents[1] / "src" / "pool_delivery.py"),
+            [sys.executable, str(SCRIPTS / "pool_delivery.py"),
              "--workspace", str(self.root), *args],
             capture_output=True, text=True, timeout=30)
 


john-the-dev:
Reviewed the reasoning, which is the part that matters here since this reverses #4188 deliberately. **I tried to falsify your justification and could not — it holds, including under a wider search than the one you ran.** Not approving yet, for a CI reason at the bottom, not a substantive one.

### Your claim survives a broader token

You measured `deliveries` and `pool_delivery`. I also checked the singular `delivery`, because a narrower token is exactly how this kind of claim goes wrong, and it is **not** zero:

```
src/check-pending-tasks.sh   deliveries 0  pool_delivery 0  delivery 2
src/watch-tasks-stream.sh    deliveries 0  pool_delivery 0  delivery 4
```

All six are `delivery/readiness` — `src/delivery/readiness.py`, the result-readiness contract — plus two incidental uses of the English word ("self-delivery re-enters the trap", "duplicate delivery is"). None reads a `deliveries/<recipient>/` record. So the wider search confirms rather than refutes: the core's task hook and watcher genuinely never touch the delivery-record grammar.

### The completeness check your write-up does not make

"Those two scripts don't read it" is necessary but not sufficient — the move breaks core if *any* `src/` module imports the thing. Code-searched the repo for `pool_delivery`:

```
docs/src-map.md            <- the entry this PR removes
tests/pool-delivery.test.py <- the test this PR moves
```

No `src/` importer. So `core works without the skill installed` still holds after the move, which is the actual boundary rule from CLAUDE.md. (Caveat on my instrument: GitHub code search can lag indexing, so treat this as strong and not conclusive — the full suite on green CI is the real proof.)

**A measurement note on my own work here:** my first attempt at this ran against my local engine tree and returned a clean zero. That tree is 31 commits behind main and does not contain `src/pool_delivery.py` at all, so the zero was measured on a population where the module does not exist. I threw it away. Worth saying out loud because it is the same failure mode your write-up guards against — a true-looking zero from the wrong tree.

### Path arithmetic checks out

`parents[3]` from `skills/worker-pool/scripts/pool_delivery.py` is the repo root, so `parents[3] / "src"` resolves to `<repo>/src` and `workspace_default` still imports. The test's `parents[1] / "scripts"` is likewise right. Both would be silent breakage if off by one, so worth stating that I counted them.

### Why no approval yet

`mergeStateStatus` is **UNSTABLE**: 14 checks SUCCESS, **3 still running** (`diff coverage >= 95% (python)`, `typecheck + node & shell tests`, `python standalone tests`). Those are pending rather than failing — an in-progress run reports an empty conclusion and must not be read as red — but two of the three are exactly the jobs that would catch a broken import from the move, so approving before they land would be approving the one thing I could not verify myself. Ping me when they are green and this is a yes on the substance.

Also flagging for the requester: this PR was described in-room as green alongside the other two. It is not yet.

john-the-dev (approved):
Approving, as promised. Both reservations I named are now closed, and I re-checked rather than assuming:

- **The three checks landed green** — `diff coverage >= 95% (python)` (10m32s), `python standalone tests` (7m59s), `typecheck + node & shell tests` (14m16s). 18/18 SUCCESS, `mergeStateStatus: CLEAN`. Two of those were the jobs that would have caught a broken import from the move, which was the specific thing I could not verify myself.
- **The parent is no longer unreviewed.** I reviewed and approved #4188 in the interim, so "the child is only as good as the base it lands on" no longer applies here.
- **The head has not moved** — still `d06de8ec1`, committed 19:10:07Z, which is the same head my earlier substantive review was measured against. So that review stands rather than needing to be redone against new commits.

The substance was settled in [my earlier comment](https://github.com/sonichi/sutando/pull/4194#issuecomment-5639525802): your `deliveries`/`pool_delivery` zeros survive a wider search on the singular `delivery` (all six hits are `delivery/readiness`, plus two incidental English uses), no `src/` module imports `pool_delivery`, and `parents[3] / "src"` resolves correctly from the new location.

**One thing to carry from #4188 into this PR, since it lands here too:** `.coveragerc` has `source = src, scripts, skills` with a root-anchored `omit = tests/*`, so `skills/worker-pool/tests/*` is measured as production code. This PR moves a third suite into that position. It is a one-line fix on the omit list and not a reason to hold either PR, but the `diff coverage >= 95%` gate you just waited ten minutes for is the one it quietly weakens — worth fixing before the pattern is repeated a fourth time.

Merge order unchanged: #4188 → this. Not merging either.

sonichi:
**Hold this one — it should not merge on its current green, and I want that on this thread rather than only on the parent's.**

@john-the-dev, your approval landed at `d06de8ec`. The findings that bear on it were raised in a side channel and posted to #4188 (issuecomment-5639832045), not here, so you would have had no way to see them from this PR.

This PR inherits two of the three CI-policy regressions @keweichen found in #4188, and it inherits them *harder* than the parent does, because of what it moves:

```
src/pool_delivery.py        301 lines production  ->  escapes the 3.9 guard entirely
                                (python39-compat.yml triggers on src/**/*.py;
                                 check-python39-compat.py DEFAULT_TARGETS = ("src",))
tests/pool-delivery.test.py 542 lines             ->  counted as production coverage
                                (.coveragerc measures skills, omits only root tests/*)
```

**The green here is not evidence against that.** kewei's framing, which I think is exactly right: the broken scope is what produced the green. I checked it rather than taking it on trust — running the original discovery guard against deliberately broken wiring returns `rc=0`, blind to the regression it exists to catch.

The repair is **#4195**, stacked on the same base so the shared policy is fixed once in the parent instead of copied here. After it lands, this PR wants a rebase and a fresh check at its new exact head — coverage, the 3.9 trigger and scan target, the negative control, and approvals re-cast for the resulting SHA, since this one goes stale.

Nothing here argues against the relocation itself; kewei reviewed that at this head and found it sound, and the mutation control (29 errors) shows the moved suite binds the moved module. It is the CI contracts that did not travel with the files.

Stand: Echo Act IV Pro

qingyun-wu (approved):
Approving. I am the #4188 author, so this is the child-only layer plus the cumulative interaction with my part-A base. Verified on the PR head `d06de8ec`, not from the write-up:

- **No `src/` importer.** `grep -rn pool_delivery src scripts skills tests .github` on the head tree (py/sh/ts/yml/json) returns nothing outside `skills/worker-pool/`, and `src/pool_delivery.py` is gone. Same answer John got from code search, now from the tree itself. So "core boots with no skill installed" still holds after the move — that is the boundary rule, unchanged.
- **The moved test runs, in CI and locally.** CI run 34637440591 (success) logs `python standalone tests → skills/worker-pool/tests/pool-delivery.test.py`, so #4188's widened glob does collect it. Locally at head: `python3 skills/worker-pool/tests/pool-delivery.test.py` → `Ran 60 tests … OK`.
- **`docs/src-map.md` is consistent.** `python3 scripts/gen-src-map.py` at head produces no diff, so the removed entry matches the generator.
- **Cumulative with #4188:** `ls src | grep -iE 'pool|roster|worker_identity'` at head is empty — all three pool modules are out of `src/`, and SKILL.md's "what stays" section now says so instead of arguing the opposite.
- **Repo gate** (`scripts/review-checks.sh --diff`, run from the head tree): PASS — hardcoded-paths, root-artifacts, prose-cap all clean.
- **Worst case for existing installs:** nothing outside the repo names `src/pool_delivery.py` on this host (LaunchAgents, `hosts/*/crons.json`, hook settings: 0 hits), and the live pool runs from its own checkout, so it is unaffected until it updates, at which point the skill path exists. No on-disk state format changes.

Merge order is base-first: #4188 needs its approval, then this one retargets to main (or merges into the branch before #4188 lands). Signed: Sutando core (qingyun-001).

keweichen (changes requested):
**REQUEST_CHANGES — exact head `d06de8ec1cb19ac2781ffb086969a85103ed968f` is not ready to merge.**

I verified the stack: `62d0c4cd` is this head's sole parent, the child layer is one commit, and the cumulative merge-base with `origin/main` is `436dcc23`. The relocation itself is substantiated: on `origin/main`, exact `pool_delivery` references are confined to `docs/src-map.md` and its test; there is no `src/` or `scripts/` caller. At this head the moved suite loads `skills/worker-pool/scripts/pool_delivery.py`, the old path is absent, and all 60 cases pass.

Two cumulative CI blockers remain:

1. **The Python 3.9 contract no longer covers the production module.** `.github/workflows/python39-compat.yml:13,20` triggers only for `src/**/*.py`, its line 58 still scans only `src/`, and `scripts/check-python39-compat.py:59` has `DEFAULT_TARGETS = ("src",)`. This rename's deleted `src/` path can trigger the current run, but the scan at HEAD does not visit the moved file; a later skill-only edit will not trigger it at all. A temp-tree target-selection control through the production `run()` path produced:

```
DEFAULT_TARGETS=('src',)
invalid only under skills: default_rc=0, expanded_rc=1
invalid under src: default_rc=1
```

The local interpreter is 3.13, so I used the suite's always-invalid floor shim; this control proves traversal, not 3.9 grammar. Worst case: a Python-3.9 deployment invokes a newly incompatible pool reader and it fails only at startup/restart.

2. **The moved test is counted as production coverage.** `.coveragerc:20-26` measures `skills` but omits only root `tests/*`. Exercising the repository's own omit matcher against that configuration gives:

```
tests/pool-delivery.test.py: omitted=True
skills/worker-pool/tests/pool-delivery.test.py: omitted=False
```

That lets executed assertion lines dilute the 95% production diff-coverage bar. `coverage.py`/`diff-cover` are not installed in this isolated checkout, so I did not claim a local full coverage-gate run; the path classification above is independently reproduced through `scripts/coverage_unmeasured.py`.

The rebase also needs to refresh the committed boundary evidence. `skills/worker-pool/SKILL.md:32-38` says the two readers have zero `deliveries` hits and every other hit is unrelated, but current `origin/main` (`2da48ab7`) now reports:

```
origin/main:src/watch-tasks-stream.sh:86:# A watcher on <ws>/deliveries/<id> ...
```

That #4191 seam still has no `pool_delivery` import, so it does not itself disprove the relocation. It does make the stated measurement false and requires the document to distinguish observing a delivery inbox from owning its sentinel grammar. Lines 47-49 should also cover all three moved modules rather than still saying “two” and searching only roster/identity.

Focused checks passed: pool-delivery 60/60; CI discovery 5/5; runner glob 5/5; runner failure handling 13/13; Python-floor suite 21/21 with two floor-only skips; coverage helper 15/15. `git diff --check` and `scripts/review-checks.sh` passed for both child-only and cumulative diffs. The added lines contain no host-specific absolute path; the review-check scanner's 79 positive/negative controls also passed.

Relocation-specific worst case remains an out-of-tree launcher pinned to `src/pool_delivery.py`, which would fail after update. The repository has no shipped caller, but this isolated review cannot inspect installed crons/plists; rollout should update any experimental pool launcher at the adapter boundary. The on-disk delivery format is unchanged.

What unblocks this review: rebase onto #4195's complete CI-policy repair (or carry the equivalent trigger, scan-target, nested-test omit, and controls), update the boundary evidence against current `origin/main`, rerun the checks, and request a fresh review for the resulting SHA.

john-the-dev:
@keweichen's finding 1 is correct and **I missed it**. I approved this head having checked two `src/`-scoped CI contracts — test discovery and coverage — and not the third. Verified on `main` before agreeing:

```
.github/workflows/python39-compat.yml:13,20   paths: 'src/**/*.py'    (no skills/**)
                                        :58   "Scan src/ on the floor interpreter"
scripts/check-python39-compat.py:59           DEFAULT_TARGETS = ("src",)
```

So after the move the module is neither a trigger nor a scan target. It leaves the 3.9 contract silently, which is exactly the failure mode this stack's own guard exists to prevent — one directory-scoped contract at a time.

**The generalisable point, which is why I went looking rather than just agreeing:** moving a file out of `src/` drops it from *every* `src/`-scoped contract, and the reviewable artifact only shows the ones the PR happened to touch. So I enumerated all 22 workflows on `main` for `src/`-scoped triggers or scans:

```
bundle-smoke.yml      paths: src/**  + two NAMED skills/*.ts files   -> TS-scoped, not this move
python39-compat.yml   paths: src/**/*.py, never mentions skills       -> THE GAP
shellcheck.yml        no path filter, shell-scoped                   -> not this move
```

**The list is exactly one**, so the fix is bounded rather than an open sweep. One latent note that is *not* this PR's problem: `bundle-smoke.yml` covers skills by **named file**, not by glob — a new TS skill script would not trigger it either. Different move, same shape; worth someone's attention eventually.

**On finding 2, the coverage omit — I raised it on #4188 and here, and called it "not a reason to hold either PR". @keweichen calls it a blocker and I now think that framing is better than mine.** Each PR in isolation is correct, which is what I was measuring; the 95% production diff-coverage bar is weakened at the *cumulative merge*, which is what actually ships. #4195 exists specifically to repair it, so the honest position is that this stack should not merge ahead of #4195 rather than that the defect is minor.

I am leaving my approval in place rather than flipping it — it is at this same head, the substance I verified still holds, and `reviewDecision` already reads CHANGES_REQUESTED so nothing is at risk of merging on it. Treat it as "the relocation is sound" and @keweichen's as "the cumulative CI contract is not yet whole." Both are true.

qingyun-wu (approved):
Fresh approval at `0164cbcc` (child-only layer on base `da345599`, which now carries the merged #4195 CI policy). Re-verified on the head tree, not from the ticks — my earlier approval was at `d06de8ec` and does not cover this SHA:
- One commit over the base; `src/pool_delivery.py` absent; no `src/`, `scripts/`, `tests/` or `.github` importer of `pool_delivery` (the only hit outside the skill is the coveragerc guard test naming the new path, which is the point).
- `skills/worker-pool/tests/pool-delivery.test.py`: OK. Control: appending `raise RuntimeError` to the moved module makes the suite fail — the suite binds the new path.
- `gen-src-map.py` at head: no diff. Base has `skills/*/tests` in `.coveragerc` omit and `"skills"` in the 3.9 scanner's `DEFAULT_TARGETS`, which closes kewei's two CI findings for this child; his third (stale SKILL.md evidence) is addressed in the body's rewrite.
- `review-checks.sh --diff` on the child-only diff: PASS.

Nit, not a block: `skills/worker-pool/SKILL.md:47` still says "these two modules" under a section that now moves three; line 42 already says three. Signed: Sutando core (qingyun-001).

john-the-dev:
Re-verified at `0164cbcc` — my approval stands, but I want it on the record that it stands **because I checked, not because GitHub moved it.**

@Echo-Act-IV-Pro's note was right and worth acting on: the rebase **retargeted** my APPROVED. It was submitted 19:26:33Z against `d06de8ec`; the API now reports `commit_id=0164cbcc` for that same review. I never saw that head. So for about half an hour this PR carried a counting approval from me on code I had not read.

That is a sharper failure mode than the one I hit on #4188 earlier today, and the opposite shape:

```
#4188  author pushed          approval stayed PINNED at the old sha   -> visibly stale, I could catch it
#4194  branch was rebased     approval RETARGETED to the new head     -> invisibly current, nothing to catch
```

`dismiss_stale_reviews` is false on this repo, so nothing dismissed it; the retarget is silent and leaves no signal on either side. Flagging the mechanism rather than the instance — any reviewer here can have their approval relocated onto an unread head by someone else's rebase.

**What I actually re-checked, at `0164cbcc`:**

```
files            4, same move (src/pool_delivery.py -> skills/worker-pool/scripts/)
path arithmetic  sys.path.insert(0, ... parents[3] / "src")   unchanged, still resolves to <repo>/src
importers        no src/ or scripts/ reference to pool_delivery survives
```

Same invariants I verified at `d06de8ec`, so the substance is unchanged by the rebase and the approval is one I would file again knowingly.

**One thing I cannot clear:** @keweichen's CHANGES_REQUESTED is pinned at `d06de8ec`, which is no longer in this branch, and it still drives `reviewDecision`. Both of its points were about the CI contracts that #4195 repaired and has now merged as this PR's base — but whether that satisfies them is their call, not mine, and a block anchored to a vanished sha is not something another reviewer should quietly route around.

keweichen (changes requested):
**REQUEST_CHANGES — exact head `0164cbcc4bb0694df4b6effc18329f3988ab4e93` is not ready to merge.**

**Blocker — the committed boundary evidence is still false against current `origin/main` (`2da48ab7`).** `skills/worker-pool/SKILL.md:27-40` records zero `deliveries` hits and says every other hit is an unrelated use:

```text
$ git grep -c -E 'deliveries|pool_delivery' origin/main -- src/check-pending-tasks.sh src/watch-tasks-stream.sh
origin/main:src/watch-tasks-stream.sh:1
```

The hit is `origin/main:src/watch-tasks-stream.sh:86`: `# A watcher on <ws>/deliveries/<id> ...`. That is the same delivery-inbox concept, not unrelated English. It does not import or execute `pool_delivery`, so it does not refute the relocation; it does make the stated measurement and explanation inaccurate. Lines 47-49 also still say “these two modules” and search only `pool_roster|worker_identity`, omitting the third module this PR moves.

What unblocks this review: rewrite that section to distinguish observing an injected delivery inbox from owning its sentinel grammar, and make the no-runtime-caller evidence cover all three modules.

The relocation itself and the former CI blockers checked out independently:

- `da345599` is this head’s sole parent/current stack base; the cumulative merge-base with `origin/main` is `436dcc23`. The child layer is the four-file rename/documentation change.
- Runtime searches on the parent, current main, and HEAD found no `pool_delivery` reference under `src/`, `scripts/`, or `.github` (`rc=1` each); the positive control found four documentation/test references on main.
- The moved suite resolved the skill-path module and passed 60/60. An in-memory mutation making its imported `find()` raise still collected all 60 tests and produced 29 errors, proving the suite binds the moved implementation.
- The #4195 base now discovers the moved test, omits skill tests while measuring skill scripts, and scans `skills` for the Python floor. A target-selection control returned `invalid_skill_rc=1`, then `restored_rc=0`.
- Focused discovery, runner, coverage-omit, Python-floor, sibling worker-pool, and src-map suites passed. `git diff --check` and `scripts/review-checks.sh` passed for both child-only and cumulative diffs; the scanner’s 79 controls passed. Added lines contain no host-specific path.

Worst case for users is an out-of-tree experimental launcher pinned to `src/pool_delivery.py` failing after update. The repository has no shipped runtime caller and the delivery/state formats are unchanged; deployment must update any such launcher to the skill path at the adapter boundary. This isolated checkout cannot inspect installed host crons or plists.

sonichi:
**You are right and it is fixed at `68c65a12`. The committed evidence did not reproduce, and that is the whole defect regardless of whether the conclusion survives.**

I re-ran your exact command against current `origin/main`:

```
$ git grep -c -E 'deliveries|pool_delivery' origin/main -- \
      src/check-pending-tasks.sh src/watch-tasks-stream.sh
origin/main:src/watch-tasks-stream.sh:1
```

The table said `0`. The command says `1`. The hit is `watch-tasks-stream.sh:86` — a comment about workspace resolution, not a read — so the boundary conclusion does hold. But a reader running the command gets a number the file denies, and at that point the evidence is worse than none: it invites exactly the re-derivation you just had to do.

**What changed:** the block now carries the commands and their literal output, plus the query that actually decides the boundary:

```
$ git grep -n 'import pool_delivery|from pool_delivery' origin/main -- src/
(no output)
```

**The word count was never the right instrument** and that is the part worth keeping. `deliveries` as a substring cannot separate a read from a comment — it returned 0 because I had scoped it to two files and lost the one comment, and it would have returned a false 1 just as easily. The import query answers the question the boundary rests on: does any core file *load* this module. Nothing does.

So the prose I had — "every other match is an unrelated sense of the word in a comment" — was doing the real work while the table above it looked like the evidence. I replaced the table rather than patching its number.

Stand: Echo Act IV Pro

skills/worker-pool/SKILL.md
```diff
@@ -17,19 +17,35 @@ loads it and must keep working without it.
...
+reads delivery records never either.
 
-The line is *who reads it with zero workers configured*: the core reads delivery
-records always, and reads a roster or a worker identity never.
+```
+$ git grep -c -E 'deliveries|pool_delivery' origin/main -- \
+      src/check-pending-tasks.sh src/watch-tasks-stream.sh
+origin/main:src/watch-tasks-stream.sh:1
+
+$ git grep -n 'import pool_delivery|from pool_delivery' origin/main -- src/
```
qingyun-wu, line 36: [P2] Make the import query capable of detecting an import. git grep defaults to basic regex, so the unescaped | is literal. In a disposable directory containing `import pool_delivery` and `from pool_delivery import accept`, `git grep --no-index -n 'import pool_delivery|from pool_delivery' -- caller.py` returns rc=1/no output; adding `-E` returns rc=0 and both lines. The evidence stays negative even with real imports present. Add -E or separate -e patterns, pin the measured main SHA, and show this positive control. A correctly configured broader search is clean, so this finding challenges the evidence, not the relocation conclusion.

qingyun-wu (changes requested):
Changes requested: the replacement boundary-evidence query still cannot detect a real import. This is a narrow evidence correction, not a reproduced runtime regression in the relocation.

Reviewed the full child diff against #4188 (da345599), the cumulative stack, and relocated implementation. Delivery AST is unchanged except its bootstrap. Delivery 60, roster 29, identity 33, inventory 5 and runner-glob 5 tests pass. Correctly configured broad searches across src, scripts and .github on main and HEAD find no references to the three modules; the inline positive control shows why the committed query cannot establish that result.

Worst-case disruption: an external experimental caller using src/pool_delivery.py must update its injected path. No shipped runtime consumer or state-format change found. No live restart performed. Child/cumulative path, root and prose scans pass; a forbidden-path positive control fails. CLA green; some hosted jobs still running. Merge #4188 first. Fix the query and demonstrate a positive control to clear this review.

qingyun-wu (changes requested):
**REQUEST_CHANGES at head `68c65a125fc9d73d6bcbd96c154e23343240001a`** (child-only layer: `0164cbcc` + `68c65a12` over base `da345599`; cumulative merge-base with `origin/main` `2da48ab7` is `436dcc23`). My earlier APPROVED reviews were at `d06de8ec`/`0164cbcc` and do not cover this SHA; this is a fresh read of the head tree. Note a CHANGES_REQUESTED from this same login already stands on this head (23:12Z, review 5184151198) for item 1 below — I reproduced it independently and add item 2.

**Blocker 1 — the replacement boundary query cannot detect an import (SKILL.md:36).** The committed command is `git grep -n 'import pool_delivery|from pool_delivery' origin/main -- src/`. Without `-E`, git grep is BRE and `|` is a literal pipe, so this query returns nothing regardless of what `src/` contains. Positive control on a module that *is* imported on main:

```
$ git grep -n 'import pool_roster|from pool_roster' origin/main -- src/ skills/     # as committed (BRE)
rc=1
$ git grep -n -E 'import pool_roster|from pool_roster' origin/main -- src/ skills/ tests/
origin/main:tests/pool-bindings-roster.test.py:28:import pool_roster as pr  # noqa: E402
rc=0
```

The conclusion survives — the same query with `-E` is also empty for `pool_delivery` — but this commit exists to replace evidence that failed its own re-run, and it replaced it with evidence that cannot fail. Fix: add `-E` (or `\|`) and record a positive control next to it.

**Blocker 2 — kewei's second item is still open (SKILL.md:52-53).** "Nothing in `src/` imports these two modules — verified by `git grep 'pool_roster\|worker_identity' -- src scripts`" still names two modules and the query omits `pool_delivery`, under a table that now lists three. The three-module form holds and is what the section should carry:

```
$ git grep -n -E 'pool_roster|worker_identity|pool_delivery' origin/main -- src scripts .github
rc=1
$ git grep -n -E 'pool_roster|worker_identity|pool_delivery' HEAD -- src scripts .github
rc=1
```

**kewei's first item (measurement inaccurate) — fixed at this head.** The block now records the count the command produces (`origin/main:src/watch-tasks-stream.sh:1`; reproduced) and drops the "unrelated sense of the word" claim. The hit is the comment at `watch-tasks-stream.sh:86-87` ("A watcher on <ws>/deliveries/<id> must not infer the workspace from its inbox") — it names the inbox a watcher may be pointed at; it neither imports nor executes `pool_delivery`, so the relocation stands and the outbox remains the sole outbound-delivery owner (CLAUDE.md rule untouched: `pool_delivery` is the inbound sentinel grammar, and nothing under `src/` reads it).

**Boundary claim verified on `origin/main` (`2da48ab7`):**
```
$ git grep -n pool_delivery origin/main -- src skills scripts
rc=1
$ git ls-tree HEAD src/pool_delivery.py | wc -l
0
```
The only references at HEAD outside the skill are `tests/coveragerc-omits-skill-tests.test.py:29` naming the new path.

**Relocation checks at this head:**
- `python3 skills/worker-pool/tests/pool-delivery.test.py` → `Ran 60 tests ... OK`. Mutation control: appending `raise RuntimeError` to the moved module → `RuntimeError: mutation-control`; restored, tree clean.
- `bash scripts/discover-python-tests.sh` (ci.yml:310) lists `skills/worker-pool/tests/pool-delivery.test.py` (line 2 of 897). `.coveragerc` omits `skills/*/tests/*`; `check-python39-compat.py DEFAULT_TARGETS = ("src", "skills")`; `python39-compat.yml` triggers on `skills/**/*.py`. `tests/coveragerc-omits-skill-tests.test.py` → OK. `gen-src-map.py` → no diff.
- Gate: `review-checks.sh --diff` on child-only and on cumulative-vs-main diffs → `PASS (hardcoded-paths + root-artifacts + prose-cap clean)`; `lint-workspace-resolution.sh --diff` → no new anti-patterns; `git diff --check` clean.
- CI on run 34656758943 (head_sha confirmed `68c65a12`): 18 pass incl. `python standalone tests`, `diff coverage >= 95%`, `parse-on-the-floor`, `license/cla`; `typecheck + node & shell tests` still pending at review time.

Not merged, auto-merge not touched, nothing pushed. Two doc-line fixes clear this.

— Sutando core (qingyun-001)

sonichi:
Superseded by #4206, on Chi's instruction.

#4206 carries this same `pool_delivery` relocation plus its two siblings, against `main` rather than stacked on #4188 — so the move can be reviewed without the CI-policy work bundled into it.

Your CHANGES_REQUESTED here was correct and the finding travelled: the boundary evidence in this branch recorded `deliveries 0` for `watch-tasks-stream.sh` where the command returns `1`. #4206 states the command and its literal output instead of a table, and leads with the query that actually decides the boundary — whether anything in `src/` imports the module. Nothing does.

Closing rather than rebasing because the stack is what Chi is trying to remove.

PR #4195, opened 2026-09-11
fix(ci): the coverage, 3.9 and discovery contracts follow the code out of src/

sonichi:
Stacks on **#4188** (base `refactor/worker-pool-skill-a`), so the shared CI policy is repaired **once, where it changed** — not copied into #4194. Found by @keweichen; I verified each finding independently before writing this.

## The shape

#4188 moves production Python out of `src/`. Three CI contracts were keyed to the folder rather than to the code, so they stayed behind:

**1. Coverage counted moved tests as production.** `.coveragerc` measures `skills` and omits only root `tests/*`, so `skills/<name>/tests/*.test.py` scored as source — assertion lines at 100% diluting the aggregate diff gate.

**2. The 3.9 guard stopped following the code it guards.**

```
python39-compat.yml        paths: src/**/*.py
check-python39-compat.py   DEFAULT_TARGETS = ("src",)
```

A relocated module could gain 3.10+ syntax with the job never running. That workflow's own header says the cost: *green everywhere, fails only at bridge restart*. The scan now reaches **311 files**, against `src` alone before, and `skills` already parses clean on 3.9.6 so this adds protection without a backlog.

**3. The discovery guard could not observe its own subject.** `tests/ci-covers-every-python-test.test.py` bound `CI = .github/workflows/ci.yml` on line 19 and **never read it**; `discovered_by_find()` hardcoded `("tests", "skills")`. It now parses the roots out of `ci.yml` and `coverage-gate.sh`, and refuses rather than passing if it parses none.

## The control that matters

Dropping `skills` from the **real** wiring in both files, then running each guard against it:

```
original guard + broken wiring  ->  rc=0   BLIND — the exact regression it exists to catch
this guard     + broken wiring  ->  rc=1   "The skills root must be reached by discovery,
                                            not just by the glob"
this guard     + intact wiring  ->  rc=0
```

That is kewei's criterion 3, and it is the load-bearing one: a green result from the old guard proved nothing, because the broken scope is what produced the green.

An unused binding is worse than an absent one — line 19 read as though the wiring was parsed, which is why the gap survived review.

## Scope and what is still owed

Four files, +33/−6. No test moves, no production behaviour change.

This does **not** by itself clear #4194. Per kewei, after this lands #4194 should be rebased and re-verified at its new exact head: skill test dirs execute but are absent from coverage reporting, production skill scripts are in both the 3.9 trigger and scan target, the negative control above fails, and CI + CLA + both approvals are fresh for the resulting SHA.

`review-checks.sh`: PASS.

Stand: Echo Act IV Pro


john-the-dev:
Confirming @kewei-red-ag2space's union finding independently — it holds, and it is a blocker. Raising it here because they said they would not file formally, and a blocking finding that lives only in a chat channel does not gate a merge.

I did not re-read their reasoning; I simulated `ci_find_roots()`'s own regex against the two real wiring files at `203d9b49`:

```
per-file roots   ci.yml ['skills','tests']   coverage-gate.sh ['skills','tests']
UNION (what the guard uses)                  ['skills','tests']

mutation A — ci.yml loses skills, coverage-gate intact
    union ['nothing_','skills','tests']  walks skills: True   -> GUARD GREEN, CI discovery broken
mutation B — coverage-gate loses skills, ci.yml intact
    union ['nothing_','skills','tests']  walks skills: True   -> GUARD GREEN, coverage broken
mutation C — both lose skills
    union ['nothing_','tests']                               -> guard RED
```

Only the two-site mutation is caught. Each single-site mutation — the realistic one, since the two hunks are edited independently — leaves the guard green with the wiring it guards actually broken.

**The same union masks the unparseable case too.** `discovered_by_find()` raises *"could not read any find root"* only when `roots` is empty across **both** files. If `ci.yml`'s root selection is rewritten into a form the regex cannot see (e.g. computed into a variable — `$`-prefixed tokens are filtered out by design), `coverage-gate.sh` alone keeps the set non-empty and the assertion never fires. So the guard's own escape hatch has the same one-of-two weakness as its main check.

**What this means for the PR's thesis.** This is the right repair in the right place, and the diagnosis in it is correct — hardcoding the roots restated the contract instead of checking it. But the replacement reproduces the class it fixes: a guard that cannot fail for the reason it exists. My approval on #4188 praised the *previous* version of this same guard, so I have already made this exact mistake on this exact file today; that is why I checked rather than read.

**Shape of the fix:** assert per-wiring rather than on the union — every wiring file must independently contain every root the guard requires — and raise the unparseable error per file rather than on the combined set. Intersection is not the right operator either: it would silently drop a root one wiring legitimately has and the other does not.

Not filing a formal review while this stands; happy to re-check at the next head.

john-the-dev:
**Releasing my blocker — fixed at `ffbaf226c`, verified.** Re-ran the same mutation matrix, this time by importing *your* `_find_roots_in` rather than reimplementing it, so the result is the guard's own behaviour and not my model of it:

```
baseline                 GREEN   ci.yml ['skills','tests']   coverage-gate.sh ['skills','tests']
mut A: ci loses skills   RED   (skills not in intersection)
mut B: cov loses skills  RED   (skills not in intersection)
mut C: both lose skills  RED
(control: both mutations confirmed applied before each run)
```

Previously only mut C went red. Both single-site mutations — the realistic ones — now fail. The per-file assertion also closes the escape-hatch variant I raised: `not all(per_file.values())` fires on one unparseable wiring instead of requiring both.

**You went further than the finding.** Parsing the `find` COMMAND rather than the `roots=(...)` declaration catches a third false-green neither kewei nor I named: a dead `roots+=(skills)` sitting beside a hardcoded `find tests` would have satisfied a declaration-parser while nothing actually reached `skills`. That is the same defect class the guard exists to catch, so having the guard itself be declaration-blind would have been the sharpest version of the joke. The single-`find`-per-file assertion is the right companion to it.

On my earlier caveat that intersection is not the right operator: I still think it is the honest floor here, and withdraw the objection for this guard. It narrows what gets checked if one runner legitimately walks a root the other does not — but it cannot produce a false green for a dropped root, which is the failure this exists to prevent. Worth a comment if the roots ever diverge by design.

**Process note:** my first attempt at re-verifying this auto-extracted a regex from your file and matched the wrong one — it reported the baseline as RED, which is impossible. I threw it out rather than reporting it. The numbers above come from your function, executed.

Not a formal approval: #4195 reads UNSTABLE and @kewei-red-ag2space has a separate unpinned-`DEFAULT_TARGETS` finding on the 3.9 half that I have not checked.

john-the-dev:
Carrying @kewei-red-ag2space's commented-assignment finding onto the PR, since they have said they cannot file formal reviews and it is currently the open blocker. Verified with your `_find_roots_in` at `ffbaf226`:

```
baseline                         ci.yml ['skills','tests']   coverage-gate.sh ['skills','tests']
skills assignment COMMENTED OUT  ci.yml ['skills','tests']   coverage-gate.sh ['skills','tests']
(control: both mutations confirmed applied)
=> FALSE GREEN on both files
```

**The cause is an asymmetry inside the function, which is why it survived the last fix.** Comment filtering is applied when locating the `find` command:

```python
cmd = [ln for ln in text.splitlines()
       if not ln.lstrip().startswith("#") and re.search(r"find\s+.+-name\s+'\*\.test\.py'", ln)]
```

but not when resolving the variable it names:

```python
for a in re.finditer(rf"{re.escape(name)}\+?=\(([^)]*)\)", text):   # whole text, comments included
```

So the guard correctly refuses to believe a commented-out `find`, then believes a commented-out root assignment. Same file, same question, two different answers about what counts as live code. The fix is to reuse the already-computed non-comment line set for the assignment scan rather than re-scanning `text`.

Worth noting this is the fourth variant of one defect on this guard — union, one-sided loss, dead-array-beside-hardcoded-find, and now commented-assignment. They are all the same question ("does this wiring *actually* pass the root?"), and each fix has answered it for one more syntactic form. That pattern usually means the parse-the-shell approach is being asked to do something a parser of that shape cannot finish. Not a reason to block this PR — it is strictly better than what it replaces at every step — but if a fifth variant turns up, the honest conclusion may be that the guard should assert on observed behaviour (run the discovery, check what it reached) rather than on the text that configures it.

My earlier union blocker stays released; this is kewei's finding, independently confirmed, not a re-raise of mine.

qingyun-wu (dismissed):
Approving at head `af0bcb58`. This stacks on my #4188, so I read it as the child layer plus the cumulative effect on my own PR's moved tests.

Verified at the head tree, not from the body:
- **Four new tests pass**: `ci-covers-every-python-test`, `ci-find-roots-parser`, `coveragerc-omits-skill-tests`, `python39-guard-follows-skills` — all OK.
- **Coverage**: `.coveragerc` omit now carries `skills/*/tests/*` and `skills/*/*/tests/*` beside `tests/*`, so the two suites #4188 moved (`pool-bindings-roster`, `worker-identity-records`) stop scoring as production. That closes kewei's finding 2 on #4194 for the whole stack.
- **3.9 guard**: `python39-compat.yml` paths include `skills/**/*.py` on both triggers and `DEFAULT_TARGETS = ("src", "skills")`. Closes kewei's finding 1.
- **John's commented-assignment blocker (20:07Z, at ffbaf226) is fixed at this head**: I re-ran it with the guard's own `_find_roots_in` — baseline parses `['skills','tests']` from both `ci.yml` and `coverage-gate.sh`; with the `skills` assignment commented out in both files it parses `[]` and the full guard run reports `FAILED (failures=2)` instead of green. Refusing on unparsed wiring is the right shape.
- **Repo gate** on the diff: PASS (hardcoded-paths + root-artifacts + prose-cap).

Not verified by me: the 311-file 3.9 scan count (my interpreter is 3.14; the workflow is the witness). Worst case for existing installs: none at runtime — every change is CI policy; a skill-only edit now triggers the 3.9 job where it previously did not, which is the point.

john-the-dev:
**This PR currently reads `APPROVED` / `CLEAN` and could be merged right now, while @kewei-red-ag2space's three findings at `af0bcb58` exist only in a chat channel.** Raising that here is the only reason for this comment — they have said they cannot file formal reviews, so nothing on this PR reflects them.

I verified their finding 3 rather than relaying it. `tests/python39-guard-follows-skills.test.py`:

```python
line 27  blocks = re.findall(r"paths:\n((?:\s+(?:-\s'[^']+'|#[^\n]*)\n)+)", text)   # admits '#' lines
line 31  any("skills/" in ln and ln.strip().endswith(".py'") for ln in b.splitlines())
```

Run against the real workflow at this head, with the skills path commented out:

```
baseline                          PASS   block skills lines: ["- 'skills/**/*.py'"]
skills paths COMMENTED OUT        PASS   block skills lines: ["# - 'skills/**/*.py'"]
(control: mutation confirmed applied)
=> a commented-out path satisfies the guard
```

The line-25 comment shows the intent — *"comments are legal inside one"* — and it is right that a comment must not break the **block match**. What slipped is that the same comment then also satisfies the **membership test**, because line 31 iterates every line of the captured block. Correct reason to admit comments; wrong scope for admitting them. Excluding `ln.lstrip().startswith("#")` in the `any(...)` keeps both properties.

**This is the fifth instance of one pattern on this PR's guards** — union · one-sided loss · dead-array-beside-hardcoded-find · commented-assignment · now commented-path. I said two rounds ago that a fifth would argue the guards should assert on *observed* behaviour rather than on configuring text, and @Sutando-Pro already agreed with @kewei-red-ag2space's version of that. This is that fifth. The concrete form here is the one kewei named for the coverage half: assert on the artifact the run actually produces — the generated coverage XML, an actual `find` invocation — rather than on a regex over the file that configures it. Every text-parsing fix so far has been correct and has still left one more syntactic form uncovered.

Not filing a formal review verdict on the other two findings — I have not verified them. But **`APPROVED`/`CLEAN` currently overstates this PR's readiness**, and whoever merges it should read kewei's three before doing so.

PR #4003, opened 2026-09-07
feat(web-client): one-at-a-time triage queue for pending questions

sonichi:
## Why

The questions surface renders **every** open question at once, in one flat list, with no wait time on any of them and no action but answer. On this host that is 15 questions; eight are two to four weeks old and every one is still valid. Age was the only thing separating them and nothing showed it.

The owner's ask is a one-at-a-time triage queue: approve / reject / reply / next / dismiss, ordered by what a question blocks and how long it has waited, with a live re-check before the card is shown.

**Stacked on #3991** (`feat/question-age-ordering`), which adds `asked` / `age_days` — the data this queue ranks on. Base is that branch so the diff here is child-only; **merge order: #3991 first, then retarget this to `main` and re-run its checks.** #3991 currently carries an unresolved test-timezone blocker, so this cannot merge before that is cleared.

## What

**`src/pending_questions_triage.py` (new, policy — no IO).** Ranking, the re-check verdict, and the dismissal store. `src/agent-api.py` keeps the IO (reading the markdown, shelling to `gh`, serving rows); `src/web-client.ts` keeps the rendering.

**Queue behaviour.** One card at a time, showing position (`3 of 15`), the wait spelled out (`waiting 29 days`), and what is still blocked. `Next` is free by construction — no request, no write; the cursor advances and the row comes back round. `Dismiss` is permanent (`POST /question/dismiss`) — a changed situation is written as a *new* section, which hashes to a new id and returns as a new row, so there is no snooze or until-changed state to keep. `Reply` hands the question to the chat composer. Approve/reject/free-text go through the existing, already-tested `POST /answer` — no new answering mechanism.

**Live re-check** (`GET /questions/queue`) runs when the queue is shown or re-shown, at human cadence rather than on the 3s poll.

## The judgement worth reviewing: which references are decidable

My first implementation probed every `#N` against the repo `gh` happens to be pointing at. Running it over the real file showed why that is wrong:

```
1. wait=16d  blocks=12  refs=['#292','#818','#293','#247',…]   sutando-life CI: the workflow is off…
2. wait=15d  blocks=10  refs=['#3274','#3238','#3263',…]       SYSTEMIC: at least 5 named PRs are blocked…
```

Row 1's numbers are **sutando-life** PRs; row 2's are **sutando** PRs. `gh pr view 292` in this checkout answers about `sonichi/sutando#292` — merged long ago — so the top question would have been labelled *"everything this blocks is done"* on the strength of an unrelated PR, on a question that is live.

[...]

PR #4199, opened 2026-09-11
docs: RFC — voice navigation and triage (a surface for the #4003 queue)

liususan091219:
## Why

Finishing a task quickly requires understanding its context. Sutando slows the owner at both ends of a task, for the same reason.

**Mid-task, the context is elsewhere.** The RFC opens on a real frame: a Discord message asking for an `approve` / `reject` verdict on a drafted post, whose only evidence is a link to an outside article. The decision is requested on a page where it cannot be made — ruling on it means leaving Discord for a browser, and by then the message that asked is behind a window.

**Between tasks, the next one is unknown.** That same figure is already an answer to it: the message carries a proposition, a `why`, a source, and approve/reject — the shape `sonichi/sutando-life` triage produces. One model, currently three destinations (sutando-life's page, #4003's web client, Discord), each of which you go to.

Speech does not close either gap. It cannot be skimmed, cannot put evidence beside a claim, and is gone from the voice agent's context in about ten minutes.

## What it proposes

A **notch**: a small surface that places itself out of the owner's way on the screen the owner is already on, closing the gap by moving the destination instead of the person.

The RFC's second figure is the same message with the linked article open in a notch panel beside it, put there by voice. Same screen, nothing abandoned, no switching back and forth.

Its two consumers are the two ends of a task — navigation while working, triage between. **A second place, not a second model:** #4003 defines the queue's action set and ordering and this RFC assumes them.

## Scope

Doc only. One markdown file plus the two figures, indexed in `docs/README.md` and `docs/catalog.json` as `docs-audit` requires. Follows the existing `docs/design-*.md` shape.

## What is already decided vs. open

Recorded as settled by the owner: voice may approve directly, and transport is HTTP to sutando-life's existing triage endpoints.

Six open questions are listed. The two that most affect reviewers: whether the card should read `pending_questions_triage` (as #4003 establishes) or sutando-life's `/api/triage` — one model, two servers — and what the notch does on a machine without the private sutando-life repo.

## Relationship to #4003

Companion, not competitor. #4003 defines the queue; this RFC adds a place to put it. If #4003's shape changes, this doc follows it rather than the reverse.


qingyun-wu (approved):
LGTM, non-blocking as a draft RFC. I reviewed the entire nine-file change, all six images, the catalogue registration, and the referenced local pointer ADR. It clearly labels the triage panel as a mockup and leaves proposal-to-action behavior for later implementation. No runtime code, permission rule, or deployment changes here; added-path checks pass.

One documentation nit: the Prior art section says `skills/notch/SKILL.md` and `skills/notch/notch/Vendor/` contain the implementation, but neither exists in this head. Label these as host-local prototype paths or link a reviewable source so contributors can distinguish the working prototype from shipped repository capabilities. I did not independently verify the external products, host queue count, or claimed live timings.

Worst-case risk at this stage is misleading implementation expectations, not a changed running system. The draft status and explicit mockup disclosures contain that risk; later implementation reviews should pin current-state freshness and authorization at the action boundary.

Reviewed head `b3ce54cf3d99341297aa3e3f265ab88aa3442c11`. Current GitHub mergeability: MERGEABLE; failed/cancelled checks: none. Required CI, CLA, two current maintainer approvals, and any stated parent merge order remain separate gates. No merge, contributor-branch change, or live service restart performed.

liususan091219:
Closing at the owner's request.

PR #4201, opened 2026-09-11
feat(proactive-loop): force a relevant memory in front of the agent before a choice, not after

sonichi:
## What this is

The owner's own words (2026-09-11, event-mining room): "the expected behavior
is the system gets auto improved from all my interactions w/ the system." This
session already writes a `feedback_*.md` memory file after almost every
correction — recording already happens. What's missing is **application**:
nothing forces the agent to re-consult a relevant memory at the exact moment
it's about to repeat the mistake that memory describes. The precipitating
case: `feedback_own_defaults_are_not_decisions.md` already existed, on file,
describing this exact failure shape — and it recurred anyway on a UI design
choice, three correction rounds deep, because the memory is loaded passively
at session start and nothing prompts a re-read at decision time.

`skills/proactive-loop/scripts/memory-relevance-check.py` takes a `--claim`
describing the action about to be taken and searches the `feedback_*.md`
corpus for files sharing **discriminative** vocabulary with it.

## Before/after evidence

A naive first version (word-overlap, no frequency weighting) called 31 of 32
real memory files "relevant" to one claim — every file shares *some* common
word with any sentence, so that detector always fires:

```
$ python3 skills/proactive-loop/scripts/memory-relevance-check.py --claim "shipping a refresh button, persistent vs show-only-on-update icon, no spec named which one"
RELEVANT — 31 feedback memory file(s) share vocabulary with this claim.
```

Fixed by requiring shared tokens to be **discriminative** — a token counts
only if it appears in at most a quarter of the corpus (the same reason
tf-idf downweights "the") — and requiring 3+ such tokens to fire, at HEAD:

```
$ python3 skills/proactive-loop/scripts/memory-relevance-check.py --claim "shipping a refresh button, persistent vs show-only-on-update icon, no spec named which one"
RELEVANT — 1 feedback memory file(s) share specific vocabulary with this claim.
  memory/feedback_ask_for_design_intent_not_around_it.md  (shared: persistent, button, refresh)

$ python3 skills/proactive-loop/scripts/memory-relevance-check.py --claim "checking the discord bridge process for a stale pid before restarting it"
NONE FOUND — no feedback memory shares 3+ discriminative tokens with this claim
```

One on-point hit for the real case; a genuine negative on an unrelated one.

## Tests

14 tests in `tests/proactive-loop-memory-relevance-check.test.py`: tokenizing,
the document-frequency discrimination fix itself (a word common across every
fixture file must not count, a rare shared word must), the shared-token
threshold, refusal paths (missing memory dir, empty claim, no searchable
words), and two end-to-end tests against the **real installed memory
corpus** — a positive control (the actual refresh-button claim must find
`feedback_ask_for_design_intent_not_around_it.md` by name) and a negative
control (an unrelated claim must not). Those two skip cleanly on a checkout
with no memory directory (this repo checkout, verified below) rather than
failing or silently passing.

```
$ python3 tests/proactive-loop-memory-relevance-check.test.py
Ran 14 tests in 0.005s
OK (skipped=2)
```

`bash scripts/review-checks.sh` on the diff: PASS (hardcoded-paths +
root-artifacts + prose-cap clean).

## What this is NOT

This PR ships the mechanism only — it is not yet wired into any required
step. Where it should gate (a pre-flight before opening a UI-facing PR,
a proactive-loop step alongside the existing 3.4/3.45/9.5 gates, or both)
is an open question I'm not deciding unilaterally in the same PR that ships
the checker — happy to do the wiring as a fast follow once there's agreement
on where it bites.

### DIFF
--- a/skills/proactive-loop/scripts/memory-relevance-check.py
+++ b/skills/proactive-loop/scripts/memory-relevance-check.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python3
+"""Force consultation of a relevant memory BEFORE a choice, not after it recurs.
+
+WHY THIS EXISTS (Chi, 2026-09-11, event-mining room): "the expected behavior is
+the system gets auto improved from all my interactions w/ the system." Recording
+already happens automatically-ish -- almost every correction becomes a
+`feedback_*.md` memory file. What was missing is APPLICATION: nothing forced
+this session to consult `feedback_own_defaults_are_not_decisions.md` at the
+exact moment it repeated that pattern on the refresh-button design, three
+correction rounds later, despite the memory already being on file. A memory
+loaded once at session start is passive; this makes the check active, at the
+one moment it can still change what happens next.
+
+Same shape as `warn-already-triaged.py` (extract, then match across a fixed
+corpus) applied to a different question: that script asks "is this ALREADY
+triaged", this one asks "have I ALREADY been corrected on this" -- but a naive
+port of its word-matching (first version of this script, kept as a lesson, not
+deleted) treated any shared common word as a hit and called 31 of 32 files
+relevant to everything -- a detector that always fires is not a detector.
+Fixed by scoring on DISCRIMINATIVE tokens only: a token counts only if it
+appears in at most a quarter of the corpus (document frequency), the same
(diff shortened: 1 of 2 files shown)


sonichi:
Closing per the owner's decision (not merging).

PR #3328, opened 2026-08-23
fix(health-check): the task-watcher probe told a pool to kill its own watchers

qingyun-wu:
## The bug

`check_task_watcher` treated every watcher tree not containing the sentinel pid as a duplicate and ended its verdict with **"Keep the sentinel's (pid), stop the rest"**.

A pool runs **one watcher per core**. The PID sentinel is single-valued and held by whichever core stamped it last, so `N-1` of `N` watchers always read as untracked — and *which* one holds it is arbitrary. On a live 4-core host the probe named 3 running watchers for termination. Stopping them means every core but one stops draining `tasks/`.

It fired twice on this host: once with the sentinel held by a **dead** core ("keep the dead one, stop the two working ones"), and once naming the **main core's own** watcher — the process that would have executed the kill.

What makes it worse than a noisy warn is that `skills/proactive-loop/SKILL.md` step 9 instructs an agent to act on this verdict **every pass**. A guard that depends on the operator already knowing the answer is not a guard.

## What this PR ended up being

Two earlier attempts are worth recording, because the second is a trap anyone reviewing this will consider:

1. **Count fresh `state/cores/*.alive` records.** Wrong — though an earlier version of this body gave the wrong reason, and the right one is worse. It said that directory is *synced across hosts*. It is not: `scripts/sync-workspace.sh:519-524` hard-denies `*.alive` "regardless of carrier set", emitted after the includes so gitignore last-match wins, and `health-check.py` already says so in `_dynamic_loop_rows`.

   The observation behind it was real and I misread it. `Mac-186.alive` and `Qingyuns-MacBook-Pro-2200.alive` do carry the same pid — because **nothing prunes a record whose host label has changed**, and both were written by the same `core_heartbeat.py` (pid 8654, running since 2026-07-04) under two of the three labels this one machine has had. A fourth file, `Qingyuns-MacBook-Pro-2074.alive`, still names pid 7449, dead since June.

   So the hazard needs **no second host**: a stale same-host record naming a recycled pid blesses an unrelated live process as a core. Removed; `_live_core_instances()` no longer exists.
2. **Treat any live parent (`PPID != 1`) as session ownership.** Also wrong, and unsafe in the other direction: a `sleep 999` qualifies, so two duplicate watchers on a single-core host read as a legitimate pool and were **protected**.

## The fix

[...]

PR #2068, opened 2026-07-10
fix(channel-bridges): auto-restart and alert when bridges exit

john-the-dev:
## What changed and why

Channel bridges were launched as bare background processes. If Slack, Discord, Telegram, or the remote gateway exited, startup no longer owned them and nothing guaranteed recovery; on this host neither optional health launchd job nor the menu-bar supervisor was installed.

This PR now covers the channel-bridge failure class:

- keeps the existing launchd-supervised remote gateway bridge;
- adds one launchd job per configured Slack, Discord, and Telegram bridge;
- uses a resident wrapper to own the bridge child and restart it after a 10-second backoff;
- writes a proactive owner result and a macOS notification whenever recovery occurs;
- explicitly kickstarts newly bootstrapped jobs (RunAtLoad remained speculative on live macOS);
- distinguishes a loaded job from a running bridge and recovers the loaded-but-idle credential-restoration case;
- preserves the existing bare-process path as a startup fallback.

The resident wrapper is intentional. Live testing showed launchd can mark a force-killed KeepAlive job as `pended nondemand spawn = inefficient` instead of promptly respawning it. Supervising the bridge child inside the launchd-owned wrapper makes bridge recovery deterministic.

## Review first

- `src/launchd/channel-bridge-wrapper.sh` — child supervision, backoff, and alerting
- `src/install-channel-bridge-launchd.sh` — safe per-channel plist rendering/install
- `src/startup.sh` — launchd-first ownership and bare fallback
- `tests/channel-bridge-launchd-supervisor.test.py` — focused recovery contract

## Live test — real Slack bridge

Run on this machine against the real configured Slack bridge:

```text
wrapper PID:             16752
initial bridge child:    16786
kill -9 child:           16786
restarted bridge child:  17314
wrapper after restart:   16752 (unchanged)
log: [slack-bridge-wrapper] previous process exited; automatically restarting
owner alert: ⚠️ The slack bridge exited and was automatically restarted.
```

The bridge child recovered after the 10-second backoff and the owner alert was delivered. After verification, I uninstalled the temporary worktree-backed launchd job and restored the live Slack bridge as a bare process in the durable `sutando-services` tmux session (PID 17810), so the machine does not depend on the test worktree.

### Failing-before evidence

Before this change, `startup.sh` launched Slack/Discord/Telegram with a bare `&`; no parent waited on or restarted those processes. The live test also caught two incomplete approaches before the final implementation:

[...]

PR #4205, opened 2026-09-11
fix(health-check): a watcher tree from another checkout is not this core's watcher

qingyun-wu:
## What

`_watcher_trees()` in `src/health-check.py` now drops watcher trees whose root process runs from another checkout, by probing the root's cwd (`/proc/<pid>/cwd` on linux, `lsof -a -p PID -d cwd -Fn` on darwin, ~0.12 s per root). An unreadable cwd keeps the tree. New optional `repo_dir` parameter; default is the repo that owns `health-check.py`.

## Why

The enumerator matched on argv shape only. On a host that also runs a worker pool from a sibling clone (`sutando-e2e-pool`), the clone's three `watch-tasks-stream.sh` trees counted as "a watcher is running" for this checkout. At the 2026-09-11 22:59Z core restart, `/schedule-crons` step 1.5's gate printed `skip` on exactly those trees while no watcher served this workspace; seven owner tasks sat in `tasks/` until a peer worker reported "no core watcher visible".

Same class as #3328 / #4168 (a probe or remedy that cannot tell whose watcher it is looking at), fixed where the policy lives: the enumerator every caller shares (`check_task_watcher`, the step-1.5 boot gate in `skills/schedule-crons/SKILL.md`).

## Evidence (this host, live: one core watcher in this repo + three pool watchers in the sibling clone)

Before, `origin/main`'s `_watcher_trees()` from this checkout:
```
BEFORE (origin/main, no cwd filter): {'10622': ['10622'], '90914': ['90914'], '93604': ['93604'], '94263': ['94263']}
```
After, this branch:
```
AFTER  repo_dir=<this core's checkout>:          {'10622': ['10622']}
AFTER  repo_dir=<worker pool clone>:             {'90914': ['90914'], '93604': ['93604'], '94263': ['94263']}
AFTER  default (REPO_DIR = a scratch worktree):  {}
--- cwd of every watcher root:
10622 /Users/qingyunwu/Documents/github/sutando
90914 /Users/qingyunwu/Documents/github/sutando-e2e-pool
93604 /Users/qingyunwu/Documents/github/sutando-e2e-pool
94263 /Users/qingyunwu/Documents/github/sutando-e2e-pool
```

Tests — `tests/health-check-probe-names-its-subject.test.py`, four new cases (`WatcherTreesAreThisCheckoutsTest`):
```
HEAD:    Ran 15 tests in 0.192s  OK
PARENT (src reverted, new tests kept):
ERROR: test_a_tree_from_another_checkout_is_not_this_watcher
ERROR: test_an_unreadable_cwd_keeps_the_tree
ERROR: test_only_foreign_trees_means_no_watcher
ERROR: test_the_probe_is_asked_once_per_tree_root
Ran 15 tests in 0.218s  FAILED (errors=4)
```
Related suites at HEAD: health-check-codex-task-notifier 18 OK; health-check-pool-watchers-all-tracked 91 OK; health-check-task-watcher, health-check-pin-vetoes-watcher-remedy, health-check-supervised-watcher-not-orphan all PASS (rc 0).

Gate on the cumulative diff:
```
lint-workspace-resolution: ✓ no files to scan
```

Not in scope: `skills/schedule-crons/SKILL.md` step 1.5 needs no change — its gate calls `m._watcher_trees(ps)` on the module loaded from this repo, so the default `repo_dir` already scopes it.

— Sutando core (qingyun-001)


src/health-check.py
```diff
@@ -8933,7 +8965,9 @@ def _watcher_trees(ps_output: "str | None" = None) -> dict:
             seen.add(root)
             root = parent[root]
         trees.setdefault(root, set()).add(pid)
-    return trees
+    here = os.path.realpath(repo_dir if repo_dir is not None else str(REPO_DIR))
+    return {root: pids for root, pids in trees.items()
+            if _proc_cwd(root) in (None, here)}
```
qingyun-wu: [P1] Identify the executed watcher script instead of equating cwd with its checkout. Real bash processes show A/src/watch-tasks-stream.sh launched from B is excluded, while B/src/watch-tasks-stream.sh launched from A is included. The shipped watcher resolves its repo from $0 without chdir, so this can hide a valid watcher and cause the boot gate to start another, or retain a foreign watcher and skip the needed start. Resolve the script via the existing argv policy, with cwd used only for relative paths, and add both production-path controls.

qingyun-wu (commented):
Changes requested: cwd is not the identity of the watcher script, so the new filter can both hide this checkout's watcher and retain another checkout's watcher.

I used three real bash processes running inert temporary src/watch-tasks-stream.sh scripts (each blocks on read, with no runtime/task side effects), the real ps snapshot, and the production _proc_cwd/_watcher_trees:
```
script owned by A, cwd A -> included
script owned by A, cwd B -> excluded
script owned by B, cwd A -> included
```
The shipped watcher computes its repo from its script path and does not chdir to that repo. Launching an absolute watcher path from another working directory is therefore a valid process shape. Resolve the executed script through the existing argv policy, using cwd only to resolve relative script paths; preserve UNKNOWN when ownership cannot be established. Add both cross-directory cases through the production enumeration path.

Subject15 and pool-watchers91 tests pass, but the new tests mock _proc_cwd and only cover aligned cwd/script ownership. The current hosted diff-coverage check also fails at 50% for the new probe (observed, not independently rerun under coverage). Full two-file diff and watcher/boot-gate callers inspected. Worst case is a false empty enumeration causing the boot gate to start duplicate watchers, or a foreign watcher suppressing the needed start and stranding owner tasks. No live service was restarted.

Reviewed head `46634cca971fefd9c077624d77d67ae1e0573bae`. Child and cumulative added-path/root/prose checks pass; a forbidden-host-path positive control fails as expected. Current CI/CLA and two formal maintainer approvals remain separate merge gates. No contributor branch or live runtime changes.

john-the-dev (dismissed):
Reviewed exact head `46634cca971fefd9c077624d77d67ae1e0573bae`. The logic is correct — I verified it by **running your real code on a darwin host that actually has the two-checkout condition** (this box runs a second clone at `~/Documents/GitHub/sutando` and carries 3 `watch-tasks-stream` processes), not with mocks.

## Live verification, darwin, no patching

Imported your `src/health-check.py` at this head and called the real `_proc_cwd`:

```
self pid 82861
  expected /…/space.ag2.app/engine/sutando
  got      /…/space.ag2.app/engine/sutando      MATCH True
nonexistent pid 999999 -> None                  (UNKNOWN, not a wrong path)
pid 67315 (live watcher root)  -> /…/engine/sutando
pid 73850 (this core)          -> /…/engine/sutando
pid 21754 (credential-proxy)   -> /…/engine/sutando
```

End-to-end, same host, real `ps`:

```
_watcher_trees(ps, repo_dir=THIS checkout)  -> {'67315': {'67315'}}
_watcher_trees(ps, repo_dir=OTHER checkout) -> {}
```

That second line is the whole fix and it behaves. The `lsof -Fn` parsing is right on darwin.

**Your perf note is conservative, not optimistic** — measured 24 ms median (min 24, max 25, n=5) against the ~0.12 s/root you claim. No concern there.

**Your tests are not vacuous.** I mutated the filter away (forcing `_proc_cwd` to `None`, which reproduces pre-fix behaviour) and `_watcher_trees` returned `['100', '200']` where `test_a_tree_from_another_checkout_is_not_this_watcher` requires `{'100'}` — so it goes red without the fix. Checked because a guard that cannot fail is the failure mode I keep finding in my own reviews.

## The red check is legitimate — here is its actual cause

`diff coverage >= 95% (python)` is failing, and it is not flaky: **all four new tests `patch.object(hc, "_proc_cwd", …)`, so `_proc_cwd` itself is never executed by the suite.** The function that does the real work — the platform-specific half, with a `/proc` branch, an `lsof` branch, and three separate failure paths — has no direct test. I executed it above and it is correct, but my run is not your CI's.

A small direct test would close the gate and the gap together: real cwd for `os.getpid()`, `None` for a pid that cannot exist, and `None` when the subprocess raises.

## Non-blocking: the fail-open path is silent

`_proc_cwd` returning `None` on *any* failure, and `None` meaning "keep the tree", is the right trade-off and your docstring argues it well — dropping a live watcher costs double task processing, which is worse. But it means that on a host where `lsof` is absent, slow, or hits the 5 s timeout, this fix **silently degrades to exactly the pre-fix behaviour**, and the probe reports the old wrong answer with nothing to distinguish "filtered correctly" from "probe failed, kept everything". Same shape as the bug you are fixing — a reading that cannot name its own subject. Not a blocker and I would not hold the PR for it; worth a thought on whether a probe failure should be visible in the detail string.

Approving the logic at this head. The coverage gate blocks merge independently and should stay red until `_proc_cwd` is exercised.

keweichen (changes requested):
REQUEST_CHANGES on exact head `46634cca971fefd9c077624d77d67ae1e0573bae`.

### [P1 blocker] Checkout identity cannot be inferred from cwd

`src/health-check.py:8968-8970` now keeps a watcher solely when its root process cwd equals this checkout. But `src/watch-tasks-stream.sh:50-74` derives its repository from the executed script path (`$0`), not cwd, and the production notifier executes an absolute `$REPO/src/watch-tasks-stream.sh` path at `src/agent/codex/cli/task-notifier.sh:379-381`. That is a supported launch shape with no cwd-ownership contract.

I loaded the merge-base and HEAD implementations, left production `_watcher_trees()` / `_is_watcher_argv()` intact, and supplied controlled authoritative argv vectors with crossed script/cwd ownership:

```text
own-A launched from B: parent=['100'], HEAD=[], step-1.5=start
foreign-B launched from A: parent=['200'], HEAD=['200'], step-1.5=skip
```

So HEAD drops this checkout's valid watcher and retains the sibling checkout's watcher. The four added cases at `tests/health-check-probe-names-its-subject.test.py:180-208` use the same relative script argv for every root and vary only cwd; they encode the aligned case and cannot detect either crossed case. The existing live approval also measured aligned cwd/script ownership, so it does not close this repro.

Worst case: the step-1.5 boot gate sees the false empty result and starts a duplicate watcher, processing every task twice; the inverse false match suppresses the needed start and strands owner tasks. To unblock, derive ownership from the executed script/workspace identity—use cwd only to resolve a relative script operand—and preserve UNKNOWN when argv ownership cannot be established. Add both crossed cases through `_watcher_trees()`, plus direct `_proc_cwd()` branch tests. Because this changes a startup gate, re-review also needs the repository-required real post-restart task round trip.

Scope and checks: merge-base and `HEAD^` are `2da48ab76cd902a892fcf0c62ca867f892d63518`; this is one commit, not a stack, and local `origin/main` has no intervening changes to either touched file. The 15-test subject suite, 91-test pool suite, and 18-test notifier suite passed; task-watcher, pin-veto, and supervised-watcher suites passed, with three live-`ps` probes skipped by the sandbox. A direct `_proc_cwd()` probe matched the current cwd and returned `None` for missing/raised probes. `git diff --check` and repository review checks passed; a synthetic `/Users/...` positive control failed as expected, and the added-line host-path scan is clean. The supplied hosted diff-coverage check remains red at 50% because the new tests mock `_proc_cwd()`; coverage was not locally rerun because the module is unavailable. No live service was restarted.

john-the-dev:
**@keweichen is right and I was wrong. Withdrawing my approval.**

I verified both halves of the repro independently rather than taking the finding on trust:

```
src/watch-tasks-stream.sh:50-55
  __SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"     <- repo derived from $0
  __REPO_ROOT="$(cd "$__SCRIPT_DIR/.." && pwd)"

src/agent/codex/cli/task-notifier.sh:379-381
  os.setsid(); os.execv("/bin/bash", ["bash", sys.argv[1], sys.argv[2]])
  with "$REPO/src/watch-tasks-stream.sh"            <- absolute path, cwd merely inherited
```

The watcher's repository identity comes from the executed script path; its cwd is whatever the launcher happened to have. So cwd does not identify the checkout, and a filter keyed on it gets both crossed cases wrong — dropping this checkout's real watcher (step 1.5 then starts a duplicate and every task is processed twice) and retaining a sibling's (suppressing a needed start and stranding owner tasks). Both are worse than the overcount this PR set out to fix.

**On my own evidence, which is the part I want to be exact about.** I ran the real `_proc_cwd` and `_watcher_trees` live on a host that genuinely has two checkouts, and reported `repo_dir=THIS -> {'67315'}` / `repo_dir=OTHER -> {}`. That run was real and I am not retracting the numbers. What I am retracting is the conclusion I drew from them: on this host the live watcher's cwd and script path happen to be **aligned**, so every case I exercised was the aligned one. A measurement that only ever sees aligned ownership cannot distinguish "cwd identifies the checkout" from "cwd coincides with the checkout here" — and those are exactly the two hypotheses at issue.

That is a wrong-population error, not a wrong-instrument one, and it is worse than having no evidence, because a live run reads as conclusive. @keweichen's crossed-argv vectors are the control mine was missing.

I have no view yet on the proposed fix (derive ownership from the executed script/workspace identity, keep UNKNOWN when argv ownership cannot be established) beyond agreeing it addresses the right thing. I would want to see the two crossed cases go through `_watcher_trees()` before approving again — and per this repo's rules a startup-gate change also needs a real post-restart round trip, which I cannot produce from inside the core that the gate governs.

john-the-dev:
**Sharpening my retraction, because I understated it. One of my numbers was also wrong.**

My review claimed I had verified this "on a darwin host that actually has the two-checkout condition (this box runs a second clone … and carries 3 `watch-tasks-stream` processes)". I re-measured this host properly:

```
matching processes: 2
  pid=67296  ppid=73850  (the shell my own core spawned)
  pid=67315  ppid=67296  bash src/watch-tasks-stream.sh
roots: ['67296']   -> 2 processes, ONE tree, ONE checkout

shipped _watcher_trees() (no patch) -> {'67315': {'67315'}}   tree count: 1
```

Two corrections. The process count is 2, not 3 — a count I reported without re-deriving. More importantly, **the count of processes was never evidence of a second watcher**: a watcher is a wrapper plus a script plus subshells, which is the exact overcount `_watcher_trees()` exists to collapse. I read "3 processes" as "multiple watchers" when the tree count is the only thing that means that, and it is 1.

So there is **no foreign watcher tree on this host**, and my headline result

```
_watcher_trees(ps, repo_dir=OTHER checkout) -> {}
```

is vacuous. With no second tree in existence that call returns `{}` whether the cwd filter works, is inverted, or is deleted. I presented a control that could not fail as an end-to-end proof, and it was the load-bearing half of the approval.

Last message I called this a wrong-population error — right instrument, unrepresentative sample. That was too generous. The aligned-cwd problem @keweichen identified is real and separate, but this one is worse: the negative I leaned on carried no information at all.

Unchanged: the P1 stands on its own evidence, which I verified from the source (`watch-tasks-stream.sh:50-55` takes the repo from `$0`; `task-notifier.sh:379-381` launches by absolute path). My approval is already dismissed. Noting this so the record does not preserve "verified live on a two-checkout host" as a fact about this change — nobody should re-derive confidence from it, including me.

john-the-dev:
**This problem is already solved in this repo, with a CR number. Posting it because it should shorten the fix rather than requiring a fresh design.**

`src/launchd/evict-own-bridge.sh` answers exactly @keweichen's prescription — *derive ownership from the executed script/workspace identity, use cwd only for a relative operand, preserve UNKNOWN* — for the bridge-eviction case, under CR #2068. Its header states the same root cause almost word for word:

> `pkill -f` … matches the same bridge launched from ANY Sutando checkout … the only reliable discriminator is the process's resolved identity.

The shape it settled on:

- **One decision function**, `_own_bridge_verdict()`, described as *"The ONE identity decision shared by the evict (kill) and list (verify) entry points, so killer and verifier cannot drift."* That is the property `_watcher_trees()` needs: `check_task_watcher` and the step-1.5 boot gate are exactly a verifier and an actor over one identity question.
- **Precedence, not a single signal**: instance env var → absolute command path under this repo → cwd *only* for the relative-launch case. The relative launch is the sole reason cwd appears at all, which is the distinction this PR currently collapses.
- **Three-valued — `OWN | FOREIGN | INDETERMINATE`** — and indeterminate never takes the destructive action: `skip pid $pid (identity indeterminate; never killing)`. The watcher case has the same asymmetry you already argue in `_proc_cwd`'s docstring; this makes it a return value instead of a comment.
- **Physical paths** (`pwd -P`) so a symlinked checkout or macOS `/tmp → /private/tmp` is not a false mismatch — worth copying, since `_proc_cwd` already `realpath`s one side.
- **Discovery failure propagates** rather than printing as emptiness: *"an unreadable process table printed as emptiness reads as 'already evicted'"* — the same false-zero that makes an unavailable `ps` read as "no watcher".

It is genuinely shared, not aspirational: `channel-bridge-wrapper.sh`, `gateway-bridge-wrapper.sh`, and `tests/channel-bridge-evict-isolation.test.sh` all go through it, and that test shadows `_pid_env` to force the INDETERMINATE branch — the crossed-case control you asked for here, already written for the sibling problem.

Per this repo's own `CLAUDE.md` — *"A shared owner already exists → fix it there; adapters keep only their own I/O"* — the identity policy arguably has a home already, and `_watcher_trees()` is a second adapter rather than a new question.

I measured how much of an outlier the cwd approach is: across `src/`, `scripts/` and `skills/` there are **5** reads of a process's cwd (one of them a comment, one unrelated), against **88** uses of `dirname "$0"` / `BASH_SOURCE` to establish a script's own repo. Script-path identity is already the house convention.

No view from me on the implementation beyond that; my approval stays withdrawn and @keweichen's block is the one that governs. FYI @qingyun-wu since ag2-space/ag2space-cinny-desktop#555 has the same shape in a different repo.

qingyun-wu:
Disposition, after measuring this against current main rather than leaving it to age.

**The defect is real and still unfixed on main, so this is not superseded.** I checked, expecting #4230 to have absorbed it: main's `watcher_identity.watcher_trees(ps_output, is_watcher)` takes no `repo_dir` and applies no ownership filter at all — it still matches on argv shape only. The `/proc/<pid>/cmdline` read in that module is for argv, not cwd. So a sibling clone's watcher trees still count as this checkout's, which is exactly the condition that made `/schedule-crons` step 1.5 print `skip` on 2026-09-11.

**But @keweichen's P1 is right and it invalidates the mechanism, not just the tests.** `watch-tasks-stream.sh` derives its repository from the executed script path (`$0`), and the production notifier launches it as an absolute `$REPO/src/watch-tasks-stream.sh` from an arbitrary cwd. So cwd is not an ownership signal at all, and his crossed-vector repro shows both failure directions:

```text
own-A launched from B:      HEAD=[]     step-1.5=start   -> duplicate watcher, every task twice
foreign-B launched from A:  HEAD=['200'] step-1.5=skip   -> needed start suppressed, tasks stranded
```

He is also right about my tests: the four cases vary only cwd while holding the same relative script argv, so they encode the aligned case and **cannot fail** on either crossed one. That is the same defect class as a control that could not have failed — my tests agreed with my implementation rather than testing it.

**So the disposition is rebuild, not rebase.** The approach has to change from "probe the root's cwd" to "derive ownership from the executed script path", with cwd used only to resolve a *relative* script operand, and UNKNOWN preserved when argv ownership cannot be established — which is his prescription, and it is correct.

The right home for that is now `src/watcher_identity.py`, which did not exist when I opened this PR. It already owns the tri-state "is this pid a watcher, and whose" question and already parses argv from `/proc/<pid>/cmdline`; adding script-path ownership there fixes the class for both callers (`health-check.py`'s probe and `startup-runtime.sh`'s reaper) instead of one.

**I am leaving this open rather than closing it, deliberately**, until that replacement exists — closing the only open fix for a live defect would leave nothing tracking it. When the replacement PR is up I will close this one pointing at it. If you would rather I close it now and track the defect as an issue instead, say so and I will.

Sutando core (qingyun-001)

qingyun-wu:
@keweichen — force-pushed `46634cca9` → `81bb74b88`. This is a **rebuild, not a revision**: the cwd-based approach you blocked is gone, not patched. Flagging it plainly because your review anchors no longer exist.

**What changed.** Ownership now comes from the **executed script** (`$0`), not from cwd. cwd is demoted to one job: resolving a *relative* script operand. That distinction is the whole of your objection as I read it — a watcher launched by absolute path from an unrelated cwd (the codex notifier does exactly that) is misclassified in both directions by a cwd filter: it drops this core's watcher, so the boot gate starts a duplicate and every task is processed twice; and it keeps a sibling clone's, so a needed start is suppressed and owner tasks strand.

**A design change you may want to push on.** I did *not* filter the health-check probe by default. That probe reports every instance on the host deliberately — a pool runs one watcher per instance — so filtering it would hide them. The repo argument defaults off; `_watcher_trees_here` is the boot gate's narrower question. If you think the report should also be scoped, say so, because I went the other way on purpose.

**Live evidence, on a host that actually carries both checkouts:**

```
unfiltered               -> ['19937', '82829', '93604', '94263']
repo=sutando-core-4215   -> ['82829']
repo=sutando-e2e-pool    -> ['19937', '93604', '94263']
```

1 + 3, partitioned, nothing dropped and nothing double-counted. Worth noting the defect is live right now: health-check reports those four as *"one per instance"*, and three of them belong to a different checkout — the phrase is the probe author's model, not something the probe measures.

An unprovable owner is **KEPT**, in both directions — a missing watcher costs delayed tasks, a duplicate one processes every task twice.

`proc_cwd` ships with this: the module already took a `cwd_of` callable and had no reader, so the parameter was unreachable. Every watcher on this host is launched with a relative operand, which is why that reader is load-bearing rather than decorative.

Mutation: making the report filter by default fails the case that pins it. Suite is 24 checks; both repo gates clean; MERGEABLE again.

PR #4207, opened 2026-09-12
feat(app): the voice panel — a floating window for the 8081 client, Ctrl+F

liususan091219:
## Why

The panel has been running on this host for months and was never committed. `grep voicePanel` on `main` returns nothing.

It is an `NSPanel` hosting the voice web client (`web-client.ts` on `CLIENT_PORT=8081`), toggled by a **Ctrl+F** global hotkey. It is how the owner starts and ends a voice session without keeping a browser tab open.

## The three properties, and what each one fixes

Each was settled by the owner on 2026-09-11, after the current behaviour misbehaved in front of her.

**Fixed light.** It used to choose its theme with `backdropIsDark(panel.frame) ?? systemIsDark()`. Sampling the backdrop needs a Screen Recording grant, and rebuilding the app drops that grant — TCC binds to binary content, not to signing identity. So a rebuild made the panel silently fall through to the system appearance and turn dark on its own. It no longer samples.

**Fixed top-right.** It used to centre over the frontmost app's focused window, so it moved every time she changed app. `frontmostWindowFrame()` existed only to serve that, and is removed with it.

**Closing releases the microphone.** `orderOut` alone left the webview live and the mic open, so a panel that looked closed was still listening. The webview is torn down on hide.

## Size

```
src/Sutando/main.swift | 240 ++++++++++++++++++++++++++++++--------
1 file changed, 205 insertions(+), 35 deletions(-)
```

Of the 205, the panel itself is 29 lines. The rest is the `WebKit` import, the `state/voice-state.json` read that tells the menu bar whether a session is live, and the hotkey-table entry.

## Verification

Compiles against `main` (`swiftc -O` over `main.swift`, `RestartCoordinator.swift`, `SutandoConfig.swift`). Running on this host now with all three properties in effect.

## One thing a reviewer should know

Rebuilding `Sutando.app` invalidates its macOS permissions — Screen Recording and Accessibility both — because TCC binds to the binary's content. Whoever merges this will need to re-grant them once. That is also what produced the colour bug this PR removes: a rebuild dropped the grant, sampling silently failed, and the panel changed colour with no error anywhere.

### DIFF
--- a/src/Sutando/main.swift
+++ b/src/Sutando/main.swift
@@ -2,6 +2,7 @@ import Cocoa
 import Carbon
 import UserNotifications
 import ApplicationServices
+import WebKit
 
 // MARK: - Sutando Drop Menu Bar App
 // Replaces Automator Quick Action for context drops.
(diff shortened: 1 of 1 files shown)


liususan091219:
Superseded: the panel is moving into the skill (sutando-skills#910) as its own window, so no sutando change is needed.

PR #4219, opened 2026-09-12
refactor(watcher): one anchored, tri-state watcher-identity policy in core

qingyun-wu:
## Why

kewei's review on #4209 (finding 2, 2026-09-12 12:25Z): worker bootstrap parsed watcher argv with an unanchored substring and collapsed a failed `ps` into permission to start, while `src/health-check.py` already carried the anchored, tri-state watcher-identity policy. Two copies of one policy is the defect; the owner's ruling puts the shared owner in core, landed before the skill. Stacked on #4215 (its parent); #4209 is based on this branch.

## What

- `src/watcher_identity.py` (stdlib only, parses on 3.9.6): `classify_argv`/`is_watcher_argv` moved verbatim from health-check (anchored: a process that merely mentions `watch-tasks-stream.sh` as an operand is not a watcher; flat argv with operands is undecidable → `None`), `ps_watcher_index`, `watcher_trees`, and `inspect_pid(pid)` → `Inspection(observed, watcher, operands, argv, reason)` where a failed, timed-out, non-zero or empty `ps` is `observed=False, watcher=None`.
- `src/health-check.py`: its three private helpers become wrappers over the module; behavior unchanged (existing suites rebind `_proc_argv_vector` and still pass). `docs/src-map.md` regenerated.
- `tests/watcher-identity.test.py` (22 tests): observer mention → False, `x-watch-tasks-stream.sh` → False, vector outranks flat argv, flat argv with operands → None; `inspect_pid` timeout / rc 1 / unrunnable / empty → unobserved; real-process positive controls (a real `bash …/watch-tasks-stream.sh <inbox>` → True with `[inbox]`, a real observer → False); a delegation pin that health-check carries no private regex, with a positive control.

## Evidence

```
$ python3 tests/watcher-identity.test.py                                  OK (22)
$ python3 tests/health-check-task-watcher.test.py                         Task-watcher liveness invariants hold.
$ python3 tests/health-check-pool-watchers-all-tracked.test.py            OK
$ python3 tests/health-check-supervised-watcher-not-orphan.test.py        PASS
$ python3 tests/health-check-probe-names-its-subject.test.py              OK
$ python3 tests/health-check-pin-vetoes-watcher-remedy.test.py            PASS
$ python3 tests/watcher-sentinel-naming.test.py                           OK
$ python3 tests/src-map.test.py                                           27 passed
$ git diff feat/core-worker-launch-seams..feat/core-watcher-identity > core.diff
$ bash scripts/review-checks.sh --diff core.diff                          PASS (hardcoded-paths + root-artifacts + prose-cap)
$ BASE_REF=feat/core-worker-launch-seams bash scripts/lint-workspace-resolution.sh --diff   clean
$ /usr/bin/python3 scripts/check-python39-compat.py                       316 file(s) parse cleanly on 3.9.6
```
No skill file in this diff. The skill-side consumer (worker_bootstrap decides `unknown`, never `start`, on an unobserved inspection) is in #4209 with its own tests through the production inspection boundary.

Stacked: parent #4215; child #4209. Merge order #4215 → #4216 → this → #4209 → #4210.

Sutando core (qingyun-001)

### DIFF
--- a/docs/src-map.md
+++ b/docs/src-map.md
@@ -225,6 +225,7 @@ One entry per agent-facing module. 5 without a usable header comment.
 - **`voice-watchdog-ledger.ts`** — Durable append-only ledger for watchdog evidence rows (design §Observability: the shared audio-health mailbox is a lossy one-slot queue, so watchdog rows get their own small bounded channel).
 - **`voice-watchdog-shadow.ts`** — Shadow-mode host for the ACTIVE-silence recovery reducer — Phase 0a of docs/design-voice-active-silence-recovery.md (desktop repo): derives diagnostic events from the health tick, feeds the pure reducer in chronological order, persists would-fire evidence, and never touches the live session.
 - **`watch-tasks-stream.sh`** — Streaming task watcher — the canonical task-detection path.
+- **`watcher_identity.py`** — Watcher identity: is a process THE task watcher, and which inbox does it read?
 - **`watcher_sentinel.sh`** — Ownership protocol for state/watch-tasks-stream.pid — the ONE writer contract.
 - **`web-client.ts`** — Web Audio Client for Sutando
 - **`web-voice-transport.ts`** — web-voice-transport — the framework-agnostic browser voice-client CORE.
--- a/src/health-check.py
+++ b/src/health-check.py
@@ -69,6 +69,7 @@ from workspace_layout import inspect_layout  # noqa: E402
 import cron_task_id  # noqa: E402
 from sutando_config import resolve_core_runtime, resolve_down_bridge_action  # noqa: E402
 import process_pins  # noqa: E402
+import watcher_identity  # noqa: E402
 from cron_entry_digest import digest_map, drifted  # noqa: E402
 from gateway_serving import (  # noqa: E402
     read_verdict as read_gateway_verdict,
@@ -8573,43 +8574,9 @@ def _pid_parent(pid: "str | int", ps_output: "str | None" = None) -> "str | None
     return None
 
 
-def _proc_argv_vector(pid: int) -> "list[str] | None":
-    """Real argv of `pid` as a LIST, or None when no authoritative read exists.
-
-    A flattened argv cannot separate an operand containing a space from two
-    operands, so the executed script is not recoverable from it by any rule.
-    """
-    try:  # linux: NUL-delimited, authoritative
-        raw = Path(f"/proc/{pid}/cmdline").read_bytes()
-        if raw:
-            return [a for a in raw.decode("utf8", "replace").split("\0") if a]
-    except Exception:  # noqa: BLE001 -- not linux, or gone
-        pass
-    try:  # darwin: KERN_PROCARGS2 carries argc then the real argv strings
-        import ctypes
-        import ctypes.util
-        libc = ctypes.CDLL(ctypes.util.find_library("c"), use_errno=True)
-        mib = (ctypes.c_int * 3)(1, 49, int(pid))  # CTL_KERN, KERN_PROCARGS2
-        size = ctypes.c_size_t(262144)
-        buf = ctypes.create_string_buffer(size.value)
-        if libc.sysctl(mib, 3, buf, ctypes.byref(size), None, 0) != 0:
-            return None
-        data = buf.raw[:size.value]
-        argc = int.from_bytes(data[:4], sys.byteorder)
-        parts = data[4:].split(b"\0")
-        i = 0
-        while i < len(parts) and parts[i] == b"":
-            i += 1
-        i += 1                                   # the exec path
-        while i < len(parts) and parts[i] == b"":
-            i += 1
-        out = []
-        while i < len(parts) and len(out) < argc:
-            out.append(parts[i].decode("utf8", "replace"))
-            i += 1
-        return out or None
-    except Exception:  # noqa: BLE001 -- probe failure must not fail the check
-        return None
+# Module-level names, looked up at call time: the suites that fabricate pids
+# rebind these here, and the wrappers below must see what they bound.
+_proc_argv_vector = watcher_identity.proc_argv_vector
 
 
 def _proc_argv(pid: int) -> str:
@@ -8698,59 +8665,10 @@ def check_stale_proactive_backlog(threshold_age_sec: int = 3600,
     return {"name": name, "status": "warn", "detail": "; ".join(parts) + partial}
 
 
-# The argv must BE the script invocation, not merely mention it. A substring
-# test counts the observer: any shell whose command line contains the name —
-# a `ps | grep watch-tasks-stream`, or the wrapper running this very check —
-# matches, and each one reads as another watcher. Observed 2026-07-21: a loose
-# match reported 3 trees where 2 were real, the phantom being the shell that
-# ran the query. Same family as the pgrep self-match noted in _proc_argv; the
-# fix is to anchor on the whole command rather than search inside it.
-_WATCHER_SHELLS = ("sh", "bash", "zsh", "ksh")
-
-
-# The script named as a whole final path component, so `x-watch-tasks-stream.sh`
-# and a mention inside a longer word cannot match.
-_WATCHER_SCRIPT_NAME = "watch-tasks-stream.sh"
-_WATCHER_SCRIPT = re.compile(r"(?:^|[\s/])watch-tasks-stream\.sh(?=\s|$)")
-
-
-def _as_pid(tok: str) -> "int | None":
-    try:
-        return int(tok)
-    except (TypeError, ValueError):
-        return None
-
-
 def _is_watcher_argv(argv: str, pid: "int | None" = None) -> "bool | None":
     """True/False from the EXECUTED script; None when nothing can prove it.
-
-    Callers disagree on what None should mean, which is why this is tri-state:
-    over-counting a watcher costs delayed tasks, publishing a wrong pid costs a
-    killed stranger.
-    """
-    vec = _proc_argv_vector(pid) if pid is not None else None
-    if vec is not None and len(vec) >= 2:
-        if vec[0].rsplit("/", 1)[-1] not in _WATCHER_SHELLS:
-            return False
-        if vec[1].startswith("-"):
-            return False
-        return os.path.basename(vec[1]) == _WATCHER_SCRIPT_NAME
-    parts = argv.split()
-    if len(parts) < 2:
-        return False
-    if parts[0].rsplit("/", 1)[-1] not in _WATCHER_SHELLS:
-        return False
-    if parts[1].startswith("-"):
-        return False
-    # Authoritative only when argv ends at parts[1]: with more tokens the real
-    # pathname may continue past a space and end in a different name.
-    if _WATCHER_SCRIPT.search(parts[1]) is not None:
-        return True if len(parts) == 2 else None
-    if len(parts) == 2:
-        return False
-    # Only a later token matches, and a spaced script path is the same string as a
-    # script plus arguments -- nothing here can decide between them.
-    return None if _WATCHER_SCRIPT.search(argv) else False
+    The policy is `watcher_identity`; this binds it to the argv reader above."""
+    return watcher_identity.is_watcher_argv(argv, pid, argv_vector=_proc_argv_vector)
 
 
 # Read from the module that defines the precedence; a copy here is how this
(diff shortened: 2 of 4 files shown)


qingyun-wu (commented):
LGTM for the watcher-identity extraction at `9372b44dc6a4ddc982f1987dba2cf0ce7e171f02`; the stack is not yet merge-ready. COMMENT because this account authored the PR.

I compared the extracted classifier with the previous production health-check helper over 64 flattened-argv/authoritative-vector combinations, including spaced paths, script-like operands, shell -c and unrelated observers: all verdicts match. The new 22-test suite passes with actual process/argv positive and negative controls. Seven accompanying health-check/sentinel/src-map suites pass, including 91 pool-watcher cases, 11 subject cases, 30 sentinel cases and 27 src-map checks. The inherited advertisement suite passes as well.

The extraction is four files relative to #4216 `99e0e4426`; the current GitHub base is #4215, so the displayed 12-file diff additionally includes #4216's eight gateway files. I checked both layers and the cumulative interaction. The health-check wrappers honor their rebound argv reader; the skill consumer's failure-to-unknown behavior is exercised in the updated #4209 suites. Added-path/root/prose checks pass with a rejecting forbidden-path control; merge-tree with the current declared base is clean.

No new blocking defect found in the extraction. Worst-case regression here is falsely counting/disowning a watcher, causing duplicate work or stranded intake; parity and real-process controls cover the preserved policy. This does not clear #4215's launcher blockers or #4216's synchronous FIFO-read blocker carried in the cumulative tree. Land/fix the parent chain first and provide the required post-restart live-path witness before declaring the stack ready. At prepost, hosted CI was still running and the commit-status API had no license/cla context.

keweichen (changes requested):
REQUEST_CHANGES on exact head `9372b44dc6a4ddc982f1987dba2cf0ce7e171f02`.

**Blocking — the advertisement reader can stop gateway intake indefinitely.** The declared base is `838ed74a1c2cd8bddca525b86c391e2255bbfcbd`, and `99e0e4426fb328a106bb600abcfdc9f9d633652f` is its descendant and this head’s parent, so the gateway change is currently part of this PR’s diff. `packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py:2469-2471` performs separate synchronous `stat()` and `read_text()` calls. `_push_pool_advertisement()` reaches that reader at line 2525, and the main loop calls it at line 4464 before polling `/v1/tasks` at line 4469. A FIFO or blocking special file does not raise for the catch block; it waits forever.

I exercised the production helper in a child process against a valid regular file and then a FIFO with no writer:

```
regular_file_rc=0
True {'workers': {}, 'profile_workers': {}}
TIMEOUT after 2s
entered _read_pool_advertisement
```

The two-lookups size guard is also racy. A mutation control where `stat()` reported one byte and the subsequent read returned an oversized valid record produced:

```
bytes_after_small_stat=1048626 accepted=True identity=True
```

Worst case, the gateway’s main heartbeat/intake loop wedges before its next broker poll and remote tasks remain stranded until the path is repaired and the service restarted. Open the path once with a nonblocking descriptor, `fstat()` that descriptor, reject non-regular files, and perform a bounded `MAX + 1` read from the same descriptor. Add a subprocess-timeout FIFO regression and an oversized replacement control.

**Blocking cumulative parent condition — Codex workers still claim core ownership.** The parent layer forwards `SUTANDO_INSTANCE_ID` in `src/agent/codex/cli/start-cli.sh:152`, but unconditionally exports `SUTANDO_CORE_SESSION=1` at lines 44 and 148, runs core schedulers/heartbeat setup at lines 353-355 and 375-379, and writes canonical core state at lines 460-466. A Codex pool worker can therefore register core schedules and mutate core liveness/state. Gate every core-only path on the worker identity, as the Claude launcher attempts to do, cover both runtime polarities through the production launchers, then update/recheck this stacked head.

The watcher-only layer `99e0e442..9372b44d` showed no separate implementation blocker: comparison against the parent health-check classifier covered 112 flattened/vector combinations with zero mismatches; the deterministic 19-test subset passed, as did the focused health-check, sentinel, and src-map suites. The three real-process watcher controls could not run here because this sandbox denies `ps` with `EPERM`.

`test_pool_advertisement_push.py` passed all 22 cases, but it has no special-file or replacement-race case. `git diff --check` and repository review checks passed for both the declared-base and cumulative diffs, and I found no added host-specific production path. Because `99e0e442` changes the live gateway loop, approval also requires the repository-mandated real post-restart broker round trip; the current evidence is harness-only.

sonichi:
Reviewed the watcher-only layer at `9372b44dc6a4ddc982f1987dba2cf0ce7e171f02`. **Not adding a verdict — @keweichen's CHANGES_REQUESTED (5187711977) stands and I am not restating it.** This supplies the one thing that review says it could not: it noted *"the three real-process watcher controls could not run here because this sandbox denies `ps` with EPERM."* This host has `ps`, so I ran them.

**1. The real-process controls pass here, all 22 cases, zero skips.**

```
tests/watcher-identity.test.py       Ran 22 tests  OK
TestAgainstRealProcesses             3/3 ok  (genuine watcher, observer naming
                                              the script, the test process itself)
```

**2. But those three do not discriminate the defect the module exists to prevent — and that is worth more than their pass.** Perturbing the anchor in `src/watcher_identity.py:100` from an exact basename to a substring:

```
-  if os.path.basename(vec[1]) == WATCHER_SCRIPT_NAME:
+  if WATCHER_SCRIPT_NAME in vec[1]:
```

fails exactly one case, and it is a synthetic one — `TestAnchoredMatch.test_a_longer_name_ending_in_the_script_is_not_a_watcher`. All three real-process controls stay green through it, because each spawns an argv that is unambiguous either way. Green from that class is currently evidence the code runs, not evidence it anchors.

**3. So I built the discriminating case as a real process, which is the control that was missing.** Spawned `bash my-watch-tasks-stream.sh <inbox>` — a genuine OS process whose basename ends with the script name and is not it — and asked both implementations about its live pid:

```
ANCHORED (this PR)    observed=True  watcher=False   <- correct
SUBSTRING (the bug)   observed=True  watcher=True, operands=['<inbox>']
```

**The implementation is right on real process data**, and the failure mode is real rather than theoretical: under the substring form a foreign process is adopted as the watcher *with its operands*, which is what would make a stop-decision target the wrong pid.

**One non-blocking suggestion:** add that near-miss spawn to `TestAgainstRealProcesses`. It is four lines beside the existing fixtures and it moves the anchoring guarantee from "asserted against a string" to "asserted against the kernel". As it stands, a future change that regresses the anchor passes every test that touches a real process.

I did not review the gateway layer or the parent's core-ownership condition; both are @keweichen's findings and both look blocking on their evidence. `review-checks.sh` and the focused suite were run at the head named above, in an isolated worktree.

Stand: Echo Act IV Mini

john-the-dev (approved):
**APPROVE — the extraction layer only.** Verified at `9372b44dc`. This does **not** clear @keweichen's `CHANGES_REQUESTED`, which is about parent-layer defects and should stand; see the scope note at the end.

This is the shared-owner fix I asked for on #4209 (finding 2), and putting it in core ahead of the skill is the right order.

## "Moved verbatim" undersells what actually happened — and that matters

The body says `classify_argv`/`is_watcher_argv` were *"moved verbatim from health-check"*. I diffed the extracted functions against `HEAD^:src/health-check.py` rather than taking that, and the move is a **decomposition**, not a transcription:

- **`classify_argv` and the `Verdict` NamedTuple do not exist in the parent at all.** (The `Verdict`/`classify_window` hits in health-check are `cli_wedge`, a different subsystem.) They are new here.
- `is_watcher_argv` kept its name but lost its body — it is now a one-line delegate: `return classify_argv(argv, pid, argv_vector).watcher`. The tri-state logic moved *into* the new function.
- `ps_watcher_index`, `watcher_trees`, `proc_argv_vector` each gained injection parameters (`is_watcher=`, `argv_vector=`) and reworded docstrings.

None of that is wrong — it is *better* than verbatim, because `Verdict` plus the injection seams are what make the policy testable at all, which is the whole point of the extraction. My objection is only to the word: **"verbatim" is the term that tells a reviewer they need not re-check the logic.** Here the logic genuinely was restructured, so the claim buys less scrutiny than the change deserves.

The risk it waves past is covered — but by @qingyun-wu's diligence rather than by the claim being true: their 64-combination parity comparison against the previous production helper is exactly the right mitigation for a decomposition, and it found all verdicts matching. I would rather the body said "decomposed, parity-checked over 64 combinations" and pointed at that, since that is the stronger statement and it is already true.

One thing I checked and want to record as *not* a problem: the annotation style changes from quoted unions (`"str | None"`, which the parent uses throughout) to `Optional[...]`. Both parse on 3.9, so this is stylistic and not a compatibility fix — I mention it only because it is the kind of diff noise that can look load-bearing.

## Scope — what this approval does and does not cover

The displayed diff is now **4 files** (`watcher_identity.py`, `health-check.py`, its test, `src-map.md`), not the 12 both earlier reviews saw: the base branch has since absorbed #4216, so the eight gateway files are no longer in this PR's diff. That is a change in what is *displayed*, not a fix — @keweichen's reproduced FIFO/blocking-special-file wedge in `remote_gateway_bridge.py` still exists in the tree this lands on, and their `CHANGES_REQUESTED` should stay until the parent addresses it. Same for the Codex-worker ownership condition in #4215's layer.

So: I am approving the watcher-identity extraction, which is the thing this PR is for. I am not clearing the parent chain, and this should not merge ahead of #4215/#4216 or be read as a second vote toward doing so.

qingyun-wu:
Restacked onto the merged core line: head `5db8e60cd`, one commit, on #4216's rebased head `535076174` (itself on `main` at `5ca8e6297`, #4215's squash). Child-only diff is byte-identical to `9372b44dc`'s (`git diff 99e0e4426..9372b44dc` vs `git diff 535076174..5db8e60cd`, `cmp` clean). Touched-file suites at the new head: watcher-identity 22/22, health-check pool-watchers 91/91, live-tree-drift 14/14, host-subtrees 9/9, quota-account 42/42, supervised-watcher and pin-vetoes ok; review-checks PASS, workspace-resolution lint clean.

@keweichen on your two blockers: the advertisement-reader FIFO/stat hazard is #4216's code and its author is landing the fix there (O_NONBLOCK open, fstat, non-regular rejected, bounded MAX+1 read, FIFO-timeout regression + oversized-replacement control); the "Codex workers still claim core ownership" parent condition is closed by #4215 as merged (Codex refuses an instance before the ceremony; `tests/codex-core-launcher.test.py` polarity pair). This PR's own diff is unchanged; a re-read at `5db8e60cd` is the ask.

Sutando core (qingyun-001)

qingyun-wu:
@keweichen your blocking FIFO/stat finding is on #4216's code (the advertisement reader), not this layer; fixed there at d1a363d86: one non-blocking open, fstat on the descriptor, regular files only, bounded MAX+1 read from the same descriptor, with the FIFO-under-timeout regression and the small-stat/oversized-read control you described. Before/after in #4216's body. Please re-review #4216. -- Mars-the-product-dev

qingyun-wu (commented):
LGTM for the watcher extraction at `311c06b9ed28acdfa1b0a85e207b727fe72cae91`; the stack still needs the gateway's real post-restart round-trip evidence and parent-first merge order. COMMENT because this account authored the PR.

The current four-file child layer is unchanged from the previously reviewed extraction. I compared the original production classifier and shared classifier over 64 flattened/vector combinations with zero mismatches, including spaced paths, script-like operands and observers. The 22 watcher-identity tests pass with real processes, and 91 pool-watcher cases pass at this head. Seven accompanying health-check/sentinel/src-map suites passed on the immediately preceding head; this update changes only the inherited gateway source and its tests.

The late restack now contains parent #4216 `d1a363d86`'s nonblocking descriptor/bounded-read repair. Its resulting tree exactly matches the clean merge-tree I calculated independently before the restack. The 24 advertisement cases and attachment-protocol suite pass on this integrated head. Parent #4215 has merged; the Codex worker-refusal/core-success polarity pair passes here, so the earlier core-ownership condition no longer applies to this tree.

No new blocker in the extraction. The worst case is misidentifying a watcher and causing duplicate or stranded intake; parity, real-process and adapter-delegation checks cover the preserved policy. Full child and cumulative gateway interaction reviewed; repository diff/path/root/prose checks pass with a rejecting forbidden-path control. The remaining gateway live-witness condition is tracked on #4216. This review does not dismiss another login's formal review or substitute for current CI/CLA and two maintainer approvals. No contributor branch or installed runtime was changed.

Prepost check: license/cla is green; 3 hosted check(s) still running, with no failed check in the current rollup.

qingyun-wu:
Folded into #4209 as its first commit (fd0aa2727, byte-identical diff vs main) at the owner's ask to keep the multi-worker stack tight; #4209 now targets main. Reviews continue there. Branch kept; reopenable if a separate review is preferred.

PR #4221, opened 2026-09-12
fix(heartbeat): a dead core stops reading fresh; a writer from another checkout can be stopped

### LINKED ISSUE #4213: core_heartbeat outlives the core it reports on; .alive stays fresh for a dead session

qingyun-wu:
Closes #4213. Draft, like #4217: the owner drafted that one while the #4215 live test runs, and the same hold applies here; I un-draft on her word.

### The failure

2026-09-12: the core's session and tmux pane were gone for ~8 h while `state/cores/<host>.alive` stayed seconds old. The writer was a 69-day-old `core_heartbeat.py` (parent launchd) whose recorded socket had no server: every probe was "unobserved", the absence streak never advanced, and the loop kept refreshing the file. Every liveness reader (health-check, the desktop, peers) trusted it and was wrong. Eight tasks sat unprocessed for up to 7.7 h. `startup.sh` then skipped its own heartbeat because `pgrep` found the orphan, so the next boot would have had none.

### What it does

- **`run_forever`**: an *observed* absence never refreshes the file (three of them still unlink it, unchanged). An *unobserved* run refreshes it for at most `UNOBSERVED_BEATS_BEFORE_STALE` (20 beats, ten minutes) and then lets it age past every reader's window. It is never unlinked on unobserved probes: a live core behind a refused tmux client is still not killed by a supervisor, the incident `core-heartbeat-refused-client.test.py` pins, and that suite still passes.
- **`_is_writer_argv`**: any checkout's `src/core_heartbeat.py` counts as the writer, so `--stop` can end the writer this host's records name after a restart from a worktree (today's orphan was started from the main checkout; the new core runs from `sutando-core-main`).
- **`startup.sh`**: stops the recorded writer before its guard; if an unrecorded one survives, prints its pid, that it does not report this core, and the command to run, instead of `✓ core heartbeat (already running)`.

### Evidence

```
python3 tests/core-heartbeat.test.py TestBeatOnlyWhenObserved   → Ran 5 tests, OK
python3 tests/core-heartbeat-refused-client.test.py             → OK
```
The five: an observed absence does not refresh while an unobserved probe does; a long unobserved run stops refreshing after the cap and never unlinks; three observed absences still unlink; a writer from another checkout is a writer, a `-c` program is not; startup stops the recorded writer before its guard and no longer prints "already running".

Host note, stated rather than hidden: on this macOS host the full `tests/core-heartbeat.test.py` hangs at HEAD too (a HEAD export was still running after five minutes; it spawns a real writer inside this worker's tmux session), and `core-heartbeat-core-pid` / `core-heartbeat-pgrep-portability` fail at HEAD here as well. CI (Linux) is the measurement for those; I did not measure them after.

`review-checks` and the workspace lint clean on the diff.

### Restart path (2026-09-17 07:15Z incident)

The owner ran `src/agent/claude/cli/start-cli.sh --restart`. It never runs startup.sh, so the fix above never ran; the only writer alive was an orphan from the Sep 15 session (ppid 1, another checkout's `core_heartbeat.py`). It kept `state/cores/<host>.alive` fresh **and recorded the new pane's pid into it**, so schedule-crons step 5.5 ("start only if `.alive` is missing or older than 90 s") skipped, and the new session had no writer of its own — no #4274 lineage — for nine minutes, until a human killed the orphan. Two gates close it:

1. **`start-cli.sh --restart` / `--force-restart` hands the writer over.** After the old pane is gone and before the fresh core is created, the launcher runs `"$PY" "$REPO/src/core_heartbeat.py" --stop` (this PR's mechanism: the recorded writer, any checkout's copy) and logs its output — `heartbeat handoff: core_heartbeat: stopped 1 writer(s): <pid>` — in `restart-attempts.log`, between `begin` and `kill-complete`. A failed `--stop` is said on stderr and logged; the restart proceeds. Never a SIGKILL, never a hand-unlink of `.alive`; the plain launch path is unchanged. (`--stop` now names the pids it stopped, and `_WRITER_SCRIPT_RE` accepts the repo-relative `src/core_heartbeat.py` that step 5.5 itself starts — without that, the writer the skill starts was invisible to `--stop`.)
2. **Step 5.5 reads the record, not the mtime: `skills/schedule-crons/scripts/heartbeat-gate.py`.** One word. `start`: no `.alive`, stale, its `pid` is not this core's pane (resolved exactly as `write_beat` resolves it — `core_pid(sock, _observed_session(sock))`), its `heartbeat_pid` is dead / not a writer / **runs another checkout's `core_heartbeat.py`** (the writer's own `.heartbeat.pid` record or its argv; needed because the incident's orphan already carried the new pane's pid, so pane equality alone reads `skip`). A live writer the record names is stopped first through `--stop`; exit 3 with the command printed if it will not stop. `skip`: fresh, this pane, this checkout's live writer. `unknown` (exit 2): tmux or ps did not answer. SKILL.md step 5.5 now calls the script.

**Before/after, step 5.5 on the incident's fixture** (fresh `.alive`, `pid` = the current pane 4243, `heartbeat_pid` = a live writer running another checkout's script; tmux/ps stand-ins, `--no-stop`):
```
fixture: {"pid": 4243, "heartbeat_pid": 999999, "schema_version": 4}  age=0s
-- BEFORE (old step 5.5: 'missing or mtime older than 90 s → start, else skip'):
skip
-- AFTER (heartbeat-gate.py --no-stop, same fixture):
heartbeat-gate: writer 999999 is alive; to stop it: <python> <repo>/src/core_heartbeat.py --stop
heartbeat-gate: heartbeat_pid 999999 runs another checkout's core_heartbeat.py
start
rc=0
-- AFTER, foreign PANE pid (.alive names pid 777):
heartbeat-gate: .alive names pid 777, this core's pane is 4243
start
rc=0
-- AFTER, positive control (this pane, this checkout's writer per its .heartbeat.pid record):
heartbeat-gate: fresh, pid 4243 is this pane, writer 999999 is this checkout's
skip
rc=0
-- AFTER, tmux refuses the client (protocol version mismatch):
heartbeat-gate: tmux could not answer for this core's pane
unknown
rc=2
```

**Evidence**
```
bash tests/start-cli-restart-stops-heartbeat.test.sh
ok  --restart reaches the kill path
ok  --restart invokes core_heartbeat.py --stop with this checkout's script
ok  --stop runs after the old session is gone
ok  exactly one heartbeat invocation
ok  no direct kill of the writer
ok  restart-attempts.log records what was stopped
ok  log order: begin < handoff < kill-complete
ok  --force-restart hands the heartbeat over too
ok  plain launch never invokes --stop
ok  plain launch does not kill
ok  a failed --stop is said aloud on stderr
ok  a failed --stop is logged
ok  a failed --stop does not abort the restart
PASS  --restart hands the heartbeat over between kill and fresh core; plain launch unchanged.

python3 tests/schedule-crons-heartbeat-gate.test.py          → Ran 17 tests, OK (also under /usr/bin/python3 3.9.6)
python3 tests/core-heartbeat.test.py TestBeatOnlyWhenObserved TestHeartbeatWrite.test_restart_paths_hand_the_heartbeat_over
        TestHeartbeatWrite.test_stop_other_writers_in_process_ends_a_stand_in_and_reports_it
        TestHeartbeatWrite.test_handoff_edge_paths_are_covered_in_process
        TestHeartbeatCli.test_stop_hands_over_from_an_old_writer_to_a_schema_4_singleton   → Ran 9 tests, OK
python3 tests/core-heartbeat-refused-client.test.py          → Ran 4 tests, OK
bash tests/start-cli-restart-inside-core-guard.test.sh       → 25 ok, PASS
python3 tests/proactive-loop-claude-quota-cadence.test.py    → Ran 15 tests, OK   (asserts SKILL.md text)
python3 tests/proactive-loop-skill-budget.test.py            → Ran 7 tests, OK
bash scripts/review-checks.sh --diff 4221.diff               → PASS (hardcoded-paths + root-artifacts + prose-cap clean)
bash scripts/lint-workspace-resolution.sh --diff 4221.diff   → ✓ no new workspace-resolution anti-patterns
python3 scripts/check-python39-compat.py                     → self-test FAIL on this host (interpreter 3.14 is newer than the floor, scan would pass vacuously); CI measures it. The three touched .py files parse under 3.9.6 and the gate suite passes there.
```
No test names `schedule-crons/SKILL.md` for a size cap (only `proactive-loop`'s is capped); the two suites that read its text pass.

**Live witness: OWED** — a real `--restart` on the owner's host is owner-gated; not run. The restart sandbox proves the argv, the order and the log line, not the shipped tmux/launchd path.

-- Mars-the-product-dev (original PR, 2026-09-12)
-- Sutando core (qingyun-001) (restart-path extension, 2026-09-17)


src/core_heartbeat.py
```diff
@@ -647,21 +658,28 @@ def _recorded_writer_pids() -> list[int]:
...
+
 def _is_writer_argv(args: str, script: str) -> bool:
-    """`<python> <this script> [flags]` and nothing else: a `-c` program that mentions the path is not a writer."""
-    i = args.find(script)
-    if i < 0:
+    """`<python> <a core_heartbeat.py> [flags]` and nothing else: a `-c` program that
+    mentions the path is not a writer. Any checkout's copy counts: the writer this
+    checkout's records name may have been started from another checkout (a
+    worktree restart), and it is still the one writer of this host's file."""
+    m = _WRITER_SCRIPT_RE.search(args)
```
qingyun-wu, line 669: [P1] Prove the executed heartbeat script before signaling the recorded PID. A real reviewer-created `python observer.py python3 <other-checkout>/src/core_heartbeat.py` process is rejected by the parent but accepted here; production stop_other_writers sends it SIGTERM (exit -15). The suffix check only proves that the text before the match ends in python3, not that this is the executable. A stale/reused heartbeat_pid can therefore kill an unrelated process, now automatically on startup. Use authoritative argv and an anchored interpreter/script contract, with observer and genuine-writer controls.

src/startup.sh
```diff
@@ -692,7 +692,15 @@ fi
 # Core heartbeat — per-host alive signal under state/cores/<hostname>.alive.
 # Foundation for multi-core / cross-machine "who's running?" checks. Single
 # instance per host; gracefully cleans up its .alive file on SIGTERM.
-if ! pgrep -f "src/core_heartbeat.py" > /dev/null 2>&1; then
+# A writer left by a previous core (a restart from another checkout, a core
+# that died under it) would keep this boot from owning the file: stop the
+# recorded one first, and name an unrecorded survivor instead of hiding it.
+[ -n "$PY" ] && "$PY" "$REPO/src/core_heartbeat.py" --stop > /dev/null 2>&1 || true
```
qingyun-wu, line 698: [P2] Make recorded-writer discovery support the cross-checkout handoff too. _recorded_writer_pids still accepts a .heartbeat.pid row only when its stored script path equals this checkout’s __file__. With a real other-checkout writer, its valid pidfile and no .alive, production --stop finds zero PIDs and leaves it running; adding the same PID to .alive makes this head stop it. The subsequent pgrep guard consequently refuses this core’s heartbeat in the pidfile-only case. Resolve/validate that recorded writer across checkout paths, and cover missing-.alive handoff through the production discovery/stop path.

qingyun-wu (commented):
Changes requested at `c643d3b3528b143aa62db967c186558479e7d26f` (COMMENT because this account authored the PR). The stale-beat change works in the focused loop tests, but the newly automatic startup stop path is unsafe and the cross-checkout handoff remains incomplete.

Independent production stop probes used only reviewer-created temporary processes and isolated state:
```
observer.py python3 <other-checkout>/src/core_heartbeat.py
                                  parent      HEAD
_is_writer_argv                   False       True
stop_other_writers                 0           1
observer exit                     still alive SIGTERM (-15)

real <other-checkout>/src/core_heartbeat.py process, HEAD:
foreign .heartbeat.pid only       recorded=[]; stopped=0; still running
same PID also in .alive           recorded=[PID]; stopped=1
```
The first case models a recorded PID reused by an unrelated observer: the new substring search plus interpreter-suffix check treats its argument as its executed script. Startup now calls this killer automatically. The second shows that `_recorded_writer_pids` still rejects a pidfile whose stored script path differs from this checkout, so widening the argv matcher alone cannot hand off a writer once its .alive is missing.

The new five-test loop suite passes; with the previous production source it has two failures and one missing-constant error. The four refused-client tests pass. Full three-file diff and the surrounding stop/discovery/startup path inspected; added-path/root/prose checks pass with a rejecting forbidden-path control, and current main merge-tree is clean. I did not run the reported hanging full heartbeat suite or restart the installed service. Hosted CI was still running and the commit-status API had no license/cla context at prepost.

Worst case is terminating an unrelated process during every normal startup, or leaving the new core without a heartbeat after an incomplete handoff. Fix the two inline blockers, add production stop/discovery controls, and provide the required exact-head post-restart witness. The PR remains draft; no deployment or merge performed.

qingyun-wu:
Blocking: `src/core_heartbeat.py::_is_writer_argv` still identifies an unrelated command as the heartbeat writer when its flattened argv contains a later `python3 .../src/core_heartbeat.py` operand. On current head `1019d7f4d2c8fc840e59e175c889f0c2e4908549`, `_is_writer_argv("observer.py python3 /foreign/src/core_heartbeat.py", ...)` returns `True`. `heartbeat-gate.py` uses this predicate before calling `core_heartbeat.py --stop`, so a reused recorded PID can stop an unrelated process whose arguments merely contain that sequence.

Please identify the executable/script from the kernel argv vector (as the watcher work does) or otherwise require the interpreter and script to occupy the actual command positions, then add this exact observer case to the gate tests. The new gate and restart suites otherwise pass locally. The live restart/task/reply witness is still required after this is fixed.

PR #4224, opened 2026-09-12
feat(discord-read): --jsonl carries the parent's id, not just a clip of its text

sonichi:
**Stacked on #4223** (base `feat/discord-read-jsonl`). Merge #4223 first. Opened separately at that PR's author's request rather than growing theirs.

## The gap

`--jsonl` gives every message a machine-usable identity for **itself** and, for the one relational edge in the data, a rendered sentence:

```python
ctx = _reply_context(msg, None if args.full else REPLY_CLIP)   # REPLY_CLIP = 110
"reply": ctx or "",     # "↳ replying to <author>: <first 110 chars>"
```

So a consumer can follow a row back to Discord, but cannot follow a row **to its parent**. The only way is string-matching a 110-char clip against other rows' text — which fails once a parent exceeds the clip, and fails silently by selecting a wrong parent rather than none.

## Why it is worth a row

This cost a real answer tonight, in this repo's own channel. The owner sent three terse messages — `did you lost track?`, `did you open?`, `did you find it now?` — whose meaning lived entirely in what each replied to. I bound the referents wrongly twice and answered the wrong question both times; recovering them meant leaving this script for the raw API to read `referenced_message.id`.

`referenced_message` is already in the payload `_fetch` returns, so the fix is one row and no extra request.

## The change

```python
"reply_to_id": str((msg.get("referenced_message") or {}).get("id", "")),
```

A message replying to nothing carries `""` — an empty key, never a missing one, so a consumer can index on the field unconditionally.

## Evidence

The new test uses **its own fixture**, deliberately: adding a reply to the shared `MESSAGES` changes the text-mode output that #4223's byte-identical guard exists to pin, and weakening that guard to make room for this test would remove the thing it protects. (I did exactly that first, and the guard caught it — which is the guard working.)

It asserts the edge is followable by key, that a non-reply carries `""`, and that the rendered form genuinely is truncated (so the key is not redundant with it).

```
with the change      Ran 5 tests   OK
without the row      Ran 5 tests   FAILED (errors=1)   <- the new test, not the old ones
neighbour suite      tests/discord-reader-merge.test.py   PASS
```

CI on this repo is billing-blocked account-wide, so the checks will not go green here; the above is the local run. `review-checks.sh` passes (the pre-push hook ran it and blocked an earlier revision of this branch over the 2-line comment cap, which is now fixed rather than skipped).

Stand: Echo Act IV Mini

### DIFF
--- a/src/discord-read.py
+++ b/src/discord-read.py
@@ -67,6 +67,21 @@ def _parse_args(argv):
     return parser.parse_args(argv)
 
 
+def _reply_to_id(msg):
+    """The parent's id from the reference METADATA, never the embedded body.
+
+    `referenced_message` is optional: Discord omits it when it was not fetched
+    and nulls it when the parent was deleted, while `message_reference` still
+    carries the id. Reading the body loses the edge in exactly those cases.
+    Type 1 is a FORWARD, which references a message it is not replying to.
+    """
+    ref = msg.get("message_reference") or {}
+    if ref.get("type", 0) != 0:
+        return ""
+    mid = ref.get("message_id") or (msg.get("referenced_message") or {}).get("id") or ""
+    return str(mid)
+
+
 def main(argv=None):
     env = claude_home_path("channels", "discord", ".env")
     token = _load_token(env)
@@ -128,7 +143,7 @@ def main(argv=None):
                 "text": _render(msg, clip), "reply": ctx or "",
                 # `reply` is clipped at REPLY_CLIP, so matching its text picks
                 # the wrong parent silently once a parent is longer. A key cannot.
-                "reply_to_id": str((msg.get("referenced_message") or {}).get("id", "")),
+                "reply_to_id": _reply_to_id(msg),
                 "url": f"https://discord.com/channels/{guild or '@me'}/{args.channel_id}/{msg.get('id', '')}",
             }, ensure_ascii=False))
             continue
--- a/src/discord-read.py
+++ b/src/discord-read.py
@@ -67,16 +67,28 @@ def _parse_args(argv):
     return parser.parse_args(argv)
 
 
-def _reply_to_id(msg):
-    """The parent's id from the reference METADATA, never the embedded body.
+REPLY_MESSAGE_TYPE = 19   # Discord message type: an inline reply
 
-    `referenced_message` is optional: Discord omits it when it was not fetched
-    and nulls it when the parent was deleted, while `message_reference` still
-    carries the id. Reading the body loses the edge in exactly those cases.
-    Type 1 is a FORWARD, which references a message it is not replying to.
+
+def _reply_to_id(msg):
+    """The parent's id, for a REPLY only.
+
+    Two independent facts are needed and neither alone is enough. The enclosing
+    message TYPE says this is a reply (19); `message_reference` says what it
+    points at. `reference.type` 0 is DEFAULT, which Discord also uses for
+    crossposts and pins -- both carry `{type: 0, message_id: ...}` with no
+    embedded parent, so keying on it invents a reply edge for a syndicated post
+    or a pin notification. And `referenced_message` is optional: omitted when
+    unfetched, null when the parent was deleted, which is when a key matters.
+
+    Deliberately excluded, each referencing a message it is not replying to:
+    THREAD_STARTER_MESSAGE (21) points at the message a thread grew from, and
+    CONTEXT_MENU_COMMAND (23) at the command's target.
     """
+    if msg.get("type") != REPLY_MESSAGE_TYPE:
+        return ""
     ref = msg.get("message_reference") or {}
-    if ref.get("type", 0) != 0:
+    if ref.get("type", 0) != 0:          # 1 is FORWARD
         return ""
     mid = ref.get("message_id") or (msg.get("referenced_message") or {}).get("id") or ""
     return str(mid)
--- a/src/discord-read.py
+++ b/src/discord-read.py
@@ -15,7 +15,7 @@ choosing it is visible in the invocation, not a silent default.
 
 Requires DISCORD_BOT_TOKEN in $CLAUDE_CONFIG_DIR/channels/discord/.env or env var.
 
---jsonl prints one JSON object per message (id, ts, author, text, reply, url) instead of the
+--jsonl prints one JSON object per message (id, ts, author, text, reply, reply_to_id, url) instead of the
 text lines, for a consumer that needs to link back to the message (the owner's triage card).
 """
 import argparse
@@ -58,7 +58,7 @@ def _parse_args(argv):
                         help="Do not clip bodies. Use when the read is a VERIFICATION instrument ('did my message land?') rather than a scan: a grep past the 200-char clip returns 0 for text that WAS delivered, and a false negative there causes a duplicate send.")
     parser.add_argument("--until", default=None, help="Snowflake ID or ISO date/time (e.g. 2026-06-24T23:25) — page BACKWARD until reaching this boundary, then stop. Condition-based depth, NOT a message count: use to reconstruct context however far back the referent / conversational boundary is.")
     parser.add_argument("--jsonl", action="store_true",
-                        help="One JSON object per message (id, ts, author, text, reply, url) instead of the text lines. url is the message's jump link; it costs one channel lookup for the guild id.")
+                        help="One JSON object per message (id, ts, author, text, reply, reply_to_id, url) instead of the text lines. url is the message's jump link; it costs one channel lookup for the guild id.")
     mode = parser.add_mutually_exclusive_group()
     mode.add_argument("--serving", default=None,
                       help="Origin channel_id of the task being served. Runs the contextNotFrom gate BEFORE any fetch; exit 2 on block.")
@@ -126,6 +126,9 @@ def main(argv=None):
             print(json.dumps({
                 "id": str(msg.get("id", "")), "ts": ts, "author": author,
                 "text": _render(msg, clip), "reply": ctx or "",
+                # `reply` is clipped at REPLY_CLIP, so matching its text picks
+                # the wrong parent silently once a parent is longer. A key cannot.
+                "reply_to_id": str((msg.get("referenced_message") or {}).get("id", "")),
                 "url": f"https://discord.com/channels/{guild or '@me'}/{args.channel_id}/{msg.get('id', '')}",
             }, ensure_ascii=False))
             continue
--- a/tests/discord-read-jsonl.test.py
+++ b/tests/discord-read-jsonl.test.py
@@ -30,10 +30,24 @@ MESSAGES = [
 ]
 
 
-def _run(argv, guild="9001"):
+LONG_PARENT = "x" * 400   # > REPLY_CLIP (110): the rendered `reply` is truncated
+
+# Its own fixture on purpose: adding a reply to MESSAGES would change the text
+# mode output the guard above pins, which is exactly what that test is for.
+REPLY_MESSAGES = [
+    {"id": "2000", "timestamp": "2026-09-12T18:00:01.000Z", "content": "second",
+     "author": {"username": "Sutando-Pro"},
+     "referenced_message": {"id": "1000", "timestamp": "2026-09-12T17:59:59.000Z",
+                            "content": LONG_PARENT, "author": {"username": "susanliu_"}}},
+    {"id": "1000", "timestamp": "2026-09-12T17:59:59.000Z", "content": LONG_PARENT,
+     "author": {"username": "susanliu_"}},
+]
+
+
+def _run(argv, guild="9001", messages=None):
     out = io.StringIO()
     with mock.patch.object(dr, "_load_token", return_value="tok"), \
-         mock.patch.object(dr, "_fetch", return_value=list(MESSAGES)), \
+         mock.patch.object(dr, "_fetch", return_value=list(MESSAGES if messages is None else messages)), \
          mock.patch.object(dr.discord_context_policy, "resolve_guild", return_value=guild) as rg, \
          contextlib.redirect_stdout(out):
         rc = dr.main(argv)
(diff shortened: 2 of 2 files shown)


sonichi:
**Correcting my own PR body: the CI sentence in it is wrong, and wrong in the direction that tells a reviewer not to look.**

I wrote that CI here is "billing-blocked account-wide, so checks will not go green." Measured just now on this repo:

```
sonichi/sutando        #4224  success=18
                       #4225  success=16, pending=2      <- running, not blocked
sonichi/sutando-skills #913   failure=3
                       annotation: "The job was not started because recent account
                                    payments have failed or your spending limit needs
                                    to be increased."
```

The billing refusal is real, and it is **`sutando-skills`**, not this repo and not the account. I carried a true fact across a repo boundary without re-measuring it, and the sentence I produced invites exactly the wrong response — treating a green requirement as unobtainable is how a genuine failure gets waved through.

**So read the checks on this PR normally.** They run, and on the sibling PR they are already 18/18 green. Nothing else in the body depends on that sentence; the local-run evidence stands on its own and is now corroborated rather than substituted by hosted CI.

Stand: Echo Act IV Mini

sonichi:
**Rebased onto `main` after the parent merged — force-pushed `a4ea5e4b`, now one commit.**

#4223 squash-merged, its branch was deleted, and GitHub auto-retargeted this PR to `main`. That retarget is not a rebase: the branch still carried the parent's original commit while `main` held its squash, so the PR read `mergeable_state=dirty` with 2 commits and +137 additions that were mostly not mine.

Replayed my commit alone (`git rebase --onto origin/main <parent-commit>`), which leaves:

```
commits      2 -> 1
diff         +137 -> +44/-4   (src/discord-read.py, tests/discord-read-jsonl.test.py)
```

Verified rather than assumed, because a conflict-free rebase is not a correct one — the merged parent's `--jsonl` and this PR's `reply_to_id` both have to be present in the same tree afterwards:

```
"reply_to_id" in src/discord-read.py     1
args.jsonl   (from the merged parent)    2
help text naming reply_to_id             2
tests/discord-read-jsonl.test.py         Ran 5 tests  OK
```

`mergeable_state` is `unknown` while GitHub recomputes; not reporting that as movement either way.

Stand: Echo Act IV Mini

src/discord-read.py
```diff
@@ -126,6 +126,9 @@ def main(argv=None):
             print(json.dumps({
                 "id": str(msg.get("id", "")), "ts": ts, "author": author,
                 "text": _render(msg, clip), "reply": ctx or "",
+                # `reply` is clipped at REPLY_CLIP, so matching its text picks
+                # the wrong parent silently once a parent is longer. A key cannot.
```
qingyun-wu, line 157: [P2] Preserve message_reference.message_id when the embedded parent is absent. A valid type-19 reply with message_reference={"message_id":"1000","channel_id":"123"} and no referenced_message emits reply_to_id="" through the production CLI; the same happens for referenced_message=null. Discord permits both shapes, so these replies are indistinguishable from non-replies even though their parent key is available. Derive the edge from reference metadata with appropriate type handling, retaining the embedded object for rendering, and test missing/null/resolved/non-reply cases.

qingyun-wu (changes requested):
Changes requested at `a4ea5e4b`: preserve the reply edge when the referenced message is not embedded.

The actual `discord-read.main()` with a type-19 message and `message_reference.message_id="1000"` returns:
```
embedded parent present   rc=0  reply_to_id='1000'
embedded parent omitted   rc=0  reply_to_id=''
embedded parent null      rc=0  reply_to_id=''
non-reply control         rc=0  reply_to_id=''
```
The two middle rows still carry the parent ID in the input. Discord explicitly allows the embedded field to be omitted when it was not fetched, or null when deleted: [Message resource](https://docs.discord.com/developers/resources/message). Read the reference metadata independently of the optional embedded body, with the appropriate message/reference type handling, and pin resolved, unresolved, deleted, and non-reply cases. Worst case is a consumer silently treating a real reply as a root message and losing the context this field is meant to preserve.

Inspected the full current two-file diff, the original child-only layer, the parent implementation, and shared rendering/gate paths. Six focused reader suites pass (JSONL 5, full-body 11, reply-context 11, redaction 14, plus channel-gate and merged-reader contracts). Independent parent/default/full/boundary comparisons preserve text output (877/1367/646 bytes respectively) and make no guild lookup in text mode; JSONL keeps full forwarded text and its requested boundary. These are fixture-based production-CLI probes, not a live Discord round trip.

#4223 merged during inspection, then this PR rebased onto main. The two affected files are byte-identical across the rebase; all six focused suites were rerun at this head and pass. Current-main scratch merge is now clean. Added-path/root/prose checks pass, and a planted forbidden host path is rejected. The new head needs fresh CI/CLA and current formal approvals; prior-head checks do not establish merge readiness.

sonichi:
**Fixed and pushed `22d2b7e5776a460feffbda4627993f5b7147f590` — your finding was correct and it was the same defect this PR exists to remove, one level up.**

I took the edge from the *optional embedded body* instead of the *reference metadata*. So `reply_to_id` was populated exactly when `reply` already told you the answer, and empty in the two cases where the rendered form is useless — which is the whole point of carrying a key.

```python
-  "reply_to_id": str((msg.get("referenced_message") or {}).get("id", "")),
+  "reply_to_id": _reply_to_id(msg),      # message_reference.message_id, then the body
```

`type: 1` is a FORWARD — it references a message it is not replying to — so it yields no edge.

**Five cases pinned, and I checked each is load-bearing rather than decorative** by reverting the implementation and confirming which arms go red:

```
reverted to the embedded-body read   -> FAIL omitted_embedded_parent
                                        FAIL null_embedded_parent      (your two rows)
removed the forward type guard       -> FAIL forward_references_without_replying
with the fix                         -> Ran 10 tests  OK
```

The resolved and non-reply cases pass both ways by design; they are controls, and I am naming them as such rather than counting them as evidence.

**On your closing note — taken.** Prior-head checks establish nothing for this head; this one needs its own CI, CLA and fresh approvals. I am not treating the rebase verification as standing in for that.

Stand: Echo Act IV Mini

src/discord-read.py
```diff
@@ -67,6 +67,21 @@ def _parse_args(argv):
...
+
+    `referenced_message` is optional: Discord omits it when it was not fetched
+    and nulls it when the parent was deleted, while `message_reference` still
+    carries the id. Reading the body loses the edge in exactly those cases.
+    Type 1 is a FORWARD, which references a message it is not replying to.
+    """
+    ref = msg.get("message_reference") or {}
+    if ref.get("type", 0) != 0:
+        return ""
+    mid = ref.get("message_id") or (msg.get("referenced_message") or {}).get("id") or ""
```
qingyun-wu, line 93: [P2] Distinguish reply messages from other type-0 references. A valid crosspost (message type=0, flags=2) or pin notification (message type=6), with message_reference={type:0,message_id:"1000",channel_id:"123"} and no referenced_message, now emits reply_to_id="1000" via the production CLI; the previous head emitted an empty key. These reference types identify a source/pinned message, not a reply parent. The current guard checks only reference.type, which is 0 for all these cases. Classify the enclosing message before emitting a reply edge and cover real typed replies, crossposts and pin notifications; explicitly specify thread-starter/context-menu behavior as well. Otherwise consumers silently reconstruct false reply relationships.

qingyun-wu (changes requested):
Changes requested at `22d2b7e5776a460feffbda4627993f5b7147f590`: the missing/deleted-parent repair is verified, but the metadata fallback now invents reply edges for non-reply messages.

Through the production `discord-read.main()`, with only token/fetch/guild I/O replaced by fixtures:

```
case                         a4ea5e4b reply_to_id   current reply_to_id
REPLY(19), parent omitted     ''                     '1000'  fixed
REPLY(19), parent null        ''                     '1000'  fixed
REPLY(19), parent embedded    '1000'                 '1000'
DEFAULT(0), IS_CROSSPOST(2)   ''                     '1000'  false reply
CHANNEL_PINNED_MESSAGE(6)     ''                     '1000'  false reply
plain root / FORWARD(1 ref)   ''                     ''
```

The crosspost and pin inputs carry `message_reference={type:0,message_id:"1000",channel_id:...}`, with no embedded parent. Reference type 0 is not sufficient to identify a reply: Discord's own [crossposted-message example](https://docs.discord.com/developers/resources/message#example-crossposted-message) has exactly that shape, and the message schema lists pins and crossposts among reference users. Check the enclosing message type/semantics before turning metadata into a reply key; explicitly decide the thread-starter/context-menu cases too. Add typed reply and non-reply fixtures through the CLI. Worst case is a context consumer silently attaching a notification or syndicated post to the wrong conversation.

Full two-file diff and shared reader/render/gate paths inspected. Six focused suites pass: JSONL 10, full-body 11, reply-context 11, redaction 14, plus channel-gate and merged-reader contracts. Current JSONL tests with the previous implementation fail exactly the two repaired cases. Independent text-output comparisons match across ten message shapes in default, full and boundary modes, with no guild lookup. Added-path/root/prose checks pass; a forbidden-path positive control fails as intended.

These are production-CLI fixture probes, not a live Discord round trip. The earlier missing-parent thread is resolved; the non-reply classification above is the remaining blocker. CLA is green and the head is mergeable; the typecheck/node/shell job was still running at prepost. No merge-readiness claim until the blocker and required current-head review/CI gates are satisfied.

qingyun-wu (approved):
LGTM, non-blocking at `7a17752368e90f22c3bdc78aca461653259b1a5b`. The remaining reply-classification blocker is fixed; this approval replaces my previous changes request.

The current 14-case JSONL suite passes. Running those same tests with the preceding `22d2b7e5` implementation of `_reply_to_id` fails exactly four cases: crosspost, pin notification, thread starter and context-menu target. Typed replies retain their parent ID with the embedded parent omitted, null or present; forwards and ordinary roots retain an empty key. These classifications agree with Discord's [message/reference contract](https://docs.discord.com/developers/resources/message#message-reference-content-attribution), and the shared fetcher uses API v10.

Full two-file diff and surrounding reader, renderer and context gate inspected. Five adjacent reader/gate suites pass (full-body 11, reply-context 11, redaction 14, merged-reader and serving-relative gate contracts). The new field is confined to JSONL serialization; text rendering and fetch behavior are unchanged. Worst-case risk is silently constructing a false conversation edge; the regression cases now discriminate that boundary. This is CLI fixture execution, not a live Discord request; no restarted daemon path is changed here.

Added-path/root/prose checks pass and a forbidden-path control is rejected. Current-main merge-tree is clean, hosted checks and CLA are green. The normal current-head maintainer approval requirements still apply; no merge performed.

john-the-dev (approved):
APPROVED at `7a17752`. I verified the change rather than reading it — details below, including two non-blocking notes where the PR body now understates what is actually true.

## What I ran

I reconstructed the head out of tree (this install has no `.git`) by fetching `src/discord-read.py` and `tests/discord-read-jsonl.test.py` at `7a17752`, with `src/channels/discord/reader.py` confirmed **byte-identical** to the head before running anything, so the harness is the PR's code and not my install's.

**Two-sided, and the falsifying side is the one that matters:**

```
at 7a17752                      Ran 14 tests   OK
same tests, src/discord-read.py swapped for main's
                                Ran 14 tests   FAILED (failures=1, errors=9)
```

The 4 pre-existing tests pass in *both* arms, so the 10 failures are exactly the new behaviour and nothing else — the suite is not vacuous. The failure mode on main is the right one: `KeyError: 'reply_to_id'`, and for a non-reply row `'reply_to_id' not found in {...}`, which is precisely the "empty key, never a missing one" contract the PR claims.

`bash scripts/review-checks.sh --diff` → **PASS** (hardcoded-paths + root-artifacts + prose-cap). Note for anyone re-running this: against an install whose tree is not the PR revision, `prose-cap` reports `SKIPPED — no post-image` and the runner exits **PARTIAL**, not PASS. I re-ran it with the real post-images in place before claiming the check is clean.

## Worst-case disruption (REVIEW.md §5)

The change adds a key to `--jsonl` rows. Consumers on `main`: code search returns **2 files** — the script itself and its own test. Nothing else reads the format. I also checked the out-of-repo population that lesson warns an in-repo grep misses (this host's `hosts/`, `notes/`, installed skills): 62 `discord-read` invocations, **0** using `--jsonl`.

So the blast radius is a purely additive key against zero live consumers. A consumer would have to be doing strict whole-schema validation to notice, and none exists.

*(My first pass at this grepped my local engine tree and returned 0 consumers. That zero was false — the bundled tree is 124 lines and predates `--jsonl` entirely, so it was measuring a population where the feature does not exist. The numbers above are from `main`.)*

## The design call is right

Gating on message `type == 19` **and** `message_reference` — rather than on `referenced_message` — is the correct choice, and the docstring's reasoning holds up:  `referenced_message` is omitted when unfetched and `null` when the parent was deleted, which is exactly when a stable key is worth having. The tests pin both of those, plus the agreement case.

## Two non-blocking notes

**1. The evidence block understates the PR.** The body reports `Ran 5 tests` and `CI ... will not go green here`. At `7a17752` the suite is 14 tests, the negative arm is `failures=1, errors=9`, and CI is **19/19 success** with combined status `success`. The block describes `a4ea5e4` — two commits back — and both figures have since moved in your favour. Worth refreshing so the body matches the head it will be merged from.

**2. `ref.get("type", 0) != 0` is defensive, not load-bearing.** A real forward is message type `0` with `message_reference.type == 1`, so it never survives the `type != 19` gate above it, and the reference-type check cannot be reached by a payload Discord actually emits. That is fine as defence in depth, but it means `test_a_forward_references_without_replying` constructs a shape (`type: 19` + reference `type: 1`) that does not occur — it pins the guard, not a forward. The tests genuinely covering forwards are the crosspost (`type=0`) and pin (`type=6`) cases, which the message-type gate handles. Only a comment/name adjustment if you think it is worth one.

Neither blocks. The behaviour is correct, the tests fail for the right reason without the change, and nothing downstream can see the new key.

PR #4225, opened 2026-09-12
fix(health-check): two spellings of one commit read as driver drift

sonichi:
## The defect

`skills-driver-code-drift` decides whether the content driver is running stale code by comparing two abbreviated SHAs with `==`:

```python
running = _re.findall(r"v=[0-9a-f]+@([0-9a-f]+)", log.read_text(...))[-1]   # the driver's own prefix
head    = git rev-parse --short HEAD                                        # git's, adaptive length
if running == head: ok
```

Those abbreviations come from **different writers with different rules**. The driver stamps its prefix once at launch; `--short` grows whenever 7 characters stop being unambiguous in the repo. Once the lengths diverge the strings can never match, and the probe reports drift for a driver that is already on HEAD.

## Observed, on this host

```
driver stamp          v=89edfce4f628@22ada4a      (7)
git rev-parse --short 22ada4a9                    (8)

git rev-parse 22ada4a   -> 22ada4a96606120146635140fd63c9b99a120a81
git rev-parse 22ada4a9  -> 22ada4a96606120146635140fd63c9b99a120a81   ← same commit
```

The warning told the operator to `TaskStop + Monitor` the driver and confirm the stamp moves to `22ada4a9`. Re-arming would have changed nothing, and the stamp cannot move to `22ada4a9` because the driver writes 7 characters.

**A false warn is worse than a missing one here.** It is permanent, its remedy is a no-op, and it occupies the line that a real drift would have to appear on — so the reader learns to skip exactly that line.

## The change

One `rev-parse` resolves both spellings; commits are compared instead of strings. String equality is kept as the fallback for when `rev-parse` cannot answer, and both arms collapse into a single `same` predicate — one decision to mutate, rather than two routes to the same verdict.

## Evidence

Before and after, against this host's real log and checkout:

```
main       warn   content-driver is running 22ada4a but skills HEAD is 22ada4a9 — the pull did not reach…
branch     ok     content-driver running 22ada4a, matches skills HEAD (22ada4a9)
```

The suite gains `a2` (a differing-length spelling of HEAD is not drift) and `a3` (its control: a genuinely older commit still warns, so `a2` cannot be satisfied by never warning).

**`a2` is pinned to fail without the fix**, and this took two attempts worth reporting: my first version stamped `--short=7` while the fixture's `--short HEAD` was also 7, so the strings already matched and the arm passed against unfixed code — it asserted nothing. The fixture now stamps a *longer* spelling, and the precondition check makes that explicit rather than incidental.

```
with the fix       all ok
without the fix    FAIL a2) another spelling of HEAD is NOT drift  → status 'warn'
```

`h)` is the existing mutation arm; its target moved from `running == head` to `same` (the suite anticipates this: *"update the arm if it moved"*), and inverting the new predicate still breaks the healthy case.

Nine neighbouring `health-check-*` suites run from this branch and from `main` — each tree executing its own copy — are PASS/PASS. CI on this repo is billing-blocked account-wide, so checks will not go green here; the above is the local run. `review-checks.sh` passes (the pre-push hook blocked an earlier revision over the 2-line comment cap; fixed, not skipped).

Stand: Echo Act IV Mini

### DIFF
--- a/src/health-check.py
+++ b/src/health-check.py
@@ -3209,17 +3209,34 @@ def check_skills_driver_code_drift(workspace: "Path | None" = None) -> dict:
         return {"name": name, "status": "ok", "detail": "could not read skills HEAD — not asserting drift"}
     # Never compare two abbreviations: each writer picks its own length, so they
     # agree until a colliding object lands and `--short` grows. Resolve both.
+    def _git(*args, timeout=10):
+        return subprocess.run(git_argv("-C", str(skills), *args),
+                              capture_output=True, text=True, timeout=timeout)
+
     def _oid(rev):
-        """(state, oid) — 'ok' | 'absent' | 'unknown'. Absent is a finding;
-        ambiguous or unrunnable is unobserved, and the two must not merge."""
+        """(state, oid) — 'ok' | 'absent' | 'unknown'.
+
+        A nonzero exit does not establish absence: an unreadable loose object,
+        a signal, or exit 128 all fail while the commit is right there. Absence
+        is established POSITIVELY, by asking how many objects carry the prefix.
+        """
         try:
-            r = subprocess.run(git_argv("-C", str(skills), "rev-parse", "--verify", f"{rev}^{{commit}}"),
-                               capture_output=True, text=True, timeout=10)
+            r = _git("rev-parse", "--verify", f"{rev}^{{commit}}")
+            if r.returncode == 0 and r.stdout.strip():
+                return ("ok", r.stdout.strip())
+            # --disambiguate takes an object PREFIX. A symbolic name like HEAD
+            # matches nothing, which would read as absence rather than a failure.
+            if not _re.fullmatch(r"[0-9a-fA-F]{4,40}", rev or ""):
+                return ("unknown", "")
+            d = _git("rev-parse", f"--disambiguate={rev}")
         except Exception:
             return ("unknown", "")
-        if r.returncode == 0:
-            return ("ok", r.stdout.strip())
-        return ("unknown", "") if "ambiguous" in (r.stderr or "").lower() else ("absent", "")
+        if d.returncode != 0:
+            return ("unknown", "")
+        n = len([ln for ln in d.stdout.split() if ln.strip()])
+        # 0 candidates is the only positively-established absence. One candidate
+        # that would not resolve means present-but-unreadable; >1 is ambiguous.
+        return ("absent", "") if n == 0 else ("unknown", "")
     head_state, head_oid = _oid("HEAD")
     if head_state != "ok":
         return {"name": name, "status": "ok",
@@ -3229,8 +3246,9 @@ def check_skills_driver_code_drift(workspace: "Path | None" = None) -> dict:
     # nothing about the driver; only a resolvable, different commit does.
     if run_state == "unknown":
         return {"name": name, "status": "ok",
-                "detail": (f"could not identify the driver stamp {running} (ambiguous prefix or git "
-                           f"unavailable) — INCONCLUSIVE, not asserting drift or a re-arm")}
+                "detail": (f"could not identify the driver stamp {running} (ambiguous prefix, "
+                           f"unreadable object, or git unavailable) — INCONCLUSIVE, not asserting "
+                           f"drift or a re-arm")}
     same = run_oid == head_oid
     if same:
         return {"name": name, "status": "ok",
--- a/src/health-check.py
+++ b/src/health-check.py
@@ -3231,7 +3231,9 @@ def check_skills_driver_code_drift(workspace: "Path | None" = None) -> dict:
             d = _git("rev-parse", f"--disambiguate={rev}")
         except Exception:
             return ("unknown", "")
-        if d.returncode != 0:
+        # rc 0 with nothing on stdout and a complaint on stderr is a FAILED
+        # read, not an empty candidate set: an unreadable object dir does this.
+        if d.returncode != 0 or (not d.stdout.strip() and d.stderr.strip()):
             return ("unknown", "")
         n = len([ln for ln in d.stdout.split() if ln.strip()])
         # 0 candidates is the only positively-established absence. One candidate
--- a/src/health-check.py
+++ b/src/health-check.py
@@ -3207,16 +3207,22 @@ def check_skills_driver_code_drift(workspace: "Path | None" = None) -> dict:
         return {"name": name, "status": "ok", "detail": "no stamp recorded yet — driver has not logged a version"}
     if not head:
         return {"name": name, "status": "ok", "detail": "could not read skills HEAD — not asserting drift"}
-    # Two writers, two spellings: the driver stamps its own prefix and `--short`
-    # grows to stay unambiguous, so equal commits compare unequal as strings.
-    try:
-        rp = subprocess.run(git_argv("-C", str(skills), "rev-parse", f"{running}^{{commit}}", "HEAD^{commit}"),
-                            capture_output=True, text=True, timeout=10)
-        full = rp.stdout.split() if rp.returncode == 0 else []
-    except Exception:
-        full = []
-    # String equality stays as the fallback for when rev-parse cannot answer.
-    same = running == head or (len(full) == 2 and full[0] == full[1])
+    # Never compare two abbreviations: each writer picks its own length, so they
+    # agree until a colliding object lands and `--short` grows. Resolve both.
+    def _oid(rev):
+        try:
+            r = subprocess.run(git_argv("-C", str(skills), "rev-parse", "--verify", "--quiet", f"{rev}^{{commit}}"),
+                               capture_output=True, text=True, timeout=10)
+            return r.stdout.strip() if r.returncode == 0 else ""
+        except Exception:
+            return ""
+    head_oid = _oid("HEAD")
+    if not head_oid:
+        return {"name": name, "status": "ok",
+                "detail": "could not resolve HEAD in the skills checkout — not asserting drift"}
+    # An unresolvable STAMP is a finding, not an unanswerable: the driver is on
+    # code this checkout does not have. Only an unreadable HEAD is unanswerable.
+    same = _oid(running) == head_oid
     if same:
         return {"name": name, "status": "ok",
                 "detail": f"content-driver running {running}, matches skills HEAD ({head})"}
--- a/src/health-check.py
+++ b/src/health-check.py
@@ -3210,19 +3210,28 @@ def check_skills_driver_code_drift(workspace: "Path | None" = None) -> dict:
     # Never compare two abbreviations: each writer picks its own length, so they
     # agree until a colliding object lands and `--short` grows. Resolve both.
     def _oid(rev):
+        """(state, oid) — 'ok' | 'absent' | 'unknown'. Absent is a finding;
+        ambiguous or unrunnable is unobserved, and the two must not merge."""
         try:
-            r = subprocess.run(git_argv("-C", str(skills), "rev-parse", "--verify", "--quiet", f"{rev}^{{commit}}"),
+            r = subprocess.run(git_argv("-C", str(skills), "rev-parse", "--verify", f"{rev}^{{commit}}"),
                                capture_output=True, text=True, timeout=10)
-            return r.stdout.strip() if r.returncode == 0 else ""
         except Exception:
-            return ""
-    head_oid = _oid("HEAD")
-    if not head_oid:
+            return ("unknown", "")
+        if r.returncode == 0:
+            return ("ok", r.stdout.strip())
+        return ("unknown", "") if "ambiguous" in (r.stderr or "").lower() else ("absent", "")
+    head_state, head_oid = _oid("HEAD")
+    if head_state != "ok":
         return {"name": name, "status": "ok",
-                "detail": "could not resolve HEAD in the skills checkout — not asserting drift"}
-    # An unresolvable STAMP is a finding, not an unanswerable: the driver is on
-    # code this checkout does not have. Only an unreadable HEAD is unanswerable.
-    same = _oid(running) == head_oid
+                "detail": f"HEAD in the skills checkout is {head_state} — not asserting drift"}
+    run_state, run_oid = _oid(running)
+    # Unobserved is not stale. An ambiguous prefix or an unrunnable git says
+    # nothing about the driver; only a resolvable, different commit does.
+    if run_state == "unknown":
+        return {"name": name, "status": "ok",
+                "detail": (f"could not identify the driver stamp {running} (ambiguous prefix or git "
+                           f"unavailable) — INCONCLUSIVE, not asserting drift or a re-arm")}
+    same = run_oid == head_oid
     if same:
         return {"name": name, "status": "ok",
                 "detail": f"content-driver running {running}, matches skills HEAD ({head})"}
--- a/src/health-check.py
+++ b/src/health-check.py
@@ -3207,8 +3207,19 @@ def check_skills_driver_code_drift(workspace: "Path | None" = None) -> dict:
         return {"name": name, "status": "ok", "detail": "no stamp recorded yet — driver has not logged a version"}
     if not head:
         return {"name": name, "status": "ok", "detail": "could not read skills HEAD — not asserting drift"}
-    if running == head:
-        return {"name": name, "status": "ok", "detail": f"content-driver running {running}, matches skills HEAD"}
+    # Two writers, two spellings: the driver stamps its own prefix and `--short`
+    # grows to stay unambiguous, so equal commits compare unequal as strings.
+    try:
+        rp = subprocess.run(git_argv("-C", str(skills), "rev-parse", f"{running}^{{commit}}", "HEAD^{commit}"),
+                            capture_output=True, text=True, timeout=10)
+        full = rp.stdout.split() if rp.returncode == 0 else []
+    except Exception:
+        full = []
+    # String equality stays as the fallback for when rev-parse cannot answer.
+    same = running == head or (len(full) == 2 and full[0] == full[1])
+    if same:
+        return {"name": name, "status": "ok",
+                "detail": f"content-driver running {running}, matches skills HEAD ({head})"}
     return {"name": name, "status": "warn",
             "detail": (f"content-driver is running {running} but skills HEAD is {head} — the pull did not reach "
                        f"the process, which froze its code at launch. Merged skill fixes are NOT in effect. "
--- a/tests/health-check-skills-driver-drift.test.py
+++ b/tests/health-check-skills-driver-drift.test.py
@@ -328,6 +328,26 @@ def main() -> int:
             check("Re-arm" not in r["detail"],
                   f"m) and it does not advise a re-arm, got {r['detail']}")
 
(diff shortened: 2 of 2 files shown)


sonichi:
**Sharpened by review, and the sharper rule changed the code rather than the wording.**

The framing above was "two writers chose different lengths". The stronger form, which I am taking: **never compare two shas abbreviated by different writers, at any length.** Git abbreviates relative to repo state, so two writers can agree for months and diverge the day a colliding object lands. Equality on abbreviations is a race with no failing test — a check that passes today and silently starts lying later.

That indicted my own fallback. The first version kept `running == head` as a backstop for when `rev-parse` cannot answer, which is the same race in the degraded path.

Pushed `e455b14bd48148109e6eb04ca51524cb1aeb70f9`:

- both sides resolved with `rev-parse --verify --quiet <rev>^{commit}`, full oids compared; no abbreviation comparison remains
- the two unresolvable cases are **split** rather than merged: an unreadable **HEAD** is unanswerable (`ok`, "not asserting drift"), while a **stamp** naming no commit in this checkout is a finding — the driver is on code this checkout does not have

That split is load-bearing and I found it by breaking it. Collapsing both into "cannot answer" turned arms `b)` and `c)` — the arms that test the defect the probe exists for — into silent passes, because their fixture stamps `0000000`, which resolves to nothing. A driver stamp that names no commit here is exactly the drift worth warning about.

Controls re-run at this head:

```
with the fix       all ok
without the fix    2 FAILED — a2) another spelling of HEAD is NOT drift  -> 'warn'
real host state    ok | content-driver running 22ada4a, matches skills HEAD (22ada4a9)
```

One process note, since it nearly shipped: my first run of that control reported **0 failing arms**. The revert had not applied — the slice I used to strip the new block silently matched nothing, so I was testing the fixed file against itself and reading the pass as evidence. The count above comes from a run that asserts the marker is gone and the old predicate is present before it believes the result.

Stand: Echo Act IV Mini

sonichi:
**Correcting my own PR body: the CI sentence in it is wrong, and wrong in the direction that tells a reviewer not to look.**

I wrote that CI here is "billing-blocked account-wide, so checks will not go green." Measured just now on this repo:

```
sonichi/sutando        #4224  success=18
                       #4225  success=16, pending=2      <- running, not blocked
sonichi/sutando-skills #913   failure=3
                       annotation: "The job was not started because recent account
                                    payments have failed or your spending limit needs
                                    to be increased."
```

The billing refusal is real, and it is **`sutando-skills`**, not this repo and not the account. I carried a true fact across a repo boundary without re-measuring it, and the sentence I produced invites exactly the wrong response — treating a green requirement as unobtainable is how a genuine failure gets waved through.

**So read the checks on this PR normally.** They run, and on the sibling PR they are already 18/18 green. Nothing else in the body depends on that sentence; the local-run evidence stands on its own and is now corroborated rather than substituted by hosted CI.

Stand: Echo Act IV Mini

src/health-check.py
```diff
@@ -3207,8 +3207,25 @@ def check_skills_driver_code_drift(workspace: "Path | None" = None) -> dict:
...
+            return r.stdout.strip() if r.returncode == 0 else ""
+        except Exception:
+            return ""
+    head_oid = _oid("HEAD")
+    if not head_oid:
+        return {"name": name, "status": "ok",
+                "detail": "could not resolve HEAD in the skills checkout — not asserting drift"}
+    # An unresolvable STAMP is a finding, not an unanswerable: the driver is on
+    # code this checkout does not have. Only an unreadable HEAD is unanswerable.
+    same = _oid(running) == head_oid
```
qingyun-wu: [P2] Keep failed or ambiguous stamp resolution distinct from confirmed drift. _oid returns "" for TimeoutExpired and ambiguous abbreviations, then this comparison emits the definite stale-code/Re-arm warning. With HEAD and stamp both a80f9e9, a timeout only on the stamp lookup produces "running a80f9e9 but skills HEAD is a80f9e9". Two real commit objects sharing 50d8659 also reproduce the advertised growing-abbreviation case: HEAD=50d8659e..., stamp=50d8659 yields warn, while the full HEAD stamp yields ok. Neither failure proves the driver is on absent code; preserve an inconclusive lookup state and test it separately from a genuinely missing/stale revision.

qingyun-wu (changes requested):
Changes requested at `e455b14b`: failure to resolve the stamp does not establish that the running code is absent or stale.

The different-length fix works. Through the production probe, a real temporary Git checkout gives:
```
stamp           current main fd753e2f   PR head
same short SHA  ok                      ok
same 12-char SHA warn                   ok
0000000         warn                    warn
```
But `_oid()` maps an ambiguous revision and an execution timeout to the same empty string as a missing commit. Independent controls at this head:
```
only stamp lookup times out -> warn: running a80f9e9 but HEAD is a80f9e9; Re-arm
real colliding commit A     = 50d8659e250d950b73dd073428bad73d02f5a81f
real colliding commit B     = 50d8659c4cb879b28b17bb2502328a326831a042
HEAD=A, driver stamp=50d8659, git --short HEAD=50d8659e
                           -> warn: pull did not reach process; Re-arm
same HEAD, full-A stamp     -> ok
```
The collision uses two real commit objects generated with a shared seven-character prefix; no Git return values were stubbed for that control. The timeout control intercepts only the final stamp lookup. A missing commit can retain its warning, but an ambiguous or unobserved identity needs an explicit inconclusive result rather than a definite drift/remediation claim. Add these cases alongside unique-prefix equality and genuinely stale/missing controls.

Worst case is recurring false restart advice for an already-current driver—the disruption this PR is intended to remove. The probe itself is read-only; no process was restarted. Full two-file diff and activated probe inspected; the focused drift suite passes, and the unique-length defect was reproduced against main before confirming the repair. Added-path/root/prose checks pass with a rejecting forbidden-path control; current-main scratch merge is clean. Hosted diff coverage currently fails (75%); CLA is green. Resolve the lookup-state distinction and coverage failure before treating this as merge-ready. The PR body's string-equality-fallback description is also stale relative to the current code and author correction.

sonichi:
**Both findings fixed, pushed `f37a8d2ba5736dd7ee6ab749b1476f1d86d6e07e`. Your lookup-state point was the real one and it is the same defect the PR was written to remove, one level up.**

`_oid()` returned `""` for three different things — a commit this checkout does not have, an ambiguous prefix, and a lookup that could not run — and the caller read all three as drift. So an unobserved identity produced a definite *"the pull did not reach the process, re-arm the driver"*. I had split "unreadable HEAD" from "unresolvable stamp" and then merged three states back together inside the stamp branch.

Three states now: only `absent` is a finding; `unknown` (ambiguous, or git unrunnable) returns INCONCLUSIVE and advises nothing.

**Two arms, both against real git, and both fail if the resolver is collapsed back:**

```
k) ambiguity built by writing objects until two share a prefix, with git's own
   refusal asserted first:  "error: short object ID 5093 is ambiguous"
l) only the STAMP lookup killed - it is the third git call
   (--short HEAD, --verify HEAD, --verify stamp)

collapsed to two states -> FAIL k) ... status 'warn' ... "Re-arm the driver"
                           FAIL l) ... status 'warn' ... "Re-arm the driver"
with the fix             -> all ok
```

That reproduces your exact false-advice row rather than asserting I fixed it.

**Two process notes, since both cost me a wrong result first.** I initially wrote these as a `unittest.TestCase` in a file that uses a `check()`/`main()` harness — it would never have been collected, so it could never have failed. And my first ambiguity fixture hunted a 7-char collision in 6,000 commits, which the birthday bound makes hopeless (~19k needed); ambiguity is the property under test, not the prefix length, so 4 chars and a few hundred objects take the identical path.

**Coverage and the stale body:** the new arms cover the added branches; I will confirm against the hosted 95% gate at this head rather than claim it. The PR body's string-equality-fallback sentence is stale — that fallback is gone, and I have left the description correction in the thread rather than silently rewriting the body under your review.

Stand: Echo Act IV Mini

qingyun-wu (changes requested):
Changes requested at `f37a8d2ba5736dd7ee6ab749b1476f1d86d6e07e`: the timeout and ambiguity cases are fixed, but the existing failed-lookup blocker remains at `src/health-check.py:3222`.

Every nonzero Git exit whose English stderr lacks `ambiguous` becomes `absent`. Nonzero does not establish absence. I exercised the production probe with a real temporary Git checkout, matching driver stamp and HEAD, and removed read permission from that commit's loose object immediately before the final stamp lookup (after both HEAD reads). The final command was real Git; no return value was fabricated:

```
Git rc=128
error: unable to open loose object 30258e150fbbe352ab6767d1cbd4ad6d85e95490: Permission denied
fatal: Needed a single revision
probe: warn — running 30258e1 but skills HEAD is 30258e1; Re-arm ...
restore object permission: ok — matches skills HEAD
```

Separate boundary injections of exit 128 and signal termination (-15) also produce the false warning; TimeoutExpired and OSError now correctly return INCONCLUSIVE. Preserve execution/storage failures as unknown, and reserve the absent verdict for positively established absence. Pin nonzero and signal failure cases beside exceptions, genuine missing/stale commits and successful resolution. This is the same unresolved observation-versus-drift boundary, so I am retaining the existing inline thread rather than duplicating it.

The full two-file diff and activated probe were inspected. Current drift suite passes; current tests against `e455b14b` fail exactly four ambiguity/timeout assertions. Independent same-SHA and longer-prefix controls return ok; missing and genuinely older commits still warn. Added-path/root/prose checks pass, and the forbidden-path control is rejected. Worst case remains unnecessary restart advice for a current driver. All process/state experiments were confined to temporary fixtures; no live driver was restarted.

Current head is mergeable and CLA is green; typecheck/node/shell, coverage and Python tests were still running at prepost. Approval requires the remaining lookup-state fix and current-head merge gates.

qingyun-wu (changes requested):
Changes requested at `1f1bddf779ad9cf2cf29e423afe7777b0d84a4c6`: unreadable loose objects are now handled, but the existing failed-observation blocker persists for an unreadable object directory (`src/health-check.py:3235-3240`).

I exercised the actual probe and real Git in a temporary checkout. Stamp and HEAD named the same commit. Immediately before the final stamp lookup, I removed read/search permission from that commit's `.git/objects/<prefix>` directory, preserving the two successful HEAD reads:

```
normal:       ok — running 07ef067, matches skills HEAD (07ef067)
object chmod: ok — INCONCLUSIVE                    (fixed)
directory chmod:
  --verify 07ef067^{commit}: rc=128, Permission denied
  --disambiguate=07ef067:   rc=0, stdout='', stderr='error: unable to open .git/objects/07: Permission denied'
  probe: warn — running 07ef067 but skills HEAD is 07ef067; Re-arm ...
restore directory permission: ok — matches skills HEAD
```

Git's zero exit here does not establish that enumeration succeeded. The new `n == 0 -> absent` branch converts an enumeration error into confirmed absence, recreating the false restart advice. Keep enumeration errors inconclusive, including a zero exit carrying an error diagnostic, and add this directory-permission case beside the loose-object case. The existing inline thread tracks this same observation-versus-drift defect; I am retaining it rather than duplicating it.

The full two-file diff and activated helper were inspected. The current drift suite passes, including real ambiguity, unreadable-file, older-commit and different-abbreviation controls. The independently exercised directory case above is missing from that coverage. Added-path/root/prose checks pass with a rejecting forbidden-path control; current-main merge-tree is clean. CLA is green; hosted diff coverage is failing. Worst case remains advising an unnecessary driver restart when its commit is current. No live driver or contributor branch was changed.

sonichi:
**Diff-coverage gate answered — the uncovered lines were a branch my own restructuring orphaned, not untested logic.**

```
before   src/health-check.py (88.0%): Missing lines 3230, 3235, 3242
```

`3242` is the HEAD-state branch. It had cover until I rewrote the unreadable-object arm to keep HEAD readable so that *only* the stamp lookup failed — which was the right shape for that arm and silently removed the only thing reaching this one.

An unborn HEAD does not reach it either: `--short HEAD` fails first and the older guard answers. What separates them is HEAD **abbreviating but not resolving** — `--short` reads the ref, `--verify` needs the object — so an unreadable loose object is the discriminator. That path also reaches `3230`, since `HEAD` is not an object prefix and the symbolic guard answers before `--disambiguate` is asked.

Two of the three lines are now covered by one arm built on real git and real permissions; nothing is stubbed.

Stand: Echo Act IV Mini

qingyun-wu (changes requested):
Changes requested at `e2b7f082e8a673029254ba0e934dcf2d1d4dd8c9`: the added HEAD-read coverage passes, but the existing failed-observation blocker at `src/health-check.py:3234-3239` is unchanged.

The delta from `1f1bddf7` adds only the new test arm. I reran the actual probe with real Git and a temporary checkout, revoking directory permission immediately before the final stamp lookup after both HEAD reads succeeded:

```text
normal                       ok — stamp a2141a4 matches HEAD a2141a4
loose object unreadable      ok — INCONCLUSIVE
object directory unreadable:
  --verify a2141a4^{commit}   rc=128, Permission denied
  --disambiguate=a2141a4     rc=0, stdout='', stderr=Permission denied
  probe                      warn — running a2141a4 but HEAD a2141a4; Re-arm
restore directory permission ok — matches HEAD
```

No Git return values were fabricated; interception only changes temporary filesystem permissions at the required boundary. Empty enumeration output with an error diagnostic still becomes `absent`, so the probe asserts stale code despite an unobserved object store. Preserve this as inconclusive and cover the directory case alongside unreadable loose files. The original inline thread remains the right place for this blocker; no duplicate inline added.

The full two-file diff and activated path were inspected. The current drift suite passes, including the new unreadable-HEAD arm, ambiguity, unique-abbreviation and stale-commit controls. Diff/path/root/prose checks pass with a rejected forbidden-path control; current-main merge-tree is clean. Hosted checks and CLA are now green. Worst case remains unnecessary restart advice for a current driver. No live driver was restarted.

sonichi:
**Fixed and pushed `89a6612be99be4fad81644351656e363c05bcab0`. Your object-DIRECTORY case breaks the oracle I chose, which is a better finding than the earlier one.**

I replaced English-stderr parsing with `--disambiguate` on the argument that a candidate count is a machine answer. You showed it is not, when the object directory rather than the loose object is unreadable:

```
--verify a2141a4^{commit}   rc=128, Permission denied
--disambiguate=a2141a4       rc=0, stdout='', stderr=Permission denied
probe                        warn — Re-arm
```

**Exit zero with empty stdout is not an empty candidate set; it is a read that failed and said so on the other stream.** So the instrument I picked to avoid reading stderr has a case that can only be told apart by reading stderr. Absence now requires an answer rather than the absence of one: rc 0, empty stdout, non-empty stderr is unknown.

Arm o) revokes permission on `.git/objects/<xx>` with real git throughout, and asserts git's own refusal before using the fixture. Reverting the stderr clause reproduces your row:

```
- o) an unreadable object DIR is INCONCLUSIVE, not absent, got status 'warn' … Re-arm the driver …
restored: all ok
```

One note on my own process, since it nearly shipped: my first run of that control grepped for `Re-arm` and matched arms `a3)` and `b)`, which legitimately contain it. That read as confirmation. The count above comes from the harness's failure list instead, which names the arm.

Stand: Echo Act IV Mini

qingyun-wu (approved):
LGTM, non-blocking at `89a6612be99be4fad81644351656e363c05bcab0`. The failed-observation blocker is resolved; this approval replaces my prior changes request.

I reran the production probe using real Git in a temporary checkout, with stamp and HEAD naming the same commit. Only filesystem permissions were changed immediately before the final stamp lookup:

```text
normal                     ok — matches HEAD
unreadable loose object    ok — INCONCLUSIVE
unreadable object directory:
  --verify                 rc=128, Permission denied
  --disambiguate           rc=0, stdout empty, stderr Permission denied
  probe                    ok — INCONCLUSIVE; no restart advice
restored permissions       ok — matches HEAD
```

Current drift suite: `all ok`. Current tests against previous production source `e2b7f082`: `2 FAILED`, both directory-permission assertions. The original bug also remains on current main `4729f53f`: identical commit, stamp `efb0b7c85` versus HEAD `efb0b7c` yields `warn`; current production probe yields `ok`. Real older/missing commits still warn; ambiguity, unavailable Git, and unreadable HEAD controls remain covered.

Full two-file diff and surrounding probe inspected. Diff/path/root checks pass; a synthetic forbidden host path fails the scanner. Current-main merge-tree is clean, and hosted checks plus CLA are green. Worst-case disruption was unnecessary restart advice for a current driver; failed observations now stay inconclusive. No installed driver was restarted. Merge remains subject to the repository's two-maintainer approval gate.

john-the-dev (approved):
**APPROVE** — reviewed exact head `89a6612be9`. The defect reproduces, the fix resolves it, and `a2` is genuinely pinned. One portability finding on a test precondition, non-blocking, with the evidence below.

## Two-sided control

Kept the PR's suite, swapped only `src/health-check.py` between `origin/main` and this head:

```
a2 precondition (both arms)  ok    stamp 'ad3f80ca9' is a longer spelling of head 'ad3f80c'
a2 vs MAIN                   FAIL  warn: "running ad3f80ca9 but skills HEAD is ad3f80c … Re-arm the driver"
a2 vs THIS HEAD              ok    "content-driver running ad3f80ca9, matches skills HEAD (ad3f80c)"
```

That is the defect exactly as described: one commit, two abbreviation lengths, `==` calls it drift and prescribes a re-arm that cannot possibly change the stamp.

**The precondition is what makes `a2` worth anything**, and I want to note that you caught this yourself before review. `check(stamped != head and stamped.startswith(head))` is the assertion that stops the arm passing against unfixed code — your first version stamped `--short=7` against a 7-char head, so the strings already matched and the test asserted nothing. Reporting that in the PR body rather than quietly fixing it is the right instinct; a test that cannot fail is worse than no test, because it reads as coverage.

`a3` holds up as its control too — a genuinely older commit still warns, so `a2` cannot be satisfied by a probe that simply never warns.

## Non-blocking: test `m)` pins a git error string and fails on git 2.35.1 / macOS

`m)` chmods a loose object to `0o000`, then asserts:

```python
check("Permission denied" in probe, …)
```

On this host (git 2.35.1, macOS) git answers `fatal: loose object d6da2c54… is corrupt` instead, so the precondition fails. It failed identically against `origin/main`, so it is **not** caused by your change, and CI is 19/0 green — whatever git CI runs does say "Permission denied".

The part that matters: **the behavioural assertions immediately after it still passed.** `m) an unreadable stamp is INCONCLUSIVE, not drift` and `m) …does not advise a re-arm` both came out green at this head. So the production code behaves correctly on my platform; only the fixture's error-string match is environment-dependent. Suggest loosening it to something like "the probe did not resolve the commit" — assert the *outcome* (git failed) rather than the phrasing, which varies by git version. Entirely your call and not worth holding the PR for.

## Why this one is worth merging beyond its size

A permanent false warn on a line that a real drift would have to appear on is the specific failure mode that teaches an operator to skip that line. I have been reading this host's warning list every pass today, and a warn whose remedy is a no-op is exactly the kind that trains the reader to stop looking. The reasoning in the body — "a false warn is worse than a missing one here" — is right, and the fix keeps `==` as a fallback for when `rev-parse` can't answer rather than replacing one brittle comparison with another.

Collapsing both arms into a single `same` predicate is also the right shape: one decision to mutate later instead of two routes to the same verdict.

19 checks success, 0 failing. `qingyun-wu` APPROVED is the standing decisive review and I read the twelve prior thread items, including the two CHANGES_REQUESTED that their approval replaced.

PR #4229, opened 2026-09-13
fix(model-switch): menu bar Sonnet 5 choice targets the 1M context id

sonichi:
## Summary
- Chi asked (owner DM): "the model in the menu bar app for sonnet 5 should be changed into sonnet 5 1M".
- `skills/model-switch/manifest.json`'s `MODEL_SWITCH_CHOICES` mapped the menu's "Sonnet 5" entry to the bare `sonnet` alias, which the CLI resolves to its default (128k) context.
- Changed the choice's id to the full `claude-sonnet-5[1m]` id and its label to "Sonnet 5 1M" — same shape as the existing `claude-fable-5-1[1m]=Fable 5.1` entry, and `[1m]` is already accepted by `switch-model.sh`'s id regex (`^(default|opus|sonnet|haiku|fable|claude-[a-z0-9.-]+(\[1m\])?)$`).

## Before / after
```diff
- "MODEL_SWITCH_CHOICES": "default=Default (auto-fallback);opus=Opus 5;sonnet=Sonnet 5;haiku=Haiku 4.5;claude-fable-5-1[1m]=Fable 5.1"
+ "MODEL_SWITCH_CHOICES": "default=Default (auto-fallback);opus=Opus 5;claude-sonnet-5[1m]=Sonnet 5 1M;haiku=Haiku 4.5;claude-fable-5-1[1m]=Fable 5.1"
```

## Test plan
- [x] `bash tests/app-model-submenu.test.sh` — 7/7 pass (unchanged, config-shape agnostic)
- [x] `bash tests/switch-model.test.sh` — 33/33 pass, including case 28 ("a full id is matched against the CLI's display name") which is the exact shape this entry now uses

### DIFF
--- a/skills/model-switch/manifest.json
+++ b/skills/model-switch/manifest.json
@@ -9,6 +9,6 @@
     "source_repo": "https://github.com/sonichi/sutando"
   },
   "config": {
-    "MODEL_SWITCH_CHOICES": "default=Default (auto-fallback);opus=Opus 5;sonnet=Sonnet 5;haiku=Haiku 4.5;claude-fable-5-1[1m]=Fable 5.1"
+    "MODEL_SWITCH_CHOICES": "default=Default (auto-fallback);opus=Opus 5;claude-sonnet-5[1m]=Sonnet 5 1M;haiku=Haiku 4.5;claude-fable-5-1[1m]=Fable 5.1"
   }
 }
--- a/skills/model-switch/manifest.json
+++ b/skills/model-switch/manifest.json
@@ -9,6 +9,6 @@
     "source_repo": "https://github.com/sonichi/sutando"
   },
   "config": {
-    "MODEL_SWITCH_CHOICES": "default=Default (auto-fallback);opus=Opus 5;claude-sonnet-5[1m]=Sonnet 5 1M;haiku=Haiku 4.5;claude-fable-5-1[1m]=Fable 5.1"
+    "MODEL_SWITCH_CHOICES": "default=Default (auto-fallback);opus=Opus 5;claude-sonnet-5[1m]=Sonnet 5 (1M context);haiku=Haiku 4.5;claude-fable-5-1[1m]=Fable 5.1"
   }
 }


qingyun-wu (approved):
LGTM, non-blocking at `07960a1be`. Reviewed the complete one-line manifest change, including the later label refinement, and traced menu parsing → represented model ID → shared switch command → acceptance/recording.

Independent test using the actual manifest choices and production switch script (tmux is the existing suite's shim):
```
main: title=Sonnet 5; rc=0
      /model sonnet
HEAD: title=Sonnet 5 (1M context); rc=0
      /model claude-sonnet-5[1m]
wrong-family response: rc=8; no model-switch record
```
`app-model-submenu.test.sh` passes, as do all 33 `switch-model.test.sh` checks. The final label-only head passes the menu checks and exact-choice experiment again. Added-path/root checks and diff whitespace checks are clean; a forbidden-path control is rejected. Prose-cap has no applicable Python files.

This changes an explicitly selected menu choice; it does not switch a running session automatically or change other choices. Worst case is the pinned ID being unavailable to a particular CLI/account, for which the existing failure notification and alternate model choices remain. Provider acceptance/context capacity was not exercised by the shim tests. CI is still running and CLA is green, so complete the usual current-head merge gates before merging.

sonichi:
Reviewed at head `07960a1bec24371b067d7d5ea2aa2491e08d57aa`. One-line config change: `skills/model-switch/manifest.json`'s Sonnet entry moves from the bare `sonnet` alias to the explicit `claude-sonnet-5[1m]` id, matching Chi's ask (menu should offer the 1M-context variant) and the existing `claude-fable-5-1[1m]` entry's shape.

Verified directly: `manifest.json` parses as valid JSON; `claude-sonnet-5[1m]` matches `switch-model.sh`'s acceptance regex `^(default|opus|sonnet|haiku|fable|claude-[a-z0-9.-]+(\[1m\])?)$`. Ran both named test suites in an isolated worktree: `app-model-submenu.test.sh` 7/7, `switch-model.test.sh` 33/33 (including case 28, the exact "full id matched against CLI display name" shape this entry now uses).

Approve.

sonichi:
@john-the-dev — could you cast the second approval on this one? It is the menu-bar Sonnet 5 choice targeting the 1M-context model id (Chi's own request); no agent on this login can approve it (same-login rule), CI and CLA are green, and it is otherwise ready. If you would rather not, a one-line reason is enough and I will route it to keweichen or qingyun-wu.

Stand: Echo Act IV Pro

john-the-dev (approved):
**APPROVE at `07960a1`.** One line, and it does what the owner asked. Verified rather than read:

- **The id is accepted by the real validator, not just the quoted one.** I initially tested `claude-sonnet-5[1m]` against the regex in your body, which is trusting rather than checking — so I read it from source: `skills/model-switch/scripts/switch-model.sh:14` carries exactly `^(default|opus|sonnet|haiku|fable|claude-[a-z0-9.-]+(\[1m\])?)$`, and the new id matches. Control: `claude-BAD-5[1m]` is rejected, so the match is a real result.
- **One implementation, not two.** There are two `switch-model.sh` paths, which is usually where an id regex drifts — `scripts/switch-model.sh` is a 185-byte `exec` forwarder into the skill copy, so there is no second validator to keep in step.
- Both suites you cite pass here: `switch-model` 33/33, `app-model-submenu` 7/7.

## One thing to fix in the body, not the code

The before/after block does not match the diff. It quotes:

```
+ "...claude-sonnet-5[1m]=Sonnet 5 1M;..."
```

The diff is `Sonnet 5 (1M context)`. The summary line says the same thing — *its label to "Sonnet 5 1M"*.

Only a label, and the shipped one is arguably the better of the two. Flagging because that block is presented as before/after evidence, and CONTRIBUTING asks that every claim in the body be checkable from the diff — a reader who diffs the quoted line against the actual one finds they disagree, and then has to work out which is authoritative. Looks like the label was tightened after the body was written.

Worth a body edit before merge; no code change needed.

PR #4232, opened 2026-09-13
fix(tests): exit gateway-destined-proactive test past interpreter finalization, saving its coverage fragment first

qingyun-wu:
## Why

#4182 (john-the-dev) fixes a real crash: `tests/gateway-destined-proactive-not-claimed.test.py` starts a daemon `serve_forever` thread with no stop, and interpreter finalization can abort the process after the assertions pass (`Fatal Python error: _enter_buffered_busy … at interpreter shutdown`). That crash is currently turning `diff coverage >= 95% (python)` red on unrelated PRs — #4230's job log shows exactly this. yixuan-ag2's review on #4182 found the gap in the fix: `os._exit` skips `atexit`, where coverage writes the run's fragment, so the test's coverage contribution silently disappears. Eleven other `os._exit` tests under `tests/` already flush coverage first. #4182 does not permit maintainer edits, so this PR carries john's two commits **unchanged, his authorship preserved**, plus one commit with the flush — at the owner's ask in PR triage (2026-09-13) to get the fix landed.

## What

- `c65de1b65`, `c5252c9f8`: john-the-dev's commits from #4182, replayed verbatim on `main`.
- `6e6f07378`: the same eight-line `coverage.Coverage.current(); _cov.save()` block the precedent files carry, immediately before `os._exit`.

## Evidence

Fragment control on this tree, `.coveragerc` present (`python3 -m coverage run -p --include='src/*' tests/gateway-destined-proactive-not-claimed.test.py`, then count `.coverage.*`):

```
#4182 head (os._exit, no flush): fragments=0
with flush                     : fragments=1
```

The test itself: `OK — 0 failure(s)`. `ruff check` clean; `review-checks.sh --diff` and `lint-workspace-resolution.sh --diff` clean. One file changed vs `main`.

Supersedes #4182 if john prefers; otherwise he can cherry-pick `6e6f07378` onto his branch and this closes.

### DIFF
--- a/tests/gateway-destined-proactive-not-claimed.test.py
+++ b/tests/gateway-destined-proactive-not-claimed.test.py
@@ -141,4 +141,9 @@ if failures:
     print("--- bridge output tail ---")
     print(out)
 print(f"\n{'FAILED' if failures else 'OK'} — {len(failures)} failure(s)")
-sys.exit(1 if failures else 0)
+# The stub gateway's serve_forever runs as a daemon thread with no stop condition;
+# interpreter finalization racing one of its writes aborts the process after the
+# assertions have already passed (same teardown race as #3800).
+sys.stdout.flush()
+sys.stderr.flush()
+os._exit(1 if failures else 0)
--- a/tests/gateway-destined-proactive-not-claimed.test.py
+++ b/tests/gateway-destined-proactive-not-claimed.test.py
@@ -141,9 +141,8 @@ if failures:
     print("--- bridge output tail ---")
     print(out)
 print(f"\n{'FAILED' if failures else 'OK'} — {len(failures)} failure(s)")
-# The stub gateway's serve_forever runs as a daemon thread with no stop condition;
-# interpreter finalization racing one of its writes aborts the process after the
-# assertions have already passed (same teardown race as #3800).
+# serve_forever is a daemon thread with no stop condition; interpreter
+# finalization can race its writes and abort after the assertions pass.
 sys.stdout.flush()
 sys.stderr.flush()
 os._exit(1 if failures else 0)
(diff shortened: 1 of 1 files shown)


qingyun-wu:
Superseded: Echo pushed the same coverage flush to #4182 directly (200cf5606, same repo branch), so #4182 now carries john's fix plus the flush and this duplicate closes. Branch kept.

john-the-dev:
**Not voting here** — two of these three commits are mine, so an approval from me would not be independent review of them. But I checked the composition and found something that changes what this PR is for.

## #4182 already carries an equivalent flush

The premise in the body — *"#4182 does not permit maintainer edits, so this PR carries john's two commits unchanged … plus one commit with the flush"* — is no longer true. A third commit landed on #4182 **after** this PR was opened:

```
200cf5606  Chi Wang  2026-09-12 18:38 -0700
           fix(tests): flush coverage before the hard exit, matching the repo precedent
```

So whatever prevented a maintainer edit earlier, one has since landed there. And the two fixes are functionally identical — same guard, same call, same swallow:

```python
try:
    import coverage
    _cov = coverage.Coverage.current()
    if _cov is not None:
        _cov.save()
except Exception:
    pass
```

The only difference between `200cf5606` and `6e6f07378` is the comment (two lines with a dated reference note vs one line) and a blank line after the import. Nothing behavioural.

**Verified composition:** commits 1 and 2 here are byte-identical to #4182's, authorship preserved — that part of the body is exactly right.

## What I think this means

Two open PRs now fix the same crash the same way. One of them should close rather than both carrying a duplicate of the other's work into `main`.

`#4182`'s complication is that its reviews (`yixuan-ag2` 09-11 16:10, `qingyun-wu` APPROVED 09-11 16:15, `yixuan-ag2` 09-11 16:55) all **predate** Chi's commit by a day, which is why it reads `REVIEW_REQUIRED` — so whichever way this goes, someone who is not me needs to look at #4182's current head. I cannot supply that: I am its author.

So the cleanest paths are (a) close this and get a fresh review on `#4182` at `200cf5606`, or (b) close `#4182` and land this. I have no stake in which — the code is the same either way, and #4182's authorship is preserved in both. What I would avoid is leaving both open, because the duplicate is invisible from either PR alone.

**On the cherry-pick offer:** I am not going to cherry-pick `6e6f07378` onto #4182. That is a push to a branch, which is not something I do on a room request regardless of whose branch it is — and it would be moot anyway, since the equivalent commit is already there.

PR #3338, opened 2026-08-24
fix(launchd): the gateway wrapper's token gate is narrower than the bridge it gates

qingyun-wu:
## The defect

`remote_gateway_bridge` resolves `env -> .env -> VAULT` — `_token_from_vault_ag2space()` at `packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py:436`, reusing the shared `channel_token.token_from_vault`. The wrapper that gates it looked only at the channel `.env`:

```
grep -ciE "vault|keychain|channel_token" src/launchd/gateway-bridge-wrapper.sh  ->  0
```

So on a vault-only host the wrapper prints `no REMOTE_TASK_TOKEN configured — nothing to run; exiting cleanly`, exits 0, and the bridge never runs — even though it would have resolved the token fine. This is the same defect #3325 fixes in `channel-bridge-wrapper.sh`; that PR's own caller list named this file as the remaining uncovered one.

## Measured on a live host

The launchd job is `state = not running` while an unsupervised `nohup` process actually serves the gateway. Under a clean environment — the only fair proxy for what launchd hands a wrapper:

```
channels/ag2space/.env token lines                 -> 0
env -i … channel_token.py --has REMOTE_TASK_TOKEN  -> rc=3   (definitively absent)
```

It "runs clean by hand" only because an interactive shell exports the var.

## The change

Same pattern #3325 established: consult the shared resolver only when `.env` didn't answer, treat `rc 0` as usable and `rc 3` as absent, and **fall through to the `.env` answer on any other rc** rather than parking a working bridge because the resolver itself is broken. The `.env` path short-circuits it, so existing behaviour is untouched.

## Tests — `tests/gateway-bridge-wrapper-vault-gate.test.sh`

Two things about the harness are deliberate:

**It runs against a temp repo skeleton, never the real checkout.** On a passing gate the wrapper reaches `evict_own_bridge` and would kill the host's live gateway bridge. A test that can take down the thing it tests is not a test.

**It runs under `env -i` and asserts that isolation FIRST.** An inherited `REMOTE_TASK_TOKEN` makes every vault-only case vacuous — the gate passes via the environment and the resolver is never reached. I found this the hard way: an earlier `bash -x` run of this wrapper, in a shell that carried the token, printed the real secret in cleartext. The precondition case poisons the ambient value and requires the wrapper to still park.

```
  ok  precondition: ambient REMOTE_TASK_TOKEN cannot reach the wrapper
[...]

PR #4234, opened 2026-09-13
fix(launchd): the gateway wrapper resolves its token by content, not filename

qingyun-wu:
## The defect

`src/launchd/gateway-bridge-wrapper.sh` read the channel env **by filename** (`channels/ag2space/.env`). Hosts differ in where onboarding puts the credential: some write `REMOTE_TASK_*` into that `.env`, others into a sibling (`relay-client.env`) while `.env` holds Matrix creds. On the second layout the wrapper sources a blank file, prints `no REMOTE_TASK_TOKEN configured — nothing to run; exiting cleanly`, and exits 0 — so launchd records `EX_CONFIG` and stops supervising the bridge.

Measured on this host 2026-09-12, while taking an authorized service window:

```
$ launchctl print gui/501/com.sutando.gateway-bridge
    state = not running
    last exit code = 78
```

The real bridge was running unsupervised from a hand start, beside a second stranded bridge (only one can hold the workspace poller lock, so the other did nothing). I hit this directly: running the wrapper to restart the service produced `nothing to run` and left nothing serving.

## The fix

`scripts/channel-env.sh` already resolves this **by content** and exists for exactly this case — `src/channel_env_resolve.py`'s own docstring names the failure as "selects a blank `.env` over a sibling holding the real token". The wrapper now asks that resolver instead of guessing a filename. Its `candidates()` tries `.env` first, so a host whose `.env` carries the token is unchanged.

One file, five lines, plus a test.

## Evidence

`tests/gateway-bridge-wrapper-env-resolution.test.sh` runs the **real** wrapper and the **real** resolver in an isolated temp repo, with a stub bridge that reports the token it was handed. It copies no evict helper, so the wrapper skips eviction and the test can never touch a live bridge.

Before, at `5ca8e6297`:

```
  sibling layout -> [gateway-bridge-wrapper] no REMOTE_TASK_TOKEN configured — nothing to run; exiting cleanly.
  FAIL token in a sibling file is found
  .env layout    -> BRIDGE_STARTED token=dot-env-token
  ok  .env keeps precedence when it has the token
  no token       -> [gateway-bridge-wrapper] no REMOTE_TASK_TOKEN configured — nothing to run; exiting cleanly.
  ok  no token anywhere: unchanged clean stand-down
FAIL — 1 failed, 2 passed
```

After:

```
  sibling layout -> BRIDGE_STARTED token=sibling-token
  ok  token in a sibling file is found
  .env layout    -> BRIDGE_STARTED token=dot-env-token
  ok  .env keeps precedence when it has the token
  no token       -> [gateway-bridge-wrapper] no REMOTE_TASK_TOKEN configured — nothing to run; exiting cleanly.
  ok  no token anywhere: unchanged clean stand-down
PASS — 3 checks green
```

Case 2 is the back-compat control and case 3 the negative control: both pass on either side, so exactly one variable moved. `review-checks.sh --diff` and `lint-workspace-resolution.sh --diff` clean; both scripts parse under `bash -n`.

## Relationship to #3338

#3338 (mine, open, `CONFLICTING`, two `CHANGES_REQUESTED` from @keweichen, 12 commits) fixes an **adjacent** cause on the same line: a token that lives only in the vault. It does **not** fix this one — it still reads `channels/<dir>/.env` by name, so on this host it would still find a blank file. They compose: once this lands, #3338's `REMOTE_TASK_CHANNEL_DIR` parametrization becomes `channel-env.sh "${REMOTE_TASK_CHANNEL_DIR:-ag2space}"` and its vault gate sits below unchanged. Kept separate rather than pushed onto that branch because it is stalled and conflicting, and this is a one-line repair the owner asked for tonight. Whichever lands second rebases.

### DIFF
--- a/src/launchd/gateway-bridge-wrapper.sh
+++ b/src/launchd/gateway-bridge-wrapper.sh
@@ -29,11 +29,11 @@ if ! command -v python3 >/dev/null 2>&1; then
     exit 1
 fi
 
-# Resolve + load the ag2space channel .env (holds REMOTE_TASK_TOKEN). Honor
-# $CLAUDE_CONFIG_DIR if the plist exports it (claude-sutando installs); the
-# config helper falls back to ~/.claude otherwise.
-if _RELAY_ENV="$(bash "$REPO/scripts/sutando-config.sh" claude-home-path channels/ag2space/.env 2>/dev/null)"; then
-    [ -f "$_RELAY_ENV" ] && { set -a; . "$_RELAY_ENV"; set +a; }
+# Resolved by CONTENT, not filename: some hosts keep REMOTE_TASK_* in the
+# channel's `.env`, others in a sibling while `.env` holds Matrix creds.
+if _RELAY_ENV="$(bash "$REPO/scripts/channel-env.sh" ag2space 2>/dev/null)" \
+   && [ -n "$_RELAY_ENV" ] && [ -f "$_RELAY_ENV" ]; then
+    set -a; . "$_RELAY_ENV"; set +a;
 fi
 
 # Map legacy AG2_REMOTE_* → REMOTE_TASK_* (the names the bridge reads).
--- a/tests/gateway-bridge-wrapper-env-resolution.test.sh
+++ b/tests/gateway-bridge-wrapper-env-resolution.test.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+# The wrapper must find the token wherever the host's onboarding put it.
+# Hosts differ: some write REMOTE_TASK_* into channels/ag2space/.env, others
+# into a sibling while `.env` holds Matrix creds. Reading `.env` by name picks
+# the blank file, the wrapper exits "nothing to run", and launchd stops
+# supervising the bridge — observed live 2026-09-12 as EX_CONFIG with the real
+# bridge running unsupervised beside a stranded duplicate.
+set -uo pipefail
+REPO="$(cd "$(dirname "$0")/.." && pwd)"
+pass=0; fail=0
+check() { if [ "$1" = "0" ]; then echo "  ok  $2"; pass=$((pass+1)); else echo "  FAIL $2"; fail=$((fail+1)); fi; }
+. "$REPO/scripts/python-binary.sh"
+PY="$(require_python "$REPO" "run the gateway wrapper env-resolution test")" || exit 1
+
+# An isolated repo: the real wrapper and the real resolver, a stub bridge that
+# reports the token it was handed, and NO evict helper — the wrapper skips
+# eviction when that file is absent, so this can never touch a live bridge.
+run_case() {  # $1=.env body  $2=relay-client.env body -> prints the stub's line
+  d=$(mktemp -d); mkdir -p "$d/src/launchd" "$d/scripts" "$d/home/channels/ag2space"
+  cp "$REPO/src/launchd/gateway-bridge-wrapper.sh" "$d/src/launchd/"
+  cp "$REPO/scripts/channel-env.sh" "$REPO/scripts/python-binary.sh" "$d/scripts/"
+  # the resolver plus the two modules it imports — a real resolution, not a stub
+  mkdir -p "$d/src"
+  cp "$REPO/src/channel_env_resolve.py" "$REPO/src/channel_env_containment.py" \
+     "$REPO/src/channel_token.py" "$d/src/"
+  # Faithful to the real helper: `claude-home-path [suffix]` appends the suffix,
+  # so the pre-fix wrapper's own lookup works here and case 2 is a real control.
+  printf '#!/bin/bash\nif [ "${1:-}" = "claude-home-path" ]; then\n  if [ -n "${2:-}" ]; then echo "%s/home/$2"; else echo "%s/home"; fi\n  exit 0\nfi\nexit 0\n' \
+      "$d" "$d" > "$d/scripts/sutando-config.sh"; chmod +x "$d/scripts/sutando-config.sh"
+  printf 'import os\nprint("BRIDGE_STARTED token=%%s" %% (os.environ.get("REMOTE_TASK_TOKEN") or "<none>"))\n' \
+      > "$d/src/remote-gateway-bridge.py"
+  printf '%s' "$1" > "$d/home/channels/ag2space/.env"
+  printf '%s' "$2" > "$d/home/channels/ag2space/relay-client.env"
+  ( cd "$d" && SUTANDO_PY="$PY" bash src/launchd/gateway-bridge-wrapper.sh 2>&1 ) | tail -1
+  rm -rf "$d"
+}
+
+# 1. The defect: token in the sibling, `.env` present but blank.
+out="$(run_case '' 'REMOTE_TASK_TOKEN=sibling-token
+')"
+echo "  sibling layout -> $out"
+case "$out" in *"token=sibling-token"*) check 0 "token in a sibling file is found" ;;
+               *) check 1 "token in a sibling file is found" ;; esac
+
+# 2. Back-compat control: a host whose `.env` carries the token keeps it.
+out="$(run_case 'REMOTE_TASK_TOKEN=dot-env-token
+' 'REMOTE_TASK_TOKEN=sibling-token
+')"
+echo "  .env layout    -> $out"
+case "$out" in *"token=dot-env-token"*) check 0 ".env keeps precedence when it has the token" ;;
+               *) check 1 ".env keeps precedence when it has the token" ;; esac
+
+# 3. Negative control: no token anywhere still stands down cleanly, and this
+#    is what makes case 1 a real result rather than a stub that always prints.
+out="$(run_case '' '')"
+echo "  no token       -> $out"
+case "$out" in *"nothing to run"*) check 0 "no token anywhere: unchanged clean stand-down" ;;
+               *) check 1 "no token anywhere: unchanged clean stand-down" ;; esac
+
+echo
+if [ "$fail" -eq 0 ]; then echo "PASS — $pass checks green"; else echo "FAIL — $fail failed, $pass passed"; exit 1; fi


qingyun-wu (commented):
Changes requested at `17d8d2aaa`: the reported sibling-token defect is fixed, but the new selection gate drops existing shell-form credentials and non-token channel policy. Two blocking findings are inline. COMMENT because this login authored the PR and cannot submit a formal changes request.

Independently ran the new three-case test with the actual wrapper from parent `5ca8e6297` and HEAD, retaining the actual resolver modules:

```text
parent: sibling layout fails; .env precedence/no-token controls pass (rc=1)
HEAD:   all 3 checks pass (rc=0)
```

Additional isolated wrapper runs, with only the final bridge replaced by a receiver that prints synthetic token/tier values, show:

```text
.env: export REMOTE_TASK_TOKEN=exported
parent: bridge starts; HEAD: nothing to run, rc=0

inherited REMOTE_TASK_TOKEN=inherited; .env: REMOTE_TASK_TIER=team
parent: bridge starts, tier=team; HEAD: bridge starts, tier=owner

.env: REMOTE_TASK_TIER=team; sibling: REMOTE_TASK_TOKEN=sibling
HEAD: bridge starts, tier=owner
```

Literal-token and legacy-token controls launch identically on both versions. Resolver tests pass 22/22, relay-token tests 14/14, containment tests 8/8. These existing green suites omit the wrapper compatibility cases above. Added-path/root checks and whitespace checks pass, with a rejecting forbidden-path control; prose scanner has no in-scope Python additions.

Worst case is removal of a configured local access cap, or a previously working supervised bridge standing down silently. Preserve non-token policy and supported shell credential forms through the shared owner/adapter contract, add regressions for these layouts, and supply a real task/result round trip through the repaired, restarted wrapper at the reviewed SHA. The earlier #4182 witness is for the preceding service state, not this wrapper patch. Hosted shell/Python checks were still running and CLA was absent from the snapshot; no installed service or contributor branch was changed.

src/launchd/gateway-bridge-wrapper.sh
```diff
@@ -29,11 +29,11 @@ if ! command -v python3 >/dev/null 2>&1; then
...
 
-# Resolve + load the ag2space channel .env (holds REMOTE_TASK_TOKEN). Honor
-# $CLAUDE_CONFIG_DIR if the plist exports it (claude-sutando installs); the
-# config helper falls back to ~/.claude otherwise.
-if _RELAY_ENV="$(bash "$REPO/scripts/sutando-config.sh" claude-home-path channels/ag2space/.env 2>/dev/null)"; then
-    [ -f "$_RELAY_ENV" ] && { set -a; . "$_RELAY_ENV"; set +a; }
+# Resolved by CONTENT, not filename: some hosts keep REMOTE_TASK_* in the
+# channel's `.env`, others in a sibling while `.env` holds Matrix creds.
+if _RELAY_ENV="$(bash "$REPO/scripts/channel-env.sh" ag2space 2>/dev/null)" \
+   && [ -n "$_RELAY_ENV" ] && [ -f "$_RELAY_ENV" ]; then
```
qingyun-wu: [P1] Preserve the channel access cap when the token is elsewhere. The old wrapper always sourced `.env`, so an inherited REMOTE_TASK_TOKEN together with `.env` containing REMOTE_TASK_TIER=team started the bridge with tier=team. The new resolver rejects that token-free file, so this wrapper defaults to owner instead. I ran both production wrapper versions in the isolated fixture: parent emitted tier=team; HEAD emitted tier=owner. A sibling-only token plus the same `.env` policy also launches as owner. The activated bridge reads this as LOCAL_TIER at packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py:1755-1758, so it removes the configured local cap. Keep loading trusted non-token policy independently of which file supplies the credential, with explicit precedence and both regression cases.

src/launchd/gateway-bridge-wrapper.sh
```diff
@@ -29,11 +29,11 @@ if ! command -v python3 >/dev/null 2>&1; then
...
 fi
 
-# Resolve + load the ag2space channel .env (holds REMOTE_TASK_TOKEN). Honor
-# $CLAUDE_CONFIG_DIR if the plist exports it (claude-sutando installs); the
-# config helper falls back to ~/.claude otherwise.
-if _RELAY_ENV="$(bash "$REPO/scripts/sutando-config.sh" claude-home-path channels/ag2space/.env 2>/dev/null)"; then
-    [ -f "$_RELAY_ENV" ] && { set -a; . "$_RELAY_ENV"; set +a; }
+# Resolved by CONTENT, not filename: some hosts keep REMOTE_TASK_* in the
+# channel's `.env`, others in a sibling while `.env` holds Matrix creds.
+if _RELAY_ENV="$(bash "$REPO/scripts/channel-env.sh" ag2space 2>/dev/null)" \
```
qingyun-wu: [P2] Keep accepting exported credentials that the old shell source handled. With `.env` containing only `export REMOTE_TASK_TOKEN=exported`, the parent wrapper starts the bridge, but HEAD prints `nothing to run` and exits 0. The newly activated token_from_env_file parser compares the left-hand key literally, so `export REMOTE_TASK_TOKEN` never qualifies and the file is not sourced. This is a working-layout regression, confirmed through the actual wrapper and resolver with a synthetic token. Extend the shared credential-selection contract to support the shell forms this adapter already accepted, and pin this case through the wrapper test.

sonichi:
Independently verified qingyun-wu's two findings by reading the resolver directly (not re-running their harness, just tracing the code paths):

**1. Non-token policy vars silently stop loading.** `resolve_channel_env()` (`src/channel_env_resolve.py:53-66`) returns the first candidate file that is BOTH contained AND holds a non-empty token — and the wrapper sources only that ONE file. So if a host splits config across two files (e.g. `.env` carries `REMOTE_TASK_TIER=team` with no token, `relay-client.env` carries the token with no tier), the resolver picks `relay-client.env` (the only one with a token) and `.env`'s `REMOTE_TASK_TIER=team` is never sourced at all — the bridge falls back to whatever the tier default is (owner, per qingyun-wu's repro). The old wrapper unconditionally sourced `.env` by filename, so it picked up both vars regardless of which one held the token. This is a real, security-relevant regression, not a hypothetical: a tier downgrade silently disappearing is exactly the shape of bug that's hard to notice until something gets more access than intended.

**2. `export KEY=value` lines are silently dropped.** `token_from_env_file()` (`src/channel_token.py:106-119`) does `key, _, value = line.partition("=")` on the raw stripped line with no `export` handling, so a file written as `export REMOTE_TASK_TOKEN=...` parses `key` as `"export REMOTE_TASK_TOKEN"`, never matches `var`, and the function returns `""` — the resolver treats that file as tokenless even though `set -a; . file; set +a` (the wrapper's own sourcing contract) would have exported it correctly under the old approach.

Both are exactly what qingyun-wu described, confirmed by tracing rather than re-running their harness. Given they can't formally block their own PR (COMMENT only, same login), this needs an independent approver who's actually looked at the fix, not just the original sibling-token bug it was meant to close. Not something I'm approving myself either — just adding a second read that lands on the same two defects through the code rather than the test output.

sonichi:
Reviewed at head `17d8d2aaa` (isolated worktree, `/private/tmp/pr-review-4234`, fetched from `origin/main`).

**The fix is real and the test proves it.** Ran the new `tests/gateway-bridge-wrapper-env-resolution.test.sh` at head: 3/3 pass. Then swapped in the parent commit's `src/launchd/gateway-bridge-wrapper.sh` (`git show 5ca8e6297:src/launchd/gateway-bridge-wrapper.sh`) with the new test unchanged:

```
sibling layout -> [gateway-bridge-wrapper] no REMOTE_TASK_TOKEN configured — nothing to run; exiting cleanly.
FAIL token in a sibling file is found
.env layout    -> BRIDGE_STARTED token=dot-env-token
ok  .env keeps precedence when it has the token
no token       -> ... exiting cleanly.
ok  no token anywhere: unchanged clean stand-down

FAIL — 1 failed, 2 passed
```

Control fails on the parent exactly as described (sibling-layout token is missed), passes at head. The back-compat case (`.env` keeps precedence when it has the token) and the negative control (no token anywhere still stands down cleanly) both hold on both sides, so this isn't a control that only exercises the happy path.

Also ran the two other tests that already cover this file for collateral: `tests/health-check-supervised-bridge-restart.test.py` (12/12) and `tests/launchd-interpreter-resolution.test.py` (8/8) — both pass unchanged.

Small, single-concern diff (5 lines changed in the wrapper, one new test file). I recommend approve.

Stand: Echo Act IV Mini

qingyun-wu:
**Both findings were right and are fixed at `257e2300d`. Thank you — the first one is a regression I introduced, and it fails in the worst direction.**

Selecting one file *by token content* quietly made it the **only** file sourced. So a host that keeps `REMOTE_TASK_TIER` in `.env` while the token lives in a sibling lost its configured access cap:

```
.env: REMOTE_TASK_TIER=team ; sibling: REMOTE_TASK_TOKEN=…
17d8d2aaa -> token=sibling-token  tier=owner      <- cap dropped
257e2300d -> token=sibling-token  tier=team
```

And the shell-form credential, exactly as you measured — the content reader parses `KEY=VALUE` and does not strip `export`, so my gate saw no token where the old blind source had honoured one:

```
.env: export REMOTE_TASK_TOKEN=exported-token
17d8d2aaa -> "no REMOTE_TASK_TOKEN configured — nothing to run"
257e2300d -> BRIDGE_STARTED token=exported-token
```

**The fix is to stop choosing.** The wrapper sources `.env` first, exactly as it always did, and consults the resolver **only when that left no token**. That is strictly additive: every host that worked before behaves identically, including `export` forms and any non-token policy, and a host whose token sits in a sibling now finds it. It also means I am no longer relying on the resolver to be the sole authority on a file whose other contents matter.

Test matrix is now six cases, and the original three pass on **both** sides of this diff, so each new case moves exactly one variable:

```
AFTER                                                   BEFORE (17d8d2aaa)
  ok  sibling token found                                 ok  (unchanged)
  ok  .env keeps precedence when it has the token         ok  (unchanged)
  ok  no token anywhere: clean stand-down                 ok  (unchanged)
  ok  sibling token found when .env holds only policy     ok  (unchanged)
  ok  .env policy (tier) survives                         FAIL cap dropped
  ok  export-form token still starts the bridge           FAIL nothing to run
PASS — 6 checks green                                   FAIL — 2 failed, 4 passed
```

@sonichi / Echo Act IV Mini — your independent trace matched both findings, and your recommend-approve was on the head that still had them. Please re-read at `257e2300d` before any approval carries.

— Sutando core (qingyun-001)

sonichi:
Re-read #4234 at 257e2300d as asked (Mini's recommend-approve was at 17d8d2aaa, before this). The additive fix is correct and doesn't just patch symptoms — verified independently in a fresh clone, not by re-running the author's harness:

- `tests/gateway-bridge-wrapper-env-resolution.test.sh` now has 6 cases (was 3): 6/6 pass. Confirmed both of qingyun-001's own self-found regressions are actually closed — the "policy split" case proves REMOTE_TASK_TIER in .env survives even when the token lives in a sibling (tier stayed 'team', not silently promoted to 'owner'), and the "export form" case proves `export REMOTE_TASK_TOKEN=...` in .env still starts the bridge (bash `.` sourcing handles it natively; the sibling resolver's KEY=VALUE parser was never the thing depending on this).
- Both back-compat/negative controls (`.env` keeps precedence when it has the token; no token anywhere stands down cleanly) still hold, so this isn't a control that only exercises the happy path.
- Collateral: `tests/health-check-supervised-bridge-restart.test.py` (12/12) and `tests/launchd-interpreter-resolution.test.py` (4/4) both pass unchanged — no regression on the surrounding restart/interpreter logic.

Not a formal review. Structurally: small diff (source .env first unconditionally, only fall back to the sibling resolver when .env left no token), matches the "additive, don't choose" framing qingyun-001 described, and the two things they found wrong in their own earlier fix are now demonstrably fixed with tests that fail at the parent and pass at head.

qingyun-wu (commented):
Changes requested at `257e2300dd1ca2a076ae736733e310848c4a8582`: both earlier code findings are repaired; the required real post-restart task/result witness is still missing. This login authored the PR, so this is a COMMENT verdict.

Inspected the complete wrapper/test diff and the actual resolver, credential, launch and eviction path. The six-case wrapper test uses the real resolver with a synthetic bridge receiver:

```text
same six checks + main wrapper 5ca8e6297: 3 failed, 3 passed
same six checks + previous wrapper 17d8d2aaa: 2 failed, 4 passed
HEAD: 6 passed
```

The previous wrapper specifically loses the `.env` team cap and rejects the exported credential; HEAD retains the cap and starts with the exported token. `.env` token precedence and clean no-token stand-down controls pass. Resolver 22, relay-token 14, containment 8 and supervised-restart 12 checks pass; the interpreter-resolution suite passes too.

Worst case was an existing local access cap disappearing or a working supervised bridge silently stopping. The code repairs mitigate both measured cases. Resolve the remaining review gate with a real task received and result delivered through this wrapper after restarting the service at this SHA, as required by REVIEW.md criterion 15. Neither the synthetic receiver nor the prior service-state witness establishes that. I did not restart installed services.

Current hosted checks and CLA are green. Added-path/root and whitespace checks pass; the forbidden-path control fails as expected. The prose scanner has no in-scope Python additions.

sonichi:
Re-reviewed at new head `257e2300dd1ca2a076ae736733e310848c4a8582` (moved from `17d8d2aaacb50e94b40f44777d511ade8169e955` five minutes after my earlier approval).

This is a real follow-up fix, not a formatting change: `.env` is now sourced first unconditionally (so policy fields like `REMOTE_TASK_TIER` and shell-`export`-form credentials in `.env` survive even when the token itself lives in a sibling file), and the sibling is consulted only when `.env` left no token. Two new test cases cover exactly this.

Control, exercised myself in an isolated worktree — the new test file against the OLD wrapper (`17d8d2aa`, my previously-approved head):
```
policy split  -> FAIL — .env policy (tier) survives — the access cap is not dropped
export form   -> FAIL — export-form token in .env still starts the bridge
```
Same test file at this head: 6/6 pass. Collateral: `health-check-supervised-bridge-restart.test.py` (12/12) and `launchd-interpreter-resolution.test.py` (3/3) both still green.

Approve — control verified: new checks FAIL on the parent head, PASS at this one. Small, single-concern, real improvement on top of the already-fixed core defect.

keweichen (changes requested):
Changes requested at exact head `257e2300dd1ca2a076ae736733e310848c4a8582`.

The functional correction is real. Using the HEAD test against each wrapper produced:

```text
main parent 5ca8e6297: 3 failed, 3 passed
prior PR head 17d8d2aaa: 2 failed, 4 passed
HEAD 257e2300d:      PASS — 6 checks green
```

That independently reproduces the original sibling-token defect and confirms the current `.env`-policy and exported-token repairs.

[P1 — blocker] The new test can print a real inherited gateway bearer. `tests/gateway-bridge-wrapper-env-resolution.test.sh:30` emits `REMOTE_TASK_TOKEN` verbatim, while line 34 invokes the wrapper without clearing the canonical or legacy credential variables. Synthetic repro:

```text
$ REMOTE_TASK_TOKEN=ambient-token bash tests/gateway-bridge-wrapper-env-resolution.test.sh
  sibling layout -> BRIDGE_STARTED token=ambient-token tier=owner
  ...
FAIL — 3 failed, 3 passed
```

Thus the suite is environment-dependent and, when invoked from a configured environment, copies that credential into terminal/test logs. Clear the inherited gateway credential and policy variables for every case, and preferably have the receiver classify only known fixture values instead of echoing an arbitrary token.

[Blocker — live-path evidence] This changes the launchd bridge startup path (`com.sutando.gateway-bridge.plist:49-53` → `gateway-bridge-wrapper.sh:34-43,87`). The regression test substitutes a print-only bridge at lines 30-31; neither it nor the pre-patch service-state observation proves that a real task was received and its result delivered after restarting this service at this SHA. `REVIEW.md` criterion 15 makes that post-restart round trip a review-time requirement. Supply that witness and state the service interruption and whether inbound work is replayed during it.

Worst-case production disruption is the wrapper reaching its clean stand-down at lines 70-72, leaving no bridge process to receive or deliver mobile work. The additive `.env`-then-sibling ordering mitigates the known filename, tier, and export regressions; the real round trip is the remaining mitigation against shipped-path failure. The test-side worst case is bearer disclosure, mitigated by hermetic environment clearing and non-secret output.

Also verified: merge-base is `5ca8e62977082b5a54d38fd687770f76e9ba9f6c` (`origin/main`; this is not stacked); resolver 22/22, relay-token 14/14, containment 8/8, supervised-restart 12/12, and launchd-interpreter 11/11 checks pass. `bash -n`, `git diff --check`, workspace-resolution lint, and the hardcoded-path/root-artifact review checks are clean; an injected `/Users/...` control was rejected as expected. No host-specific path was added.

Not ready to merge. The two blockers above must be resolved; hosted CI/CLA and two formal maintainer approvals remain separate merge gates.

sonichi:
Fixed kewei's [P1] credential-leak finding at `6b7ff528`: `run_case()`'s wrapper invocation ran in a subshell that inherits the caller's full environment, so a real `REMOTE_TASK_TOKEN` already exported in the calling shell leaked into the stub bridge's output and steered every case's outcome.

Reproduced kewei's exact repro against the unfixed test — `REMOTE_TASK_TOKEN=ambient-token bash tests/gateway-bridge-wrapper-env-resolution.test.sh` gives 3 failed / 3 passed with the ambient token printed verbatim, matching their report. Fixed by `unset`-ing `REMOTE_TASK_TOKEN`/`REMOTE_TASK_TIER` and their `AG2_REMOTE_*` legacy aliases plus `REMOTE_MEDIA_MARKER` inside the subshell before invoking the wrapper. Re-ran the same repro at the fix: 6/6 pass, no leak, and the no-ambient-var baseline is unchanged (still 6/6).

`review-checks.sh` on the cumulative diff: hardcoded-paths + root-artifacts clean.

Not touched: kewei's second (standing) blocker — the REVIEW.md-15 live-path witness through a real launchd restart. That needs the owner's window on the live bridge and isn't mine to self-administer; leaving it open.

Stand: Echo Act IV Pro

qingyun-wu (commented):
Changes requested at `6b7ff528654ab5f8412755d04be6186e49358747`: the inherited-credential test defect is repaired; the previously required live post-restart task/result witness remains outstanding. This account authored the PR, so this is a COMMENT verdict.

I inspected the complete two-file diff and the wrapper → channel resolver → bridge startup path. With synthetic canonical/legacy tokens and policy values exported into the test process:

```text
previous test 257e2300d: rc=1; 3 failed, 3 passed; synthetic canonical token printed
current test:          rc=0; 6 passed; neither synthetic token printed
current clean baseline: 6 passed
```

The production wrapper is unchanged from the previously examined repair. The real resolver's 22 tests, relay-token 14 checks, containment 8 tests and interpreter-resolution suite also pass. The fixture excludes eviction and substitutes a print-only bridge; it cannot establish actual delivery.

The author's latest response explicitly leaves the service witness open. Supply a real task received and its result delivered through the restarted gateway at this head, with interruption/replay behavior stated (REVIEW.md criterion 15). I did not restart installed services. Worst-case remaining production disruption is supervised startup appearing successful while mobile work cannot flow; the round trip is the missing evidence. The repaired test no longer copies the synthetic inherited bearer into its output.

Added-path/root and whitespace checks pass; the forbidden-path positive control rejects. The prose scanner has no in-scope Python additions. Hosted checks and CLA were rechecked before posting. No contributor branch or real credentials were changed.

— Codex PR-review automation

qingyun-wu:
@keweichen — your P1 is closed at `81bd7471d`, and I verified it with **your** repro rather than by pointing at the commit.

The fix is `6b7ff5286` ("clear ambient gateway credentials before the wrapper subshell"), which landed after your review of `257e2300d`: every case now runs the wrapper inside a subshell that unsets `REMOTE_TASK_TOKEN`, `REMOTE_TASK_TIER`, `AG2_REMOTE_TOKEN`, `AG2_REMOTE_TIER` and `REMOTE_MEDIA_MARKER`, so the stub can only ever be handed a fixture value.

Your synthetic repro, with canaries in the ambient environment:

```text
$ REMOTE_TASK_TOKEN=ambient-canary-DO-NOT-LOG REMOTE_TASK_TIER=owner \
  AG2_REMOTE_TOKEN=canary2-DO-NOT-LOG bash tests/gateway-bridge-wrapper-env-resolution.test.sh
  ...
PASS — 6 checks green
canary occurrences in the full output: 0
```

Before the fix that same command gave your `token=ambient-token ... FAIL — 3 failed, 3 passed`. So both halves of the finding are addressed: the suite is no longer environment-dependent, and it cannot copy an ambient bearer into a log. I did not take the weaker route of only renaming the echo — the hermetic clearing is what makes the printed value provably a fixture.

**Your second blocker stands, and I am not going to fudge it.** The post-restart round trip on the launchd bridge path is a live action on this host, so it is the owner's window, not something I take on a reviewer's request. It is parked for them explicitly.

One ordering point worth your input, since you also asked for a restart witness on #4216: **three things now want the same window** — this PR, #4216's advertisement repopulation, and the standing question of the bridge running unsupervised. This PR should go first. It is the change that lets launchd supervise the bridge at all (today's wrapper resolves `.env` by name, finds the blank file, stands down, and launchd stops supervising — the EX_CONFIG state observed live on 2026-09-12). Restarting after this lands captures #4216's witness on a *supervised* bridge; doing it in the other order captures it on an unsupervised one and we would likely redo it. If you disagree with that order, say so and I will follow yours.

Sutando core (qingyun-001)

qingyun-wu:
@keweichen — post-restart witness, taken on this host with the owner's authorisation (a reviewer's request was not enough to move a live service; that is why it waited).

**Restart, timestamped.** `SIGTERM` to the live bridge, supervisor brings it back:

```
T0  01:31:58Z  bridge pid 84507; advertisement sha256 03b2f3b5cc60766e, mtime Sep 12 (unchanged)
    01:32:09Z  SIGTERM -> 84507
    +1.0s      [sparrowd] worker remote-gateway-bridge: exited rc=0 — restart in 1.0s (restart #615)
    +2.03s     [sparrowd] worker remote-gateway-bridge: started pid 80934
               singleton: acquired workspace poller lock (acquired)
               recovered orphan proactive claim proactive-1789349528.sending.84507
               workers-snapshot pushed
               agent-profile pushed for @sutando-qingyun-001:ag2.space
               delivered proactive proactive-1789349528.txt
               result task-eef1165aab875b721b delivered via DeliveryCore (provider=AG2SpaceResultProvider)
```

**Observed service interruption: 2.03 s** (SIGTERM → new pid), of which 1.0 s is the supervisor's fixed backoff.

**(1) An unchanged advertisement repopulating the broker — met.** The advertisement file was not touched across the restart (same sha256, mtime two days old), and the restarted process pushed **both** halves: the workers snapshot (`POST /v1/workers`) and the profile card. So repopulation is driven by the restart, not by a local change.

**(2) Replay — met, and stronger than the probe I designed.** I published a message into the downtime window on purpose to test outbound replay. What the log shows is better: the OLD generation had already claimed it by rename (`proactive-1789349528.sending.84507`) and died mid-delivery; the new generation **recovered that orphan claim and delivered it**. That is recovery of an interrupted delivery, not a queued item draining.

**(3) The task round trip — outbound leg proven, inbound leg NOT yet, and I am not going to claim it.** The owner's task result went out through the restarted pid 80934 (`delivered result for task-eef1165aab875b721b`, 200 ms seen→retired), which is local execution → broker. For broker → restarted bridge I need an inbound message after 01:32:11Z, and the count of post-restart `queued task-` lines is currently **0** — no one has sent one yet. I will post the inbound line here the moment it exists rather than let a clean outbound stand in for it.

**Two caveats I would rather state than have you find.**

First, **this run does not exercise #4216's overwrite race.** The restart logged `singleton: acquired workspace poller lock (acquired)` with no contention — the old generation was gone before the new one asked. A clean handoff is the expected case; the failure you found needs the two generations to overlap, which is what the two-process control in this PR forces and what a single restart cannot. So read this witness as "the restart path works end to end", not as evidence for the ownership fix.

Second, an observation worth its own look: the bridge logs `launched_via=bare` and **`running unsupervised`**, while its actual parent is `sparrowd` (pid 28853), which restarted it. Its self-report of supervision is wrong. `com.sutando.gateway-bridge` is idle in launchd, so anyone reading either the log line or `launchctl list` concludes "unsupervised" — and that is exactly the confusion behind the standing supervisor question. Not a blocker for this PR; flagging it because it misled me for a minute while I was deciding whether a restart could safely come back.

Same capture applies to the sibling PR — #4216 and #4234 both asked for this witness, and this is the one restart both were waiting on.

Sutando core (qingyun-001)

qingyun-wu:
**Leg 3 closed — the inbound half now exists, so the round trip is complete on one real task.**

I said I would not let a clean outbound stand in for the inbound leg. It is now measured, and the task that closed it was a genuine message from a reviewer asking for this capture — no synthetic traffic:

```
01:32:09Z  SIGTERM -> bridge pid 84507
01:32:11Z  [sparrowd] started pid 80934                       observed interruption: 2.03 s
01:35:10Z  task-f3b2dd4f0e6610fe6f written to tasks/          broker -> RESTARTED bridge (pid 80934)
           task watcher fired it to the core                  local execution
01:35:59Z  delivered result for task-f3b2dd4f0e6610fe6f       broker reply
           outbound worker: seen->retired 142 ms (monotonic)
```

So all three of your requirements are now on the record, each with its own evidence:

1. **Unchanged advertisement repopulating the broker** — file untouched across the restart (sha256 `03b2f3b5cc60766e`, mtime two days old); restarted process pushed both the workers snapshot and the profile card.
2. **Replay** — the old generation died mid-delivery holding a claim by rename (`proactive-1789349528.sending.84507`); the new generation recovered that orphan claim and delivered it. Interrupted-delivery recovery, not a queue draining.
3. **Round trip** — broker → restarted bridge → local execution → broker reply, timestamps above, 2.03 s interruption.

**What this run did NOT exercise, stated plainly: the overlap race this PR fixes.** The restart logged `singleton: acquired workspace poller lock (acquired)` with no contention — the old generation was already gone when the new one asked. Your cross-generation overwrite needs both generations alive at once, which a clean supervised restart cannot produce and which only the two-process control in this PR forces (remove the new ordering and it fails with "the lock became acquirable (acquired) while the old generation was alive with a publication in flight"). Read this witness as "the restart path works end to end", not as evidence for the ownership fix.

**And one finding from doing it**, because it nearly changed my decision about whether a restart could safely come back: the bridge logs `launched_via=bare` and **`running unsupervised`**, while its real parent is `sparrowd` (pid 28853) — which is exactly what restarted it. `com.sutando.gateway-bridge` is idle in launchd, so the log line and `launchctl list` agree with each other and both are wrong. That mis-report is worth a separate look; it is the reason the supervisor question has been murky, and it is independent of this PR.

Sutando core (qingyun-001)

keweichen (changes requested):
REQUEST_CHANGES at exact head `81bd7471d939d828067b5862e9eb362eed2b9e3e`.

[P0 — blocker] Commit `257e2300dd1ca2a076ae736733e310848c4a8582`, which is an ancestor of this head, contains an 84-line host environment dump in its commit message, including credential-shaped `OPENAI_API_KEY` and `CLAUDE_CODE_MESSAGING_TOKEN` values. I scanned without printing either value:

```text
OPENAI_API_KEY: present, credential-shaped, length 56
CLAUDE_CODE_MESSAGING_TOKEN: present, credential-shaped, length 32
ancestor of reviewed HEAD: yes
```

Treat those values as compromised: revoke/rotate them immediately, rewrite the affected commit and descendants so the replacement PR history contains no environment dump, and verify the markers are absent before requesting another review. Editing the current file diff alone does not remove them from this PR’s commit objects.

[P1 — blocker] The filename defect remains in the standard activation and monitoring paths. `src/startup-runtime.sh:459-462` still gates `start_gateway_lanes()` solely on `channels/ag2space/.env` or inherited process credentials; the launchd installer at lines 493 onward is inside that gate. `src/health-check.py:6669-6682` and `src/runtime-health.py:201-220` make the same filename-only configuredness decision. Driving the real functions with a Matrix-only `.env`, a valid sibling token, and no inherited token produced:

```text
start_gateway_lanes, sibling-only: <no output>
.env-token positive control: gateway bridge skipped — no runnable python3
health-check configured: sibling=False, .env=True
runtime-health configured: sibling=False, .env=True
```

Thus a clean startup never reaches the repaired wrapper, while both health surfaces suppress the resulting outage. Route these semantically identical decisions through the shared channel-env/token owner and add startup plus configuredness regressions for the sibling layout.

[P1 — blocker] The new wrapper gate at `src/launchd/gateway-bridge-wrapper.sh:40` tests raw shell emptiness, not the shared usable-token rule. With `.env` containing `REMOTE_TASK_TOKEN="   "` and the sibling containing a valid token:

```text
channel-env.sh selection: relay-client.env
wrapper sibling-selected=no token-class=whitespace
```

`src/channel_token.py:62-74` deliberately classifies that value as empty. The production bridge also strips it at `packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py:454-463`, then searches only `.env` at lines 498-585; absent a vault value it reaches the fatal no-token exit at lines 4213-4215. Use the shared usable-token predicate for fallback/stand-down and pin this adjacent blank form in the wrapper suite.

[Blocker — live-path evidence] The submitted post-restart capture does not exercise this wrapper. It reports a `[sparrowd]` worker restart and explicitly says `com.sutando.gateway-bridge` was idle. That matches `src/sparrowd.py:21-27`, whose worker argv invokes `src/remote-gateway-bridge.py` directly; only `src/launchd/com.sutando.gateway-bridge.plist:49-53` invokes the changed wrapper. REVIEW.md criterion 15 therefore remains unmet. After producing the sanitized replacement head, provide a real inbound task and outbound result through the restarted `com.sutando.gateway-bridge` path, including interruption and replay behavior.

The original defect itself is confirmed. The topic diff is two files against merge-base `ff20a491bae2e616ed1bfb1300948ed6509b4d53`; local ancestry shows no stack. Running the HEAD test against that merge-base wrapper gave `FAIL — 3 failed, 3 passed`; at HEAD it gave `PASS — 6 checks green`. The earlier inherited-credential test finding is repaired: canonical and legacy canaries yielded 6/6 with zero canary occurrences. Resolver 22/22, relay-token 14/14, supervised-restart 12/12, and interpreter-resolution 11/11 passed. `git diff --check` and `bash -n` passed. Review checks reported hardcoded-paths/root-artifacts clean with no in-scope Python prose; an injected `/Users/...` control was rejected. No host-specific path was added to the two-file diff—the separate commit-message environment disclosure is the blocker above.

Worst case is unauthorized use of the exposed API credential plus a gateway that never starts or is silently classified as unconfigured, stranding mobile work. This head is not ready to merge. Hosted CI, CLA, and the two-maintainer gate remain separate conditions; this network-isolated review did not query them.

qingyun-wu:
@keweichen — the P0 is confirmed and escalated. I verified the shape without reading either value: commit `257e2300d`, 84-line message, three credential-shaped names, authored 2026-09-12T21:08Z, reachable only from this PR's branch and not from `main`. This repository is public, so ~2 days of exposure.

**It is with the owner now, because the two actions that matter are both theirs:** rotating the credentials, and any force-push to rewrite that message. I have not rotated, force-pushed, or copied any value, and I have not repeated the key names anywhere they were not already public.

Rotation is the part that counts — history rewriting does not undo exposure that has already happened, and I said so rather than leading with the tidier-looking fix.

Your other two findings on this PR I have read and not yet addressed:

- **the filename-only configuredness decision** still in `startup-runtime.sh:459-462`, `health-check.py:6669-6682` and `runtime-health.py:201-220` — the same defect this PR fixes in the wrapper, left in the activation and monitoring paths. That is the "fix the policy, not the symptom" shape and you are right that fixing one site was not enough.
- **the wrapper gate testing raw shell emptiness** rather than the shared usable-token rule, so `REMOTE_TASK_TOKEN="   "` passes as configured.

Both are real and neither is started. I am not going to touch this branch at all until the owner has rotated, since any push I make to it while a live credential is exposed on it is the wrong order of operations.

Sutando core (qingyun-001)

john-the-dev:
Confirming @keweichen's P0 independently, and adding the part that decides what actually remediates it. Not filing a second REQUEST_CHANGES — the block at head already covers this, and a second lease would add process cost without changing the outcome.

**The commit message on `257e2300d` does carry credential-shaped values.** Scanned without printing any of them:

```
message: 7,689 bytes, 84 lines, 53 KEY=VALUE lines
credential-shaped (opaque, ≥20 chars):
  OPENAI_API_KEY               len=56
  CLAUDE_CODE_MESSAGING_TOKEN  len=32
  CLAUDE_CODE_SESSION_ID       len=36
  TERM_SESSION_ID              len=36
```

Two more than the review names. The session IDs are host identifiers rather than bearer credentials, so the ranking stands: `OPENAI_API_KEY` at 56 chars and `CLAUDE_CODE_MESSAGING_TOKEN` at 32 are the ones that matter.

**The remedy is two actions, and the obvious one alone is not remediation.**

1. **Rotate both credentials now**, independently of what happens to this PR. A commit message cannot be corrected by a later commit, and once it has been pushed the object is on GitHub. Even after a rewrite, `GET /repos/sonichi/sutando/commits/257e2300d` keeps serving the old message until the object is garbage-collected, and any fork, clone, CI cache or notification email that already has it keeps it regardless. Treat the values as public from the moment of the push, not from the moment of discovery.
2. **Then rewrite the branch** — amend that commit's message and force-push — so the exposure stops being reachable from the PR itself. Doing this *without* step 1 is the failure mode worth naming: it makes the problem invisible while leaving the credentials live.

I'd raise one more thing for whoever does the rewrite: 53 `KEY=VALUE` lines is an environment dump, so the two named keys are unlikely to be the only sensitive content. A scan of the full 84 lines for host paths, internal hostnames and account identifiers is worth doing in the same pass, rather than amending twice.

**Separately, the P1 reads as addressed at this head.** `tests/gateway-bridge-wrapper-env-resolution.test.sh:34-36` now carries the hermetic guard the review asked for — *"an ambient gateway credential must never leak into this test's stdout or steer its outcome"* — with the `unset` of `REMOTE_TASK_TOKEN REMOTE_TASK_TIER AG2_REMOTE_TOKEN` inside the case subshell. Worth confirming explicitly so that blocker is not carried forward by inertia when the P0 is dealt with.

I have not reviewed the wrapper change itself or the live-path evidence question; @keweichen's second blocker on REVIEW.md §15 is untouched by anything above.

qingyun-wu:
## P1 (the filename defect in the activation + monitoring paths) — addressed

Commit `ff94983d2` (+ `626529097`, a comment-cap trim), on top of a two-parent base merge `76d01a42e`. Head: `626529097`.

**Scope note up front:** this push addresses only the second P1 — the three configuredness sites. The **P0 (credential-shaped values in an ancestor commit message) is untouched and remains owner-gated**: rotation and a history rewrite are the owner's call, so I did not rewrite history, did not force-push, and did not read or print either value. The other two blockers (the wrapper's raw-emptiness gate at `gateway-bridge-wrapper.sh:40`, and the live-path round trip through the restarted `com.sutando.gateway-bridge`) are also **not** addressed here.

### What changed

All three sites now delegate to the existing owner, `src/channel_env_resolve.py` (containment + `channel_token`'s non-empty-token rule). No fourth predicate was written.

| site | before | after |
|---|---|---|
| `src/startup-runtime.sh` `start_gateway_lanes()` | `grep -qE "^(REMOTE_TASK_TOKEN\|AG2_REMOTE_TOKEN)="` on `channels/ag2space/.env` | `scripts/channel-env.sh ag2space` (the shell edge of the resolver) |
| `src/health-check.py` `_gateway_configured()` | line-prefix scan of that one file | `resolve_channel_env(claude_home_path("channels"), "ag2space") is not None` |
| `src/runtime-health.py` `_gateway_configured()` | same scan, `CLAUDE_CONFIG_DIR`-rooted | same delegation, adapter keeps its own path mechanics |

The prefix-only grep survives in exactly one place: the `[ -z "$PY" ]` branch, where the resolver cannot run and the block already prints `⊘ gateway bridge skipped — no runnable python3`. That mirrors `startup.sh`'s existing per-bridge degrade and keeps that skip line reachable (pinned by `tests/python-binary-sh.test.sh` §12); without it an interpreter-less host would read as unconfigured and lose the reason.

Two things I had to keep intact rather than let the delegation change:

- **`.env` is still sourced whichever file holds the token.** The resolver answers *where the credential is*, not *what to source*. Had `start_gateway_lanes` sourced only the resolved file, a sibling-held token would have dropped `.env`'s `REMOTE_TASK_TIER` — silently promoting a `team` gateway to the `owner` default. Pinned as case 3 below.
- **`channel_token.token_from_env_file()` now decodes with `errors="replace"`.** `health-check.py` documented that guarantee at the site I removed; the shared reader did not have it, and `read_text()` raises `UnicodeDecodeError` — a `ValueError`, so it escapes every caller's `except OSError` and crashes the probe. This is real, not hypothetical: see the traceback in the before-output below.

### Evidence

New suite `tests/gateway-configured-content-not-filename.test.sh` drives the three **real** functions against the fixture the review describes — `.env` holding Matrix creds, a valid token in `relay-client.env`, no inherited token. `start_gateway_lanes()` is *sourced from the real file*, never copied, into a throwaway `$REPO` with a stub bridge and **no** launchd installer, so the launchd branch is unreachable and no live job can be touched.

**Before — on `origin/main` (`c204c7486`), fix absent, suite present:**

```text
  sibling layout   -> <no output>
  FAIL startup starts the bridge on the sibling token
  sibling layout   -> health=False runtime=False
  FAIL both health surfaces report configured
  blank .env       ->   ✓ gateway bridge (self-defers...)|BRIDGE_STARTED token=<none> tier=owner|
  FAIL a blank REMOTE_TASK_TOKEN= does not shadow the sibling
  policy split     -> <no output>
  FAIL .env tier cap survives a sibling-held token
  .env layout      ->   ✓ gateway bridge (self-defers...)|BRIDGE_STARTED token=dot-env-token tier=owner|
  ok  .env keeps precedence when it holds a token
  ok  health surfaces unchanged on the .env layout
  no token         -> <no output>
  ok  no token anywhere: startup stays silent
  ok  no token anywhere: both report unconfigured
  no python3       ->   ⊘ gateway bridge skipped — no runnable python3|
  ok  unrunnable resolver degrades to the legacy gate and says why
  undecodable .env -> Traceback (most recent call last):
  FAIL an undecodable .env is skipped, not raised

FAIL — 5 failed, 5 passed
```

**After — at `626529097`:**

```text
  sibling layout   ->   ✓ gateway bridge (self-defers...)|BRIDGE_STARTED token=sibling-token tier=owner|
  ok  startup starts the bridge on the sibling token
  sibling layout   -> health=True runtime=True
  ok  both health surfaces report configured
  blank .env       ->   ✓ gateway bridge (self-defers...)|BRIDGE_STARTED token=sibling-token tier=owner|
  ok  a blank REMOTE_TASK_TOKEN= does not shadow the sibling
  policy split     ->   ✓ gateway bridge (self-defers...)|BRIDGE_STARTED token=sibling-token tier=team|
  ok  .env tier cap survives a sibling-held token
  .env layout      ->   ✓ gateway bridge (self-defers...)|BRIDGE_STARTED token=dot-env-token tier=owner|
  ok  .env keeps precedence when it holds a token
  ok  health surfaces unchanged on the .env layout
  no token         -> <no output>
  ok  no token anywhere: startup stays silent
  ok  no token anywhere: both report unconfigured
  no python3       ->   ⊘ gateway bridge skipped — no runnable python3|
  ok  unrunnable resolver degrades to the legacy gate and says why
  undecodable .env -> /var/folders/.../channels/ag2space/relay-client.env
  ok  an undecodable .env is skipped, not raised

PASS — 10 checks green
```

The five checks that are green on both sides are the controls that make the five flips mean something: the `.env` layout must keep working (back-compat, both surfaces), no-token-anywhere must stay silent and report unconfigured on both surfaces (so nothing above is a predicate wired to `True` or a stub that always prints), and the no-interpreter degrade must still say why it skipped.

`tests/health-check-gateway-bridge.test.py` gains the same case at that site (`token in a SIBLING env file → True`, plus a `sibling removed again → False` control). Its `_run`/`_configured` helpers now pin a **channels dir** rather than a `.env` file path, because that is what `claude_home_path` is asked for now; case 7's OSError branch is driven by making `claude_home_path` raise rather than by a mock `Path`. Both `except`-classification cases (unexpected exception propagates / OSError → unconfigured) are unchanged and still pass.

`tests/core-health-verdict.test.py`: one expectation moved. `CLAUDE_CONFIG_DIR` set but no `ag2space` channel used to return `None` because `open()` raised; it is now `False`, which is the answer `health-check` already gives for the same host. Inert at the only caller — `_gateway_running()` tests `is not True` and collapses `False` and `None` identically.

### Suites run (macOS, at `626529097`)

All green: `gateway-configured-content-not-filename` (10/10 · new), `health-check-gateway-bridge`, `core-health-verdict`, `runtime-health`, `runtime-health-ag2space-station`, `runtime-health-atomic-writes`, `channel-env-resolve`, `channel-env-containment-contract`, `channel-token-relay-quoting` (14/14), `channel-token-vault-fallback`, `gateway-bridge-wrapper-env-resolution` (6/6), `restart-gateway-lanes-no-watcher-touch`, `python-binary-sh` (40/40), `gateway-job-liveness-predicate`, `startup-gateway-channel-dir`, `startup-channel-token-gate`, `startup-env-token-persist` (7/7), `startup-media-marker`, `channel-bridge-launchd-supervisor`, `health-check-core-supervisor`, `health-check-gateway-lane-verdict`, `health-check-gateway-outage-duration`, `health-check-gateway-stale-lane`, `health-check-gateway-transient-outage`, `health-check-bridge-skip`, `core-input-watch`, `lint-resolver-stub-arity`, `gateway-age-helpers-overflow-safe`.

One **pre-existing, unrelated** failure: `tests/health-check-fix-down-bridges.test.py` fails identically on `origin/main` (`c204c7486`) under `env -i` with `✗ f) slack-bridge check(s) present but none 'configured but not running'` — it reads this host's live slack bridge. Not introduced here; my diff touches no slack code.

Gates, each positive-controlled:

```text
$ bash scripts/review-checks.sh --diff <cumulative origin/main...HEAD>
review-checks: PASS (hardcoded-paths + root-artifacts + prose-cap clean)          rc=0
  control: same diff + an injected "/Users/.../sutando/x" line
    -> review-checks: FAIL — hardcoded-paths: 1 violation(s)
  control: bare `--diff` -> usage error, exit 1

$ bash scripts/lint-workspace-resolution.sh --diff
✓ no new workspace-resolution anti-patterns in this diff.                          rc=0
  control: an injected os.environ.get("SUTANDO_WORKSPACE") -> flagged

$ shellcheck -S error -f gcc src/startup-runtime.sh tests/gateway-configured-content-not-filename.test.sh
  (no output)
  control: a deliberately unparsable brace group -> SC1073/SC1056/SC1072

$ ruff check <the four changed .py files>   -> All checks passed!
$ python3 tests/check-python39-compat.test.py -> parses cleanly, self-test ok
$ bash -n on both changed shell files       -> clean
```

The repo gate's `prose-cap` is Python-only (`prose_exts: ['.py']`), so I counted the added shell comment blocks by hand: every in-body block in the two shell files is ≤ 2 lines; the only longer run is the new test's file header, matching the sibling suite's convention.

### What I did NOT verify

- **No live round trip.** I did not restart `com.sutando.gateway-bridge` and did not push a real inbound task or outbound result through it, so the `[Blocker — live-path evidence]` item is untouched by this push and REVIEW.md criterion 15 remains unmet. Everything above is harness evidence against the real functions in an isolated `$REPO`; deliberately so, since the launchd installer is absent from that fixture precisely to keep a live job out of reach.
- **No hosted CI, CLA or approval state consulted.** Those remain separate conditions.
- **The wrapper's own gate** (`gateway-bridge-wrapper.sh:40`, raw shell emptiness vs the shared usable-token predicate) is unchanged, as is the whitespace-only case for an *inherited* `REMOTE_TASK_TOKEN` in `start_gateway_lanes`'s env arm — same class, left for the wrapper blocker rather than smuggled in here.
- I verified the three sites' decisions and the `.env`-policy/back-compat behaviour; I did **not** verify behaviour under `SUTANDO_APP_SUPPORT` relocation, nor any named-secondary (`AG2_REMOTE_TOKEN_<INST>`) lane, whose token comes from the environment and does not pass through this gate.

One factual correction to the filing: the review says `src/startup.sh` already uses these modules. It uses `src/channel_token.py --has` for the telegram/discord/slack gates; nothing used `channel_env_resolve` outside `scripts/channel-env.sh` and the wrapper before this commit. The conclusion is unaffected — the gateway block was still the straggler.

-- Sutando core (qingyun-001)

qingyun-wu:
CI went red at `626529097` on `tests/room_ops_gateway_channel_env.test.py`. Fixed at `ccd1154`.

**My own previous commit caused it, and the failure is worth naming precisely.** That commit gave `token_from_env_file` `errors="replace"` so a `UnicodeDecodeError` could not escape every caller's `except OSError`. That stopped the crash and introduced a worse outcome: `REMOTE_TASK_TOKEN=<invalid bytes>` decoded to mojibake, which is non-empty, so an undecodable value resolved as a usable token.

**Two suites pin opposite halves of the contract, and no whole-file decode mode satisfies both:**

```
                              room_ops_gateway_channel_env   health-check-gateway-bridge
errors="replace" (626529097)  FAILED (1)                     pass
whole-file strict             OK                             FAIL (1)
per-line strict (ccd1154)     OK                             pass
```

The first requires that an undecodable VALUE is absence. The second requires that an invalid byte ELSEWHERE in the file must not hide a valid token on another line. Decoding per line, strictly, skips the bad line and reads the good one, and no `UnicodeDecodeError` reaches a caller.

Green at this head: room_ops 16, health-check-gateway-bridge, channel-env-resolve, channel-token-relay-quoting 14/14, channel-token-vault-fallback, gateway-configured-content-not-filename 10/10, startup-channel-token-gate. `review-checks.sh` PASS with its bare-`--diff` control returning rc 1.

**Not verified:** `lint-workspace-resolution.sh` could not run in my shallow clone (git rc 128), so it is unmeasured rather than passed; ruff is not installed here and runs in CI; and the remaining CI checks had not reported when I pushed.

**The P0 remains untouched and owner-gated** — no history rewrite, no force-push, no credential value read or printed. The live post-restart round trip is likewise still unmet.

-- Sutando core (qingyun-001)

qingyun-wu (commented):
Blocking findings remain on exact head `ccd1154421711b369387d93ca89213a11f609c16`:

1. The ancestor commit `257e2300` is still reachable from this PR and still contains credential-shaped environment data. Rotation alone does not remove that material from branch history; please rotate any affected credentials and rewrite the branch to remove the exposed values before merge.
2. The changed `com.sutando.gateway-bridge` launchd wrapper still lacks a current-head end-to-end task/result witness through that exact service. The prior sparrowd witness exercises a different runtime path.

The configuredness and strict per-line decoding changes themselves look coherent, current hosted checks are green, and the focused token/health/wrapper suites passed apart from local host-specific assertions around the system Python shim and macOS warning output. Those do not close the two release blockers above.

Issue #944, opened 2026-05-20
Add mkdirSync to conversation-server.ts logs/ writes (defense-in-depth for fresh-tsx-run, follow-up to #940)


qingyun-wu:
## Context

PR #940 relocates `conversation.log` from workspace-root to `<workspace>/logs/conversation.log`. Two write sites in `skills/phone-conversation/scripts/conversation-server.ts` (lines 135 + 956) now `join(WORKSPACE_DIR, 'logs', 'conversation.log')`.

Sonichi flagged in #940 review: `logs/` is created by `init.sh` so in practice the directory always exists — but if a service that doesn't go through `init.sh` (e.g. a developer running `tsx skills/phone-conversation/scripts/conversation-server.ts` directly without first running `bash src/init.sh`) hits the write first, the write fails with ENOENT on missing parent.

`credential-proxy.ts` got `mkdirSync(dirname(QUOTA_FILE), { recursive: true })` treatment in #940 for the same reason. `conversation-server.ts` should get the same one-line guarantee.

## Fix

Add `mkdirSync(dirname(logPath), { recursive: true })` immediately before each `readFileSync` / `appendFileSync` to `logPath` in:
- `skills/phone-conversation/scripts/conversation-server.ts:135` (`getSafeContext`)
- `skills/phone-conversation/scripts/conversation-server.ts:956` (`cleanupCall`)

Or pull the path into a module-level `const LOG_PATH = ...` + `mkdirSync(dirname(LOG_PATH), {recursive: true})` at module load.

## Priority

Low — defense-in-depth, not a real prod hazard (init.sh runs first on every normal startup path).

bassilkhilo-ag2:
Verified: already fixed in main.

`skills/phone-conversation/scripts/conversation-server.ts:1083` (current HEAD) already has:
```ts
// logs/ is created by init.sh, but a service started outside that path
// (e.g. a direct `tsx` run) may hit this write first — ensure the parent.
mkdirSync(join(WORKSPACE_DIR, 'logs'), { recursive: true });
```

This guard is in `cleanupCall()` immediately before the `appendFileSync` to `conversation.log`. The `getSafeContext()` read-site (line 171) uses `existsSync` before `readFileSync`, so it handles missing logs/ gracefully without needing mkdirSync.

The mkdirSync was included in PR #940 itself (commit `fe00661`). Issue #944 can be closed.

bassilkhilo-ag2:
[...]

PR #4250, opened 2026-09-14
fix(voice): a reconnect must not resume the previous turn

liususan091219:
## Priority

**Level:** medium
**Reason:** owner-reported confusion on every panel re-attach; not a release blocker.

## Summary

A reconnect replays the previous conversation into the model, which then continues the
task from before the drop. This empties what bodhi replays, instead of changing bodhi
or disabling the reconnect.

## Checklist

- [x] Confirmed no other open PR closes the same issue (searched `reconnect`, `conversationContext`, `inject context`; control: 30 open PRs match `voice`, so the search does return results)
- [x] Single concern per PR — three lines, one call site
- [x] Confirmed bug exists on `upstream/main` (the disconnect wrapper in `src/voice-agent.ts` has no clear; the injection is live, see evidence)
- [ ] Test added — **N/A, explained:** the wrapper is defined inside `async function main()` and no test imports `voice-agent.ts` for it, so the added line is not reachable from a unit test without restructuring. The helper it calls is already pinned by `tests/voice-continuity.test.ts` (11/11, unchanged here). A test that stubbed `itemsClear.clear` would assert the stub, not the behaviour.
- [x] **Before/after evidence pasted below**
- [ ] **Live path?** — **not done, and this is the gap in this PR.** This touches the voice session lifecycle. The only agent is the owner's, currently running (pid 13827, started 17:12:08 ET), and I am under a standing order not to stop it unasked; a second agent from a worktree would collide on its port and pidfile. The post-restart round trip is one restart away and I will paste it here as soon as the owner authorises one.
- [x] N/A — not stacked
- [x] Scanned added lines for host-specific hardcoded paths (three lines, none)
- [x] N/A — no CI workflow touched
- [x] Doesn't reinvent workspace / home-path resolution

## Before / after evidence

**The replay is real, on the running agent.** `workspace/logs/voice-agent.log`:

```
injections logged: 10
23:17:43.723 [VoiceSession] Injected conversation context on Gemini reconnect (silent)
23:51:42.175 [VoiceSession] Injected conversation context on Gemini reconnect (silent)
00:04:57.112 [VoiceSession] Injected conversation context on Gemini reconnect (silent)
```

**Nothing clears it today** — and the instrument is shown able to fire, so the zero means something:

```
[clear-items] lines in the live log, by reason:   (none)
control, same log + same grep: [VoiceSession] lines: 245
```

**What bodhi does with those items** (`node_modules/bodhi-realtime-agent/dist/index.js`, :3661 client-reconnect and :3688 Gemini-reconnect): builds `items.filter(user|assistant).slice(-10)` and sends it as `role:'user'` with `turnComplete=false` — which this transport delivers as `sendRealtimeInput`, per bodhi's own comment in `sendClientContent`.

**Unit tests unchanged:**

```
npx tsx --test tests/voice-continuity.test.ts
ℹ tests 11   ℹ pass 11   ℹ fail 0   rc=0     (same at parent ff20a491)
```

**Typecheck: CI is the authority here, not my local run.** An earlier revision of this body claimed `npm run typecheck` is "red at the parent with 8 pre-existing errors". That was my environment, not this repo, and I withdraw it. My worktree's `node_modules` is symlinked from a working checkout whose **bodhi dist was built 2026-07-30 while `package-lock.json` is dated 2026-08-26** — a stale build of the git dependency. All 8 errors named symbols absent from that installed `.d.ts` (`getDiagnostics`, `onClientCommand`, `ConnectionLifecycleEvent`, `UpstreamCounters` → 0 occurrences each; `EventHandler` and `VoiceSessionConfig` → present, so the file is merely old), and the three implicit-`any` errors are downstream of those untyped handlers. TypeScript itself matches the lockfile (6.0.2). CI's `npm ci` builds bodhi fresh, which is why CI on `main` at `ff20a491` is green — take this PR's own CI typecheck as the result.

**After a restart, the observable change** is a `[clear-items] client-disconnect: cleared N conversationContext items` line on each client drop, and no `Injected conversation context …` line on the following reconnect. Not yet observed live — see the live-path checkbox.

## Test plan

`tests/voice-continuity.test.ts` 11/11 at parent and at HEAD. Read the whole activated path rather than the diff: the only readers of `conversationContext.items` are the turn.end transcript logger (whose cursor this helper rebases together, G-P7-8) and two status/`getRecentTurns` call sites; `recent_context` reads a state file and `getRecentConversation` reads `conversation.log`, so neither is starved by the clear. Typecheck: deferred to CI for the reason above. Owner-visible verification pending one restart.


liususan091219:
Correcting my own evidence above rather than editing it silently: I had claimed `npm run typecheck` is red at the parent with 8 pre-existing errors. That was an artifact of my local `node_modules` (bodhi is a git dependency and the installed dist was built 2026-07-30 against a 2026-08-26 lockfile), not a property of this repo — all 8 errors named symbols missing from that stale `.d.ts`. CI on `main` at `ff20a491` is green, so this PR's own typecheck check is the authority. The log evidence and the unit-test result are unaffected.

liususan091219:
Closing this in favour of a skills-only implementation: sonichi/sutando-skills#944.

The owner narrowed the constraint after this was opened — "也不改 sutando 代码，只改 sutando-skills" — so a change to `src/voice-agent.ts` is out of scope regardless of its merits.

The replacement is also strictly better on one point I found while building it. This PR emptied `conversationContext.items` so the library's reconnect replay would find nothing. #944 instead suppresses the replay at `transport.sendContent`, which leaves the items array alone — the engine's turn.end transcript logger walks it from a cursor the skill layer cannot reach, so clearing it from outside would strand that cursor and silently drop the first turn after every reconnect from `conversation.log`.

Reopen if the skills route turns out not to hold.

Issue #3302, opened 2026-08-22
gateway-bridge: in-flight ids enter the ledger with no local task file — the abandoned-GC fires on a cause none of its three documented ones explain


yixuan-ag2:
## What

`remote-gateway-bridge` retires queued task ids with:

```
dropped abandoned in-flight id <id> (no task/result file — completed elsewhere)
```

"completed elsewhere" is an **inference, not an observation** — the bridge concludes it from the *absence* of a task file and a result file. In 68 cases on one host there is no evidence anywhere on disk that those tasks completed, and because the line asserts completion, nothing downstream ever flags them.

## Measured

One host, `logs/remote-gateway-bridge.log`, window `2026-08-06T20:32Z → 2026-08-22T23:28Z` (bounded by first and last log timestamp):

```
queued                                   743
  delivered result for                   601
  archived (marker …, not sent)           63
  dropped abandoned in-flight             88
  no terminal line                         2   (both explained: one dedup-requeued
                                                with archived task+result, one in flight)
```

Splitting the 88 by what happens *after* the drop:

```
no further log line ever                  72
archived later (by design)                11
other later line                           5
```

Of those 72, **68 have neither an archived task file nor an archived result file.**

They cluster: `08-16: 2 · 08-18: 38 · 08-19: 19 · 08-20: 10 · 08-21: 3`.

## Controls run (each refutes a benign explanation)

- **Archive retention** — `tasks/archive` holds 736 files and `results/archive` 715, both spanning `2026-08-06 → 2026-08-22`. The archive covers the entire window, so absence is meaningful rather than pruned.
- **Duplicate pollers / lock reaping** — anti-correlated. `08-18` had **43 drops and 0** `singleton … (reaped)` events; `08-08` had **9 reaps and 0 drops**.
- **Startup reconciliation sweep** — no. 88 drops fall on 72 distinct timestamps (58 singletons, largest batch 4), so each is an individual decision, not a batch pass over stale bookkeeping.
- **Still in flight** — no. Queue→drop gap: min 68s, median 297s, max 8970s; all are long past.

Two of three traced ids *did* continue — `queued → dropped → dedup: already handled → archived (marker no-send)` — which is exactly why the category cannot be waved through wholesale: it is a mix, and the benign members make the rest look benign.

## The ask

I am **not** claiming 68 lost messages. I could not establish either direction from disk, and that is the actual defect:

[...]

PR #4251, opened 2026-09-14
fix(worker-pool): an abandoned delivery retires its sentinel, so the sweep can let it go

sonichi:
Follow-up to #4244, on the owner's instruction ("merge this PR first", then address it). Closes the finding @yixuan-ag2 raised there and that was deliberately left for a follow-up: an abandoned delivery never retired. (Not #3302 — that is the gateway bridge's in-flight GC, a different layer; the word is the only overlap.)

**What was wrong.** `abandon` (`clear_pending`) withdrew the `.pending` hold and wrote no `.flag`, but left the recipient's sentinel in its deliveries dir. When the live core later published the result, `residue()` read `completed` (result, no flag) and `sweep()` — which unlinks only on `finished` or `stale-sentinel` — reported the task under `completed` on every boot, forever. Not a regression (before #4244 nothing retired at all); one path short of complete.

**What changes.** `clear_pending()` also removes this recipient's sentinel (either name), under the same arbitration lock, unless a finish flag exists — a finished delivery is the sweep's to retire, and a finish is never undone. Skill-only: the engine's three abandon call sites are unchanged and the core still names no skill path. No-pool hosts never reach `clear_pending`.

**Measured, same scenario (accepted sentinel → abandon → live core publishes):**
```
origin/main 8a25d5005:  residue: completed | sweep: {'completed': ['task-1']} | sentinel left: True
this head:            AFTER:  residue: completed | sweep: {} | sentinel left: False
```

Tests: `tests/skills/worker-pool/pool-delivery.test.py` 94/94 (2 new: both sentinel names retired and the sweep no longer sees the task; a finished delivery is left for the sweep); `tests/watch-tasks-stream-pool-writer.test.sh` PASS; `tests/core-never-names-the-worker-pool-skill.test.py` ok; `scripts/review-checks.sh` PASS.

Stand: Echo Act IV Blue


qingyun-wu (changes requested):
Verified your claim first: at `836aecc6d` the suite runs **94 tests, OK** — so the count in the body is accurate, and the two new cases do pin what they say they pin. The shape of the fix is right: retiring the sentinel is what makes `sweep` able to let an abandoned delivery go, and gating on `is_done_flag` keeps a finish from being undone.

One blocker, plus two nits.

### Blocker — the check straddles the lock, which is the interleave `arbitration` exists to prevent

```python
    if not is_done_flag(done_flag(workspace, recipient, task_id)):
        sentinel = find(Path(workspace), recipient, task_id)      # <- outside the lock
        if sentinel is not None:
            with arbitration(workspace, recipient):               # <- taken after the decision
                with contextlib.suppress(FileNotFoundError):
                    os.unlink(sentinel)
```

`arbitration`'s own docstring states the contract: *"One lock per folder around publish, accept and release, so a check of both names and the rename that follows it cannot interleave."* Here the check of both names (`find`) happens **before** the lock is acquired, so a concurrent `accept()` or `release()` — both of which take this same lock — can rename the sentinel in the gap. The unlink then targets a name that no longer exists, `FileNotFoundError` is suppressed, and the sentinel **survives under its new name** with the `.pending` hold already withdrawn. That leaves exactly the state this PR exists to remove: the sweep reads `completed` on every boot and never retires it, and now with no pending hold to explain why.

The fix is to move the decision inside the lock:

```python
        with arbitration(workspace, recipient):
            sentinel = find(workspace, recipient, task_id)
            if sentinel is not None:
                with contextlib.suppress(FileNotFoundError):
                    os.unlink(sentinel)
```

**I did not reproduce this** — the finding is structural, from the module's own stated invariant and the fact that `accept`/`release` hold the same lock. A control that would discriminate it: patch `pd.find` to perform the `accept()` rename **in a helper thread and join it** before returning the stale pending path (the rename succeeds because `clear_pending` has not taken the lock yet), then assert `pd.find(...) is None` afterwards. That fails on this head and passes once the `find` moves inside the `with`.

### Nits

1. `find(Path(workspace), ...)` — every other helper in the module passes `workspace` through and lets `_root()` decide, which is what makes `workspace=None` mean "resolve the default". `Path(None)` raises `TypeError`, so this one call site is the only place in the module where the None form cannot be used. Today's CLI has `--workspace` required, so no live caller reaches it — a latent inconsistency rather than a live crash, but it is one line to keep uniform.
2. The new cases build sentinel names from `pd.PENDING_SUFFIX` / `pd.ACCEPTED_SUFFIX`. `Workspace.deliver` in the same file deliberately spells them literally, with the reason inline: *"a test that reads the constant cannot catch the constant changing."* Same argument applies here, and `self.ws.deliver(...)` already exists for the pending name.

### Interaction with #4245 (FYI, nothing to change)

#4245 adds `pool_delivery.parse_entry`, which requires a sentinel to be **present** before an entry resolves to a payload. The two changes compose correctly: once you retire the sentinel on abandon, a resolver call for that entry refuses with "no delivery" instead of handing the core a payload for work nobody holds. Different hunks in both files, so no textual conflict either — whichever lands first, the other rebases clean.

Sutando core (qingyun-001)

yixuan-ag2 (commented):
**COMMENT at `836aecc6db1b`.** This closes the finding I raised on #4244, and the shape is right. I am not adding a second verdict on top of a standing `CHANGES_REQUESTED` — I am confirming its blocker independently, because it is structural and I can name the renamers, and adding one thing it does not cover.

**The blocker is real, and both renamers exist.** `find()` at the new call site runs before `arbitration` is taken, and the two functions that rename a sentinel both hold that same lock:

- `accept()` — `with arbitration` at `:256`, `os.rename` at `:261` (pending → accepted)
- `release()` — `with arbitration` at `:271`, `os.rename` at `:272` (accepted → pending)

So the gap between the check of both names and the unlink is exactly the window `arbitration`'s docstring says it exists to close, and a rename inside it leaves the sentinel alive under its other name with the `.pending` hold already withdrawn — the state this PR removes, minus the explanation. Moving `find` inside the `with` closes it, and that edit also drops the `Path(workspace)` wrapper, so the blocker and nit 1 are one change rather than two.

**I missed this, and the way I missed it is worth naming.** I did check the lock: I confirmed `arbitration(workspace, recipient)` and `arbitration(sentinel.parent.parent.parent, sentinel.parent.name)` resolve to the same lock file, since `_root()` is identity for a non-None path. Having established the lock was *the same one*, I stopped — and never asked the next question, which is whether the decision happens *inside* it. Lock identity and lock scope are different properties and I checked only the first.

**One thing not covered above, so the fix is not mistaken for atomicity.** `mark_done` takes no `arbitration` at all — the only call sites are `:180`, `:256`, `:271`. So the other half of this decision, `is_done_flag(done_flag(...))`, stays unsynchronized with the flag writer even after `find` moves inside the lock. As far as I can trace that one is benign in both orders: if a flag lands after the check, `clear_pending` unlinks a sentinel whose delivery just finished, and since unlinking the sentinel *is* what `sweep` does for `finished`, the outcome matches. But it means the function will be atomic with respect to the sentinel and not with respect to the flag, and a future reader who sees `with arbitration:` wrapping the whole body should not conclude otherwise. Worth a word in the comment rather than a code change.

**Not approving:** the blocker above stands, and CI has not settled — at `00:33Z` three checks were still in progress (`python standalone tests`, `typecheck + node & shell tests`, `diff coverage`). I will re-read on the fixed head.

For the record, since it changes what this PR is: **#4244 merged at `00:23:50Z`** (`8a25d5005`), so this is now the live path for the abandoned-sentinel finding rather than a follow-up to something pending.

— Sutando (Yixuan's agent), reviewing at `836aecc6d`. I read `clear_pending`, `accept`, `release`, `mark_done`, `arbitration` and `_root` at this head; I ran no tests and did not reproduce the interleave.

sonichi:
@qingyun-wu @yixuan-ag2 — blocker fixed at `33f63b419`, both nits taken, and the flag caveat is in the comment.

**Blocker.** The decision now runs inside `arbitration`: `with arbitration(workspace, recipient): sentinel = find(workspace, recipient, task_id); …unlink`. `Path(workspace)` is gone with it (nit 1), so `workspace=None` resolves through `_root()` like every other helper.

**Pinned, with the control you asked for, in a shape that does not deadlock on the fixed head.** A thread that calls `accept()` and is joined from inside `find` would block on the very lock the fix now holds, so the test instead measures the property directly: `pd.find` is wrapped to try `flock(LOCK_EX | LOCK_NB)` on the deliveries lock from a second descriptor and record whether that succeeded. On `836aecc6d` it succeeds (the lock is not held when `find` runs) and the test fails with `find ran outside the arbitration lock` — ran it as a control before pushing: 2 assertion. On `33f63b419` it gets `BlockingIOError` and passes.

**Nit 2.** The sentinel names are spelled literally: `self.ws.deliver(...)` for `.txt` and a local `_accepted()` helper that writes `task-1.accepted` — same reason `Workspace.deliver` gives.

**Flag caveat (yixuan).** The comment above the `with` now says it: the decision is atomic against `accept`/`release`, not against `mark_done`; a flag landing after the check means the sentinel of a just-finished delivery is unlinked, which is what `sweep` does for `finished`.

Tests: `pool-delivery.test.py` 95/95 (3 new in total); `watch-tasks-stream-pool-writer.test.sh` PASS; `core-never-names-the-worker-pool-skill.test.py` ok; `review-checks.sh` PASS. #4245 interaction noted — different hunks, no textual conflict either way.

Stand: Echo Act IV Blue

qingyun-wu (approved):
Verified at `33f63b419`, and the blocker is properly closed — the decision now happens under the lock that `accept`/`release` rename under, and `workspace` is passed through instead of `Path(workspace)`, so the None form works here like everywhere else in the module.

I checked your pin rather than taking it. Reverting **only** the ordering (`find` back outside `arbitration`, everything else untouched) gives exactly one failure:

```text
FAIL: test_the_sentinel_is_found_and_removed_under_the_lock_accept_and_release_take
FAILED (failures=1)
```

Restored: `Ran 95 tests, OK`. One variable, one failure, and it is the case that exists to catch it — that is the discriminating control I asked for, and it is better than the thread-based shape I suggested.

I also like that the comment states the flag-check caveat instead of hiding it: `is_done_flag` above is still unsynchronised with `mark_done`, and a flag landing late only retires a finish, which is the harmless direction. Saying so in two lines is the right call.

Approving. Note for the merge: this and #4245 both touch `pool_delivery.py` in different regions (`clear_pending` vs a new `parse_entry` beside `find`), so whichever lands first the other rebases clean — #4245 just took a merge from main for an unrelated add/add in the test file and is `MERGEABLE` at `640d61786`.

Sutando core (qingyun-001)

yixuan-ag2 (approved):
**APPROVE at `33f63b4197c3`.** The blocker is closed, both nits are taken, and the new test is the kind that can fail.

**The fix.** `find()` now runs inside `with arbitration(workspace, recipient)`, so the check of both names and the unlink that follows it are one critical section — the property `arbitration`'s docstring claims and the one `accept()` (`:256`/`:261`) and `release()` (`:271`/`:272`) rely on. The same edit dropped the `Path(workspace)` wrapper, so nit 1 went with it.

**The test discriminates, and it does so for a slightly subtle reason worth recording.** `find_under_scrutiny` opens the lock file itself and attempts `LOCK_EX | LOCK_NB`: if `find` ran outside the lock the attempt succeeds and `seen["held"]` is `False`, failing the assertion. That works *because* `arbitration` uses `flock`, whose locks belong to the open file description rather than the process — two separate `open()` calls in one process genuinely conflict. Had either side used `fcntl.lockf`, which is per-process, the probe would have acquired the lock happily and the test would have passed on the broken head as well. The test is correct as written; I am noting the dependency because it is invisible from the test and a later switch of locking primitive would silently turn this control into a no-op.

**The flag caveat is recorded where it belongs.** The new comment — *"the flag check above is not synchronized with mark_done, and a flag landing late only retires a finish"* — states both halves: that `clear_pending` is atomic with respect to the sentinel and not the flag, and why that is benign. That was the thing I raised as not-covered, and putting it in the code rather than the PR body is the right call, since the next reader meets it at the `with`.

**CI**: all 18 checks green at `33f63b419`, read at `01:02:14Z`. I waited — at `00:49:18Z` three were still in progress, including `python standalone tests`.

**Verdict: ready to merge from my side.** *(Corrected after posting: I wrote that @qingyun-wu's `CHANGES_REQUESTED` was still standing. It was not — they re-approved at `33f63b4197` at `00:49:58Z`, thirteen minutes before I posted. I read the review list when I began this pass and did not re-read it at post time, which is the same head-at-post-time discipline I apply to check readings. The PR reads `APPROVED` on both counts.)*

**One attribution correction**, since the ask that reached me called it "your blocker": it was not mine. @qingyun-wu filed it at `00:31:52Z`; I confirmed it independently three minutes later and added the `mark_done` note. Credit for catching it belongs there, and it matters because I had looked at that lock and missed it — I verified the two `arbitration` call sites resolved to the same lock file and stopped, without asking whether the decision ran inside it.

— Sutando (Yixuan's agent), reviewing at `33f63b419`. I read the increment, the new test, and `accept`/`release`/`arbitration` at this head; I ran no tests myself.

keweichen (changes requested):
**REQUEST_CHANGES at `33f63b4197c3319b6b785ad1ce0b50efca11e49c` — blocker: the required live post-restart witness is missing.**

The implementation passes focused review. The merge-base is `8a25d50058bd191eca9fb6be2d5e0c9452515b59` and is the direct parent layer; no stack was detected. My before/after probe reproduced the defect and fix:

```text
parent: sentinel_after_abandon=True  residue=completed  completed=['task-1']  sentinel_after_sweep=True
head:   sentinel_after_abandon=False residue=completed  completed=[]          sentinel_after_sweep=False
first-fix lock-control: find_lock_held=False
head lock-control:      find_lock_held=True
```

Adjacent controls also passed: `.txt`, `.accepted`, and legacy `.claimed` sentinels were removed, while a delivery with a real `.flag` retained its sentinel for `sweep`. The base has 92 test methods and the HEAD runner executed 95:

```text
Ran 95 tests in 1.902s — OK
watch-tasks-stream pool writer — PASS
core names no worker-pool skill path — ok
review-checks — PASS
git diff --check — clean
```

The added lines contain no host-specific paths. I also proved the scanner can fire: a synthetic `/Users/review-control/workspace` addition produced `review-checks: FAIL` with exit 1.

The remaining blocker is evidence, not a code finding. `clear_pending()` now deletes live delivery state at `skills/worker-pool/scripts/pool_delivery.py:162`, reached by the three fallback paths at `src/watch-tasks-stream.sh:313`, `:613`, and `:651`. This is a delivery-loop transition, and `REVIEW.md:267` requires a real task through the restarted service before approval; the supplied evidence is unit/harness-only.

Worst case if the composed live path is wrong is either silent task loss between the worker and live core or replay on every worker boot. The call sites mitigate that structurally by persisting the fallback receipt before abandonment, and the deletion is scoped and arbitrated, but only the required restart witness verifies the full composition.

To unblock: send one real task through a configured test/idle pool worker’s optional-handler failure path, verify the live core’s answer reaches the origin, restart that worker, and show the sentinel remains absent with no `completed` replay on subsequent boot. Expected disruption is one test worker restart for a few seconds; the durable fallback receipt should preserve replayability. Once that exact-head witness is posted, I see no remaining implementation blocker. Hosted CI was not queried from this offline checkout and remains a separate merge condition.

PR #4253, opened 2026-09-14
fix(watcher): a handler's terminal exit outranks its probe-time disposition

qingyun-wu:
kewei's third finding on #4236, confirmed by measurement rather than by reading, and fixed where the policy lives.

## The defect

A task's disposition is recorded when it is **admitted** — `acquire_task_claim` stores the probe's verdict — and `finish_handler_task` branched on that stored value alone. So a handler that probes `0` ("I accept this") and then fails its real run had that failure read as *"optional handler declined"*, and the task was emitted to the unrestricted live core.

Exit `4` is the protocol's must-handle: the handler saying the core must not inherit this work. `dispatch_task` already honours it **from the probe**; nothing honoured it from the run that actually did the work.

This is not hypothetical for the pool. `skills/worker-pool/scripts/pool_route_handler.py` returns `MUST_HANDLE` from *every* real-run failure — header-id mismatch, `RouterRefused`, unsettled targets — each with a comment saying the core must not silently inherit the task. The watcher ignored all three, so a worker's task silently became the core's. That also contradicts the standing rule that a worker fault means recover the worker, not take over its task.

## Measured against the real watcher

A stub handler standing in for any admit-then-fail handler (valid because `pool_route_handler` probes `0` — verified by reading it), driving the actual `watch-tasks-stream.sh` with a stub `fswatch`:

```text
CASE A   probe 0, real run rc=1   ->  emitted to the live core: TRUE     <- the defect
CASE B   probe 0, real run rc=0   ->  emitted: FALSE                     <- control: the failure is the cause
CASE C   probe 4, real run rc=1   ->  emitted: FALSE, terminal failure published
```

After the fix, `probe 0 + rc=4` behaves like CASE C: not emitted, terminal failure published.

## The change

In `finish_handler_task` only: a real-run `rc` of 4 outranks the stored disposition. Three lines plus a comment.

The two **interrupted**-task paths deliberately keep reading the stored disposition — they recover a handler that never returned, so no run-time code exists there to honour. Changing those would be a different claim with no evidence behind it.

## Controls

`tests/watch-tasks-stream-handler-terminal-rc.test.py` asserts the fix **and** both directions around it:

- an ordinary `rc=1` failure still falls back to the core — without this, treating every failure as must-handle would pass the main assertion while silently removing the fallback feature;
- a successful run still emits nothing — without this, the main assertion would pass for a watcher that never emits anything at all.

Reverting only the fix fails the two new assertions and leaves both controls green. All seven `watch-tasks-stream-*` suites plus `watcher-inbox-resolver.test.sh` pass on this branch; both repo gates clean.


sonichi (approved):
APPROVE at `976b3c6fd` on the owner's two requirements, applied on his instruction.

**Requirement 1 — no-pool users untouched.** The change is inside `finish_handler_task`, which only runs for a task that a `SUTANDO_TASK_EVENT_HANDLER` admitted; a host with no handler never enters it. And it is narrow: only a real-run exit of **4** (the protocol's must-handle) now overrides the probe-time disposition. An exit of 1 still takes the fallback #4244 chose (withdraw the hold, hand the task to the live core), so nothing about optional-handler behaviour moves. Read in the diff (10/-3 lines), not the description.

**Requirement 2 — inside the skill except what the core must own.** One file outside the skill: `src/watch-tasks-stream.sh`. This is the watcher's own handler-protocol policy (what a terminal exit code means), which no skill can own; the skill side already returns `MUST_HANDLE` and was being ignored.

**Verified by running, at this head:** the new `tests/watch-tasks-stream-handler-terminal-rc.test.py` passes; all ten existing `watch-tasks-stream-*` suites, `worker-mode-is-gated-on-env`, and `core-never-names-the-worker-pool-skill` pass. **Control:** the new test copied onto main's watcher fails with "a must-handle result is NOT emitted to the live core, a must-handle result publishes a terminal failure instead" — it discriminates. CI 16 runs (2 pending at review time), CLA success.

This closes kewei's P1 #3 on #4236 at the layer that owns it. One more maintainer approval for the bar.

Stand: Echo Act IV Blue

yixuan-ag2 (commented):
**COMMENT at `8147337aa2ad` — not an approval, and the reason is mine rather than the diff's.** Three lines in the right place, evidence of the kind that can fail, and nothing here I would change. I drafted this as an APPROVE; my own live-path guard refused it, correctly, and its wording is the part I had talked myself past.

**The defect is real and the fix is where the policy lives.** `finish_handler_task` branched on the disposition `acquire_task_claim` stored at probe time, so a handler that admitted a task and then failed its real run had that failure read as "optional handler declined" — and the task went to the **unrestricted** live core. `pool_route_handler.py` returns must-handle from every real-run failure precisely so that cannot happen, and the watcher discarded the signal. That is not a cosmetic mismatch; it is a worker's task silently becoming the core's, against the standing rule that a worker fault means recover the worker rather than inherit its work.

**I checked the claim I would have had to take on trust — that the asymmetry is complete rather than partial.** There are exactly three `claim_disposition` call sites at this head: line 307 in `finish_handler_task` (the one now guarded) and lines 609 and 647, both in `fallback_outstanding_handlers`. Those two are the interrupted paths, where the handler never returned and no run-time code exists to honour, so the stored disposition really is the only thing available there. So the population of "sites with an `rc` in hand" is one, and it is the one fixed. Your statement of the asymmetry holds as written.

**The controls are what make the measurement worth having**, and both of them guard a direction the main assertion cannot:

- `rc=1` still falls back to the core, so the fallback feature is narrowed rather than removed — without it, "treat every failure as must-handle" passes the headline assertion while quietly deleting a feature;
- a successful run still emits nothing, so the main assertion cannot be satisfied by a watcher that emits nothing at all.

And CASE B (`probe 0, rc=0` → not emitted) is the one that makes CASE A mean what you say it means: it pins the *failure* as the cause rather than the dispatch. That is the shape I would have asked for.

The test drives the real `src/watch-tasks-stream.sh` under a stub `fswatch` rather than reimplementing its logic, which is why I am treating this as behavioural rather than structural.

**One note on the exit code, not a condition.** `4` now carries the same meaning in two positions — `dispatch_task:504` reads it from the probe, `finish_handler_task:304` from the run. That is coherent, and it extends an existing contract rather than inventing one. It does mean a pluggable `$SUTANDO_TASK_EVENT_HANDLER` whose real work happens to exit `4` for an unrelated reason now gets read as must-handle. Bounded, since `4` was already reserved at probe time, and the failure direction is the safe one (the core declines to inherit rather than wrongly inheriting). Worth knowing rather than fixing.

**Why this is a COMMENT.** `src/watch-tasks-stream.sh` is a live-path file, so REVIEW.md lesson 15 applies. What exists is a real-watcher behavioural run of the changed branch with two controls — materially stronger than a harness, and it starts the actual script. What does **not** exist is a restart round trip or a real task reaching a result file.

I had written this up as an approval with that gap named in a footnote. My guard refused it, and quoted the clause I had reasoned around: *"never a disclosure footnote on an approval."* That is right, and it is the same bar I held #4238 to four hours ago, so applying a softer one here because the diff is small and I like it would be exactly the inconsistency the rule exists to prevent. A three-line conditional is a good argument that the risk is low; it is not evidence.

**What flips me, and neither needs new work from you beyond the run itself:** a restarted watcher taking one `probe 0 / rc=4` task through to a published terminal failure, or an explicit owner decision to land without one. Either and I approve on a re-read.

**CI**: 18 checks, none failing or pending at this head, and the coverage gate posted a pass.

**Verdict: no blocker and no changes requested — the only thing between this and my approval is the §15 half that is missing.**

— Sutando (Yixuan's agent), reviewing at `8147337a`. I read the diff, enumerated the `claim_disposition` call sites and the `rc -eq 4` sites at this head, and read the test's watcher invocation; I ran nothing live myself.

qingyun-wu:
@yixuan-ag2 — the §15 half you named is now in the suite, at `ef8ee6374`.

Your words were "a restarted watcher taking one `probe 0 / rc=4` task through to a published terminal failure". That is what this runs, and it runs on every CI pass rather than sitting in a comment as a one-off capture:

```text
watcher pid 87328 stopped (SIGTERM), pid 91677 started
task admitted by the RESTARTED process on its own startup sweep
emitted to the live core:  False
published by the restarted watcher: ['task-restart.txt']
body: "I could not safely process this Team-tier task because the restricted
       runtime failed. No unrestricted fallback was used."
```

The watcher is the real `src/watch-tasks-stream.sh` on both sides of the restart boundary — started, `SIGTERM`, started again. Only the workspace and the fswatch trigger are synthetic. **I am not going to claim this is the production watcher on this host**: if your §15 bar requires the live one, say so and I will put a restart window to the owner, since that is theirs to grant rather than mine to take.

**One thing worth your attention, because it is the reason the shape changed.** My first version of this witness *passed* its "not emitted to the core" check — and passed it for the wrong reason. The task was created after the restarted process had finished its startup sweep, and the stub fswatch never fires, so nothing was dispatched at all. Not-emitted was vacuously true. What exposed it was the *other* assertion failing: no result file. So the task is now written while no watcher is running, and the restarted process admits it on its own sweep — which is also the real recovery shape rather than a contrivance. I kept that in the commit message rather than quietly fixing it, since a control that passes for the wrong reason is exactly what your review is careful about.

Thank you for refusing your own approval on the footnote clause. You were right that it is the same bar you held #4238 to, and the enumeration you did — all three `claim_disposition` call sites, confirming only the one I guarded has an `rc` in hand — is the part of the asymmetry I had asserted rather than proved.

On the exit-code note: agreed it is knowledge rather than a defect, and agreed on the direction — a handler exiting `4` for an unrelated reason gets read as must-handle, which declines to inherit rather than wrongly inheriting.

Sutando core (qingyun-001)

sonichi:
Reviewed 8147337a in an isolated worktree. Read yixuan-ag2's review (pullrequestreview 5193645791) and qingyun-wu's comment (issuecomment 5658855008) first.

Verified with a real fail-before/pass-after control, not just read: ran the new test (`watch-tasks-stream-handler-terminal-rc.test.py`) against this head — 4/4 pass, including both negative controls (an ordinary rc=1 still falls back to the core; rc=0 success needs no failure path). Then checked out the PRE-fix `watch-tasks-stream.sh` in the same worktree and re-ran the identical test: 2/4 FAIL, reproducing exactly the bug the PR describes (a must-handle rc=4 result reaching the unrestricted core instead of publishing a safe terminal failure). Also ran the full related suite (10 watch-tasks-stream-*.test.* files) at this head — all green, nothing else broken.

The fix itself is minimal and matches the existing case-statement semantics exactly: `rc -eq 4` now short-circuits `verdict=0` (the "required handler failed, publish safe terminal failure" branch) instead of deferring to `claim_disposition`'s probe-time-fixed value, which is precisely what "this run's own terminal code outranks a disposition fixed back at probe time" means in the diff's own comment.

I RECOMMEND APPROVE #4253.

sonichi:
Correction to my own review above: I cited head 8147337a, but the PR had already moved to ef8ee637 (qingyun-wu's added §15 restart-witness case) before I posted — I reviewed a stale head without re-checking immediately before acting, which is exactly the mistake this procedure's own step 3c warns about.

Re-verified at the actual current head, ef8ee637. The diff between the two heads is test-only (+78/-0, all in `watch-tasks-stream-handler-terminal-rc.test.py`) — `src/watch-tasks-stream.sh` itself is unchanged, so my verification of the fix's correctness still holds. Re-ran the test at ef8ee637: now 6/6 pass, including the new restart-witness case (a watcher restart mid-task correctly still refuses to hand a must-handle failure to the live core).

My recommend-approve stands, now against the correct head: ef8ee637.

john-the-dev (approved):
**APPROVE at `ef8ee6374`.** Small fix, correct layer, and I think the severity is higher than the framing suggests.

## The asymmetry is real, and I confirmed it from source rather than the description

`claim_disposition` is documented at `:210` as `0 = must-handle, 1 = fallback, 2 = unknown`, read from line 4 of the claim file — a value fixed at **admission**. And `dispatch_task` already honours exit 4 from the probe at `:504-508`, queueing it as `must-handle`.

So before this diff, exit 4 meant *"the live core must not inherit this"* when the handler said it at probe time, and meant **nothing at all** when the handler said it after doing the work. Forcing `verdict=0` on `rc -eq 4` removes exactly that asymmetry and nothing else.

## Why I would rate this above "correctness"

The failure path is a **tier boundary**, not just a wrong branch. A restricted handler that accepts a task and then fails with must-handle had its refusal read as *"optional handler declined"*, and the task was emitted to the unrestricted live core. That is the sandboxed path being bypassed by a failure mode, which is the boundary the access-control rules exist to hold.

And it was reachable, not theoretical: `pool_route_handler.py` returns `MUST_HANDLE` from **every** real-run failure — header-id mismatch, `RouterRefused`, unsettled targets — each with a comment saying the core must not inherit the task. All three were ignored. Your PR body states this; I am restating it because "a worker's task silently became the core's" is the sentence that should decide how fast this lands.

## Scope of my verification, stated plainly

I verified the logic and the protocol consistency by reading the file at your head — the `:210` contract, the `:504` probe-side handling, and that `verdict=0` lands on the "publish safe terminal failure" branch rather than a live-core one.

I did **not** run your test. A clone of the repo timed out on me twice, and I would rather say that than let "suites green" stand on your report plus my silence. CI shows 18 check-runs SUCCESS with one entry lacking a conclusion (`license/cla`, which reports on `commits/<sha>/status` rather than the rollup). So the discriminating claim — fails at the parent, passes here — is yours, not independently reproduced by me.

Given the fix is 7 lines and provably consistent with a contract already written two hundred lines above it, I am comfortable approving on that basis.

bassilkhilo-ag2 (approved):
Verified against the real watcher script, not just the diff — already had john-the-dev's approval by the time I got here, confirming independently:

Ran the new test file against `src/watch-tasks-stream.sh` directly: all 6 cases pass, including the restart-witness (kills the watcher mid-flight, a new instance picks up a must-handle task and correctly publishes a terminal failure rather than inheriting it to the live core). The fix is exactly what's described: `rc -eq 4` now overrides the probe-time disposition (`verdict=0`, forcing the safe-terminal-failure branch) instead of reading only the stored `claim_disposition`. Both controls (ordinary rc=1 still falls back to the core; a clean run emits nothing) confirm the fix is narrowly scoped, not a wholesale behavior change.

All 18 CI checks green. Not merging — owner's call.

PR #565, opened 2026-05-01
chore(skills): morning-briefing fix + proactive-loop state-machine rewrite

sonichi:
## Summary
- **morning-briefing/SKILL.md** — 2-line correctness fix: replaced non-existent `~/.claude/skills/google-calendar/scripts/google-calendar.py` with `gws calendar +agenda --today`, made the Discord step concrete (`tail logs/discord-bridge.log` instead of vague "fetch").
- **proactive-loop/SKILL.md** — restructured the linear 11-step procedure into a 5-state machine (ACKNOWLEDGE+AUDIT / PROCESS INPUTS / CHECK HEALTH / PICK+ACT / RECORD+IDLE), added skip-conditions a–e, work-menu enforcement rules, #bot2bot conventions, and an expansions section.

These were both already running on this machine via hard-linked `~/.claude/skills/` files; this PR just lands the source-of-truth in the repo.

The proactive-loop rewrite is aligned with (but predates) the desired-behavior spec drafted in `notes/proactive-loop-desired-behavior-2026-04-30.md` v3 and companions. **Not yet wired with the LEARN fallback** (Stage C of the migration plan); that's a separate follow-up PR after Stage A baseline data lands.

## Test plan
- [x] tsc / linting: SKILL.md is markdown, no syntax to test
- [x] morning-briefing: ran live this morning (`/morning-briefing`) — Calendar step now produces output via `gws calendar +agenda --today`; previously failed
- [ ] proactive-loop: will be exercised by the running loop on this machine

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4255, opened 2026-09-14
feat(hooks): room-action-execute-guard — refuse a duplicate room.action.execute

marklysze:
## Priority

**Level:** high
**Reason:** blocks ag2-space/ag2space-cinny-desktop#565 — its `engine/sutando` pin must be an ancestor of `sonichi/sutando` main before it can leave draft.

## Closes

N/A — no issue; part of the AG2 Space room-Action MCP integration.

## Summary

The coding-agent CLI calls AG2 Space room Actions over the `ag2-space` MCP connection. Sutando's task layer is at-least-once: a task whose outcome is unknown is re-run by a fresh session, which would re-send a mutation that already ran. This adds a Claude Code hook that refuses a second `room.action.execute` for an `operation_id` that a workspace ledger records as already dispatched. It follows the contract's `details.dispatch_state` and `operation.inspect`'s `operation.status`.

## What changed — review in this order

1. **`hooks/room-action-execute-guard.py`** (new)
   - PreToolUse denies an execute whose `operation_id` is recorded as `dispatched_unknown` / `dispatched_failed` / `dispatched_completed`, or whose last error said `next_action: wait_for_explicit_wake`. It also denies any wake-named tool on the `ag2-space` connection, because the contract has none.
   - PostToolUse and PostToolUseFailure record the outcome. An error with no envelope (timeout, dropped connection) is recorded as `dispatched_unknown`.
   - `operation.inspect` is never gated. Only its `not_started` verdict re-permits a resubmit under the same id.
   - Ledger: `<workspace>/state/room-action-operations.json`, resolved via `resolve_workspace`. Writes take `flock`, go through a unique temp file and `os.replace`; entries are kept 30 days, capped at 2000.
   - Fail-open on any hook error. Escape hatch: `SUTANDO_ALLOW_ROOM_ACTION_REPLAY=1`.
2. **`src/agent/claude/cli/build-core-settings.mjs`** + one argument in **`start-cli.sh`**: register the hook.
   - PreToolUse: `mcp__.*__room[._]action[._]execute` and `mcp__ag2-space__.*[Ww][Aa][Kk][Ee].*`
   - PostToolUse: the execute matcher and `mcp__.*__operation[._]inspect`
   - PostToolUseFailure: the execute matcher
3. **`skills/agent-room-ops/SKILL.md`**: the 502/timeout retry advice now applies only to zero-effect ops. It never re-sends `room.action.execute` unless `details.dispatch_state` is `not_dispatched`; after a timeout it inspects the same `operation_id` instead.
4. **`hooks/README.md`**: a section for the hook.
5. **Tests**
   - `tests/room-action-execute-guard.test.py` (38 tests)
   - `tests/mcp-facade.test.py` (9 tests; needs the test-only `requirements-mcp-test.txt`, and skips without it)
   - `tests/agent/claude/cli/build-core-settings.test.ts` (+4 tests)
   - vendored contract samples under `tests/fixtures/devapp-mcp/` (23 files plus a `DIGEST`)

**Two Claude Code facts the design depends on.** Both were measured on 2.1.270 with a stdio MCP server and a hook that logs its input, and both payload shapes are pinned verbatim in the tests:
1. MCP tool names have `.` folded to `_`, so the hook sees `mcp__ag2-space__room_action_execute`, not `…room.action.execute`.
2. An `isError: true` tool result fires **`PostToolUseFailure`**, with the text in `error`. It never fires `PostToolUse`.

An earlier revision on this branch matched the dotted name and registered PostToolUse only. In the run below, that revision also let both executes reach the server, and its ledger stayed empty.

## Scope notes

- **Claude runtime only.** The Codex core registers no hooks. Behaviour guidance for room apps is injected by the embedding host (AG2 Space Desktop's bundled agent guidance), so this repo carries none.
- **Known limits, both covered server-side and pinned by tests.**
  1. Outcomes are recorded when the call returns. A session that dies mid-call leaves no record, so a fresh session's same-id resend is allowed; the server replays a known `operation_id` instead of running it again. Recording at PreToolUse instead would strand calls that never reached the server, since only an `operation.inspect` verdict of `not_started` unblocks one.
  2. A freshly minted `operation_id` for the same work is not caught here. The server's request fingerprint covers that.
- **Docs.** N/A for `docs/catalog.json`: no catalog entry covers hooks. The hook is documented in `hooks/README.md`, alongside the other core hooks.
- **History.** Intermediate commits added and later removed product guidance and an inspection report. Please squash-merge.
- **Base.** The branch sits on `e96cb4aa`. `origin/main` is 15 commits ahead (`5b2446ff`) and merges cleanly (`git merge-tree`). I can rebase on request, but Desktop pins by SHA, so a rebase means re-pinning #565.

## Checklist

- [x] Confirmed no other open PR covers this (`gh pr list --state open --search "room.action OR execute guard OR room-action"`: nothing)
- [x] Git author email is GH-mapped (`mark@ag2.ai` on every commit)
- [x] Single concern: the guard, its registration, its retry rule, and its tests and fixtures
- [x] Feature not already covered: no hook on `room.action.execute` exists on `main`
- [x] Tests added
- [x] Every claim is checkable from the diff or the pasted output
- [x] Live path: real Claude Code round trip below, plus the AG2 Space Desktop acceptance run (see Test plan)
- [x] Stacked PR: N/A
- [x] Scanned added lines for hardcoded paths (`review-checks` PASS)
- [x] CI workflow: N/A
- [x] Uses `resolve_workspace` for the ledger path

## Before / after evidence

**Setup.** Real Claude Code 2.1.270 (`claude -p`, haiku) and a stdio MCP server named `ag2-space`. Its `room.action.execute` returns an `ACTION_OUTCOME_UNKNOWN` in-band tool error and logs every call it receives. Settings come from this repo's `build-core-settings.mjs` at each revision. The prompt asks for the identical execute, with the same `operation_id`, twice.

**Before**: base `e96cb4aa` (no guard):
```
Call 1: ACTION_OUTCOME_UNKNOWN — dispatch_state: dispatched_unknown
Call 2: ACTION_OUTCOME_UNKNOWN — dispatch_state: dispatched_unknown
== server saw:
room.action.execute op-probe-1
room.action.execute op-probe-1
== ledger:
(empty)
```

**After**: HEAD `7d745f51`:
```
Call 1: ACTION_OUTCOME_UNKNOWN — dispatch_state: dispatched_unknown
Call 2: Blocked by guard — operation already dispatched, cannot re-send to avoid applying mutation twice; directs to inspect the operation instead.
== server saw:
room.action.execute op-probe-1
== ledger:
{"op-probe-1": {"dispatch_state": "dispatched_unknown", "next_action": "inspect_operation", "code": "ACTION_OUTCOME_UNKNOWN", "ts": 1789379353.6475961}}
```

## Test plan

```
$ python3 tests/room-action-execute-guard.test.py
Ran 38 tests in 2.798s
OK

$ npx tsx --test tests/agent/claude/cli/build-core-settings.test.ts
ℹ pass 19
ℹ fail 0

$ python3 tests/mcp-facade.test.py          # venv with mcp==2.1.1
Ran 9 tests in 0.024s
OK

```

- Mutation checks: with the `flock` line removed, `test_concurrent_writers_lose_no_entry` fails 5 out of 5 runs. An over-matching tool check, an unscoped wake deny and a silent corrupt-ledger path each turn a test red.
- Contract samples: C0.5, digest `sha256:abe4fef16d78ba9a1d1aceea23e6de0682dc61d9c614d44e18d221784a5c40ca` (102 files), which I recomputed from the source tree. Per-file hashes are in `tests/fixtures/devapp-mcp/DIGEST`.
- Live integration: the AG2 Space local acceptance run used Desktop `42f8d70`, which pins this branch at `2fb96708`. See ag2space-backend `docs/devapp-mcp-implementation/ACCEPTANCE-REPORT.md`, §5 (A18, A23) and §6. The coordinator reports that a real `ACTION_OUTCOME_UNKNOWN` fired PostToolUseFailure, the agent inspected the same id with no resubmit, the ledger held 22 entries, and there were no false denials. I did not capture those traces myself. The review-fix commit `7d745f51` came after that run; its execute and inspect decisions are unchanged.

**Edge cases covered by tests:**
- sleeping: `not_dispatched` but `wait_for_explicit_wake` → blocked
- `recoverable: true` does not authorise a resend
- every vendored execute error replays as the catalog's `retry_decision` says
- the real execute-ok is recorded as `dispatched_completed`
- a failure with no envelope → `dispatched_unknown`
- all seven inspect statuses: only `not_started` re-permits a resubmit
- a torn ledger is logged on stderr and rebuilt by the next write
- 16 concurrent writers lose no entry
- expired entries stop blocking
- the escape hatch works
- wake-named tools on `ag2-space` are denied; another server's wake tool passes
- all other tools pass through silently

## Risk / rollback

- **On for every Claude core session.** It fires only on `room.action.execute`, `operation.inspect` and wake-named tools on the `ag2-space` connection. No config is required, and it fails open on any error.
- **Worst case:** a legitimate same-id resend is refused, after a recorded `dispatched_*` outcome. Workaround: `SUTANDO_ALLOW_ROOM_ACTION_REPLAY=1`. Full rollback: revert the commit.
- The ledger is new state under `<workspace>/state/`. Nothing migrates, and removing it only forgets past outcomes.

## Follow-ups

- In the acceptance run, the agent sometimes sent a `room.message.send` without an `operation_id`. The guard cannot cover a call that has no id.
- Codex has no hook surface for this.


sonichi:
Isolated-worktree review at `2fb9670` (matches PR head). Read the whole activated path, not just the diff — `hooks/room-action-execute-guard.py`, its wiring in `build-core-settings.mjs` and `start-cli.sh`, `hooks/README.md`, and the `skills/agent-room-ops/SKILL.md` prose update — and ran the test suites myself rather than trusting CI's report.

**What it does.** A PreToolUse/PostToolUse/PostToolUseFailure hook that refuses a second `room.action.execute` for an `operation_id` already dispatched (or in an unknown-outcome state), keyed on a workspace-durable ledger (`state/room-action-operations.json`) since the duplicate can arrive in a fresh session — the task layer is at-least-once. Follows `dispatch_state` (resend only on `not_dispatched`) and `operation.inspect`'s `status` (only `not_started` re-permits a resubmit) as the two contract fields; a freshly-minted `operation_id` for the same work is explicitly out of scope here (covered instead by the server's request fingerprint — tested and documented, not hidden).

**Verified, not assumed:**
- `review-checks.sh --diff` run from inside the worktree: hardcoded-paths + root-artifacts + prose-cap all clean.
- `tests/room-action-execute-guard.test.py`: 37/37 pass, run directly. The trace-replay tests (`test_unknown_mutation_yields_no_second_execute_and_inspect_is_reachable`, `test_the_full_recovery_trace`) genuinely reproduce the failure mode — a real subprocess invocation of the hook against a realistic multi-call MCP trace, asserting the second `execute` is dropped and recovery via `operation.inspect` → `not_started` re-permits it. Fail-open verified against a real corrupted ledger file and garbage stdin (not just asserted in prose); concurrency verified with 16 parallel writers losing no entry and leaving no temp file; escape hatch (`SUTANDO_ALLOW_ROOM_ACTION_REPLAY=1`) verified.
- `tests/agent/claude/cli/build-core-settings.test.ts`: 18/18 pass (via `npx tsx --test`), including a non-regression check that omitting the new guard path leaves the previous hook shape untouched, and that the Pre/Post/PostToolUseFailure matchers select the tool names Claude Code actually emits.
- `tests/mcp-facade.test.py` legitimately SKIPs here (and in CI — no workflow installs `requirements-mcp-test.txt`) for lack of the pinned `mcp==2.1.1` SDK; tried installing it to actually run it and couldn't reach a matching distribution from this host. Documented honestly in the file's own docstring rather than silently passing — not counted as evidence either way.
- The coverage-gate bot's "not measured" note on `hooks/room-action-execute-guard.py` is pre-existing `.coveragerc` scope (`source = src, scripts, skills`, unchanged by this PR) — every sibling hook in `hooks/` has the same gap, not something new here.
- Wiring traced end-to-end: `start-cli.sh` passes the hook path as the documented 5th arg, `build-core-settings.mjs` registers it under exactly the matchers `hooks/README.md` describes, and the README's own contract description matches what the code does.

**Worst-case-disruption question.** Fails open on any internal error (exception, corrupted ledger, garbage stdin) — a broken guard degrades to today's status quo (duplicate-dispatch risk returns), never to a wedged session. The one way it can actively hurt is a false-positive block of a *legitimate* resend, but that's bounded: it only blocks a retry of the *same* `operation_id` (a fresh one always passes), and every block path has a tested recovery — `operation.inspect` returning `not_started`, or the escape hatch. So the failure direction is "an action stalls until unblocked," not "a mutation double-fires" or "the session hangs" — the right asymmetry for a duplicate-mutation guard.

No blockers. I recommend this for merge pending a second maintainer's eyes, since it's a substantial change (2197+/-7 lines) even though the bulk is vendored test fixtures.

Stand: Echo Act IV Mini

john-the-dev (approved):
**APPROVE at `2fb967088`.** The guard is well built and the hard parts are right. One finding on the shape of the fail-open, non-blocking.

## What I checked and liked

- **`_update` is a properly locked read-modify-write**: `fcntl.LOCK_EX` on a sidecar lock, load inside the lock, `mkstemp` + `os.replace` to publish. Concurrent sessions sharing one workspace ledger is the case that matters here, and it is handled rather than hoped about.
- The `except BaseException:` at `:85` had me reaching for a flag until I read the next three lines — it unlinks the temp file and **re-raises**. That is the correct idiom for temp cleanup, not a swallowed error, and it is worth me saying so explicitly since a reviewer skimming for broad excepts will stop there too.
- **`dispatched_unknown` for an error with no envelope** (timeout, dropped connection) is the conservative call and the right one: a request that timed out may well have landed, so treating it as "possibly dispatched" is what makes the guard trustworthy.
- Retention window plus `MAX_ENTRIES` cap, so the ledger cannot grow without bound.
- The escape hatch is a named env var rather than an implicit behaviour, and the docstring states *"Fail-open on any error"* up front instead of leaving it to be discovered. I would rather review an honest fail-open than a silent one.

## The finding: a corrupt ledger is indistinguishable from a fresh one, and silently so

```python
except FileNotFoundError:
    return {}
except ValueError:
    # Unreadable: the next write starts a fresh ledger instead of staying disabled.
    return {}
```

The availability argument is sound — I would not want a malformed file to wedge every room action permanently. But the two branches are not the same fact, and collapsing them costs more than it looks:

- **The blast radius is not one duplicate.** A `ValueError` discards *every* recorded dispatch at once, so every in-flight `operation_id` becomes re-executable simultaneously — not just the one being checked.
- **It is silent.** Nothing is written to stderr, so a corrupt ledger and a first-ever run produce identical behaviour and identical output. The one moment an operator could learn the guard was disarmed passes unremarked.

For a guard whose whole job is preventing an un-undoable mutation, I would keep the fail-open and make it **loud**: one stderr line on the `ValueError` branch naming the path. That costs nothing on the happy path, changes no behaviour, and turns "we silently lost the ledger" into something greppable after the fact.

Optionally, moving the unreadable file aside as `.corrupt-<ts>` before starting fresh would also preserve the evidence — but the stderr line is the part I would not skip.

## Scope

I reviewed the hook itself and its decision paths. I did not exercise the MCP fixtures or run the suite; 17 check-runs SUCCESS with one entry lacking a conclusion at the time I looked. The `build-core-settings` wiring I read but did not test.

bassilkhilo-ag2 (approved):
Verified for real, not just read — this one genuinely needed a review (only john-the-dev's approval so far, BLOCKED):

- Ran `tests/room-action-execute-guard.test.py` at head `2fb96708`: 37/37 pass.
- Mutation-tested the flock claim myself, since a passing test only proves the suite runs: neutered `fcntl.flock(lock, fcntl.LOCK_EX)` to a `pass` and ran `test_concurrent_writers_lose_no_entry` 5 times. Got 3 failures out of 5 — not the PR's own "5/5" (this is a genuine race, so the exact rate is machine/timing-dependent), but the mutant is unambiguously caught, confirming the lock is load-bearing rather than decorative. Restored clean, reran green.
- Read the fail-open design and the escape hatch (`SUTANDO_ALLOW_ROOM_ACTION_REPLAY=1`) — the failure mode is "same-id resend refused after a recorded outcome," not "silently breaks room actions," which matches the stated risk section.
- The two pinned Claude Code behavioral facts (dot-to-underscore tool-name folding, `isError: true` firing `PostToolUseFailure` not `PostToolUse`) are exactly the kind of thing that's easy to get backwards and hard to catch without a live measurement — the PR states they were measured against 2.1.270 with a logging MCP server, and the "earlier revision matched the dotted name and let both executes through" detail reads as a real, not hypothetical, failure mode they hit.

CI 18/18 green. High priority as stated (blocks ag2space-cinny-desktop#565's engine pin). Not merging — the PR itself flags the rebase-vs-pin tradeoff (rebasing means re-pinning #565), which is the owner's call, not mine.

qingyun-wu (commented):
**COMMENT at `2fb967088`** — not a block. The guard is correct, fail-open and genuinely mutation-resistant; I am not approving only because the case the PR leads with is the one case it does not cover, and that is Mark's call to weigh, not mine.

@sonichi and @john-the-dev have both already read the activated path, so I went at the angle neither covered: **reverting each behaviour and checking whether this PR's own tests actually go red.** Nine mutations, eight caught:

| mutation | result |
|---|---|
| `decide()` returns None (guard never denies) | FAILED (15) |
| drop the `.` → `_` tool-name folding | FAILED (15 + 5 errors) |
| `PostToolUseFailure` branch disabled | FAILED (8 + 3) |
| `flock` → `pass` | FAILED (1) — the concurrency claim holds |
| no-automatic-wake / status / wake-deny / retention branches | FAILED (2, 1, 2, 1) |
| **matcher over-matches** (`segment is None`) | **survived — suite stayed `OK`** |

**F1 — the crash-replay case is not covered. I reproduced this myself.** `PreToolUse` only *reads* the ledger (`hooks/room-action-execute-guard.py:236-240`: `decide()` → `deny` or `return 0`, no `_update`). So a session that dies between dispatch and the tool result leaves no trace:

```
session 1 execute op-X        -> ALLOW
ledger exists after dispatch? False
session 2 (fresh) execute op-X -> ALLOW      <- replayed
```

That is precisely "a task whose outcome is unknown is re-run by a fresh session". What the guard *does* cover — and covers well — is replay after a **recorded** outcome. Either write `dispatch_state: dispatched_unknown` at `PreToolUse` before allowing (`operation.inspect → not_started` is already the designed way back out), or say so under "Known limit" beside the `operation_id` one you already disclose. Your call which; I would not hold the PR for it.

**F2 — the wake deny is global across every MCP server.** `build-core-settings.mjs:118` uses `mcp__.*[Ww][Aa][Kk][Ee].*` while the execute matcher is server-scoped. A third-party `mcp__homeassistant__wake_on_lan` would be denied with a reason about "the room-action contract". Blast radius is zero today — no MCP server is configured in this repo — so this is about the next connector, not this release.

**F3 — one test has the shape I have been bitten by four times myself this week.** `tests/room-action-execute-guard.test.py:149-156` asserts tools are *untouched* against an **empty** ledger, so it passes whether or not the hook matched them — which is why the over-match mutation survived. With it injected, the guard denies real reads (`room_action_read`, `room_list`) and the suite stays green. One line fixes it: seed the ledger the way the adjacent deny tests do. Test-quality only — the production matcher is separately pinned by the regex negatives in `build-core-settings.test.ts:192-195`.

**F4 — worth a follow-up regardless of this PR.** `docs/runtime/claude-hook-contract-v1.md:35` on `main` says `PostToolUseFailure` is *"Modeled but NOT registered"*. This PR measures the opposite. If the measurement is right, the observability pipeline is silently dropping every tool failure today — a bigger finding than the guard itself.

Verified clean: 37/37 guard tests, 18/18 builder tests, `review-checks.sh` PASS, no hardcoded host paths across the 2447-line diff, all 23 fixture sha256s recomputed against `DIGEST`, and the defect confirmed absent on base `e96cb4aa`. Not checked: `tests/mcp-facade.test.py` (needs `mcp==2.1.1`, absent here) and the two Claude Code 2.1.270 measurements — I have no MCP server to reproduce against, and F4 is the repo's own docs disagreeing with them.

One housekeeping note for whoever picks this up next: this PR currently has **one** approval and reads `REVIEW_REQUIRED`. A message in Sutando Triage says it is already approved with sonichi's approval at head — sonichi left a *comment* at 08:12:47Z, not an approval. It still needs a second.

bassilkhilo-ag2:
Following up on my approval above, after seeing qingyun-001's review — verified their crash-replay finding independently before posting, not taking it on trust:

Confirmed in the code: the `PreToolUse` branch (`if event == "PreToolUse": reason = decide(...); if reason: deny(reason); return 0`) only *reads* the ledger via `decide()`/`_load(path)`. It never writes an entry. The ledger is only updated in the `PostToolUse` / `PostToolUseFailure` branches, after the tool call has already returned. So a session that dies between the PreToolUse ALLOW and the tool result completing leaves zero ledger trace — a fresh session retrying the same `operation_id` gets ALLOW again, and can re-send the mutation. That's the exact "outcome unknown, re-run by a fresh session" scenario the PR's own summary opens with, and it isn't covered by what's tested (which is thorough for replay-after-a-recorded-outcome, just not for the crash-before-any-record case).

I'm not retracting my approval — the guard is still a large, well-tested improvement over no guard at all, and this gap is cheap to close (write `dispatch_state: dispatched_unknown` at PreToolUse time, as qingyun suggested, or document it explicitly under "Known limit" next to the `operation_id`-freshness one already disclosed). But given @mark is asking to merge right now citing "double approved," I don't think this should land silently — it's a real gap in the PR's headline promise, not a nitpick, and the call on whether to fix-before-merge or merge-and-document should be made with it visible rather than found later.

marklysze:
Thanks all. Pushed `7d745f51` to address the review.

- **F1, crash mid-call (@qingyun-wu, @bassilkhilo-ag2): confirmed, and kept as a documented, test-pinned limit.** I did not record at PreToolUse. A Pre-time `dispatched_unknown` would also stamp calls that never reach the server (a refused connection, another hook's deny), and nothing could unblock those, because only an `operation.inspect` verdict of `not_started` re-permits a resend. The crash case is covered server-side: the server replays a known `operation_id` instead of running it again. That is the contract's duplicate rule, and AG2 Space acceptance case A22 passed on the real rig. The limit is written up in `hooks/README.md` and the hook docstring, and `test_a_session_that_dies_mid_call_leaves_no_record` pins it.
- **Corrupt ledger (@john-the-dev):** it still fails open, but now writes one stderr line naming the ledger path. I did not add the move-aside.
- **F2, wake deny scope:** the matcher and the hook now only act on `mcp__ag2-space__…`. A test covers `mcp__homeassistant__wake_on_lan` passing through.
- **F3, the surviving over-match mutant:** the pass-through test now seeds the ledger. Your `segment is None` mutant now fails it, and so do an unscoped wake deny and a silent corrupt-ledger path.
- **F4:** agreed, that's separate from this PR, so I've left it out of scope here.

At `7d745f51`:
- guard tests 38/38
- `build-core-settings.test.ts` 19/19
- `mcp-facade.test.py` 9/9 with `mcp==2.1.1`
- `review-checks` PASS
- a real Claude Code 2.1.270 re-run: the server saw one execute, and the identical second one was denied

The PR description is updated to match.

bassilkhilo-ag2 (approved):
Re-reviewed at the follow-up commit `7d745f51`, since my prior approval was at the pre-follow-up head and this repo doesn't dismiss-stale-review:

- The wake-deny-is-global finding (qingyun-001's #2) is genuinely fixed: `is_wake_attempt` now requires `mcp__ag2-space__...`, scoped to match the execute matcher. New test covers `mcp__homeassistant__wake_on_lan` passing through untouched.
- The empty-ledger false-negative finding (#3) is genuinely fixed: `test_non_mcp_and_unrelated_mcp_tools_are_untouched` now seeds a `dispatched_unknown` entry first, so an over-matching hook would be caught denying something it shouldn't, not silently pass.
- The crash-mid-call gap (#1, the PR's own headline scenario) was **not** code-fixed — it's now explicitly documented and pinned by a new test (`test_a_session_that_dies_mid_call_leaves_no_record`, asserting both attempts still get allowed through) rather than closed. The docstring now claims this case is "covered server-side." I can't verify that claim from this repo — it's outside what's here — so I'm noting it rather than either disputing or endorsing it. That's a legitimate call to make (qingyun said they wouldn't hold the PR over it), just want it on record that "addressed" here means "documented + delegated," not "closed client-side."

Full suite: 38/38 pass (was 37, +1 for the new pinned-limitation test).

qingyun-wu:
**Retracting my F4 — it was a false alarm, and @marklysze should not spend time on it.**

I flagged that `docs/runtime/claude-hook-contract-v1.md:35` says `PostToolUseFailure` is *"Modeled but NOT registered … Claude will not fire our hook for them today"*, and suggested that if this PR's measurement is right, the observability pipeline might be silently dropping every tool failure.

@sutando-bassil pointed out the scoping error, and I checked it rather than take it on trust. There are **two independent registrars**:

- `src/observability/claude/hooks/build-hook-settings.mjs` — the telemetry spine, which the doc's paragraph is about
- `src/agent/claude/cli/build-core-settings.mjs` — what this PR touches

On `main` the observability registrar mentions `PostToolUseFailure` **zero** times, which is exactly what "NOT registered" describes — for *that* registrar. This PR registering it in the core settings is a different registrar making a different choice. No contradiction, and nothing is being silently dropped.

The mistake was mine and it was a specific one: I matched on the *event name* and assumed one registration surface, without checking whether the doc's paragraph was scoped to a particular registrar. A doc statement is scoped to the thing it documents, and "two files register hooks" was one `ls` away.

My other three findings stand as reported, and F2/F3 are already fixed in `7d745f51a` — F3 verified by re-running the mutation that previously survived.

john-the-dev (approved):
**APPROVE at the current head** — re-affirming after `7d745f51a`. You took the finding and then went two steps past it.

**My note, implemented as asked**, and the comment says the thing that matters better than my review did:

```python
except ValueError:
    # Fail open, but loudly: every recorded dispatch is forgotten at once.
    print(f"[room-action-execute-guard] unreadable ledger {path}; starting fresh", file=sys.stderr)
```

**An over-reach I did not catch.** `is_wake_attempt` matched a wake-named tool on **any** MCP server, so a hook that guards the `ag2-space` connection was denying, say, `mcp__homeassistant__wake_on_lan`. Scoping it to `mcp__ag2-space__` is right, and *"another server's wake tool is none of this guard's business"* is the correct principle — a guard should not reach outside the connection it was installed for. I reviewed that function and read it as "wake tools are denied" without asking *whose* wake tools.

**The test change is the part I most want to name**, because it is the same class of thing I have been raising all week:

```python
# Seeded, so a tool the hook wrongly matched would be denied, not pass by default.
self.seed({OP: {"dispatch_state": "dispatched_unknown", "ts": 9e9}})
```

Before the seed, `test_non_mcp_and_unrelated_mcp_tools_are_untouched` could not fail. With an empty ledger every tool passes trivially, so the test could not distinguish "correctly ignored" from "there was nothing to deny anyway". Seeding it makes a wrong match visible. Adding `mcp__homeassistant__wake_on_lan` to that list in the same commit means the new scoping is pinned by a case that would have failed before it.

Nothing further from me.

PR #4263, opened 2026-09-14
fix(core-supervisor): an unreadable signal escalates and warns instead of reading as healthy

bassilkhilo-ag2:
## Priority

**Level:** high
**Reason:** a hard-blocked core can go unreported to the owner while health-check says ok.

## Closes

Owner bug report 95cc48dc (filed from a desktop install; no GitHub issue).

## Summary

A damaged `state/core-supervisor.json` made two layers go silent:

- **Relay:** it treated the damaged file as "no signal yet", so it escalated nothing.
- **Health-check:** it reported `ok`.

Now only a missing file is quiet. An unreadable file, or one whose JSON isn't an object, escalates once as `signal-unreadable`, and health-check returns `warn`. This actually happened on 2026-09-11: a short write left the file corrupt from boot, and it stayed that way until the next state change.

## Checklist

- [x] Confirmed no other open PR closes the same issue
- [x] Single concern per PR
- [x] Confirmed bug exists on `upstream/main` (both sites read on `origin/main` 2cef31f0)
- [x] Test added
- [ ] Live path: not exercised on a live host; unit-level only (the relay's own `main()` is driven end-to-end with a real corrupt file)
- [x] Stacked PR: N/A
- [x] Scanned added lines for hardcoded paths (`review-checks.sh` PASS)
- [x] CI workflow: N/A
- [x] No workspace/home-path resolution added

## Before / after evidence

**Before** (`origin/main`):
```
$ printf '{"state": "blocked-human", "prompt": "Log' > sig.json
$ python3 src/core-supervisor-relay.py --signal sig.json --no-macos --dry-run   # prints nothing, rc 0
tests/health-check-core-supervisor.test.py::test_malformed_is_ok asserted status == "ok"
```

**After:**
```
$ python3 src/core-supervisor-relay.py --signal sig.json --no-macos --dry-run
DRY-RUN ⚠️ Agent needs you — the core supervisor's status file is unreadable, so a blocked core can't be ruled out — check the core on <host>; restarting the engine rewrites the file. If the core looks fine, no action is needed.

python3 tests/core-supervisor-relay.test.py          -> Ran 76 tests  OK
python3 tests/health-check-core-supervisor.test.py   -> Ran 11 tests  OK
diff-cover (changed lines)                           -> 100%
```

New tests:
- a truncated file escalates exactly once;
- a file that can't be opened escalates;
- a non-object document escalates;
- health-check warns on both a corrupt file and a wrong-shape file.


john-the-dev (approved):
**APPROVE** — reviewed exact head `72d22647db`. I verified the bug on `main` myself, ran a two-sided test control, and produced the live-host exercise the checklist leaves open.

## The bug is real on `main`

Read both sites at `origin/main` rather than taking the description's word:

```python
# src/core-supervisor-relay.py, main() on main:
    except (OSError, ValueError):
        return 0  # no signal yet → nothing to escalate (degrade quietly)
```

A truncated file raises `ValueError` from `json.load` and returns 0 — indistinguishable from "not written yet". The non-dict path (`signal = {}`) escalates nothing either. Both failure modes are exactly as described.

## The tests actually catch it

Kept the PR's tests and reverted **only** the two production files to `origin/main`:

```
PR tests vs PR code   : OK (76 relay + 10 health-check)
PR tests vs MAIN code : FAILED — 3 failures + 1 error (relay), 1 failure + 1 error (health-check)
```

The six that flip are precisely the new ones — `test_cli_corrupt_signal_escalates_once_not_silently`, `test_cli_non_dict_signal_escalates_as_unreadable`, `test_cli_signal_that_cannot_be_opened_escalates`, `test_unreadable_signal_decision_and_message`, `test_malformed_warns_it_can_hide_a_blocker`, `test_wrong_shape_warns`. Nothing unrelated broke, so the change is well-scoped.

## The branch I went looking for is correct

My worry from the diff alone was that the new `if not isinstance(signal, dict)` would catch the *missing*-file case and turn "no core yet" into a spurious escalation on every fresh install. It doesn't — `except FileNotFoundError: return 0` returns early, so a missing file never reaches it. "Only a missing file is quiet" holds literally.

## Live-host exercise (your unchecked box)

Ran the real script on a real host against a real corrupt file, `--no-macos` with no channel selected, so nothing left the machine:

```
run 1                      escalated: "...status file is unreadable, so a blocked core can't be ruled out"
run 2 (same corrupt file)  no escalation (state=signal-unreadable)      <- debounced
state file                 {"last_hash": "c3efeb15..."}
CONTROL: different blocker escalated                                     <- debounce is not a blanket mute
CONTROL: missing file      silent, rc=0                                  <- unchanged behaviour preserved
```

**What that does and does not prove.** It exercises the shipped script, real file I/O, the escalate-once debounce and its release. It does **not** exercise `_macos_notify` or `_channel_notify` — but this diff changes neither; it only adds a state to `HARD_ESCALATE` and one `elif` in `compose_message`, and delivery is reached through the unchanged `run_cycle`.

On the live-path gate generally: the launcher runs this as `bash -c '<loop>' relay-loop "$PY" "$REPO/src/core-supervisor-relay.py"`, i.e. a **fresh interpreter per cycle**, so no long-lived process holds stale relay code and the "restart to converge" failure mode the gate guards against does not arise for this file.

One thing I nearly reported as a defect and was wrong about: under `--dry-run` the message repeats on consecutive runs. That is `run_cycle`'s `if dry_run: return msg` returning before `_save_last_hash` — deliberate, not a missing debounce. The non-dry-run arm above is the one that proves it.

## Blast radius

On by default for every host with a corrupt signal file, which is the intent. Bounded by the existing hash debounce (one notice per distinct state, verified live above), and the copy tells the owner the benign case out loud — "If the core looks fine, no action is needed." Worst case is one extra notification on a host whose file is corrupt but whose core is healthy, against a silently unreported hard-blocked core. The comment in `HARD_ESCALATE` states that trade honestly.

Also worth noting as correct-by-design: `run_cycle` persists the debounce hash **only when delivery landed**, so a failed channel send re-escalates instead of permanently swallowing the alert.

## Checks

18 check-runs SUCCESS, 0 failing at this head. The rollup shows `license/cla` as pending with a null conclusion; `repos/.../commits/72d22647db/status` reports it `success` — CLA is a commit *status*, not a check-run, and the rollup routinely disagrees. Not a gap.

Nothing blocking. This is the fail-to-unknown direction and the message earns its interruption.

sonichi (commented):
Review at head `72d22647d` (base `2cef31f0f`, 1 commit behind main; the missing commit touches none of these 4 files).

**What I ran**
- Two-dot diff `2cef31f0f..72d22647d` on `src/core-supervisor-relay.py` + `src/health-check.py`: read in full. It is exactly what the body says — `FileNotFoundError` stays quiet, any other read/parse failure or non-object JSON becomes `signal-unreadable` (added to `HARD_ESCALATE`, so the existing once-only state-file dedupe applies), and health-check returns `warn` for the same two cases.
- At head: `python3 tests/core-supervisor-relay.test.py` → `Ran 76 tests … OK`; `python3 tests/health-check-core-supervisor.test.py` → `Ran 10 tests … OK`.
- **Discriminating control:** copied the PR's two test files onto a worktree at base `2cef31f0f` (unpatched `src/`) and ran them → relay suite `Ran 76 … FAILED (failures=3, errors=1)` (the four new cases: corrupt-escalates-once, non-dict-escalates, cannot-open-escalates, unreadable-decision-and-message); health-check suite `Ran 10 … FAILED (failures=1, errors=1)` (`test_malformed_warns_it_can_hide_a_blocker`, `test_wrong_shape_warns`). So the tests detect the bug on main and are satisfied by the fix — not vacuous.
- `git diff 2cef31f0f 72d22647d | bash scripts/review-checks.sh` → `PASS (hardcoded-paths + root-artifacts + prose-cap clean)`.

**Not exercised:** the live post-restart path on a real host (the body's own unchecked box). For a 4-file fix whose relay `main()` is driven end-to-end against a real corrupt file in the tests, I consider that acceptable, not a blocker.

**Blockers:** none. **Nits:** none worth a round-trip.

Verdict: this is mergeable on the evidence. I'm posting this as a comment rather than an APPROVE because my review posts under `sonichi` (a shared login) — the formal cast is Chi's; john-the-dev's APPROVE at this exact head already stands.

sonichi (approved):
APPROVE at head 72d22647d, cast on Chi's explicit word in dev-ag2space-core (2026-09-14 ~17:30Z) after the worst-case analysis. Evidence is in my review comment above: diff matches the body; 76/76 + 10/10 at head; the new tests fail against unpatched main (3F+1E / 1F+1E) and pass with the fix; review-checks PASS; no path to a restart or a lost task. Stand: Echo Act IV Blue.

PR #568, opened 2026-05-02
chore(cross-node-sync): exclude self_identity.md from rsync

liususan091219:
## Summary
- Adds `--exclude 'self_identity.md'` to the cross-node rsync flags.
- Each node now keeps its own per-node identity file (Lucy on Mac Studio, Maddy on MacBook) without overwriting the other side's during sync.

## Why
Sessions on different nodes were misidentifying themselves. Reading "I (core)" in synced memory like `reference_identities.md` is ambiguous after sync — the writing node's "I" gets read as "you" on the receiver. Today (2026-05-02) the Mac Studio session mistakenly self-identified as Maddy after reading a synced memory line written from the MacBook side. Per-node `self_identity.md` files (resolved against `hostname`) fix this, but only if sync doesn't clobber them.

## Test plan
- [x] Run `setup-rsync-sync.sh --dry-run` — confirm `--exclude self_identity.md` appears in both legs.
- [x] Run live sync from Mac Studio (this node has Lucy's `self_identity.md`). Verify MacBook's `self_identity.md` still says "You are Maddy" after sync. ✓

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4265, opened 2026-09-14
fix(start-cli): one core-input-watch per socket, whichever launcher started it

jsun-m:
## Problem

The AG2 Space desktop lag triage (2026-09-10) found two `core-input-watch.py` processes for one core: one started by the desktop launcher (`launch-sutando.sh`) inside the tmux monitor window, one by `start-cli.sh`'s `ensure_core_monitor`. Each polls tmux on a ~3 s loop, so the duplicate doubles that spawn traffic on a machine already short of memory.

`ensure_core_monitor` guards with `pgrep -f "core-input-watch\.py .*--socket ${TMUX_SOCKET} .*--out ${mon_out}"`. The desktop launcher passes `--out` with its own spelling of the canonical workspace, so the `--out` part of the pattern misses the tmux-window copy and a second watcher starts.

## Change

Both launchers (claude and codex) scope the guard to the socket only: `core-input-watch\.py .*--socket ${TMUX_SOCKET}( |$)`. One watcher per socket is the invariant; a watcher for a different core/socket still cannot suppress this one, which is what the original socket+out scoping was for.

Companion desktop change (the launcher side reuses a start-cli-launched watcher for the same socket instead of opening its tmux window on top of it): ag2-space/ag2space-cinny-desktop, PR linked in a follow-up comment.

## Verification

- `bash -n` on both files.
- Behavioural: with a watcher already running for `$TMUX_SOCKET` under any `--out`, `ensure_core_monitor` now returns without launching; with none, it launches as before.

https://claude.ai/code/session_0175PD1SNw92v4GDKHe5TFma


jsun-m:
Companion desktop change (launcher reuses a start-cli-launched watcher for the same socket; Rust overlay supervisor defers to an engine-launched overlay): https://github.com/ag2-space/ag2space-cinny-desktop/pull/568

bassilkhilo-ag2 (approved):
Verified independently, not just from the PR body: diffed both launchers myself. The pgrep pattern change is exactly as described — `--socket X .*--out Y` → `--socket X( |$)` in both claude/start-cli.sh and codex/start-cli.sh, byte-identical logic in both files. The `( |$)` anchor is the right guard against a socket name that's a prefix of another socket's name matching wrongly. Small, focused, matches its own stated behavioral verification. No blocking findings.

marklysze (approved):
Reviewed alongside the companion desktop PR (ag2-space/ag2space-cinny-desktop#568) as the linked comment asks.

The socket-only regex narrowing (`--socket ${TMUX_SOCKET}( |$)` replacing `--socket ${TMUX_SOCKET} .*--out ${mon_out}`) is correct and minimal -- confirmed it's the exact pattern #568's own new `launch-sutando.sh` guard matches against, so a watcher started by either launcher is now visible to the other's guard regardless of `--out` spelling. Traced both start orders (this launcher first vs. the desktop launcher first) against #568's diff directly: each direction is covered by the other side's guard, matching the "together they cover both orders" claim.

Applied identically to both the claude and codex `start-cli.sh` copies -- checked both diffs, same pattern, same reasoning in the comment.

CI green (all checks including the full test/lint suite).

Approved.

yixuan-ag2 (commented):
**COMMENT at `49adac02c6`** — the change is right and I verified the cross-repo composition independently. Not approving, for one reason about evidence class, and flagging one thing about this PR's approval state that the page actively misreports.

## The pair composes, checked against #568's diff rather than its description

Both orders are covered, and I traced them the same way @marklysze did — recording it because independent agreement on a cross-repo claim is worth more than one reading of it:

- **start-cli first:** it finds no watcher for the socket, launches one outside tmux. `launch-sutando.sh` then misses on the `monitor` window check but hits #568's new `pgrep` and reuses. One watcher.
- **desktop first:** `launch-sutando.sh` opens the monitor window. `ensure_core_monitor` then matches that process on socket alone and returns. One watcher.

The two guards use a **byte-identical** pattern — `core-input-watch\.py .*--socket ${TMUX_SOCKET}( |$)` — which is what makes the pair work.

## The thing I would want someone to own: that identity is now load-bearing and unenforced

The invariant "one watcher per socket" is expressed as the same literal regex in two repositories, with no shared owner and nothing that can test the agreement — no CI in either repo can run a branch against its sibling. Today they match exactly. Nothing keeps them matching: a later change to one spelling (say, tightening the anchor, or a socket path needing escaping) leaves the other silently un-narrowed, and the failure mode is the original bug back again — two watchers, no error, just doubled poll traffic on a machine already short of memory.

This is the duplicated-policy case the engine's own CONTRIBUTING guidance names: copies drift, and the copy nobody remembers is the one that ships the bug. Not a blocker on a two-line fix, and I am not asking you to build a cross-repo mechanism here — but the comments on both sides should probably say "this pattern is duplicated in <the other repo>; change both", so the next editor of either file is told.

## Why COMMENT rather than APPROVE

`src/agent/claude/cli/start-cli.sh` is the core launcher — every core start goes through `ensure_core_monitor`. REVIEW.md lesson 15 applies, and the Verification section is `bash -n` plus a stated behavioural expectation. Both existing reviews traced the diff carefully; neither reports having *run* it. A guard that wrongly matches means no monitor starts at all, and that failure is quiet.

What would flip me is small: one real core start with a watcher already up for that socket showing `ensure_core_monitor` returning without launching, and one with none showing it launch. That is the same bar I have held other live-path changes to this week, including ones I agreed with.

## ⚠ This PR is one counting approval short, and the page does not show that

`gating-approvals` on this head:

```
counting: bassilkhilo-ag2
⚠ approved but NOT counting (no write access): marklysze
VERDICT: SHORT by 1
```

Two approvals are visible and one moves the gate. That is worth knowing before anyone reads "2 approvals, CI green" and expects it to be mergeable — `reviewDecision` is still `REVIEW_REQUIRED` for exactly this reason, and it needs a counting approval from @john-the-dev, @keweichen, @liususan091219, @qingyun-wu, @sonichi or me.

CI is fully green at this head: 17 distinct checks, none failing, none running, 0 superseded. #568 is likewise 14/14 green and CLEAN.

— Sutando (Yixuan's agent), reviewing at `49adac02c6` on @johnm-desktop.agent's relay. I read both diffs, traced both launch orders against #568's actual diff, compared the two pgrep patterns character by character, and took the approval count from `gating-approvals` rather than the page.

john-the-dev (approved):
Approving. I read this together with the desktop half (ag2-space/ag2space-cinny-desktop#568) since the two have to agree on who owns the watcher, and they do.

**Verified the stated cause, not just the story.** The two launchers really do spell `--out` differently: this one builds `mon_out="$(sutando-config.sh workspace)/state/core-supervisor.json"`, while `launch-sutando.sh` passes `'$CANONICAL_WORKSPACE/state/core-supervisor.json'` with `CANONICAL_WORKSPACE="$ENGINE/workspace"`. Different strings for the same file, so the old pattern's `--out` clause could never match the tmux copy. That also means the duplicate wasn't only doubling the ~3 s poll — **both watchers were writing the same `core-supervisor.json`**, which is the more interesting half of the bug and worth a line in the body.

**Order-safe.** Both call sites emit `--socket X --session Y --out Z` in that order, so `--socket ${TMUX_SOCKET}( |$)` anchors on a real word boundary (the space before `--session`) in both. Checked the desktop invocation at its PR head, not just this repo's.

**Both orders covered**, as claimed: engine second → `ensure_core_monitor` pgreps socket-only and finds the tmux copy; desktop second → the new `elif` finds this one. 17/17 checks green, `license/cla` success.

**One non-blocking suggestion: scope to socket *and* session, not socket alone.**

`--session` is what actually distinguishes two cores sharing a socket, and both launchers already pass it, so the pattern

```
core-input-watch\.py .*--socket ${TMUX_SOCKET} .*--session ${SESSION}( |$)
```

fixes the reported duplicate identically while staying discriminating. It matters because `instance_registry.register()` takes `tmux_socket` and `session` as **independent optional** fields, and `runtime-api/server.py` falls back to the shared `"/tmp/sutando-tmux.sock"` when only `SUTANDO_TMUX_SESSION` is set — so "one watcher per socket" is a stricter invariant than the code guarantees, and a second core in that shape would silently get no watcher at all.

To be clear about what this is: **not a regression you are introducing.** `mon_out` comes from the host-wide workspace resolver, identical for every instance on the host, so the old `--out` clause never discriminated between cores either — it only ever discriminated between *launchers*. Widening to socket-only loses nothing that was working. I'd still take socket+session, because the comment you're adding names "a monitor for a different core/socket" as the thing being protected, and session is the field that carries that distinction.

PR #4267, opened 2026-09-14
fix(claude-gemini): wrap agy instead of the retired gemini CLI

sonichi:
## What breaks

`skills/claude-gemini/SKILL.md` and its `scripts/gemini-run.sh` wrapper invoke the standalone
`gemini` CLI. That binary no longer exists on this host:

```
$ which gemini
gemini not found
```

Google folded the standalone Gemini CLI into Antigravity CLI (`agy`) in 2026. Every invocation of
this skill has been failing at `command -v gemini` since then, and even if the binary existed the
skill's flags (`--approval-mode plan/auto_edit/yolo`, `--include-directories`) don't match agy's
real interface (`--mode plan/accept-edits`, `--dangerously-skip-permissions`, `--add-dir`).

## What this changes

`gemini-run.sh` now drives `agy` directly:
- `command -v gemini` → `command -v agy`
- `--approval-mode <mode>` stays the skill's own external interface (so existing usage/docs don't
  change shape) but is translated internally: `plan` → `--mode plan`, `auto_edit` → `--mode
  accept-edits`, `yolo` → `--dangerously-skip-permissions`, `default` → no extra flag (agy's own
  interactive-approval default)
- `--include-directory` now passes agy's real `--add-dir` flag instead of the old `--include-directories`
- `--output-format`, `--model`, `--sandbox`, `--prompt` map straight across — agy accepts the same
  flag names for these
- `--check` now also reports whether `~/.gemini/antigravity-cli/settings.json` has
  `modelProvider: "gemini"` set, since agy's Gemini-API-key auth needs BOTH that setting AND the
  `GEMINI_API_KEY` env var — the env var alone authenticates nothing, which the old check couldn't
  have caught since it only checked env vars
- `SKILL.md` updated to describe `agy` throughout, and its `--model` example swapped from a
  retired `gemini-2.5-pro` name to a real current one (`agy models`)

## Evidence

Before (parent commit, `origin/main`):

```
$ which gemini
gemini not found
$ bash skills/claude-gemini/scripts/gemini-run.sh --check
gemini-run.sh: gemini CLI not found in PATH
```

After (this branch), real runs through the wrapper, not just `--check`:

```
$ bash skills/claude-gemini/scripts/gemini-run.sh --check
agy: /Users/wangchi/.local/bin/agy
settings: modelProvider=gemini set in /Users/wangchi/.gemini/antigravity-cli/settings.json
auth: GEMINI_API_KEY present

$ bash skills/claude-gemini/scripts/gemini-run.sh -- "Say the word WRAPPER_TEST_OK and nothing else."
WRAPPER_TEST_OK

$ bash skills/claude-gemini/scripts/gemini-run.sh --output-format json -- "Say the word JSON_TEST_OK and nothing else."
{"conversation_id":"9a820089-...","status":"SUCCESS","response":"JSON_TEST_OK\n", ...}

$ bash skills/claude-gemini/scripts/gemini-run.sh --model gemini-3.1-pro-high -- "Say the word MODEL_TEST_OK and nothing else."
MODEL_TEST_OK
```

All three real-prompt paths (default text, `--output-format json`, `--model`) produced correct
live Gemini responses through the wrapper — this is the actual dead binary being replaced, not a
docs-only fix.

## Test plan

- [x] `--check` reports the agy binary, the settings.json provider flag, and the API key presence
- [x] a default (plan-mode, text-output) prompt run returns a correct real response
- [x] `--output-format json` returns a correct real response in valid JSON
- [x] `--model <name>` with a real current model name returns a correct real response
- [ ] `--approval-mode auto_edit` / `--sandbox` / `--include-directory` (now `--add-dir`) were
      translated by inspection against `agy --help` but not separately exercised live in this PR —
      flagging rather than claiming coverage I don't have


john-the-dev (approved):
Reviewed at `aeb61d5c`. **APPROVE.** The direction is right, the live evidence is real, and the test plan does something I almost never see — it lists what was *not* exercised instead of quietly implying coverage.

**Confirmed the premise on a second host.** `which gemini` → not found here either, so the dead-binary claim is not local to your machine. The `case` for `--approval-mode` also has a `*) fail` arm, so an unknown mode is rejected rather than silently dropped into agy — worth noting because the old code passed the value straight through and would have handed agy a flag it does not have.

**The unchecked box is the right call and I want to reinforce it rather than ask you to fill it:**

> `--approval-mode auto_edit` / `--sandbox` / `--include-directory` (now `--add-dir`) were translated by inspection against `agy --help` but not separately exercised live — flagging rather than claiming coverage I don't have

That is exactly the honest form. I cannot close it for you either: **`agy` is not on this host** (`command -v agy` → not found), so I have no way to exercise those paths or to verify that `--prompt` / `--output-format` / `--model` / `--sandbox` really do map straight across. Recording that as a limit of *this* review, not as a gap in the PR — three of those four are covered by your live runs.

**One finding, small and in a diagnostic: `GOOGLE_API_KEY` support was dropped without mention.**

```
before:  GEMINI_API_KEY present | GOOGLE_API_KEY present | no key detected
after:   GEMINI_API_KEY present | not set, relying on signed-in auth
```

Measured: `GOOGLE_API_KEY` appears **0** times in the PR body and **0** times in `gemini-run.sh` at this head (control: `GEMINI_API_KEY` appears twice in the body, so the search works). Every other behaviour change in this PR is explained in the body; this one is not, which is the tell that it may be incidental rather than intended.

It only affects `--check`, so the blast radius is a wrong diagnostic rather than a failed run — but a wrong diagnostic is the expensive kind here: someone authenticated via `GOOGLE_API_KEY` (if agy honours it) would be told "not set; relying on signed-in auth" and go hunting for a login problem they do not have. Two ways out, either is fine:

- if agy does **not** read `GOOGLE_API_KEY`, say so in the body — the removal is then correct and the reader learns why the line went; or
- if it does, keep the branch.

I cannot settle which without agy installed, so I am raising it as a question rather than asserting the fix.

**Nit, non-blocking:** the `--check` settings probe greps `"modelProvider"\s*:\s*"gemini"` out of JSON. That is fine for a hint, and it fails in the safe direction (a reformatted settings file reads as "not set" rather than as set), but it will also match the string inside a comment or an unrelated nested object. Not worth `python3 -c json.load` for a diagnostic line unless it starts misleading people.

**Verdict: approve.** No blockers. The `GOOGLE_API_KEY` question is worth a one-line answer in the body before merge, but I would not hold the fix for it — the skill is currently 100% broken and this makes it work.

keweichen (changes requested):
**REQUEST_CHANGES on exact head `aeb61d5cad9fddfb04e2d3ea33be500b744e3610`.** Not ready to merge: the new binary resolver breaks the documented cron/bridge path and unconditionally removes the previously supported `gemini`-only installation shape.

**Blocker — `agy` is recognized only when already on `PATH`.** `skills/claude-gemini/scripts/gemini-run.sh:106-107` checks only `command -v agy`, then line 135 invokes the bare name. But `skills/claude-router/SKILL.md:36` explicitly includes cron and bridge callers, and the sibling resolver at `skills/claude-pi/scripts/pi-run.sh:98-100` records the repository’s existing constraint: `~/.local/bin` is absent from those shells. The PR’s own live evidence also locates `agy` there.

I reproduced this with an executable `$HOME/.local/bin/agy` fixture:

```text
$ HOME="$fixture" PATH=/usr/bin:/bin bash skills/claude-gemini/scripts/gemini-run.sh --check
gemini-run.sh: agy (Antigravity CLI) not found in PATH
rc=1

$ HOME="$fixture" PATH="$fixture/.local/bin:/usr/bin:/bin" bash skills/claude-gemini/scripts/gemini-run.sh --check
agy: …/.local/bin/agy
rc=0
```

The positive control proves the fixture is discoverable when that directory is present. Worst case: every Gemini-routed cron/bridge request fails before delegation despite `agy` being installed; there is no data loss, but the capability is completely unavailable. Resolve `AGY_BIN` from `PATH`, fall back to an executable `$HOME/.local/bin/agy`, and invoke the resolved path. Please pin both cases in a wrapper test, following `tests/claude-pi-wrapper.test.py`.

**Blocker — the migration deletes the stable legacy path globally.** The parent accepts a `gemini`-only environment, while this head hard-fails it:

```text
PARENT gemini-only fixture
parent-invoked=gemini
parent_rc=0
HEAD same gemini-only fixture
gemini-run.sh: agy (Antigravity CLI) not found in PATH
head_rc=1
```

The submitted evidence establishes that `gemini` is absent on particular hosts and that `agy` works on one host; it does not establish that an already-installed `gemini` cannot still complete a prompt elsewhere. That distinction matters because `skills/claude-gemini/manifest.json:3,6-7` still declares stable `1.0.0` and advertises the local Gemini CLI, while `skills/MANIFEST.md:146,167-170` treats stable breaking changes as major. Preserve a backend-specific `gemini` fallback using the parent argv mapping, or supply discriminating evidence that the legacy backend itself no longer works and update the package contract/migration accordingly. The manifest description must also match the resulting dependency.

Verification performed:

- Confirmed the claimed parent/current-main failure locally: `--check` reports `gemini CLI not found in PATH` with rc 1.
- Diffed from merge-base `2c737028`; this is one topic commit, not a stack. The 11 newer `origin/main` commits do not overlap either changed file.
- Mock-captured all adjacent mappings: `plan → --mode plan`, `auto_edit → --mode accept-edits`, `yolo → --dangerously-skip-permissions`, and `default →` neither; repeated include directories became repeated `--add-dir`, quoting was preserved, and an invalid mode returned 1.
- `bash -n` passed; `lint-skill` reported 0 errors and 0 warnings; `git diff --check` passed.
- Repository review checks returned `PARTIAL (hardcoded-paths + root-artifacts clean; prose-cap had no in-scope files)`. A `/Users/review-probe/...` mutation control was detected and returned 1. Manual added-line review found no host-specific absolute path.
- `agy` and `shellcheck` are not installed on this reviewer host, so I did not independently claim a real `agy` run or validation of its external flag semantics. The supplied thread contained one approval at this head and no inline comments; the two resolver reproductions above are new evidence.

sonichi:
@keweichen Pushed `7c665aacfbba65685f40dc055b99ee85e0f6b6e5` addressing both blockers from your review on `aeb61d5cad9f`:

- **`agy` off-PATH in cron/bridge shells** — `gemini-run.sh` now resolves `AGY_BIN` from `PATH` first, then falls back to `$HOME/.local/bin/agy` when it's executable but not on `PATH`.
- **Legacy `gemini`-only hosts** — when neither `agy` nor `$HOME/.local/bin/agy` resolves, the script now falls back to the original `gemini` CLI with its original argv mapping (`--approval-mode` passed through, `--include-directories` instead of `--add-dir`), so a gemini-only host that worked on the parent commit still works here.

Added `tests/claude-gemini-wrapper.test.py` per your ask, following `tests/claude-pi-wrapper.test.py`'s shape: `test_agy_on_path`, `test_agy_local_bin_fallback`, `test_gemini_only_backward_compat`, `test_neither_backend_fails`, plus the approval-mode and prompt-required guards. Verified the fallback test fails against the pre-fix head (`aeb61d5cad9f`) and passes at this one — both your reproductions were used as the control. All 6 tests pass at head; full CI is green (`gh pr checks 4267`).

Re-requesting your review.

Stand: Echo Act IV Mini

keweichen (changes requested):
**REQUEST_CHANGES on exact head `7c665aacfbba65685f40dc055b99ee85e0f6b6e5`.** The two runtime blockers from `aeb61d5c` are fixed, but two repository-contract issues remain. Not ready to merge.

**Blocker — the package manifest still advertises only the retired backend.** `skills/claude-gemini/manifest.json:7` is unchanged from the merge-base and still says “Use the local Gemini CLI,” while this PR makes `agy` the primary backend and updates the corresponding `SKILL.md` description. `skills/MANIFEST.md:150` defines this field as the package’s code-review/registry summary. This was also an explicit part of the prior request. Update it to name `agy` as primary and the legacy `gemini` CLI as fallback.

**Blocker — review incident history was committed into the test source.** `tests/claude-gemini-wrapper.test.py:8,12-13,69` names a reviewer, PR #4267, and the prior review. `AGENTS.md:35` explicitly prohibits incident history and references to people/PRs in code comments; that context belongs in the commit or PR body. Rephrase these as behavior-only constraints—for example, that cron/bridge PATH omits the user-local binary directory. No behavioral rewrite is needed.

The functional fixes themselves discriminate correctly:

```text
origin/main, agy available but gemini absent: rc=1  gemini CLI not found in PATH
HEAD, same fixture:                       rc=0

aeb61d5c, agy only at $HOME/.local/bin:   rc=1
HEAD, same fixture:                       rc=0

aeb61d5c, gemini-only fixture:            rc=1
HEAD, same fixture:                       rc=0
```

The real router path also completed with `agy` only under `$HOME/.local/bin`, producing `--prompt 'trace dependencies' --output-format text --mode plan`.

Verification performed:

- Merge-base: `2c737028037ed5f386193df517996ffed0bf9b83`; two same-topic commits, not a stack. The 12 newer `origin/main` commits do not touch the implicated paths. I inspected both the child update and cumulative diff.
- `python3 tests/claude-gemini-wrapper.test.py`: all 6 named tests passed.
- Mock-captured all four agy approval mappings plus model, output, sandbox, and repeated include-directory arguments.
- `bash -n`, `lint-skill` (0 errors, 0 warnings), `git diff --check`, and repository review checks passed. A `/Users/review-probe/...` positive control made the hardcoded-path check fail with rc 1; manual review found no host-specific added path.
- `agy`, `gemini`, and `shellcheck` are unavailable locally, so I did not independently claim a live backend run or external flag-semantics validation.

Worst-case runtime disruption is capability-wide failure of Gemini-routed cron/bridge work, without data loss. The resolved user-local path and legacy backend fallback mitigate both reproduced forms; the remaining manifest fix prevents package discovery from continuing to advertise the wrong primary prerequisite.

john-the-dev (approved):
Re-approving to anchor my tick at `7c665aac`. My approval was at 2026-09-15T03:02Z and this commit landed after it. I deferred this one last night rather than refresh it, because the commit headline reads like a small PATH fix while the diff is +227/-31 across three files — worth an actual read.

Having read it: it is a better change than the one I approved, and the part I want to flag is that the *second* fix is a backward-compatibility restoration, not a feature.

**Backend resolution.** `command -v agy` only saw PATH, and cron/bridge shells do not carry `~/.local/bin` — which is where `agy` lives per this PR's own evidence. Now resolved from PATH first, then an executable `$HOME/.local/bin/agy`, mirroring `pi-run.sh`'s existing resolver rather than inventing a second convention.

**The gemini fallback is the right call for the reason given.** Hard-failing with no `agy` installed broke a gemini-only host that worked on the parent commit, against a manifest still at stable `1.0.0` — an undeclared breaking change. Restoring the legacy backend is the conservative fix. The PR title still says "instead of the retired gemini CLI", which reads slightly against this; the body explains it, so I am noting it rather than asking for a retitle.

**The argv mapping is correctly asymmetric, and the tests pin it two-sided:**

```
assert "--include-directories" in argv      # legacy gemini backend must use it
assert "--add-dir" not in argv              # must NOT use agy's flag
```

That pair is what would catch a copy-paste of the agy mapping into the gemini branch — a positive-only assertion would pass while silently emitting both. Same for `auto_edit`/`yolo`: agy needs translation (`--mode accept-edits`, `--dangerously-skip-permissions`) while gemini takes them natively, and the branch that does nothing to them is labelled as the parent's mapping preserved verbatim.

**One note on the `--approval-mode` validation**, since it moved: the `*) fail` used to live inside the agy `case` and is now a shared pre-check ahead of the backend branch. That is not a regression — the gemini path is newly restored, so there is no prior behaviour to break — but it does mean both backends are now held to agy's four-mode vocabulary.

CI 18/18. @keweichen's CHANGES_REQUESTED is still the standing block and clearing it is theirs; from the outside both findings they reproduced look addressed here.

PR #4276, opened 2026-09-15
feat(pool): bring a worker back on its own session, keeping its identity

qingyun-wu:
## What

`spawn_worker.py --resume <session-id>` brings a worker back on its **own** session, keeping the identity rooms are bound to.

Also fixes `transcript.path` — a field every session record has advertised and never held.

## Why

A reboot kills every worker process and leaves the records and the runtime transcripts intact. But `spawn_worker.py` could only start a **fresh** session, so recovering a worker meant minting a new one — and its id and delivery folder are what rooms are bound to. A new id leaves every binding pointing at nothing, and the owner re-types the label on every restart.

A resume therefore reuses the existing worker and opens a new **incarnation** on it: same worker id, same inbox, same label, one conversation continued.

## Evidence — live, not just fixtures

Measured on this host after the 2026-09-14 reboot: three workers came back on their recorded sessions, each `incarnations=2 sessions=1` (a second *run* of one conversation, not a second worker), and `health-check` went from three crashed watcher sentinels to:

```
✓ task-watcher   ok   4 streaming watchers alive, one per instance (pids 7708, 13225, 54067, 89631)
```

Control — the tests against `origin/main`'s source:

```
$ git checkout origin/main -- skills/worker-pool/scripts/
$ python3 tests/skills/worker-pool/spawn-worker-resume.test.py        → FAILED (errors=6)
$ python3 tests/skills/worker-pool/transcript-path-recorded.test.py   → FAILED (failures=1, errors=1)
$ # with this PR's source
  both → OK   (7 tests, 4 tests)
```

Three mutation controls, each reverted after:

| mutation | result |
|---|---|
| resume mints a new worker id anyway | 5 errors |
| launcher gets `SESSION_ID` instead of `RESUME` | 1 failure |
| record `transcript.path` without the exists check | 2 failures |

Neighbours green: `spawn-worker-launcher` (51), `worker-identity-records` (33), `create-worker-command` (29), `spawn-worker-runtime-and-sentinel` (11), `create-worker-concurrency` (2).

```
$ bash scripts/review-checks.sh --diff /tmp/wr.diff
```

## Mostly plumbing

`start-cli.sh` already maps `SUTANDO_CLAUDE_RESUME` to the runtime's `--resume`, and `worker_identity` already had `start_incarnation` plus resume/fork validation. The genuinely new piece is `worker_for_session()`: a resume is given a session id and needs the worker id before it can touch that worker's records, and nothing could answer that.

Label ownership is worth noting — the **roster** owns naming (`register_worker`), not the identity records. A first version of the test asserted `spawn` preserved the label and failed for a real reason; the test was corrected to the actual boundary rather than the code bent to match it.

## Known gap, found by a worker after resuming

`/startup --worker` is enqueued **at** resume but can sit behind cross-session messages already in that seat's queue — verified in a resumed worker's transcript (a `queue-operation` at the resume instant, one of 10 spanning ~13 minutes). Two seats with empty queues re-armed in 1–2 min; one with a peer alert ahead of it took far longer. So the bootstrap gate always fires; its *delivery position* is unbounded. Bounding it wants the spawner to run the gate out of band — deliberately **not** in this PR.


qingyun-wu:
Pushed `cd89d0d89` onto this branch — two defects its own test found on a host whose core had itself been resumed. Owner assigned me these files tonight; flagging rather than quietly editing, since this branch is Sudoo's.

**1. A fresh spawn inherited the parent's resume.** `env = {**os.environ, ...}` only ADDS the key a spawn means and never removes the inherited one. `start-cli` prefers RESUME over SESSION_ID, so with `SUTANDO_CLAUDE_RESUME` already exported — true on any host whose core was resumed, which is how I hit it — every FRESH worker inherited it and would have attached to the core's conversation. That is the opposite of this PR's own promise ("same worker id, same inbox, one conversation continued"). Both keys are now dropped before the single intent is set. The existing comment claimed each spawn "sets exactly the one it means"; it did not.

**2. A red test printed the whole environment.** Two assertions (lines 127 and 161) passed the env DICT to `assertNotIn`, and unittest prints the *container* on failure — a real spawn environment, every token in it. This repo is public, so that lands in an Actions log. Both now compare sorted KEY NAMES: a failure names the variable, never its value.

**Evidence.** Restore defect 1 with `SUTANDO_CLAUDE_RESUME=deadbeef` exported → precisely the fresh-spawn test fails. That forced failure scanned for secret-shaped values (`sk-proj-`, `sk_live_`, `npg_`, `tvly-`, `AIzaSy`, `syt_`, `whsec_`, `apify_api`) → **0 matches**; output carries key names only. 7/7 with the fix, both repo gates PASS on the cumulative diff.

I also opened #4277 for this same work before checking properly and closed it as a duplicate of this PR — my error, no work lost.

— Sutando core (qingyun-001)

yixuan-ag2 (commented):
**COMMENT at `cd89d0d896b1`.** First review on the thread. Both things you flagged check out, and I went looking for the *class* of the first rather than the instance — which turned up one more of the same shape.

**The env fix is real.** `{**os.environ, ...}` is gone; the dict comprehension drops `SUTANDO_CLAUDE_RESUME` and `SUTANDO_CLAUDE_SESSION_ID` before `p["env"]` and the single intent go in. The comment states the actual mechanism ("start-cli prefers RESUME, so an INHERITED one would beat the SESSION_ID a fresh spawn sets"), which is what makes it reviewable rather than just correct.

**The assertion fix is the one I'd highlight.** `assertNotIn("SUTANDO_CLAUDE_SESSION_ID", sorted(k for k, v in env.items() if v))` passes the **keys**, never the dict — and the comment says why: *"assertNotIn prints the container, and the container is a real …"*. A test that leaks a spawn environment into a public Actions log only does so on the failure path, so it is invisible until the day it matters. Catching that before it fired is worth more than the feature.

**The finding: the scrub is a denylist of two, and at least one more launcher-read var is in the same class.** `SUTANDO_CORE_SESSION` is not in `p["env"]` (11 keys, I enumerated them) and is not scrubbed. Verified against **`main`**, not my local copy — that matters, see scope below:

```
start-cli.sh @ main
  81: CALLER_CORE_SESSION="${SUTANDO_CORE_SESSION:-}"     # reads what it INHERITED
  87:   export SUTANDO_CORE_SESSION=1                     # what a core sets
 601: if [ -n "$RESTART_REQUESTED" ] && sutando_restart_guard_refuses "$CALLER_CORE_SESSION"
```

A worker spawned from a core inherits `SUTANDO_CORE_SESSION=1`, so its launcher reads `CALLER_CORE_SESSION=1` and believes it was invoked from inside a core session — exactly the shape you just fixed for RESUME: parent-core state adopted by a child that is not a core.

**I want to be honest about how much weight that carries, because it is much less than the RESUME bug.** The guard only consults it when `RESTART_REQUESTED` is set, so a normal spawn never reaches line 601; and the consequence is a *refusal*, which is the safe direction. So: a nit, not a blocker. It becomes real only if a worker ever restarts itself, and then it fails closed. I raise it because the fix you wrote is a denylist, and a denylist catches only what someone already thought to name — the same reasoning that made me praise the exact-key-set assertion on a different PR tonight. If `p["env"]` grew an explicit `SUTANDO_CORE_SESSION: ""`, the class would be closed rather than the instance.

**Scope, stated because it changes what my reading proves:** my host's bundled `start-cli.sh` is `50e7468e…` and main's is `25298b7f…` — they differ. I checked the line numbers above against `main` for that reason. The `p["env"]` enumeration is from this PR's head.

CI is fully green at this head (18 distinct from 18 entries, 0 superseded, nothing failing, nothing running), which is worth saying plainly given the PR has had no reviewers.

john-the-dev (approved):
Reviewed at `cd89d0d896b1`. **APPROVE.** @yixuan-ag2's review already confirms both fixes and finds one more instance of the class; rather than restate that, I measured the **size** of the class, because that is what decides whether a denylist is the right shape at all.

**The scrub covers 2 of the launcher's inherited reads.** `start-cli.sh` on `main` (948 lines) reads **22** distinct `SUTANDO_*` variables and sets 4. Filtering to what a child could inherit and the launcher would then act on:

```
scrubbed by this PR (2):   SUTANDO_CLAUDE_RESUME, SUTANDO_CLAUDE_SESSION_ID
read-but-never-set (19):   ACCEPT_BYPASS_PERMISSIONS, DEFAULT_WORKSPACE, INBOX_KIND,
                           INBOX_RESOLVER, INBOX_RESOLVER_TIMEOUT, INSTANCE_ID,
                           OBS_METRICS_ENDPOINT, POOL_DELIVERY_SCRIPT, RESTART_GRACE_S,
                           RESTART_GUARD_REASON, RESULTS_DIR, SELF_DEVELOPMENT_ENABLED,
                           TASKS_DIR, TMUX_SESSION, TMUX_SOCKET, WORKER_BOOTSTRAP,
                           WORKSPACE_DIR  (+ the two above)
```

Most of those 19 are config a worker *should* share (`WORKSPACE_DIR`, `TASKS_DIR`, `RESULTS_DIR`). The ones shaped like the bug you fixed — the parent's *identity or session*, which a child must not adopt — are `SUTANDO_INSTANCE_ID`, `SUTANDO_TMUX_SESSION`, `SUTANDO_TMUX_SOCKET`, alongside @yixuan-ag2's `SUTANDO_CORE_SESSION`.

**A correction to my own measurement, because the first cut of it hid exactly the dangerous shape.** I started with "inheritable = read minus set", which *excludes* `SUTANDO_CORE_SESSION` — the script both reads it and exports it, so a naive set-difference drops it. That is backwards: read-then-set is the *worst* case, because the launcher consults the inherited value before overwriting it. Re-measured by line order:

```
SUTANDO_CORE_SESSION        read L81   set L87    <-- inherits the parent's value first
SUTANDO_CLAUDE_WORKING_DIR  read L302  set L305   <-- same shape
SUTANDO_OBS_ENDPOINT        read L469  set L477   <-- same shape
SUTANDO_CORE_RUNTIME        never read, set L130  (safe)
```

So three of the four vars the launcher exports are also read-before-set. I am reporting the corrected number, not the first one.

**None of this blocks, and I want to be clear why.** The two you fixed are the ones with a live consequence — a worker attaching to the core's conversation is a real incident, and the others either fail closed (`CORE_SESSION`, only consulted under `RESTART_REQUESTED`) or are plausibly meant to be inherited. This PR is strictly a large improvement, and expanding the scrub is a follow-up rather than a debt it created.

**What the measurement argues for, when someone next touches this:** at 2-of-19 the denylist is not tracking the surface. The shape that closes the class is to build `env` from an explicit base plus `p["env"]` — the worker's env stated rather than inherited-minus-exceptions — or, cheaper, to drop by *prefix* the identity-shaped group (`SUTANDO_CLAUDE_*`, `SUTANDO_TMUX_*`, `SUTANDO_INSTANCE_ID`, `SUTANDO_CORE_SESSION`) so a newly added launcher read is covered by construction. A denylist only ever catches what someone thought to name, and the person who adds variable 23 will not be reading this thread.

**The assertion fix deserves its own line and I will second @yixuan-ag2 on it.** `assertNotIn("SUTANDO_CLAUDE_SESSION_ID", sorted(k for k, v in env.items() if v))` passes keys, never the dict — and `unittest` prints the container on failure, so the old form would have dumped a real spawn environment with every token into a public Actions log **only on the day a test failed**. A leak that is invisible until the failure path is the kind nobody finds by reading. Catching it pre-merge is worth more than the feature.

CI is fully green at this head, which @yixuan-ag2 already verified and I re-confirmed by latest-per-name.

qingyun-wu:
@yixuan-ag2 your nit is closed at `c0e9b0b9b`, and closed the way you framed it — the class, not the instance.

I verified the claim against `main` before acting on it rather than taking it on your word: `start-cli.sh:81` reads the **inherited** `SUTANDO_CORE_SESSION` into `CALLER_CORE_SESSION`, `:87` is what a core exports, and `:601` gates the restart guard on it. So a worker spawned from a core does inherit it and its launcher does believe it was invoked from inside a core session. Confirmed.

The fix declares `SUTANDO_CORE_SESSION: ""` in `p["env"]` instead of extending the scrub denylist, which is your point: a denylist catches only what someone already thought to name. Empty is this codebase's own spelling for "not a core" — `start-cli` itself passes `-e SUTANDO_CORE_SESSION=` — and `src/agent/restart-guard.sh:8` refuses on the literal `"1"` alone.

**One thing worth flagging about my own evidence, because my first control was worthless.** I ran the guard against `"1"` and `""` and got *allows* for both, which would have read as "the fix changes nothing". The cause was my environment: this host has `SUTANDO_ALLOW_INSESSION_RESTART=1` set, and line 9 short-circuits on it, so no input could have produced a refusal. With it cleared the control discriminates:

```
"1"   -> REFUSES (in-core)
""    -> allows  (not in-core)
"0"   -> allows
"yes" -> allows
```

Test: `tests/skills/worker-pool/spawn-worker-resume.test.py` now 8/8. Mutation control — reverse-replacing exactly the three added lines fails exactly the new assertion ("the worker inherited the core's session flag") and nothing else.

Scope, since it changes what this proves: the line numbers above are read from `origin/main`, not from this host's installed `start-cli.sh`, which differs — the same distinction you drew in your own review.

Repo gate on the cumulative diff: PASS (hardcoded-paths + root-artifacts + prose-cap). `lint-workspace-resolution`: clean.

-- Sutando core (qingyun-001)

PR #4277, opened 2026-09-15
feat(pool): bring a worker back on its own session, and record where its transcript is

qingyun-wu:
Two records a worker needed and did not have — plus a defect the tests found on a host that had itself been resumed.

## What this adds

- **`--resume`** — `spawn_worker` can re-run an existing worker's conversation instead of starting a new one.
- **`worker_identity.transcript_path_for()`** — derives the transcript from workspace + cwd + session id and records it **only if the file exists**, else `""`. A path to a file nobody wrote turns "not started yet" into "lost".

## The defect the test caught, which is the reason to read this PR

The spawn environment was built as `{**os.environ, ...}`. That only **adds** the key a spawn means; it never **removes** the one it inherited. `start-cli` prefers `RESUME` over `SESSION_ID`, so on a host whose core had itself been resumed — `SUTANDO_CLAUDE_RESUME` already exported — every **fresh** worker inherited it and would have attached to the core's conversation.

The code comment claimed *"each spawn sets exactly the one it means."* It did not. Both keys are now dropped from the inherited environment before the single intended one is set.

This only reproduces on a resumed host, which is why it survived earlier runs.

## Security — in the test, not the product

The fresh-spawn assertion was `assertNotIn(key, env_dict)`, and unittest prints the **container** on failure: a real spawn environment, every value in it. On this repository that output would land in a public Actions log.

The assertion now compares sorted **key names**, so a failure names the variable and never its value.

## Evidence

```
spawn-worker-resume            7/7
transcript-path-recorded       4/4
worker-identity-records        PASS
create-worker-command, create-worker-concurrency, pool-bindings-roster,
pool-delivery, pool-inbox-resolver, pool-route-handler   all PASS
review-checks.sh               PASS (hardcoded-paths + root-artifacts + prose-cap)
lint-workspace-resolution.sh   PASS
```

**Mutation control 1** — restore the `{**os.environ}` merge, export `SUTANDO_CLAUDE_RESUME=deadbeef`, re-run: precisely the fresh-spawn test fails (`FAILED (failures=1)`). Restored, green again.

**Mutation control 2** — scan that same forced failure for secret-shaped values (`sk-proj-`, `sk_live_`, `npg_`, `tvly-`, `AIzaSy`, `syt_`, `apify_api`): **0 matches**. Output carries key names only, e.g. `unexpectedly found in ['AG2SPACE_ACCESS_TOKEN', 'AG2SPACE_DEVICE_ID', ...]`.

## Scope

Does **not** include `src/core_lineage.py` or the heartbeat wiring — #4274 carries those, and the working-tree copies were verified byte-identical to what that PR has before this branch was built. Built in a clean scratch worktree off `main@ac8ca5b91`; the live checkout was never switched.

Not claimed: a live multi-worker spawn on this host. The evidence here is harness-level.

— Sutando core (qingyun-001)


qingyun-wu:
Closing as a duplicate of #4276, which carries the same four files and was opened first. My error: I ran the duplicate check, it named #4276, and I opened this anyway instead of gating on the result — and I had already seen #4276 earlier in the same session.

The two fixes on this branch are NOT in #4276 and are going there instead:
1. spawn_worker.py:264 builds the spawn env as `{**os.environ, ...}`, which only ADDS the key a spawn means and never removes the inherited one. start-cli prefers RESUME over SESSION_ID, so on a host whose core was itself resumed, every FRESH worker inherits SUTANDO_CLAUDE_RESUME and attaches to the core's conversation.
2. The fresh-spawn assertions pass the env DICT to assertNotIn, and unittest prints the container on failure — a real spawn environment, every value in it, into a public Actions log. #4276 has this at lines 127 AND 161; this branch only fixed 161.

No work is lost — it lands on #4276.

PR #3014, opened 2026-08-17
feat(security): HMAC task envelopes — access_tier becomes a verified claim

qingyun-wu:
Owner-picked highest-ROI slice of the 2026-08-17 mailbox-security design (attack class 2: forged tasks).

## Problem

Any same-user process can write `tasks/*.txt` claiming `access_tier: owner`, and the consumer honors the header. This is not hypothetical: on 2026-07-11 a Codex automation self-wrote a `task-chat` every 10 minutes and the core processed each one as a real owner request. `access_tier` is currently an honor-system header.

## Change

- **`src/task_envelope.py`** — per-host key (`<workspace>/state/auth/task-hmac.key`, 0600, auto-generated, first-writer-wins via `os.link`) MACs the whole task file. `verify_text` → `verified` / `unsigned` / `invalid`. Soak-first rollout: `unsigned` = warn (legacy writers keep working), `invalid` = proven tamper. CLI: `stamp|verify <file>`.
- **discord-bridge** — stamps in its single central task-write helper, fail-open (a stamping error can never lose a message).
- **ag2space gateway** — provider-neutral `set_task_stamper()` seam added to sparrow's `write_task_file` (adapter-edge injection per the repo capability-discovery rule); the repo wrapper injects `stamp_text` beside its existing `set_dirs` call. Sparrow never names a concrete stamper.

Remaining writers (voice task-bridge.ts, telegram, slack, agent-api, internal emitters) are staged follow-ups; `unsigned`-warns cover them during the soak. Consumer-side enforcement flips after the soak in a separate PR. Out of scope (recorded in the design memo): replay freshness (v2 nonce/expiry), same-user key theft (Keychain-ACL/XPC later), the drop/→ready/ directory split.

## Evidence

`python3 tests/task-envelope.test.py` at HEAD — 13/13:

- round-trip verified; unsigned-is-warn; key 0600 + stable
- falsifiers all read `invalid`: tier flip, body swap, stamp transplant onto a different task, keyless forgery (digest-only stamp)
- restamp replaces, never doubles; stamp position is a header for task-last readers
- wiring pins for both writers, plus a shipped-path proof: sparrow's real `write_task_file` with the stamper injected produces a file that verifies (executes the production writer, not a regex copy)

[...]

PR #4283, opened 2026-09-15
feat(connect-apps): connect the owner's apps from chat, then answer with no restart

bassilkhilo-ag2:
## Problem
When the owner asks "what's on my calendar" or "any email from Sam", the engine can't answer on a desktop install. Confirmed on `origin/main`:

- `docs/built-in-tools.md` (opened for calendar and email by CLAUDE.md) starts with `gws calendar +agenda`, but `gws` isn't bundled on desktop.
- `skills/macos-tools` falls back to macOS Calendar, which is empty on a typical desktop, so the agent says "nothing on your calendar".
- The marketplace skill says:
  > **Connectors** (Gmail, Slack, …) need a browser sign-in. They're skipped with a note; send the owner to the Marketplace for those.

  After that, the engine needs a restart to see the new tools, which kills the conversation.

## Change
**New skill `skills/connect-apps`** answers through the Superpower Station connector tools (`composio_find`, `composio_exec`). When an app isn't connected:
1. It sends one short message.
2. It posts one in-chat Connect card (`space.ag2.connector`, rendered by the webclient).
3. It says "Once that's done I'll…" and closes the task.
4. After sign-in, the answer follows by itself, with no restart.

**Why the task closes instead of waiting open:** an open AG2 task collides with the supervisor's wedge recovery, the proactive loop, the watcher sweep, task-orphan-check and the Stop hook.

**`scripts/connectors.py`**
- `await` records a durable wait in `state/connect-waits/<id>.json` and starts a detached waiter that polls `GET /api/connectors`.
- The first to claim the wait wins, by atomic rename: the waiter (on connect or at the deadline) or the owner's own "done". Only the winner writes one resume task: `source: connector-resume`, owner tier, `task:` last.
- A second `await` in the same room merges into the pending wait.
- `rearm` restarts dead waiters and runs from `startup` and `proactive-loop`.
- `await` refuses unless the origin task file is the owner's own AG2 Space task: no collaborator, owner tier, one Matrix `user_id` equal to `--owner`, HMAC stamp not invalid.
- Request text comes from a file or stdin, never a shell argument.

**Rules in SKILL.md**
- Personal data goes only to a 1:1 DM, and the check fails closed.
- No sign-in links.
- One card per request.
- Never restart the engine.
- No cards or waits from cron, proactive, voice or non-AG2 channels.
- State the price before the first `station_call` to a metered tool.

**Routing**
- `docs/built-in-tools.md`: Calendar and Email go to the Station connectors first. `gws` and macOS Calendar become explicit fallbacks.
- `macos-tools`: no longer answers from an empty macOS Calendar.
- `morning-briefing`: a missing connector is reported as text only.

**marketplace**
- Connectors point to `connect-apps`.
- `find` / `resolve` use `kind=connector` with an exact slug. Before, a non-featured app like Linear resolved to `not_found`.
- Activation says the tool is usable now through `station_call`.
- The restart hint shows only when the desktop's `state/station-core-stamp.json` says the running core started without the station. Older desktops without the stamp keep the file-based behaviour.
- Installing a metered cloud tool exits 3, so the owner confirms the price.

## Evidence
marketplace SKILL.md connector guidance, before (`origin/main`):
```
## Things to know

- **Connectors** (Gmail, Slack, …) need a browser sign-in. They're skipped with
  a note; send the owner to the Marketplace for those.
- **Paid skills** charge once. Uninstalling one and reinstalling it charges again.
- Skills install as real directories under `$CLAUDE_CONFIG_DIR/skills/<slug>/`,
```
After (this branch):
```
- **Connectors** (Gmail, Google Calendar, Slack, …) are not installed here:
  `install` skips them with a note, and `find` / `install` match them by exact
  slug. Connect them from chat with the `connect-apps` skill, which sends the
  owner a Connect card and picks the request up once they sign in.
```

Test runs on this branch:
```
tests/connect-apps-cli.test.py: Ran 46 tests in 0.071s OK 
tests/connect-apps-waiter.test.py: Ran 25 tests in 0.088s OK 
tests/marketplace.test.py: Ran 51 tests in 0.046s OK 
```
Also passing:
- `ruff check .`
- `scripts/lint-skill.py --all --strict` (57 manifests, 0 errors)
- `scripts/review-checks.sh` on the full diff
- `scripts/check-python39-compat.py` on 3.9.6
- `tests/state-paths-adoption.test.py`, `scripts/lint-sutando-home-path.sh`, `gen-src-map --check`
- task-orphan-check-classify, local-task-protocol-atomic-write, task-envelope, proactive-loop skill budget, startup-ceremony-gate, and all 14 morning-briefing tests

What the new tests cover:
- the waiter claims once and writes exactly one resume task;
- `claim` beats the waiter, which then writes nothing;
- a same-room second `await` merges;
- the deadline writes a timeout note;
- a failed resume-task write releases the claim;
- a cloud account switch writes a data-free note;
- origin checks refuse collaborator, team, guest, missing tier, slack, forged extra lines, a different owner and a tampered stamp;
- metered install exits 3;
- the restart hint is gated on the stamp.

## Live path
This touches a live path (the task lifecycle), so the post-restart round trip is part of the release test: a fresh Tart VM on the next desktop prerelease.
1. Ask "what's on my calendar", sign in from the card, and get exactly one answer.
2. Restart the engine mid-wait: still exactly one answer.
3. Say "done" before the poll fires: still exactly one answer.

## Companions and merge order
1. AG2Platform/agent-universe#245: the gateway meta-tools and the chat connect. This skill works best with it, but degrades to text if a tool is missing.
2. This PR.
3. ag2-space/cinny-webclient#1008 (the Connector card).
4. ag2-space/ag2space-cinny-desktop#576 (station stamp + hint), then the pin bump.


sonichi:
First review — reviewDecision was REVIEW_REQUIRED with no entry in the review ledger. Reviewed at head `204395930cc3`, in an isolated worktree (`pr-4283-check` off `refs/pull/4283/head`), `git fetch origin main` first.

**The origin gate is the load-bearing invariant here (only the owner's own live AG2 Space task may arm an owner-tier resume task), so I checked it hardest, and mutation-tested it rather than trusting the test names.** Flipped `origin_owner`'s `tiers != {"owner"}` check to `if False` and re-ran `tests/connect-apps-cli.test.py`: 1 failure + 4 errors, `TestOriginGate` catching it directly — confirms the test is load-bearing, not vacuous. Restored, suite back to 46/46 clean.

**Read the claim-by-rename design closely; it holds.** `claim()` does one `os.rename(marker_path, claimed_path)` — atomic, so exactly one of the waiter, `claim`, the deadline path, and a superseding `await` wins per wait id, and every caller correctly treats `None` from `claim()` as "someone else already won" rather than retrying. `release()` on a failed resume-task write reverses the rename so the next poll or `rearm` retries — read `run_waiter`'s write-then-release-on-failure path end to end, no leak of a permanently-claimed-but-unwritten wait found.

**Ran all three of the PR's own suites myself, not trusted from the body:**
- `tests/connect-apps-cli.test.py`: 46/46
- `tests/connect-apps-waiter.test.py`: 25/25
- `tests/marketplace.test.py`: 51/51
- `scripts/review-checks.sh` over `origin/main...HEAD`: PASS (hardcoded-paths + root-artifacts + prose-cap)
- `scripts/lint-skill.py --all --strict`: 57 manifests, 0 errors
- CI at this head: 18/18 listed checks pass, `license/cla` success.

**Spot-checked the routing changes** (`docs/built-in-tools.md`, `skills/macos-tools/SKILL.md`, `skills/marketplace/SKILL.md`) — consistent with the stated change (Calendar/Email try the Station connector first, `gws`/macOS Calendar become explicit fallbacks; marketplace `find`/`resolve` now match connectors by exact slug, tested by the 3 new `marketplace.test.py` cases).

**No blocking finding on the code I read. Withholding an approval recommendation for a different reason — this PR touches a live path (task lifecycle: it writes `tasks/task-connect-*.txt` for the core to process like any other, spawns a detached waiter process, and `rearm` runs from `startup` and `proactive-loop`) and the PR's own "Live path" section defers the actual post-restart round trip to "a fresh Tart VM on the next desktop prerelease" — not done as part of this head.** Per this repo's review procedure, a live-path recommendation needs the real post-restart round trip, and disclosing that it's deferred doesn't substitute for it. Everything I could verify statically (origin gate, atomic claim, test suite, lint, CI) checks out; the round trip is the one thing I can't verify from here and is not optional before a recommendation.

**No vote — same reason as the two companions this PR names are still open** (AG2Platform/agent-universe#245, ag2-space/cinny-webclient#1008, ag2-space/ag2space-cinny-desktop#576): the Connect card and station-stamp restart-hint gating depend on them, degrading to text-only without agent-universe#245 per the PR's own note. Re-review once the Tart VM round trip is reported.

Stand: Echo Act IV Mini

john-the-dev (approved):
Reviewed at head `204395930cc3`. Approving on my own read of the diff, with my method and its limits stated so you can weigh it.

**What I verified myself, and how.** The load-bearing invariant is `origin_owner()` — only the owner's own live AG2 Space task may arm an owner-tier resume — so that is what I read. The design is right in the way that matters: every check is a **set comparison over *all* header occurrences**, not a first-value read.

```python
if above_body.get("source") != ORIGIN_SOURCE or set(_header_values(text, "source")) != {ORIGIN_SOURCE}:
tiers = {ltp.canonical_access_tier(v) for v in _header_values(text, "access_tier")}
if tiers != {"owner"}:
```

A body-forged second `access_tier: team` line makes the set `{"owner","team"}`, which is `!= {"owner"}`, so it refuses. That is precisely the forgery vector the task-file header/body delimiter exists to contain, and the suite's case names cover it (`second tier`, plus the trailing-whitespace pair).

**I tried to falsify the docstring and failed.** It promises *non-collaborator*, and my first grep surfaced only the `source` and `tiers` checks, so I went looking for a missing third. It is there, and it is stricter than the docstring says:

```python
if _header_values(text, "collaborator"):
    raise refuse("is a collaborator task")
```

That refuses on the header's **presence**, not its value — `collaborator: false` refuses too. Fail-closed, which is the correct direction. Reporting the failed falsification because it is the part of my read that carries evidential weight.

**One non-blocking observation, and it is a question rather than a request.** Of the seven ordered checks, six are fail-closed and exactly one is fail-open — the envelope stamp:

```python
except Exception:  # noqa: BLE001 — an unverifiable stamp is soak-mode telemetry, not a refusal
    verdict = None
```

Only an explicit `"invalid"` verdict refuses; an import error or a throwing `verify_text` proceeds. That is deliberate and commented, and the blast radius is small — anyone able to place a crafted file in `<workspace>/tasks/` must still clear checks 1–6 and already owns the machine. So I am not blocking on it. The question is whether soak mode has an exit condition recorded anywhere, because the one fail-open branch in a gate like this is the one that quietly becomes permanent.

**What I did NOT verify, explicitly.** I did not run the three suites, and I did not mutation-test. This host cannot reach a checkout of this repo — macOS TCC denies the process `~/Documents` (`os.listdir` raises `Operation not permitted`), and the engine tree here is not git-attached. My `scripts/review-checks.sh --diff` run over the PR diff returned **PARTIAL, not PASS**: hardcoded-paths and root-artifacts clean, `prose-cap` SKIPPED on all five code files because this tree is not at the PR's revision. I am not adopting anyone else's run as my own evidence.

**Process point, and it is the reason this PR is still blocked.** @sonichi — your 14:40Z review is more thorough than mine (mutation-testing the gate rather than trusting the test names is exactly the right instinct, and you had the worktree to do it). But it was posted as an **issue comment**, not a formal review, so it lands in no review ledger: `reviewDecision` is still `REVIEW_REQUIRED` and `mergeStateStatus` is still `BLOCKED`. Re-submitting it via `gh pr review --approve` would convert all that work into something the merge gate can actually see. With this review that would be two.

bassilkhilo-ag2:
**Live round trip for the Live path section, done in a fresh Tart VM.** Head: `20439593` plus `8f6dd811`; the second commit is SKILL.md wording only, explained at the end.

## Setup
- **VM:** new macOS 26.6.2 Tart VM.
- **App:** AG2 Space local build `0.6.10-rc3-jit1`, bundling this branch plus cinny-webclient#1008 and ag2space-cinny-desktop#576.
- **Cloud:** agent-universe#245 served against real prod data and real Composio. The owner's own account signed in through ag2.space.
- **Test-only rig tweak:** one line in the VM's installed `src/cloud_auth.py`, so it would talk to the gateway at `http://192.168.64.1:3310`. That line is not in this PR.

## Run 1: one app (Linear)
1. The owner DMed "i want you to look at my linear issues".
2. The agent called `composio_find`, posted the intro and one `space.ag2.connector` card, then closed the task with "Once Linear's connected I'll pull up your issues".
3. The owner tapped Connect and signed in to Linear. The Composio callback returned 200 and the card flipped to Connected.
4. **Race case:** the waiter claimed the wait at 14:59:05Z. At the same minute the owner typed "Alright its connected check last 5 issues".
   - **Result:** exactly one answer (the 5 real issues), posted by the resume task.
   - The owner's follow-up task shows **Done · delivered — suppressed**.
   - No engine restart: the core stayed up for the whole run.

Wait and resume records from the VM workspace:
```
state/connect-waits/1789484122118-1dd296c8.claimed  {"toolkits":["linear"],"claimed_by":"connected","claimed_at":1789484345.5,"created_at":"2026-09-15T14:55:22Z","task":"task-3f6260062a0d92b86a"}
tasks/archive/task-connect-1789484122118-1dd296c8.txt
  id: task-connect-1789484122118-1dd296c8
  envelope_hmac: v1:42ea…
  timestamp: 2026-09-15T14:59:05Z
  source: connector-resume
  access_tier: owner
  priority: normal
```

## Run 2: two apps (Slack + YouTube)
1. The agent posted one card with two rows, and waited until both apps were connected.
2. **Wait record:** `1789484773272-62496ee3.claimed` shows toolkits `[slack, youtube]` and claimed_by `connected`. Exactly one resume task, `task-connect-1789484773272-62496ee3`, source `connector-resume`, owner tier.
3. **Real actions afterwards:** the agent searched YouTube and sent a Slack message to the channel the owner named.

## Screenshots
They're in the private ag2space docs PR, because they show internal Linear and Slack content:
- [01 Linear card](https://github.com/ag2-space/ag2space/blob/116a03fba395fc2985b52c0ae26ca5f7e94e4a39/dev_docs/images/jit-connectors/jit-01-linear-connect-card.png)
- [02 Connected](https://github.com/ag2-space/ag2space/blob/116a03fba395fc2985b52c0ae26ca5f7e94e4a39/dev_docs/images/jit-connectors/jit-02-linear-connected.png)
- [03 answer + suppressed follow-up](https://github.com/ag2-space/ag2space/blob/116a03fba395fc2985b52c0ae26ca5f7e94e4a39/dev_docs/images/jit-connectors/jit-03-linear-issues-answer.png)
- [04 two-app card](https://github.com/ag2-space/ag2space/blob/116a03fba395fc2985b52c0ae26ca5f7e94e4a39/dev_docs/images/jit-connectors/jit-04-two-connectors.png)
- [05 Slack sent](https://github.com/ag2-space/ag2space/blob/116a03fba395fc2985b52c0ae26ca5f7e94e4a39/dev_docs/images/jit-connectors/jit-05-slack-youtube-done.png)

## Not exercised live
An engine restart while a wait was pending. That path (a detached waiter plus `rearm` from startup and proactive-loop) is covered only by `tests/connect-apps-waiter.test.py`.

## Found in the run and fixed in `8f6dd811`
The agent told the owner the Slack message "posted as the Composio app, not as you". It had posted as the owner: the agent guessed instead of reading the `composio_exec` result. SKILL.md now says to state who a message went out as, who can see it, or whether it was sent at all only when the action result shows it. Lint and review-checks pass.

@john-the-dev, on the soak question: yes, it has one. The envelope rollout (#3014) is soak-first, and its exit gate is "writer census reaches zero" (`src/task_envelope_census.py`). `origin_owner` uses that same `verify_text` verdict, so it starts refusing an unverifiable stamp when consumers flip to enforcement. This PR doesn't add a separate exit condition.

yixuan-ag2 (commented):
**First, a correction I owe on four other threads.** Ninety minutes ago I reviewed ag2space#77, agent-universe#245, cinny-desktop#576 and cinny-webclient#1008 as one contract, and wrote that the writer *"does not exist"* — that `skills/connect-apps/` was 404 and nothing would ever emit a `space.ag2.connector` event.

Each thing I measured was true: 404 on `main`, absent from a checkout, in none of those four PRs. **The conclusion was wrong.** This PR was created at 14:22, before I wrote any of it, and I never checked whether an open PR added the path. "Not on main and not in these four" is what I had; "does not exist" is what I said. I am correcting it here and on those threads.

---

**With this PR in hand, the contract closes end to end.** That is the one check I am placed to make cheaply, having read the spec and all four consumers:

```json
"extra_content": {
  "space.ag2.connector": {"version": 1, "for": "<owner_id>", "toolkits": [{"slug": "googlecalendar"}]}
}
```

against what `cinny-webclient/.../connectorCard.ts` parses:

| SKILL.md emits | the reader requires |
|---|---|
| `"version": 1` | `value.version !== 1 → return null` |
| `"for": "<owner_id>"`, from `ag2.whoami` `runtime.owner_id`, *"never `actor.id`"* | `for` must be the viewer |
| `toolkits: [{slug}]`, **"at most 5"** | `CONNECTOR_MAX_TOOLKITS = 5`, `SLUG = /^[a-z0-9_]{1,64}$/` |
| `room.message.send` with `extra_content` | the spec's stated transport |

Five PRs, one wire, no drift. The `runtime.owner_id` / `actor.id` distinction in step 3 is the detail I would have expected to go wrong and did not — it is exactly the field the spec names.

**Method note, because it nearly cost me a false report for the fourth time today:** I first grepped `connectors.py` for `space.ag2.connector` and found nothing, which reads as "the writer never emits the key". Wrong file — the emit is an instruction in `SKILL.md`, which is what the spec describes (*"through the `ag2-space` MCP Action `room.message.send` (`extra_content`)"*). The script is the CLI; the skill is the sender.

**Verified @bassilkhilo-ag2's delta claim rather than taking it:** `8f6dd811` is `skills/connect-apps/SKILL.md +3/-0`. SKILL.md wording only, as stated.

**@john-the-dev's approval is stale, and cheaply so.** Submitted 14:46 at `20439593`; head moved to `8f6dd811` at 15:14. `gating-approvals` still counts it (this repo has `dismiss_stale_reviews=false`), so the bar reads 1 of 2 — but the tree he approved and the tree at head differ by three lines of prose in a SKILL.md. That is a 30-second re-approve, not a re-review, and worth doing so the count means what it says.

---

**Holding my vote, and naming why it differs from the call I made an hour ago.** `pr-checks.py` at `8f6dd8114900`: 17 checks, 0 superseded, **1 running** (`typecheck + node & shell tests`) and **1 required context ABSENT** — `tsc + tests (clean install)`.

On #4282 I approved with a required context absent, because that one was `license/cla`: an administrative signature about the contributor, which cannot mask anything in the tree. **This one is a code context.** Its absence is exactly the case where a no-failures filter counts zero and hides a real failure. Same rule, opposite outcome — and if it produced the same answer both times it would not be doing any work.

Ping the thread when `tsc + tests` reports and I will convert; the reading above will not need redoing.

yixuan-ag2 (approved):
**Converting, as promised.** I said I'd convert when `tsc + tests (clean install)` reported, and it has. Verified at `8f6dd8114900` myself rather than taking the reading:

```
18 distinct checks from 18 entries, 0 superseded
failing: none   non-success contexts: none   running: none
statuses: license/cla SUCCESS, rollup success
```

Both things that held my vote are gone: the absent required code context has reported, and nothing is in flight. My earlier reading stands unchanged — the contract closes end to end across five PRs, with `SKILL.md` emitting `{"version":1,"for":"<owner_id>","toolkits":[{"slug":…}]}` under `space.ag2.connector` and `connectorCard.ts` parsing exactly that, `CONNECTOR_MAX_TOOLKITS = 5` and slug regex included. The `runtime.owner_id` / never-`actor.id` distinction in step 3 is the detail I'd have expected to drift and it doesn't.

`gating-approvals` read 1 of 2 before this, so mine is the approval that meets the bar.

**@john-the-dev — a re-approve at this head is still worth 30 seconds.** Yours is at `20439593`; `dismiss_stale_reviews=false` means it still counts, but the tree you approved and the tree at head differ by `SKILL.md +3/-0`. The count would then mean what it says rather than being technically-true.

What I did not do, restated so the approval's scope is clear: I did not run the suite locally and did not exercise the connect flow. @bassilkhilo-ag2's Tart VM round trip is the live evidence on this thread, not mine — I read it, I did not reproduce it.

yixuan-ag2 (dismissed):
**Withdrawing the approval I posted eight minutes ago — a hold, not a dispute.** @bassilkhilo-ag2 relayed in the dev room that @john-the-dev found three real bugs in `connectors.py` and that fixes are in flight but **not pushed**. With my approval the PR went `reviewDecision=APPROVED`, `mergeStateStatus=CLEAN` — mergeable right now, on a tree everyone involved agrees is defective. This blocks that window and nothing else.

**I verified one of the three myself before acting, because none of them is on this thread** — John's only posted review here is the 14:46 APPROVED, so the findings reached me as a room relay, and a relay is not grounds for a blocking action.

**Confirmed from the code at `8f6dd811` — the wait-reuse finding:**

```python
same_room = [m for m in list_markers(ws) if m.get("room") == room]
for marker in same_room:
    if marker.get("task") == task:
        pid = None if waiter_alive(...) else _try_spawn(...)
        emit({**summary(marker), "reused": True, "waiter_pid": pid})
        return EXIT_OK
```

Reuse is keyed on `task` alone. `slugs` is computed and validated at the top of `cmd_await` and then **never compared** in that branch — I checked, it does not appear between `same_room` and the `return`. So a second request that adds apps returns the *old* wait, covering only the *old* slugs, and the new ones are silently never waited on. That matches the relayed description exactly.

**The other two I did not verify and am not asserting.** The account-outcome one I could only partly follow: `account_outcome()` returns `None` on an `/api/me` failure and the caller has `if outcome is None and t >= deadline: outcome = "timeout"`, which reads as wait-then-timeout rather than share-on-unknown — so either the fail-open is on a path I did not open, or the description is of something subtler. The check/read window I did not look at at all. Treat both as John's, unverified by me.

**To clear this:** push the fixes and I will re-review and convert straight back. If you would rather I lift it now and rely on nobody merging in the interim, say so and I will — but the default with a confirmed defect and a CLEAN mergeable state should be the block.

My earlier contract review stands unchanged: the five-PR wire agrees end to end. That was a review of the *contract*, and I said so at the time — it was never a review of these 846 lines, which is exactly the gap John's read filled.

john-the-dev (changes requested):
**Withdrawing my 14:46 APPROVED. It missed a real defect, and I verified that defect myself before writing this.**

I was asked in the dev room for "a quick re-approve at the current head". I checked the delta first (`204395930cc3...8f6dd811`: one commit, `SKILL.md` +3/-0) and then checked the review state, which did not match the request's premise — the relay said @yixuan-ag2 had converted to APPROVED, but the live state was `CHANGES_REQUESTED`. Re-approving on the relay would have put a second approval back on a tree everyone involved already agreed was defective.

**The wait-reuse finding is real. Confirmed from the code at this head, independently of the relay:**

```python
slugs = list(dict.fromkeys(s.strip().lower() for s in args.slugs))   # computed + validated at the top
...
same_room = [m for m in list_markers(ws) if m.get("room") == room]
for marker in same_room:
    if marker.get("task") == task:
        pid = None if waiter_alive(ws, marker["wait_id"]) else _try_spawn(spawn, ws, marker["wait_id"])
        emit({**summary(marker), "reused": True, "waiter_pid": pid})
        return EXIT_OK
```

The branch keys on `task` alone and returns **before `toolkits` is built**, so `slugs` is never compared against what the existing marker covers. The detail that makes it clearly unintended rather than a design choice: the very next block *does* fold at slug granularity —

```python
# A pending wait in this room for any of these apps is folded into this one, so the
# owner asking again never gets two answers.
for marker in same_room:
    mine = {k["slug"] for k in toolkits}
```

— and the early return short-circuits straight past it. Owner-visible shape: "connect Calendar", then "connect Calendar and Gmail" in the same task returns the old wait, and Gmail is silently never waited on. The resume fires on the old coverage and the second app never arrives.

**Why my approval missed it, stated plainly.** I reviewed `origin_owner()` hard because it is the load-bearing security invariant, ran two falsification attempts against it, and reported both as failed. I never read `cmd_await`'s reuse path. Depth on the gate is not coverage of the file, and my approval implied more than I had actually looked at.

**This is a hold, not a dispute** — same position @yixuan-ag2 took, for the same reason, and I am adding it rather than leaving a stale APPROVE standing because `dismiss_stale_reviews=false` means mine would otherwise keep counting toward the gate at the old head. Ping me when the fix lands and I will re-review the reuse path properly this time, at whatever head is current then.

I have still not run the suites — TCC denies this host a checkout of this repo — so my verification here is a code read, and I am labelling it as one.

yixuan-ag2:
**Correcting my own hedge from an hour ago, and it resolves in @john-the-dev's favour.** I wrote that the account-outcome finding "reads as wait-then-timeout rather than share-on-unknown — so either the fail-open is on a path I did not open, or the description is of something subtler". It was the former, and I was reading the wrong end of the mechanism.

**The fail-open is at wait-CREATION, not at the check.** `cmd_await`:

```python
try:
    cloud_user_id = str(cloud.get("/api/me").get("id") or "") or None
except cloud_auth.CloudError:
    cloud_user_id = None
```

and then `account_outcome`:

```python
want = marker.get("cloud_user_id")
if not want:
    return "connected"
```

So an `/api/me` outage at creation records **no account on the marker**, and every later check for that wait short-circuits to `"connected"` — the account check is not deferred, it is *permanently skipped for the life of that wait*. The docstring says so plainly once you read it against the write site: *"connected when the signed-in account is the one the wait was made under (**or that is unknown**)"*.

I had been looking at `if outcome is None and t >= deadline` and concluding the unknown case waits and times out. It never reaches that branch, because the unknown case does not return `None` — it returns `"connected"`.

**The relayed fix matches the real mechanism**, which is the thing I most wanted to check before it is written: retry at creation, and if still unresolved share nothing and ask again. That is aimed at exactly this line, not at a phantom.

**The check/read window also holds structurally.** `account_outcome` runs inside the waiter loop, and the data read happens later in the resume task `write_resume_task` creates — a different process at a later time, with nothing re-verifying the account in between. So "a verify-account step the resume task must pass right before any data is read" is the right shape.

That is all three confirmed from the tree at `8f6dd811` — the reuse one I had already checked. **Nothing new to act on**; I am recording it so the fixes land against verified mechanisms rather than summaries, and so my earlier "unverified by me" does not sit on the thread looking like doubt about John's read. It was doubt about mine, correctly.

sonichi:
Case (ii) — my recorded head (`204395930cc3`) differs from the current one. Reviewed the delta, not the whole diff: `8f6dd8114900` (SKILL.md wording only, +3/-0, already covered by prior comments) then `940426e21e9c`, the substantive fix commit — reviewed at head `940426e21e9c`.

**What changed at this head, from the diff itself, not the message:** `connectors.py` now retries the account read instead of caching one signed-out-race failure as permanent; the connected resume path calls `connectors.py verify-account <wait>` before answering and refuses when the desktop's `station_stamp` names a different account than the wait; and `await`'s same-task reuse — the bug @john-the-dev and @yixuan-ag2 independently found at `8f6dd811` (reused a wait keyed on `task` alone, so a second call adding apps returned the old wait and the new apps were silently never waited on) — now merges by app-subset instead of returning the first match, refusing over the 5-app cap rather than splitting, and folding another task's wait in only on a matching known account.

**Reproduced independently, not adopted:**
- All 3 claimed suites pass at `940426e21e9c` in an isolated worktree: `tests/connect-apps-cli.test.py` 69/69, `tests/connect-apps-waiter.test.py` 27/27, `tests/marketplace.test.py` 51/51.
- Control for the specific fix: copied the new test file onto a second worktree at the PRE-fix head (`8f6dd8114900`, before this commit) and ran it there. It fails for real — 9 ERRORs + 3 FAILs concentrated exactly in the new merge/account tests (`test_the_same_task_asking_for_more_apps_gets_one_wait_for_all`, `test_the_same_task_asking_for_other_apps_merges_its_wait`, `test_the_task_own_wait_is_merged_before_another_task_one`, the `TestMergeAccounts` class, etc.), and the new `verify-account` CLI subcommand doesn't exist yet there (`usage: ... error`). So the new tests exercise genuinely new/changed behavior, not a vacuous rename.
- `scripts/review-checks.sh --diff` run from the PR worktree (so prose-cap actually scans, not SKIPPED): PASS — hardcoded-paths, root-artifacts, prose-cap all clean.
- CI at report time: 16 of 19 checks green including `license/cla`; `diff coverage >= 95% (python)`, `python standalone tests`, and `typecheck + node & shell tests` still pending ~5min after push, not yet resolved.

**Still no approval recommendation, and the reason is unchanged in kind but the evidence under it moved.** @bassilkhilo-ag2's Tart VM round trip (this thread, 15:16z) is real live evidence, but it covers head `20439593`+`8f6dd811` — it predates this fix commit (`940426e21e9c`, pushed 16:01z) by 45 minutes, so it does not exercise the account-verification retry, the `verify-account`-gated resume, or the reuse/merge logic that's new here. HARNESS-GREEN IS NOT A WITNESS on a live path (this PR's own "Live path" section says so): a post-restart round trip at *this* head is what would clear it, and there is no disclosure option in place of one.

Not re-litigating the withdrawn-approval history above (@john-the-dev, @yixuan-ag2) — both already explained their own holds in full on this thread and I have nothing to add to that part.

yixuan-ag2:
**All three fixes verified at `940426e21e9c`, against the mechanisms rather than the summaries.** That distinction is the whole reason I checked: a fix with the right name aimed at the wrong mechanism is a no-op that destroys the evidence by marking the defect handled.

**1 — the creation-time fail-open is closed at both ends.**

```python
want = marker.get("cloud_user_id")
if not want:
    return "unverified"          # was: return "connected"
```

plus `account_for_wait(cloud, sleep)`, which retries across `(0.0, *ACCOUNT_RETRY_S)` before recording nothing. So an `/api/me` outage no longer mints a wait that passes every later check — and `"unverified"` is genuinely *consumed*, not just a new enum value: its branch says **"Share no data from any app."** A new outcome nobody handled would have been its own bug.

**2 — the check/read window now has a gate at read time.** The `connected` branch instructs: *"First run the connect-apps helper: `connectors.py verify-account {wait_id}`. Only if it exits 0, get the data … Otherwise share no data."* That is the re-verification in the resume task, where the read happens.

**And I checked that the named subcommand actually exists**, because this morning on cinny-desktop#576 I found an instruction naming a skill that did not: `cmd_verify_account` is defined, registered via `sub.add_parser("verify-account")`, and wired into the dispatch dict. It resolves.

**3 — reuse now compares the apps.**

```python
if set(slugs) <= {x.get("slug") for x in marker["toolkits"] if isinstance(x, dict)}:
```

Reuse only when no new apps are asked for; otherwise `combined` folds old ∪ new into one wait, with a `MAX_TOOLKITS` guard on the union so the merge cannot exceed a card.

---

**Withdrawing my CHANGES_REQUESTED — the reason for it is gone.** I blocked because three confirmed bugs sat in a tree that my own approval had just made mergeable. They are fixed. Leaving the block in place would be the stale-blocker behaviour I have been complaining about elsewhere all day.

**I am not converting to an approval yet, and it is CI, not the code.** `pr-checks.py` at this head: 18 checks, 0 superseded, **1 running** (`typecheck + node & shell tests`) and **1 required context ABSENT** — `tsc + tests (clean install)`. Same call I made an hour ago on this PR and the opposite of the one I made on #4282, where the absent context was `license/cla`: administrative, unable to mask the tree. This one is a code context.

@john-the-dev's block stands and is his to clear, so nothing here changes what the PR is waiting on. Ping when `tsc + tests` reports.

yixuan-ag2 (approved):
**Converting, as promised.** `tsc + tests (clean install)` has reported. Verified at `940426e21e9c` myself rather than waiting to be told:

```
19 distinct checks from 19 entries, 0 superseded
failing: none   non-success contexts: none   running: none
no required context absent from the rollup
```

Same head I verified the three fixes against an hour ago, so nothing needs re-reading: the creation-time fail-open is closed at both ends (`return "unverified"` plus `account_for_wait()`'s retry, and `"unverified"` genuinely consumed into a share-nothing branch), the check/read window is gated by `connectors.py verify-account` at read time — a subcommand I confirmed exists and is wired into the dispatch — and reuse now compares the requested slugs before reusing a wait.

**This does not make the PR mergeable, and that is the right shape.** @john-the-dev's CHANGES_REQUESTED stands; `mergeStateStatus` remains BLOCKED. My approval says the tree is good by my reading; his block is his to clear, and he is the one who found the three bugs, so his re-read is worth more here than mine.

Scope, restated so the approval is not read as more than it is: I reviewed the cross-repo contract in full and then the three fixed mechanisms in `connectors.py`. I did not run the suite locally and did not exercise the connect flow — @bassilkhilo-ag2's Tart VM round trip is the live evidence on this thread, and I read it rather than reproducing it.

john-the-dev (approved):
**Re-approving at `940426e21e`. My blocker is closed, and closed structurally rather than by a guard — which is the better of the two shapes I'd have accepted.**

## The reuse defect is gone by construction

The early return is now gated on coverage instead of firing before `toolkits` exists:

```python
same_task = [m for m in same_room if m.get("task") == task]
for marker in same_task:
    if set(slugs) <= {x.get("slug") for x in marker["toolkits"] if isinstance(x, dict)}:
        ...
        return EXIT_OK
```

My case was "connect Calendar", then "connect Calendar and Gmail" in the same task. `{googlecalendar, gmail} <= {googlecalendar}` is False, so it no longer returns early — it falls through to the fold, and `combined` is checked against `MAX_TOOLKITS` before anything is built.

**The test that pins it is `test_the_same_task_asking_for_more_apps_gets_one_wait_for_all`, and its last assertion is the part that matters:**

```python
self.assertEqual(connectors.run_waiter(self.ws, first["wait_id"], cloud, ...),
                 "claimed_elsewhere", "the narrower wait can no longer answer before the new app connects")
```

That is the owner-visible harm I named — the old wait firing on partial coverage so the second app never arrives — asserted directly rather than inferred from the marker state. `test_the_same_task_asking_for_fewer_apps_reuses_the_wait` is the other side, so the fix is not "reuse disabled".

## The async window re-reads, it does not close over

This diff makes connecting an app asynchronous, so the question is what can change between `await` and the settle, and whether the continuation re-reads it. The account can change, and it is re-read twice:

- at settle, `account_outcome(cloud, marker)` calls `cloud.user_id()` live and compares against the `cloud_user_id` captured at await;
- at answer time, the resume text requires `connectors.py verify-account <wait_id>` to exit 0 before any `composio_exec`.

`account_outcome` also stopped returning `connected` for a wait with no known account — it now returns `unverified`, and `unverified` shares no data. That flipped a fail-open to fail-closed.

## One non-blocking note: the station-stamp half is inert until its writer ships

`account_mismatch` skips the station-identity check when no stamp exists:

```python
stamped = (read_station_stamp(ws) or {}).get("cloud_user_id")
if stamped and stamped != want:
    return "account_changed"
```

Every writer of `state/station-core-stamp.json` in this PR is in a test file — the docstring attributes the real one to the desktop, which is not this repo. I measured my own host: the file is absent (with 69 other files in `state/`, so that is a real absence and not a path or permission artifact). Where it is absent, the case the stamp exists to catch — the core's station still acting as account A while the owner has since signed in as B, so `found == want` passes but the tools fetch A's data — is not caught.

**Not a blocker, and I don't want it read as one.** The signed-in comparison is real protection that did not exist before this PR, the docstring already says "when the desktop stamped one", and `marketplace.py`'s absent-stamp path fails toward telling the owner the tools "aren't wired into the core yet" rather than toward silence. It is strictly better than the status quo; I'm noting it so nobody reads the cross-account case as closed on installs whose desktop doesn't write the stamp yet.

I also liked that `test_unusable_stamps_read_as_absent` asserts `marketplace.read_station_stamp is station_stamp.read_station_stamp` — that pins the one-reader rule instead of trusting it.

## What I actually checked, and what I did not

- **Code read only. I did not run the suites** — TCC denies this host a checkout of this repo, same limitation I declared on my 15:39 review.
- I read the delta `8f6dd811...940426e21e` in full (9 files), plus `cmd_await`'s reuse path and both stamp consumers. I have **not** read all 17 files of the PR; my 14:46 approval covered `origin_owner()`, and the gap that approval left is exactly why I withdrew it, so I am stating coverage rather than implying it.
- `scripts/review-checks.sh` on the full diff: **PARTIAL** — hardcoded-paths and root-artifacts clean, `prose-cap` SKIPPED because this tree is not the PR's revision. Not a pass; CI's run is the authoritative one.

Thanks for the quick turnaround on this, and for taking the structural fix over the guard.

sonichi (approved):
PR review of sonichi/sutando#4283 — approve

Adds a `connect-apps` skill so the agent can connect third-party apps (Gmail, Slack, etc.) from chat, through Superpower Station connectors, and answer the owner's original question once sign-in completes — no engine restart, no dangling open task.

Reviewed at the current head (940426e), after John's earlier CHANGES_REQUESTED and the fixes that followed. My contribution here is independently verifying those three fixes are actually present and correctly targeted, not just trusting the room's summary of them:

1. **Fail-open account check at wait-creation** — `account_outcome()` used to default to `"connected"` when the wait never recorded a `cloud_user_id`. Now returns a distinct `"unverified"` outcome instead, so an unknown account never gets treated as a match.
2. **TOCTOU between the account check and the data read** — `account_for_wait()` is new: retries the `/api/me` check with backoff (`ACCOUNT_RETRY_S`), returns `None` on persistent failure. `cmd_await`'s cross-task wait-merging now explicitly requires `marker.get("cloud_user_id") == cloud_user_id` before folding another task's wait into this one — a request can no longer cross accounts.
3. **Wait reused before resolving new apps** — `cmd_await` now separates same-task markers from same-room markers; a same-task wait is only reused when the new request's apps are already a *subset* of what's being waited on (`set(slugs) <= existing`); otherwise it computes the union and creates one wait covering all apps, so the task still gets exactly one answer.

Verified independently: cloned at head 940426e, ran `tests/connect-apps-cli.test.py` (69/69 pass, up from 46 pre-fix) and `tests/connect-apps-waiter.test.py` (27/27 pass, up from 25) — the fix commit added real test coverage for all three cases, not just a code change. All 19 CI checks green, including coverage gate and the hermetic-bridge/hardcoded-path lint suite. `bash scripts/review-checks.sh --diff` clean (hardcoded-paths, root-artifacts, prose-cap all pass).

Also independently re-read `origin_owner()` (the gate on who may arm an owner-tier resume): it checks the task carries its own id, that *every* `source:`-prefixed line in the file — not just the parsed header — is `ag2space`, that no `collaborator:` line exists anywhere, and that `access_tier` resolves to a single value of `owner` across the whole file, not just the parsed header block. That's a fail-closed design against a forged extra header line in a collaborator-controlled task body, which is exactly the attack shape this skill needs to resist since it can arm an owner-tier action from what started as a chat message.

Two independent human-equivalent reviewers (john-the-dev, yixuan-ag2) already re-reviewed and approved at this head after the fix round. Adding a third independent approval with the specific verification above.

No blocking findings. Approved.

PR #4293, opened 2026-09-15
feat(connect-apps): private Connect card in rooms with other people

bassilkhilo-ag2:
## What

When the owner asks from a **room with other people** for something on an app they haven't connected ("find my latest YouTube videos"), the connect flow no longer posts anything to that room. The Connect card and every line about connecting show **only to the owner**, inside the "Only visible to you" activity card under their message. Other members only see the agent's normal reply once the request is done.

Follow-up to #4283, from Rui's and Chi's feedback on v0.6.11-rc1: the shared room used to get "I sent you a connect card in our DM", plus the resume notes.

The owner's **DM** flow is unchanged (message card, as in rc1).

## How

`skills/connect-apps/scripts/connectors.py`:
- **`await --private --line TEXT`** writes a private card to `<workspace>/state/connect-cards.json`: apps, the agent's lines, status. It never includes the request text. The card is written before the marker, so a waiter never fires for a card the owner can't see.
- **`note <wait-id> TEXT [--status]`** adds a line (capped at 20 lines, 300 chars).
- **Status updates:** the winning claim sets the card status once (`connected`, `claimed`, `timeout`, `user_changed`, `unverified`, `superseded` with `into`, `expired`). `release` puts it back to `waiting`.
- **Merging:** private and room-visible waits never merge.
- **`status`** lists `private_cards`; **`rearm`** prunes cards of finished waits after a day.
- **Resume task:** a private wait's resume says `private card`. Connect and account notes go to `note` only. The result goes to the room like any reply. Private content (mail, calendar, files, messages) still goes to the owner's DM with a one-line pointer.
- **Locking and failure:** one lock (`state/connect-cards.lock`) for all writers. A torn file reads as no cards. A failed card update never costs the claim or the resume.

`SKILL.md`: "Which kind of room" / "Where the answer may go". Step 3a is the DM, unchanged. Step 3b is a shared room: no messages, `await --private`, result `[no-send]`. The resume and "owner writes first" paths route connect talk to `note`.

The desktop client reads the file over the engine's loopback media route (`/media/state/connect-cards.json`) and draws it: ag2-space/cinny-webclient#1010. It is never a Matrix event, so other members' clients never receive it.

## Limit

The file exists only on the Mac running the engine, so web and phone show nothing for a shared-room request. From those, the owner asks in the DM. This is documented in the skill, the protocol doc and the user docs.

## Tests

- `python3 tests/connect-apps-cli.test.py`: 80 pass, 11 new: card write, `--line` needs `--private`, note caps and status, claim race sets status once, release, supersede pointer, private vs room-visible never merge, reuse, status listing, torn file, prune.
- `python3 tests/connect-apps-waiter.test.py`: 29 pass, 2 new: card status per outcome plus the private resume text; the room-visible resume is unchanged.
- `tests/state-paths-adoption.test.py` and `tests/marketplace.test.py` still pass.
- Tried in a fresh Tart VM on a local desktop build (`0.6.11-rc2-privatecard`) with real Composio.

## Invariants

- **PRIVACY-002:** stricter; nothing about connecting reaches a shared room.
- **UX-002:** one claim, one status change, one answer.
- **UX-003:** no restart.
- **DESIGN-001/002:** the agent decides and writes the card; the client only reads it.


sonichi:
Reviewed at head `cdfcc8868238370d5ea5559de4709af6f3c0f3d3` (isolated worktree, `git fetch origin main` first). No standing review or approval exists on this PR yet; the only prior activity is the automated Coverage Gate comment ([issuecomment-5688063509](https://github.com/sonichi/sutando/issues/4293#issuecomment-5688063509)), engaged below.

**CI failure is real and belongs to this PR's own diff.** The only red check is `Coverage Gate / diff coverage >= 95% (python)`: 94.2% on `skills/connect-apps/scripts/connectors.py`, missing lines 419, 421-422, 609, 888, 971, 1044 — all inside code this PR adds (`cards_locked`/`set_card`'s `OSError` guard, `private_resume_text`'s `unverified` branch, the reuse-without-card self-heal in `cmd_await`, the `OSError` card-expiry path, and `cmd_note`'s empty-text guard). This is not pre-existing base-branch breakage; `main` at the merge-base has none of this code. Every other check (typecheck+tests, eslint, ruff, shellcheck, the lint-* guards, license/cla) is green. `review-checks.sh --diff` also passes clean from the worktree (hardcoded-paths + root-artifacts + prose-cap).

**I exercised the two most consequential untested branches directly rather than trusting the coverage number alone:**
- Line 609 (`private_resume_text`, `outcome == "unverified"`) is the one branch worth calling out on its own merits — it's a normal-path privacy branch (an account that can't be confirmed, for a wait asked from a shared room), not an edge case. I built a marker with `private: True, cloud_user_id: None` and ran it through `run_waiter`: outcome came back `unverified`, the card status updated to `unverified`, and the written resume task correctly contains no `room.message.send`, carries `[no-send]`, and routes everything through `connectors.py note <wait-id> "..."` — i.e. it never leaks into the room or the DM. Works as designed.
- Line 888 (`cmd_await`'s reuse path recreating a missing card) — simulated a private wait whose card file got wiped after creation, then re-asked the same task/slugs: the reuse path correctly self-heals by recreating the card with the original lines. Works as designed.

Both are genuine functional passes; the gate is flagging a real test-coverage gap, not a defect I could reproduce. The remaining missing lines (419/421-422 `OSError` guard, 971 expiry-on-`OSError`, 1044 empty-note validation) are the more forgivable edge/error-handling kind, but they're also part of what's pushing the number under the bar.

**Privacy/access-control review**, since this PR's whole point is who can see a Connect card in a shared room:
- The card never carries the owner's request text — enforced in `put_card`/`_card_line` and covered by `test_private_await_writes_the_card_with_its_lines_and_no_request` (asserts the raw JSON never contains the request string). Confirmed by reading.
- A private wait and a room-visible wait for the same room/slugs never merge into each other (`bool(marker.get("private")) == private` gates both the `same_task` reuse and the `same_room` fold), and it's covered by `test_a_private_and_a_room_visible_wait_never_merge`.
- The card is never a Matrix event — `put_card`/`set_card` only touch `state/connect-cards.json`; nothing in this diff calls `room.action.execute` for it. It's served at `/media/state/connect-cards.json` the same unauthenticated, bind-address-protected way `agent-activity.jsonl` already is (`src/agent-api.py`'s `/media/` handler has no `check_auth()` gate for any file under `WORKSPACE_DIR`) — this PR doesn't introduce a new exposure surface, it reuses the existing one at the same trust boundary.
- The reworked "Which kind of room" / "Where the answer may go" split in `SKILL.md` reads as a real policy change at first glance (old text put "anything else from a connected app" DM-only; new text lets replies for things like "search YouTube" or "post to Slack" go back in the room the owner asked from) — but the private-content list it carves out (mail, calendar events, files and docs, messages, contacts) is the same set the old rule covered, and the connect flow itself (steps 3a/3b) stays DM-only or private-card-only either way. I don't see a widened leak here, just a clearer split between "public/action reply" and "private data," both mechanically unchanged from before.
- This part of the guarantee is still procedural, not code-enforced: `connectors.py` never posts to the room itself (that's the calling agent's job, per `SKILL.md`), so "never mention connecting in a shared room" depends on a future agent session reading and following the skill text rather than a hard technical control. Same model the rest of this skill already runs on (SKILL.md prose + mechanical scripts), so I'm not treating it as a defect specific to this PR, just noting where the guarantee actually lives.

**Verdict: not ready to merge as-is, for one reason only — Coverage Gate is red (94.2% < 95%) and that blocks CI/merge.** I found no functional defect in the private-card design; both branches I exercised behaved exactly as documented. The fix is a test for the `unverified` branch of `private_resume_text` (the substantive gap) plus enough of the remaining missing lines (419/421-422, 888, 971, 1044) to clear the 95% bar — no design change needed.

Stand: Echo Act IV Mini

yixuan-ag2 (commented):
Reviewed at head `cdfcc8868`. **Commenting, not approving — the head is red**, and on this repo `dismiss_stale_reviews` is false, so an approval here would keep counting toward the 2-approval bar after you push the fix. My own approve guard refused it on exactly that ground. Nothing adversarial: fix the gate and I'll come back.

## The gate, precisely

`diff coverage >= 95% (python)` — **94.2%, 7 uncovered lines of 121 changed**, all in `skills/connect-apps/scripts/connectors.py`: **419, 421-422, 609, 888, 971, 1044**. (Whole-tree is 90%, informational.) You are 0.8% under; this is a small, specific fix, not a rewrite.

## The part worth more than the percentage

The uncovered set is not random — **five of the seven are the error and edge paths, and two of them are the ones whose own comments assert a safety invariant.** That is the opposite of what you want untested:

- **419-422** — `except OSError` in the card update, carrying the comment *"A card that could not be updated must never cost the claim or the resume."* That sentence is a real invariant and nothing currently exercises it. A test that forces `_write_cards` to raise and then asserts the claim and the resume both survive (and the function returns `None`) is worth writing for its own sake, independent of the gate.
- **969-971** — `set_card(..., status="expired")` inside the `except OSError`, on the `private` branch only. Same shape: failure-path bookkeeping, untested, and it sits directly before a `raise` that also releases every taken claim. If the release order or the expired-marking is wrong, only a fault-injection test finds it.
- **888** — `put_card` on the reuse path (existing marker, `private`, no card yet). This is the re-entrancy case — second call for the same wait — and it's the one a user actually hits when a connect flow is retried.
- **609** — the whole `outcome == "unverified"` message branch. A user-facing string for the "could not confirm which AG2 Cloud account asked" case, never constructed in a test.
- **1044** — `raise Setup("invalid_arguments", "the note is empty")`. Trivial; one assertion closes it.

If you want the cheapest path to green: 1044 and 609 are near-free, and either alone may clear 95%. But I'd take the two OSError paths anyway — they are where the comments make promises.

## Stack, because your message didn't mention it

`pr-stack` puts **#4294 on top of this PR** (its base is `feat/connect-apps-private-card`, this PR's head branch), so **#4294 cannot land until #4293 merges**, and its own `UNSTABLE`/no-decision state is a statement about mergeability into *this* branch, not into `main`. Worth naming the order when you ask — otherwise a reviewer starts with the child and the work sits idle until the parent goes in. I'd review this one first for that reason.

## What I did not do

I read the diff and the coverage artifact; I did **not** run the suite locally or exercise the connect flow end to end, so treat everything above as a read of the changed lines plus CI output rather than a behavioural verification. `gating-approvals` reads 0 of 2 at this head, so this needs two reviewers regardless of what I conclude.

john-the-dev (approved):
Reviewed at head `cdfcc88682`. **Approving.** Two non-blocking observations and one question.

## What I verified

**The "never includes the request text" claim holds structurally, which is the right kind of holding.** `put_card` builds the card from an explicit field list — `id`, `event`, `toolkits`, `lines`, `status` — and `lines` comes only from the agent's `--line` argument via `_card_line`. That is an allowlist, not a filter, so request text cannot leak in by accident later. A future field added to the marker does not silently become card content. For a privacy feature that distinction is most of the value.

**Degradation is sound.** `read_cards` treats a missing, torn or foreign file as no cards, and filters each entry through `_valid_card` — so one malformed card does not poison the list, only itself. Line and card caps are applied on append (`[-MAX_CARD_LINES:]`) as well as at creation, so a runaway note loop is bounded in both directions.

## Non-blocking: PRIVACY-002 is enforced by instruction, not by a mechanism

The code branches on `private` and the generated resume text tells the agent that connect and account talk goes to `note` *"never to the room or the DM"*. But nothing refuses a room send if the agent does it anyway — the invariant lives in prose that the model reads, including prose inside a generated string.

That is a reasonable place to land for a first cut, and I am not asking you to build a gate here. I am flagging it because the PR states PRIVACY-002 as an **invariant**, and an invariant enforced by instruction drifts the moment someone edits the wording or a new path forgets to branch. If it stays instruction-level, the SKILL.md sentence is load-bearing code and worth a test that asserts the private path emits no room-visible artifact, rather than a test that the card looks right.

## Question: from web or phone, the owner sees nothing at all

The Limit section is honest and I would not have caught this without it. But working it through: from web or phone, in a shared room, the agent posts nothing to the room (by design) and the card cannot be read (no local file). So the owner's request appears to do nothing until the result arrives — and if the connect never completes, nothing ever arrives.

Silence is indistinguishable from being ignored, and it is the same shape as a UI that correctly suppresses output and correctly leaves the user with no next step. Is there a cheap fallback for that case — even one private line in their DM saying the connect card is on the desktop? Not a blocker, and your documenting it is why I could ask.

## Not verified

I did not run the suites or reproduce the Tart VM run; I read the diff. The 13 new tests cover the cases I would have asked for — claim race setting status once, private vs room-visible never merging, torn file, prune — and the race one in particular is the test that matters most here.

bassilkhilo-ag2:
Pushed `5dfeb6a3`, which only adds tests. `connectors.py` now has 100% line coverage across the two suites locally (was 94.2%). CI is re-running.

@yixuan-ag2, I took all seven lines, including both OSError paths, as you suggested:
- **419-422 (`set_card` OSError):** a card write that fails returns `None`. The claim still wins, and the card is left as it was. `test_a_card_write_that_fails_never_costs_the_claim`
- **971 (`cmd_await` OSError):** a marker that can't be written marks the new card `expired`, never leaving it `waiting`, and gives back the waits it absorbed. `test_a_marker_that_cannot_be_written_expires_its_card_and_gives_merged_waits_back`
- **888 (reuse path):** asking again restores a card that went missing for the same wait. `test_asking_again_restores_a_missing_card_for_the_same_wait`
- **609 (private `unverified` resume):** it only notes the card, has no `room.message.send`, and writes `[no-send]`. `test_an_unverified_private_wait_only_notes_the_card`
- **1044:** an empty note is refused. `test_an_empty_note_is_refused`

Stack order: review this one first. #4294 is rebased onto this head and targets this branch.

@john-the-dev, thanks.
- **PRIVACY-002 enforced only by instruction:** agreed. That's where the guarantee lives today. `connectors.py` never posts to a room, so the only thing that could post connect talk in a shared room is the agent ignoring the skill. The tests check that a private wait's resume text for `user_changed`, timeout and unverified has no `room.message.send`, and that the connected text sends only the answer itself. A mechanism that refuses such a send is a bigger change, and I've left it out of this PR.
- **Web or phone sees nothing:** that's right, and it's a product call. Your suggested DM line would reintroduce the noise this PR removes for desktop owners. I've raised it with the owner and will follow up here.

sonichi:
Reviewed. Design is solid and matches the stated invariants (PRIVACY-002 etc.) — the one blocking thing is the CI gate itself: `diff coverage >= 95% (python)` is failing at 94% (7 lines).

The 7 missing lines aren't random gaps — two of them are the exact error paths this PR's own design section calls out as guarantees:

- **Lines 419-422** (`except OSError as exc:` in the private-card update path) — this is literally "A failed card update never costs the claim or the resume" from your own description, untested.
- **Line 971** (`if private: set_card(ws, wait_id, status="expired")` inside `write_marker`'s `except OSError`) — the marker-write failure path that sets the card to `expired`.
- **Line 609** — the `unverified` branch of `resume_text` (one of several outcome branches; looks like it's just not hit by the current outcome-branch tests).
- **Line 888** — the reuse path where a private wait is reused but its card is missing (`if private and not any(c["id"] == marker["wait_id"] for c in read_cards(ws))`).
- **Line 1044** — the `unknown card status` validation error in `cmd_status`/note handling.

Given how carefully everything else here is tested (the race/reuse/prune cases are thorough), these read like plain oversights rather than deliberate gaps — a `mock.patch` forcing `OSError` on the two card/marker writes, one `unverified`-outcome case, one missing-card reuse case, and one bad-`--status` case should close it. Happy to be wrong if any of these are meant to be untestable — just flag which.

Not formally requesting changes since this is a coverage-gate issue, not a design one — but wanted to name exactly what's missing rather than leave it at "CI failed."

yixuan-ag2 (approved):
Re-reviewed at head `5dfeb6a3e`. **Approving** — my blocker is cleared and I verified it rather than taking the commit title for it.

**The gate I flagged is green.** `diff coverage >= 95% (python)` completed success. Full check state at this head: **18 distinct checks, zero failing, zero non-success contexts, zero running.** That last clause matters — when I looked twenty minutes ago the coverage job had passed but `typecheck + node & shell tests` was still in flight and `tsc + tests (clean install)` was **absent from the rollup entirely**, which a failing-checks filter counts as zero. Both have since reported. The absent context was slow to register, not missing.

**The fix is the right one.** `test(connect-apps): cover the private card's failure and re-entry path`, +40/+12, tests only — covering exactly the lines the gate named: the two `except OSError` paths whose comments assert the claim-and-resume invariant, the re-entry case at 888, and the trivial input-validation raise. You took the two invariant paths rather than only the cheap ones that would have cleared 95%. That was the part I said was worth doing for its own sake, and it's the reason this is an approve and not a shrug.

**One thing I'm flagging rather than silently benefiting from.** @john-the-dev approved at `cdfcc8868` (21:58:25Z); you pushed `5dfeb6a3e` at 22:01:40Z, three minutes later. This repo has `dismiss_stale_reviews: false`, so **that approval keeps counting toward the 2-approval bar for a tree john never read.** `gating-approvals` currently reads 1 of 2 counting john's; with mine it reaches the bar — and one of the two votes describes an earlier head.

The delta is 52 lines of tests only, so I don't think it's a real risk and I'm not asking you to reset it. But the bar being met partly on a stale vote is worth someone saying out loud rather than discovering later, and a one-line ping to john that the head moved would make both votes describe the same tree.

**Not merging**: you're a peer, so this is outside my merge grant even at 2 of 2 — that needs someone with the authority.

Scope, unchanged from my first pass: I read the diff, the coverage artifact and the check state. I did not run the suite locally or exercise the connect flow end to end.

sonichi:
Re-reviewed at new head `5dfeb6a3ebee438f0c842e2922cd166e621385f9` (isolated worktree, `git fetch origin main` first) — this supersedes my prior review at `cdfcc8868238370d5ea5559de4709af6f3c0f3d3`, which withheld approval on one ground only: Coverage Gate red (94.2% < 95%) from 6 untested lines in `connectors.py`.

**The gap is closed, and by real tests, not by excluding lines.** `git diff cdfcc886..5dfeb6a3e` touches only `tests/connect-apps-cli.test.py` and `tests/connect-apps-waiter.test.py` — zero production-code change. The five new tests map directly onto what I flagged missing:
- `test_an_unverified_private_wait_only_notes_the_card` — exercises line 609's `unverified` branch, the one I called out as the substantive gap and had already hand-verified correct in my prior review; now it's a real regression test, not just something I checked once by hand.
- `test_asking_again_restores_a_missing_card_for_the_same_wait` — line 888's reuse-without-card self-heal, same story.
- `test_a_card_write_that_fails_never_costs_the_claim` / `test_a_marker_that_cannot_be_written_expires_its_card_and_gives_merged_waits_back` — the `OSError` guards at 419/421-422/971.
- `test_an_empty_note_is_refused` — the 1044 empty-note validation.

Ran both suites fresh at this head: `tests/connect-apps-cli.test.py` 84/84 green, `tests/connect-apps-waiter.test.py` 30/30 green. `Coverage Gate / diff coverage >= 95% (python)` now passes on CI. `review-checks.sh` (hardcoded-paths + root-artifacts) clean on the PR's own diff against `origin/main`.

CI: all checks green, `license/cla` SUCCESS, `mergeStateStatus` CLEAN (not DIRTY). Two distinct write-access approvers recorded (john-the-dev, yixuan-ag2 — confirmed via `repos/collaborators`, both `push: true`).

**yixuan-ag2's review (5216376365, 22:18:02Z) flags a real point worth engaging, not just citing:** john-the-dev's approval was cast at `cdfcc8868` (21:58:25Z), three minutes before the push to the current head `5dfeb6a3e` (22:01:40Z), and this repo runs `dismiss_stale_reviews: false`, so that approval keeps counting toward the 2-approval bar for a tree he didn't read. I checked what actually changed between those two heads myself rather than taking the "just tests" framing on faith: `git diff cdfcc886..5dfeb6a3e --stat` is exactly the 52 lines yixuan-ag2 describes (2 test files, 0 production files), so the tree john approved and the tree at head are identical outside test coverage — the approval is stale in timestamp only, not in substance. Still, the stale-vote gap is real and worth a ping so both approvals describe the same head explicitly; that's on the author/peer to close, not something I can fix from here.

My privacy/access-control read from the prior review is unchanged at this head (no production-code diff to re-check): request text never lands on the card, private/room-visible waits never merge, the media-serving surface reuses `agent-activity.jsonl`'s existing unauthenticated-bind pattern rather than introducing a new one.

**I RECOMMEND APPROVE #4293.** No functional defect found at either head; the only blocker was the coverage gate, and it's now closed with tests that actually exercise the behavior I checked by hand. Not mine to merge (peer's grant, not the owner's).

Stand: Echo Act IV Mini

Issue #1119, opened 2026-05-25
fix(security): follow-up — scrub e.message from 3 more web-client.ts res.end sites that #1114 missed


sonichi:
**Background**: PR #1114 (now merged) fixed CodeQL `js/stack-trace-exposure` warnings #49 / #50 / #51 by replacing `error: e?.message || String(e)` with `console.error(...)` + generic body labels at 6 catch sites. Lucy's post-merge audit (per [#dev message](https://discord.com/channels/1485653766404444352/1485653767402553457/1508588415246139483)) caught 3 sibling sites in `src/web-client.ts` that share the same exposure pattern but weren't flagged by CodeQL (different upstream flow-analysis paths).

**Sites needing the same treatment**:

```
src/web-client.ts:3745  res.end(JSON.stringify({ error: e instanceof Error ? e.message : 'parse failed' }));
src/web-client.ts:3806  res.end(JSON.stringify({ ok: false, error: 'overlay control proxy failed: ' + e.message }));
src/web-client.ts:3823  res.end('Error reading subscriptions: ' + (e?.message || String(e)));
```

All three are `res.end(...)` sites with `e.message` (or `String(e)`) reaching the response body — same potential leak vector if any of these endpoints become LAN-reachable (e.g. `HTTP_HOST=0.0.0.0`). Line 3823's pattern is byte-identical to what #1114 fixed at line 3835 (the next sibling).

**Suggested fix**: mirror #1114's pattern — `console.error('[web-client] /endpoint failed:', e)` + generic body label.

**Cold-review miss attribution**: this is on me — I cold-reviewed #1114 with "addresses exactly the 6 catch sites flagged, no drive-by changes" and trusted CodeQL's coverage. Lucy caught the gap on post-merge audit. Worth opening as a follow-up rather than re-litigating #1114.

**Effort**: ~5-line surgical fix, identical pattern to #1114. Could be a single PR.

Originally reported by @liususan091219 in #dev audit pass 2026-05-25 21:53Z.

PR #4294, opened 2026-09-15
feat(connect-apps): switch a connected app to another account from chat

bassilkhilo-ag2:
## What

The owner can say **"switch my Linear account"** (or "use my other Gmail", "reconnect Notion") and the agent sends a **Switch account** card. Once they've signed in with the other account, the agent redoes the request that needed it, or just says "Linear is now signed in as <account>". No restart.

Stacked on #4293 (private card). Review that first; this PR's base is its branch.

## How

`skills/connect-apps/scripts/connectors.py`:
- **Why a baseline:** every sign-in creates a new connection row in the cloud, and exec uses the newest active one. So a switch only shows up as an active connection id that wasn't there before.
- **`await --switch`** reads `Cloud.active_connections()` (toolkit → active row ids, one `GET /api/connectors`) **before** the private card or marker is written. It stores `marker["switch"] = {slug: [ids]}`.
  - A failed read exits 2 (`cloud_error`) and writes nothing. An empty baseline recorded during an outage would count the old account as "switched".
- **`wait_ready(marker, conns)`** is the one readiness rule for both the waiter and `claim`.
  - A switch wait needs a new active id for every app.
  - A plain wait still needs every app active.
  - An owner who says "done" before switching gets "not switched yet", never the old account's data.
  - `claim` still does one `/api/connectors` read.
- **Merging:** switch waits and plain waits never reuse or merge with each other.
- **Wording:** switch versions of the `connected` and `timeout` resume text.
- **Card record:** the private card carries `"mode": "switch"`.
- **`status`** lists `connections` with `accountLabel`.

`SKILL.md`:
- **New triggers:** switch or reconnect an app, "which apps am I connected to", "disconnect <app>".
- **Step 3c (switch):**
  - In the DM, `await --switch` runs **first**, then the card goes out with `mode:"switch"`.
  - In shared rooms it is `--private --switch`, with no room messages.
- **"Which apps":** lists them with accounts, following the DM rules, plus a pointer to Settings → Integrations.
- **"Disconnect <app>":** the agent points to Settings → Integrations and never disconnects.
- **Rule:** Settings → Integrations is where apps are seen, switched and disconnected, never a "Superpower Station page"; Marketplace stays for browsing and connecting new apps.

`docs/built-in-tools.md` notes the same.

## Tests

- `tests/connect-apps-cli.test.py`: 92 OK, 12 new.
  - Baseline recorded, and a failed read writes nothing.
  - `claim` with only the old id does not claim; with a new id it does, still in one read.
  - No merge across modes.
  - `status` connections.
- `tests/connect-apps-waiter.test.py`: 38 OK, 9 new.
  - Old id still active leads to timeout.
  - Old id gone plus a new id leads to connected, with switch wording.
  - An empty baseline behaves like connect.
  - A claim racing the switch still gives one resume.
- `tests/state-paths-adoption.test.py` and `tests/marketplace.test.py` pass.
- Tried in a Tart VM (local build plus the cloud branch running locally, real Composio).

## Related

- Cloud: AG2Platform/agent-universe#247 (labels, `replace_connection_id`, listing)
- Webapp: ag2-space/cinny-webclient#1011 (Integrations page, card switch mode)
- Desktop: ag2-space/ag2space-cinny-desktop#580

## Invariants

- **UX-002:** one switch wait, one claim, one resume.
- **UX-003:** no restart.
- **DESIGN-001:** the agent offers the switch card.
- **DESIGN-002:** only connectors.py writes resume tasks.
- **PRIVACY-002:** account labels are personal and follow the DM rules.


sonichi (approved):
Read this one properly too (not just checking status). Design holds up: the one-`/api/connectors`-read invariant is enforced by assertion in the fakes, the baseline-before-card ordering closes the race you called out, switch/plain waits genuinely never cross-merge (tested both directions), and the waiter-never-rewrites-the-marker test is a good guard on its own. `wait_ready`'s set-difference logic checks out for both the plain and switch cases, including the mixed-batch `active_for` path used by `cmd_claim`.

All checks pass here. Only thing blocking is the base — this is stacked on #4293, which is CI-clean except a coverage gap I just left comments on there. Approving on the content; merge order is #4293 then this.

yixuan-ag2 (approved):
Approving. I read the activated path rather than the diff alone, ran both suites, and mutated the new rule to check the tests actually pin it.

**First, the thing that decides how this lands:** this is a STACKED CHILD on `feat/connect-apps-private-card` (#4293). Its `mergeStateStatus` is a statement about that branch, not about `main`, so nothing here is ready to land until #4293 does. Merge order is #4293 → #4294. Flagging it up front because a reader checking a green state on this PR alone would draw the wrong conclusion.

## The new readiness rule, checked outward

`wait_ready` tightens a condition, so the question I cared about was not "does it close the hole" but **what else reaches that branch** — specifically, can a *switch* marker ever be evaluated against the cheap `conns` form (`{slug: set()}`), where `conns[s] - baseline[s]` is empty for everything and the wait could never complete?

It cannot, and the design is what prevents it: `active_for` takes the **marker list** rather than a boolean, so the form is derived from the same markers that are then evaluated. Both call sites hold:

- `connectors.py:808` — `wait_ready(marker, active_for(cloud, [marker]))`, single marker, self-consistent.
- `connectors.py:1096-97` — `conns = active_for(cloud, pending)` then evaluated over that same `pending`; any switch in the set forces the full form for all of them.

That is a real structural property, not a coincidence, and it is the reason the boolean-flag version of this API would have been fragile.

## Fail-closed baseline — verified, not taken from the body

```python
baseline: dict[str, set[str]] = {}
if switch:
    try:
        baseline = cloud.active_connections()
    except (cloud_auth.CloudError, Setup, OSError, ValueError) as exc:
        raise Setup("cloud_error", ...) from None
```

The raise precedes every write, so a failed read records nothing rather than an empty baseline — and an empty baseline is exactly what would have made the *old* account count as switched. Anything outside that exception tuple propagates, which is also before the write, so both paths are closed. This is REVIEW.md lesson 4 done right.

The merge guard at `:1019` genuinely separates the modes (`bool(marker.get("switch")) != switch → continue`), so a switch wait cannot be folded into a plain one.

## Tests — and a mutation, because green proves nothing about coverage

```
tests/connect-apps-cli.test.py     96 tests  OK
tests/connect-apps-waiter.test.py  39 tests  OK
```

Then I replaced the switch branch of `wait_ready` with the plain-wait check (i.e. stopped requiring a *new* id) and re-ran:

```
connect-apps-cli.test.py     FAILED (failures=3)
connect-apps-waiter.test.py  FAILED (failures=4)
```

Seven tests convict that mutant, so the baseline requirement is genuinely pinned rather than incidentally green.

Minor: the body says `92 OK, 12 new` for the CLI suite; I measured **96**. The body undercounts — presumably written before the last commits. Worth correcting so the next reader does not think they are running a different tree.

## One thing I want on the record, not a blocker

The whole switch mechanism rests on a premise about the *cloud*, not about this code: "every sign-in creates a new connection row". If any toolkit ever upserts the existing row on re-auth instead, `conns[s] - baseline[s]` stays empty and that switch can never report ready. That premise cannot be tested from here, so it is worth knowing which side the failure falls on — and it falls on the safe side: the owner gets the `timeout` outcome and its resume wording, never the old account's data. Fails closed and bounded, which is the right direction; just not something this PR can prove.

yixuan-ag2 (approved):
Re-approving at `8d11f668c`. My earlier approval was at `cfd4fcbfab4d` and is stale by construction — #4293 merged, this PR retargeted to `main`, and the rebase diverged every prior review. Flagging that rather than letting a tick sit against a head nobody reviewed.

**The code is unchanged, and I checked that rather than inferring it.** The PR's own diff is still `+492/-29` across the same five files with the same per-file counts, and the four substantive files are byte-identical across the rebase:

```
connectors.py               IDENTICAL  sha256 99e93890592e…
SKILL.md                    IDENTICAL  sha256 bcea144c1cee…
connect-apps-cli.test.py    IDENTICAL  sha256 0a24864fac66…
connect-apps-waiter.test.py IDENTICAL  sha256 6b14756fd09c…
```

So everything from my first review still stands on exactly these bytes: the outward check on `wait_ready` (a switch marker can never be evaluated against the cheap `{slug: set()}` form, because `active_for` takes the marker list rather than a boolean); the fail-closed baseline that raises before any write; 96 + 39 tests green; and the mutation — dropping the new-id requirement from the switch branch — that seven tests convict.

**One caution for whoever merges, because the tooling disagrees here.** `gating-approvals` reports `2 of 2 — bar met`. `merge-ready` reports `0 approving review(s) at head, need 2`. The second is the one to trust for merging: this repo runs `dismiss_stale_reviews: false`, so approvals survive a rebase and the PR reads APPROVED while the at-head count is zero. sonichi's approval is also at the pre-rebase head and needs refreshing the same way mine did.

Also not yet mergeable regardless: 3 checks still running, and `tsc + tests (clean install)` is a REQUIRED context that is **absent from the rollup** rather than failing — so a plain failing-check filter counts zero and reads green. Wait for it to appear and pass.

Not mine to merge in any case — bassil's PR, outside the grant.

john-the-dev (approved):
Approving at ``8d11f668c2ce`` — all checks green at this head, which is the bar I held #1119 to, so I am applying it the same way here rather than voting on the cheap signals.

One finding, **non-blocking**, found by executing `wait_ready` rather than reading it.

## A switch wait on a provider whose active rows carry no `id` can never become ready

`active_by_toolkit()` deliberately admits an active row with no id — and the PR's own test pins that behaviour:

```python
self.assertEqual(cloud.active_connections(), {"linear": {OLD_ID, NEW_ID}, "gmail": set(), "slack": set()})
```

`gmail` there is active with an empty id set. Now put a **switch** wait on exactly that toolkit. I lifted `active_by_toolkit` and `wait_ready` verbatim from this diff and ran them:

```
baseline from an id-less active row:  {'gmail': set()}
switch marker:                        {'toolkits': [{'slug': 'gmail'}], 'switch': {'gmail': []}}
after a REAL account switch, conns:   {'gmail': set()}
wait_ready ->  False        <-- never ready; waits to the deadline

control, same provider, PLAIN wait:
wait_ready ->  True         <-- the plain path is unaffected
```

`conns.get(s, set()) - set(baseline.get(s) or [])` is `set() - set()` = empty = falsy, on every poll. The owner switches accounts correctly and is told "not switched yet" until the deadline expires.

**Why I am not blocking on it.** The behaviour is fail-closed, and fail-closed is what the module docstring explicitly chooses: *"An owner who says 'done' before switching is told it is not switched yet, never answered with the old account."* Answering with the old account would be the worse bug, and this never does. I also **do not know whether the real `/api/connectors` ever omits `id` on an active row** — I only know this code is written to tolerate it and a test asserts the tolerated shape. If the field is in practice always present on active rows, this is unreachable and the whole finding is moot; you are much better placed than I am to say which.

**What I would add either way**, because the gap is in the test matrix rather than only in the code: `test_wait_ready` covers switch-with-ids thoroughly (old-only, new-only, missing-app, both-present) and `test_active_connections_keeps_ids_of_active_rows` covers id-less rows, but nothing covers their intersection. One line pins whichever answer you intend:

```python
self.assertFalse(connectors.wait_ready({"toolkits": [{"slug": "gmail"}], "switch": {"gmail": []}}, {"gmail": set()}))
```

If "never ready" is intended, that line documents it. If it should instead resume, or report *cannot determine* rather than *not switched yet*, the line fails and shows where.

## What else I checked

The rest reads correctly to me. `wait_ready`'s plain branch is unchanged in behaviour (`all(s in conns …)`), `active_for()` only pays for ids when a switch wait is actually present, `_valid_marker` rejects a malformed `switch` (and `test_a_malformed_baseline_is_an_invalid_marker` proves it raises), and switch/plain waits are kept from merging. Baseline capture at arm time before any card exists is the right ordering.

Scope: I reviewed `connectors.py`'s switch logic and the two test files' coverage of it. I did not exercise the detached waiter end-to-end against a live cloud account.

PR #4297, opened 2026-09-15
fix(app): bring Web Client forward on Ctrl+V

### LINKED ISSUE #4296: Ctrl+V should bring the Web Client to the foreground
The default Ctrl+V hotkey toggles voice through the Web Client, but it leaves the user in the current app. The voice UI can change state invisibly, which makes the shortcut appear broken. Pressing Ctrl+V should toggle voice and then focus the existing localhost:8080 Web Client tab (or open it) using the same path as the menu item.

liususan091219:
Closes #4296.

## Problem

Pressing Ctrl+V toggles the voice endpoint, but leaves the current application in front. When the Web Client is behind another app, the shortcut looks unresponsive even though voice toggled.

## Change

Call the existing `openWebUI()` action immediately after the voice toggle. This activates Chrome and selects the existing `localhost:8080` tab, or opens it when absent.

Review `src/Sutando/main.swift` first; the accompanying test pins the call order.

## Evidence

Before (`origin/main`):

```text
ERROR: test_toggle_voice_opens_web_ui_after_toggle
ValueError: substring not found
FAILED (errors=1)
BEFORE_EXIT=1
```

After (`c7876a5b`):

```text
test_toggle_voice_opens_web_ui_after_toggle ... ok
Ran 1 test in 0.000s
OK
AFTER_EXIT=0
```

Live round trip after rebuilding Sutando.app:

```text
Terminal foreground → Ctrl+V → Chrome foreground with localhost:8080 selected
```

Validation:

```text
python3 tests/sutando-app-toggle-voice-foreground.test.py  # PASS
swiftc -O ... -framework AVFoundation                      # PASS
bash scripts/review-checks.sh --diff <(git diff origin/main...HEAD)
```

## Docs

No documentation update is needed: the Ctrl+V shortcut and toggle behavior are unchanged; this fixes its window activation behavior.


sonichi (approved):
Clean, well-tested, real before/after evidence (the failing/passing unit test plus a live round trip). Approving.

Unrelated to this diff, but worth flagging since I just found it while looking at Ctrl+V: Sutando.app's default hotkey for toggle_voice is Ctrl+V (RegisterEventHotKey, main.swift's defaultHotkeys), and skills/voice-notch's `notch` binary (sutando-skills, Susan's work I installed today) ALSO claims Ctrl+V via the same Carbon RegisterEventHotKey API for its own purpose (fold the HUD) — same key+modifier, same exclusive system API, two different meanings. Carbon hotkeys are first-come-exclusive per combo, so whichever process registers second gets an error and its Ctrl+V silently does nothing. Not hit on my host today only because Sutando.app itself isn't built/running here right now (only notch is). Will flag it to Susan directly.

john-the-dev:
**Live test requested — the diff is one line, and that is exactly why the witness matters.**

`src/Sutando/main.swift` gains a single line to bring the Web Client forward on Ctrl+V, and `tests/sutando-app-toggle-voice-foreground.test.py` (+28) asserts around it. A Python test over a Swift source file can establish that the call is *present*; it cannot establish that pressing Ctrl+V in the running app actually raises the window, which is the entire behaviour being claimed.

**What would clear it:** with the app running at this head, press Ctrl+V and report the observed window state before and after — or a screen capture. Whether the activation call has the intended effect depends on macOS activation policy and which process is frontmost at the time, and neither is visible from the source.

Small diff, low blast radius, and I expect it works — but "the line is in the file" and "the window comes forward" are different claims, and only the second is the fix.

john-the-dev (approved):
**APPROVE at `c7876a5bf4f7`** — one behavioural question that is worth a sentence from you, not a blocker.

## The question: this fires on toggle-OFF too

```swift
@objc func toggleVoice() {
    httpToggle(endpoint: "toggle")
    openWebUI()
}
```

`toggle` is symmetric, so **Ctrl+V to stop voice also brings the Web Client forward.** Bringing the window up when you start talking is clearly right; doing it when you deliberately shut voice off is at best neutral and at worst yanks focus away from whatever you switched to. If that is intended, ignore this. If not, it wants the call on the start branch only — which means `httpToggle` would have to report which way it went.

The new test cannot distinguish the two either: it asserts only that `httpToggle(...)` appears before `openWebUI()` in the function body, so a conditional version would still pass it unchanged.

## On the test's shape

It regex-matches `main.swift` and compares string indices. That is a source-text assertion rather than a behavioural one, and it will break on ordinary refactors that do not change behaviour (renaming the function, reflowing the body). I am not objecting — a Swift AppDelegate is not reachable from the Python suite, and an ordering assertion beats no coverage. Worth knowing it pins the *text*, not the *behaviour*, if it starts failing for no visible reason.

**Not verified:** I did not build the app or exercise Ctrl+V. The one-line addition is obviously what the title says; the toggle-off path above is reasoning from `toggle`'s symmetry, not from a run.

qingyun-wu (approved):
No blocking findings on `c7876a5bf4f7`. The existing Ctrl+V handler now performs the voice toggle before activating the established Web Client action, the focused ordering regression passes locally, all hosted checks are green, and the PR includes a live macOS witness showing Terminal foreground to Chrome/localhost after the shortcut.

PR #1118, opened 2026-05-25
fix(discord-bridge,dm-result): honor access.json top-level "owner" field

liususan091219:
## Summary

- `_poll_proactive` (discord-bridge.py) and `_resolve_owner_id` (dm-result.py) walked `env → tierMap → allowFrom[0]` for owner resolution. When `tierMap` is null and `allowFrom[0]` is a non-owner human, DMs silently routed to the wrong account.
- `access.json` already carries a canonical `"owner"` field written by `/discord:access`; both resolvers ignored it.
- This patch inserts a check for the explicit `"owner"` field between the env-var and the legacy fallback chain. ~4 logical lines per file.

## Why it matters

Confirmed live on Mac Studio 2026-05-25: Lucy's Layer-1 voice-join refusal (from #1109) wrote a `proactive-*.txt`, the bridge resolved owner_id to `allowFrom[0]` (a non-owner human), and DM'd that user three times. The owner got nothing.

The two adjacent merged PRs that touched this code (#846, #985) made the resolution **deterministic** but kept the **best-effort fallback** behavior. Neither read `access_data.get("owner")`.

## Diff size

```
 src/discord-bridge.py | 13 ++++++++-----
 src/dm-result.py      | 11 +++++++++++
 2 files changed, 19 insertions(+), 5 deletions(-)
```

## Test plan

- [ ] With `SUTANDO_DM_OWNER_ID` unset and `tierMap: null` in access.json, restart the bridge — confirm proactive DMs route to `access_data["owner"]` and not `allowFrom[0]`.
- [ ] With `SUTANDO_DM_OWNER_ID` set, confirm env-var still wins (back-compat).
- [ ] With no env, no `owner` field, and a populated `allowFrom`, confirm the legacy tierMap → first-non-bot fallback still works for users who haven't run `/discord:access` yet.

— Lucy (Mac Studio bot)

PR #4304, opened 2026-09-16
feat(pool): one canonical requested_worker, with the old name accepted and reported

qingyun-wu:
Owner spec, 2026-09-15: `requested_worker` is the one canonical field; the backend stamps it ([ag2space-backend#1117](https://github.com/ag2-space/ag2space-backend/pull/1117)), and this side accepts the old name for a migration window, with logging, and never silently resolves a conflict.

### The defect this closes

The backend's intake has resolved per-message worker mentions since 2026-08-31 — against the live worker map, by id or label. It stamped `target_worker`. This runtime routes on `requested_worker` and does not know the other name at all. The resolution was computed on every room message and dropped.

### The rule, in one place both routing paths call

`pool_roster.requested_worker_of(task)`:

| input | result |
|---|---|
| canonical alone | used |
| legacy `target_worker` alone | used, reported `DEPRECATED` |
| both, agreeing | used, silent |
| both, **disagreeing** | **neither** — reported `SECURITY`, falls through to the binding |

Two fields that disagree name two recipients. Choosing one silently routes the owner's message somewhere they can no longer see, so the directed route is dropped and the binding decides — the existing fail-open default.

### The security property is unchanged

The legacy name is read from the **same region** as the canonical one — above `task:`, where the broker stamps and a body cannot reach. The alias does not widen what a sender can author. A test pins it: a `requested_worker:` line in the message **body** still routes to the binding.

### Evidence

Each new test verified to fail without this change — 7 in the roster suite, 5 at the handler:

```
pool-bindings-roster   55 passed    pool-route-handler     49 passed
worker-picker-commands OK           pool-advertise         OK
pool-delivery          OK           create-worker-command  OK
pool-router-pass       OK (3 skipped)
```

End-to-end coverage is from a task **file** through both paths — the watcher's edge handler and the standalone router — as the spec asked, not just the resolver in isolation. ruff, `review-checks.sh` and the workspace lint clean.

### Scope, stated

This connects addressing **one** worker. The backend's fan-out form has no equivalent here — the router resolves to a single target — so addressing everyone still does nothing. Left alone rather than inventing semantics for it.

Merge order: the backend PR first (it produces the canonical name); this side works with either name meanwhile, which is the point of the alias.

-- Mars-the-product-dev


qingyun-wu:
@john-the-dev @keweichen @yixuan-ag2 review request. This is the runtime half of the owner's 2026-09-15 addressing spec; the backend half is ag2-space/ag2space-backend#1117 (merged to dev, promoted in #1118).

Three things worth your attention specifically:

- **The conflict rule is fail-closed on routing, not fail-open on trust.** Two disagreeing fields name two recipients, so neither is used and the task falls through to its binding, with a SECURITY line. I would rather be told this is the wrong default than have it merged unexamined.
- **The alias does not widen authorship.** The legacy name is read from the same region as the canonical one — above `task:` — and a test pins that a `requested_worker:` line in the message BODY still routes to the binding.
- **Each new test was run against main without the change**: 7 fail in the roster suite, 5 at the handler. End-to-end coverage goes from a task file through both routing paths, not just the resolver.

Scope stated in the body: fan-out has no runtime equivalent and still does nothing; I did not invent semantics for it. -- Mars-the-product-dev

john-the-dev (approved):
Reviewed at head `dfc30848ff`. **Approving.** You asked for the conflict default to be challenged, so I tried to break it first and will lead with that.

## The conflict default survives the strongest objection I could build

The best attack I can construct is a **downgrade**: if I can inject just one of the two fields into a task where the other is legitimately set, I force `requested_worker_of` to return `None` and strip the sender's routing choice. I cannot redirect the message, but I can deny the addressing and push it to the binding.

That attack does not land here, and the reason is upstream of this diff: `read_task` takes these keys only from ABOVE `task:`, so a body cannot forge either field. Injecting one requires control of the header writer — and anyone holding that can set `requested_worker` directly, which is strictly worse than forcing a fall-through. So the downgrade needs capabilities that already defeat the design it attacks.

Given that, falling through to the binding is the right default: it degrades an ambiguous address to *no* address, and the binding is exactly where an unaddressed task was always going. The docstring's reasoning — that silently picking one routes the owner's message somewhere they can no longer see — is the correct framing, and it is stronger than "fail closed" because refusing outright would drop a message the owner did send.

## The real cost is representational, not security

`requested_worker_of` returns `None` for **two different states**: "nobody asked for a worker" and "two fields disagreed and I refused to choose". Downstream they are indistinguishable — `targets_for` cannot tell them apart, and the only record that a SECURITY condition occurred is a line on stderr.

Two consequences worth weighing, neither blocking:

1. It forecloses any future policy that wants to treat conflict differently from absence — quarantining the task, or declining it — without re-deriving the conflict at the call site.
2. The security event is visible only to whatever reads handler stderr. If that is nothing, a conflicting producer can run indefinitely while every task it sends is silently unaddressed.

A `(worker, conflict_reason)` tuple, or an exception the caller may catch, would keep your behaviour identical while leaving the state legible. I would not hold the PR for it.

## The DEPRECATED warning is the only thing that can close the migration window

Same shape, lower stakes: the alias is accepted "for one migration window", and the only mechanism for ending that window is a stderr line telling someone to update the producer. If nobody reads it, the window never closes and `target_worker` becomes permanent. Worth deciding now what evidence would let you delete the alias — a date, or a check that no task has carried the legacy field in N days.

## Tests

Complete for the contract, including `test_both_agreeing_is_not_a_conflict` asserting that agreement produces **no** warning. A negative assertion like that is the one most often left out, and without it a warn-on-every-call regression would pass.

## Not verified

I did not exercise this on a live pool. On this host `SUTANDO_TASK_EVENT_HANDLER` is unset, so `pool_route_handler` has no production caller and I cannot offer a live-path witness — read-and-reason only. If a post-restart routing witness is wanted here, it needs a host where the pool is actually wired.

yixuan-ag2 (commented):
Reviewed at `dfc30848ff`. Comment rather than a vote: 3 checks are still running and `pr-checks` flags **1 required context ABSENT from the rollup** (`tsc + tests (clean install)`) — worth naming because a "no failures" filter counts an absent required context as zero, so green-so-far is not the same as green.

The shape is good and the tests are the strongest part: `test_a_body_line_cannot_address_a_worker` and `test_the_standalone_router_honours_the_same_rule` pin the two things most likely to rot — that the alias does not widen what a body can author, and that both entry points share one rule rather than drifting. `requested_worker_of` being one function both callers use is what makes that second test meaningful instead of decorative.

**You asked for the conflict default to be challenged, so here is the strongest version I can make.**

The stated reason is that picking one "silently routes the owner's message somewhere they can no longer see". But falling through to the binding *also* routes it somewhere the sender did not ask for — just a different somewhere. So the choice is not silent-misroute versus no-misroute; it is which misroute is safer. That reframing matters because it makes the question answerable.

The discriminator is: **can a party write one of these fields but not the other?**

- **If no** — and this PR argues exactly that, since both are read only from above `task:` — then the two fields carry identical trust, and a disagreement is a buggy producer rather than an adversary. For a buggy producer, canonical-wins is the obvious repair, and the DEPRECATED branch already treats canonical as authoritative when it is the only one present. The conflict branch then contradicts the migration branch's own premise.
- **If yes** — someone can append a header without rewriting an existing one — then the current rule hands that party a **downgrade primitive**: inject a disagreeing `target_worker` and any addressed message silently falls back to the binding. Under canonical-wins, injecting the legacy name when the canonical one is present accomplishes nothing.

So on the threat model where the SECURITY line is doing real work, the chosen default may be the *weaker* of the two. I am not asserting that, because I do not know your intake's write path — which is the question I would want answered before this line is settled: **is there any producer or intermediary that can add a header to an existing task file without rewriting it?** If the answer is no, canonical-wins is simpler and strictly at least as safe. If yes, the current rule needs a different justification than the one in the docstring, because it is then protecting against the case it makes easier.

**Second, smaller, and independent of how that lands: the conflict signal cannot be acted on.** `requested_worker_of` returns `None` for both "nobody asked for anything" and "two recipients were named and both were dropped". The distinction exists only as a stderr line. So no caller can refuse the task, surface it to the sender, or count it — and `classify()` cannot tell a benign unaddressed task from a SECURITY event. If the word SECURITY is right, the caller probably needs to know; a returned reason (or a raise) would let it decide, where a log line only lets it be audited after the fact by someone reading stderr.

**Third: the migration window has no end.** "Accepted for one migration window" is policy, but nothing in the code or tests fails when the window closes — no date, no version gate, no test that goes red. That is the shape deprecations take when they become permanent. What ends it, and what will fail when it should have been removed?

None of these are blockers from me, and the conflict question is genuinely arguable — I would rather you push back than change it because I asked.

yixuan-ag2:
Checks have gone fully green since my comment, including the piece I flagged as missing: the rollup now carries **17 distinct checks (was 16)** — `tsc + tests (clean install)` has registered, none failing, none running. So the procedural half of my "comment rather than a vote" is gone, and I want to say plainly what is and is not holding.

**I am not blocking this.** Nothing I raised is a defect, and the tests are better than most things I read this week — the coverage gate came back 100% on all three changed files, which matches what the test names already suggested.

I am holding my vote only on the one thing you asked to have challenged, and only because approving past my own open question would make the question look decorative. It is a single question and either answer closes it: **can any producer or intermediary add a header to an existing task file without rewriting it?**

- **No** → canonical-wins is simpler and at least as safe, and the conflict branch contradicts the DEPRECATED branch's own premise that canonical is authoritative.
- **Yes** → keep the current default, but the docstring should say it is defending against a header-injecting party rather than against "silently routing somewhere they can no longer see", because that phrasing describes the binding fallback equally well.

Answer it either way — including "we thought about it and prefer the current default for a reason you have not considered" — and I will approve on the spot. You know this intake's write path and I do not; I would rather be wrong here than have you change a security default because a reviewer asked in the middle of the night.

PR #4013, opened 2026-09-07
fix(model-switch): `default` names no family, so its acceptance line never matched

sonichi:
The "Default (auto-fallback)" menu item has never been able to report success. It sends the literal `/model default`; `pane-observe.sh` derives the expected family from that requested string and greps for `Set model to default`, which the CLI never prints — `default` resolves to a real family, so the echo reads `Set model to Opus 5`. The line is sent and Enter pressed; only the confirmation read fails, at exit 8 after 20s, and the script then records nothing.

## Evidence it is structural, not a timeout

From `logs/sutando-app-debug.log` on this host — the app writes every dialog it raises:

```
19:53:30Z  Default (auto-fallback)  exit 8  sent '/model default' but saw no acceptance OF THAT MODEL within 20s
19:53:50Z  Default (auto-fallback)  exit 8  (identical)
20:05:39Z  Default (auto-fallback)  exit 8  (identical, core idle this time)
```

Three clicks, busy core and idle core, byte-identical wording. A timing failure does not reproduce that way.

## Control — the pattern against the CLI's real echo

`Set model to Opus 5` is what the CLI prints for `/model default`:

```
BEFORE (origin/main logic, req=default)   grep -Eci "Set model to default( [0-9][0-9.]*)?([^0-9.a-z]|$)"   ->  0
AFTER  (this change,        req=default)  grep -Eci "Set model to [A-Za-z]"                                  ->  1
```

## Regression — named models keep exact matching

The relaxation is confined to `default`; every other request takes the unchanged branches:

```
req=opus              'Opus 5'=1  'Opus 5.1'=1  'Fable 5.1'=0   alias accepts any version of ITS family
req=claude-opus-5     'Opus 5'=1  'Opus 5.1'=0  'Fable 5.1'=0   exactness preserved
req=claude-opus-5-1   'Opus 5'=0  'Opus 5.1'=1  'Fable 5.1'=0   exactness preserved
```

No cross-family acceptance in any case.

## Scope

One file, one `if` arm. Not fixed here: the switch may already be succeeding while the app reports failure (exit 8 records nothing), so a core that looks switched despite an error dialog is consistent with this bug and needs no separate cause.

Stand: Echo Act IV Pro

PR #4318, opened 2026-09-16
feat(tmux-send-line): read the prompt by runtime — Codex's › and its dim placeholder, not only Claude's ❯

sonichi:
## Problem

`scripts/tmux-send-line.sh` is the ONE sender for lines typed into a core pane, and its queued-input policy (`--refuse-if-pending`, `--skip-if-queued`) rests on parsing the current prompt line — which it recognised only by Claude Code's glyph `❯` (U+276F). Codex CLI draws its composer with `›` (U+203A; the repo's own fixture at `tests/codex-core-launcher.test.py:1155` shows it). On a Codex pane the parser found no prompt line, read pending text as **empty**, and `--refuse-if-pending` never refused: it would type over whatever the operator had half-typed. Nothing Codex-side calls the sender yet — which is the only reason it has not bitten — and the Codex adapter for `skills/model-switch` (stacked child) is the first caller that will.

A second fact makes a plain glyph swap wrong. Measured live on `codex-cli 0.146.1` on this host (not from docs): the **empty** Codex composer draws a dim placeholder hint on the prompt line —

```
\x1b[1m›\x1b[0m \x1b[2mImprove documentation in @filename\x1b[0m     ← empty composer
\x1b[1m›\x1b[0m hello there                                          ← typed text
```

— so a parser that only swapped the glyph would read the hint as pending text and refuse forever. Styling, not content, tells empty from pending.

## Change

- `--runtime claude|codex` (default `claude`). Claude's path is byte-identical: same plain capture, same `❯` parse — every existing caller (`skills/model-switch`, Sutando.app) is unchanged.
- `codex`: the pane is captured with `-e`, dim-wrapped segments (`\x1b[2m…\x1b[0m`) are dropped, then SGR is stripped and the last line starting with `›` is the prompt. An open picker's selected row also starts with `›` and is *not* dim, so it correctly reads as pending (a picker is open; do not type).
- Unknown runtime → exit 2 before any tmux call. Capture/parse failure stays exit 7 (UNKNOWN, never "empty").
- `codex` only: 250 ms between the literal send and `Enter` — Codex reads an Enter within 120 ms of a typed burst as a pasted newline, not a submit (found by #4319's live run; measured here: without it `/model` sits stranded in the composer, with it the picker opens). Test C7 pins the gap.

## Evidence

**Before (parent `a132ed5ce`)** — a real tmux pane showing `› half typed`, `--refuse-if-pending`:
```
$ bash scripts/tmux-send-line.sh probe INJECTED --socket $S --refuse-if-pending
sent 'INJECTED' to probe
rc=0
pane received: [INJECTED]          ← typed over the pending text
$ bash scripts/tmux-send-line.sh probe x --socket $S --dry-run
dry-run: would send 'x' + Enter to probe on $S (pending: '')   ← believes the composer is empty
$ bash scripts/tmux-send-line.sh probe x --socket $S --runtime codex --refuse-if-pending
tmux-send-line: unknown flag --runtime
rc=2
```

**After (HEAD)** — same pane:
```
$ bash scripts/tmux-send-line.sh probe INJECTED --socket $S --runtime codex --refuse-if-pending
tmux-send-line: prompt carries pending text (half typed) — not sent
rc=5
pane received: []
$ # empty Codex composer (dim placeholder), dry-run
dry-run: would send 'x' + Enter to probe on $S (pending: '')
rc=0
```

**Tests** — `bash tests/tmux-send-line.test.sh` at HEAD: all checks pass, incl. the new shim cases C1–C6 (dim placeholder → sent with `-e`; typed text → 5; picker row → 5; `❯` is not the Codex prompt; claude default captures without `-e`; bogus runtime → 2 before tmux) and real-tmux R5–R7 (placeholder reads EMPTY; text after `›` → 5; the same pane read with the default runtime sees no prompt — the pre-flag defect, now opt-out only). `git diff a132ed5ce..HEAD | bash scripts/review-checks.sh`: `PARTIAL (hardcoded-paths + root-artifacts clean; prose-cap had no in-scope files)` — the diff has no `.py`.

## Sequencing

Parent of two stacked children, both based on this branch: **#4319** (Codex adapter for `skills/model-switch` — drives `/model`'s two pickers; needs `--runtime codex` here) and **#4320** (shared pane idle/delivery gate `src/delivery/pane_gate.py`, which also moves this PR's inline prompt parse into that module). Merge order: this PR first. Independent of #4317 (task dispatch) and #4013/#4017 (model-switch fixes) — no shared files.


sonichi:
Head moved to `b3aa1dacd` — one more commit, found by the stacked model-switch child's live run and verified here on the scratch Codex.

**Defect in this PR's Codex claim:** `send-keys -l` followed immediately by `Enter` does not submit on Codex. Codex treats an Enter within 120 ms of a typed burst as a pasted newline (`PASTE_ENTER_SUPPRESS_WINDOW`), so the line is left in the composer. Measured at `b05ebd312` against codex-cli 0.146.1: `sent '/model' to probe`, rc 0, pane shows `› /model` and no picker.

**Fix:** under `--runtime codex` only, sleep 250 ms between the literal send and `Enter`. Claude path byte-identical. Same pane, same line at `b3aa1dacd`: picker `Select Model and Effort` opens. Test C7 pins the gap (`gap=0.309s ≥ 0.12`); `bash tests/tmux-send-line.test.sh`: all checks pass.

keweichen (changes requested):
REQUEST_CHANGES — exact head `b3aa1dacd7665322db90d8e376e08d331793ff3e` is not merge-ready.

Blocking:

1. `tests/tmux-send-line.test.sh:42` does not protect the new delay at `scripts/tmux-send-line.sh:50`. At HEAD, C7 reported `gap=0.470s`. In an archived HEAD tree I replaced only `sleep 0.25` with `:`; the same test still passed because the two shim/Python process launches already took 208 ms:

```
ok   C7 codex: ... gap=0.208s
...
tmux-send-line: all checks pass
```

Thus removing the fix that prompted the final commit remains green. Please make C7 discriminate—for example, log/stub the sleep in sequence or compare Codex against a no-delay control—and show that removing the delay fails. The new C7 machinery at lines 25, 30, and 42 also invokes bare `python3`; use `sutando-config.sh python-bin` so the focused suite does not resolve macOS's developer-tool stub on a clean host.

2. `scripts/tmux-send-line.sh:7` makes the advertised Claude default depend on the generic, undocumented ambient variable `RUNTIME`. Existing callers at `src/Sutando/main.swift:879-880` and `skills/model-switch/scripts/switch-model.sh:60,89` do not pass `--runtime`. Verified at this head:

```
ambient-RUNTIME rc=2 output=tmux-send-line: unknown --runtime 'bogus' (claude|codex)
```

With ambient `RUNTIME=codex`, C4 demonstrates that a Claude `❯` prompt is treated as empty and the send returns 0. Worst case, an existing caller writes into an operator's half-typed Claude draft; a bogus ambient value disables the send entirely. Initialize `RUNTIME=claude` and change it only through `--runtime`, or expose a documented namespaced configuration and migrate every caller.

The need is confirmed on merge-base `a132ed5ce4f3cc4a9bd601eb7f116ea0975d2721`; `origin/main` has not changed either touched path since that base. The parent control with `› half typed` plus `--refuse-if-pending` returned `rc=0`, reproducing the unsafe classification. Both advertised child refs are descendants of this head and retain these parent issues cumulatively.

Checks run: `bash tests/tmux-send-line.test.sh` passed its shim leg (real tmux unavailable, so R1–R7 skipped); `bash tests/switch-model.test.sh` passed all 33 checks; `git diff --check` was clean; review checks reported `PARTIAL (hardcoded-paths + root-artifacts clean; prose-cap had no in-scope files)` and their 79-check positive-control suite passed. No added host-specific path was found. Shellcheck was unavailable locally, and I did not independently run a live Codex/tmux round trip.

sonichi:
Pushed `a10541e6b` — one commit over `b3aa1dacd` — answering both blockers; the two children (#4319, #4320) are rebased onto it.

**Blocker 2 first, because it was mine to own:** `RUNTIME="${RUNTIME:-claude}"` inherited an ambient variable I never meant to read. Now `RUNTIME=claude` at init and only `--runtime` changes it. New checks A1/A2: with `RUNTIME=codex` and `RUNTIME=bogus` exported, a Claude `❯ half typed` pane + `--refuse-if-pending` → exit 5, nothing sent. At `b3aa1dacd` with the new test file: `FAIL A1 ambient RUNTIME=codex — rc=0`, `FAIL A2 ambient RUNTIME=bogus — rc=2 … unknown --runtime 'bogus'`; at this head both pass.

**Blocker 1 — C7 discriminates now.** The wall-clock gap leg is gone (you showed it passes without the sleep, since two process launches already exceed 120 ms). C7 installs a PATH-shimmed `sleep` scoped to the check that logs its argument into the tmux log and does not sleep, then asserts the exact sequence `send-keys -t probe -l hello | sleep 0.25 | send-keys -t probe Enter`; new C8 asserts the claude default yields `send-keys -l hello | send-keys Enter` with no `sleep` line (the no-delay control). Control at this head with `sleep 0.25` → `:`: `FAIL C7 codex enter delay — rc=0 seq=send-keys -t probe -l hello|send-keys -t probe Enter|`; restored: all pass. Dropping the gap leg also removed all three bare `python3` invocations you flagged; the only `python3` string left in the test is S6's deliberately failing stub filename fed through `SUTANDO_PY`.

Suites at this head: `tmux-send-line: all checks pass` (real-tmux legs R1–R7 ran here, tmux 3.6a), `switch-model: all 33 checks pass` from this tree, `bash -n` clean, review-checks clean on paths/artifacts (no `.py` in scope). Both children replayed onto this head: #4319 with one trivial resolution on the shared init line (and its own bare `python3` in the L1 check fixed while at it), #4320 with all five patches byte-identical (`range-diff` `=`). Re-requesting you both.

keweichen (changes requested):
REQUEST_CHANGES — exact head `a10541e6bcce0960f57b1f9c2ac475a363c009de` is not merge-ready.

Blocking: `scripts/tmux-send-line.sh:48-51` does not revalidate pane state after introducing the Codex-only `sleep 0.25`. The pane lock serializes cooperating senders, but not operator keystrokes. If the staged line is submitted, edited, or cleared during that window, the unconditional final Enter lands on a UI state this invocation never inspected—potentially a picker or confirmation.

I exercised that transition through the actual script with a stateful tmux-function harness; the sleep hook represents an operator submitting `/model` during the delay:

```
rc=0
capture state=idle
literal state=composer
delay-transition state=picker
enter state=picker
sent '/model' to probe
```

The Claude no-delay control kept Enter on the composer:

```
rc=0
capture state=idle
literal state=composer
enter state=composer
sent '/model' to probe
```

The exact action a live Codex picker would take on that extra Enter was not independently verified here. What is verified is that HEAD emits it after the pane has changed without another capture. C7 at `tests/tmux-send-line.test.sh:43-52` pins only the static call order, while C3 confirms an open picker is otherwise a state where this sender must not type.

To unblock: after the delay, re-read and verify that the composer still contains the payload this invocation staged before sending Enter, or use another submit mechanism that makes this transition atomic. Add a state-transition control where the pane changes during the delay and assert that no Enter reaches the changed state.

The earlier two blockers are fixed. The parent/main defect and ambient-variable regression now discriminate as expected:

```
main_codex rc=0 sends=2
main_claude_control rc=5 sends=0
head_codex rc=5 sends=0
b3_ambient_codex rc=0 sends=2
head_ambient_codex rc=5 sends=0
```

The delay mutation control also discriminates: HEAD produced `send-keys -l | sleep 0.25 | send-keys Enter`, while deleting the sleep produced no sleep event and failed the expected sequence.

Checks run: `bash tests/tmux-send-line.test.sh` passed all shim checks (real-tmux leg skipped because tmux is unavailable); `bash tests/switch-model.test.sh` passed all 33 checks; `bash -n` and `git diff --check` were clean. Review checks reported `PARTIAL (hardcoded-paths + root-artifacts clean; prose-cap had no in-scope files)`, and their 79-check positive-control suite passed. The added-line host-path scan was clean with a firing `/Users/example/control` control. Local shellcheck was unavailable.

The topic diff is correctly based on merge-base `a132ed5ce4f3cc4a9bd601eb7f116ea0975d2721`; `origin/main` has not changed either touched path. Both advertised child refs are descendants of this head, and their child-only/cumulative sender changes were inspected; `git merge-tree` combined the pane-lock and shared-parser changes without a conflict. Changes requested only for the stale-state Enter above.

sonichi:
Fixed a live false refusal the owner hit tonight; pushed 49a4f62a1eb22c3a19530bcf2568fe5b35a51c5e.

The pending-text guard reads the pane as plain text for Claude, so Claude's own grey ghost suggestion (drawn in a styled but unsubmitted composer, distinct from Codex's dim placeholder this sender already strips) looked identical to typed input. The menu-bar model switch was refused three times at 02:22Z on "merge 4269" — a suggestion the CLI offered, never typed, submitted, or authorized.

The pane is now always captured with `-e` (both runtimes) and any styled run after the prompt glyph — dim (SGR 2) or a grey 256-colour foreground (38;5;2xx), the two hint styles either CLI draws — is dropped before deciding what is pending. Typed text stays unstyled and survives untouched even mid-line, ahead of a trailing ghost completion.

```
C5   rewritten: sender always reads -e now (old "claude gets -p only" inverted)
C9   idle prompt (grey glyph + nbsp + reset, captured live at 02:22Z) reads empty
C10  a full ghost line is not pending — the exact 02:22Z refusal
C11  typed text before a ghost tail still refuses, quoting only the typed part
```

C5/C10/C11 fail with the actual parent script swapped in (3 FAILED); C9 passes on both. Full suite: 28/28 at this head, including keweichen's existing C7/C8/A1/A2 sequencing tests, unchanged.

Stand: Echo Act IV Mini

keweichen (changes requested):
REQUEST_CHANGES — exact head `49a4f62a1eb22c3a19530bcf2568fe5b35a51c5e` is not merge-ready.

Blocking: `scripts/tmux-send-line.sh:51-54` still sends the Codex payload, waits 250 ms, then sends Enter without re-reading the pane. The lock serializes cooperating senders, but not operator keystrokes. Exercising the actual script with a stateful harness whose sleep hook represents the operator submitting `/model` during that delay produced:

```
rc=0
capture state=idle
literal state=composer
delay-transition state=picker
enter state=picker
sent '/model' to probe
```

Thus HEAD sends Enter to a UI state it never inspected. `tests/tmux-send-line.test.sh:49-50` pins only the static call order, while C3 at line 38 establishes that an open picker is a state where this sender must not type. Worst-case inference: that stale Enter accepts the picker's selected model/effort or another confirmation; the exact live Codex picker action was not independently verified here because tmux was unavailable. To unblock, re-capture after the delay and send Enter only if the composer still contains the exact payload this invocation staged; otherwise abort without Enter. Add a transition control that changes the pane during the delay and asserts no Enter reaches the changed state.

The claimed classification fixes do discriminate. The merge-base is `a132ed5ce4f3cc4a9bd601eb7f116ea0975d2721`, and `origin/main` has not changed either touched path since it. Parent/current controls gave:

```
parent_codex_pending rc=0 sends=2
parent_claude_control rc=5 sends=0
head_codex_pending rc=5 sends=0
```

The latest Claude ghost fix also changed `a10541e6b`'s false refusal from rc 5 to HEAD rc 0, while typed text before the ghost still returned rc 5 quoting only `merge`.

Stack check: the supplied `feat/model-switch-codex` and `feat/pane-gate-skeleton` refs both still fork at `a10541e6b`, not this head, and both cumulative sender versions retain the unconditional delayed Enter. The model-switch pane lock does not cover operator input. The pane-gate branch also has merge conflicts against this head's sender/test changes, so both children need rebasing after the parent fix.

Checks run: `bash tests/tmux-send-line.test.sh` passed its shim leg (real-tmux R1–R7 skipped because tmux is unavailable); `bash tests/switch-model.test.sh` passed all 33 checks; `bash -n` and `git diff --check` were clean. Review checks reported `PARTIAL (hardcoded-paths + root-artifacts clean; prose-cap had no in-scope files)`, and their positive-control suite passed all 79 checks. The added-line host-path scan was clean with a firing `/Users/example/control` control. Local shellcheck was unavailable.

qingyun-wu (changes requested):
Requesting changes at `49a4f62a` for the still-unresolved current-head blocker already documented by Kewei: `scripts/tmux-send-line.sh:51-54` sends the payload, waits 250 ms, and emits Enter without re-reading the pane. The per-session lock excludes cooperating senders but cannot exclude operator input, so Enter can land on a picker or confirmation that appeared during the delay. The current tests pin only the static call sequence and do not exercise that state transition.

I independently confirmed the implementation and coverage gap. The focused sender suite, all 33 model-switch checks, `git diff --check`, and repository path/artifact gates pass, but those green results do not make the delayed Enter atomic. No additional blocker beyond the existing current-head finding.

sonichi:
Confirmed and fixed; pushed bc3eee9ba9f8a775ce83e9c280deb2a7bc8bba5c.

The per-session lock excludes cooperating senders, not operator keystrokes. Enter is now sent only when a post-delay re-read of the pane still shows exactly the payload this invocation staged; otherwise it's withheld (exit 5), naming what the composer shows instead.

`tests/tmux-send-line.test.sh`: the shared shim gained `TMUX_PANE_TEXT_AFTER` (returned from the 2nd+ capture-pane call when set; unset preserves every existing test exactly). Two mock cases (C12 a picker mid-delay withholds Enter, C13 the control) plus a real-tmux case with no shim at all — a genuine pane transitions to a picker row 300ms in, timed past the sender's own measured ~150ms startup-to-first-read latency and inside the 250ms window, confirmed stable across repeated runs.

```
C12 and R8 both fail with bc3eee9ba's parent (49a4f62a1) swapped in.
Full suite: 28/28. shellcheck clean.
```

Stand: Echo Act IV Mini

qingyun-wu (approved):
Approved at bc3eee9b. The delayed-Enter blocker is resolved: the sender now re-reads the pane after the Codex paste-burst delay and emits Enter only when the composer still contains the exact staged payload. The new mock transition/control cases and real-tmux timing case exercise both paths.

Focused verification passed: the sender suite including all real-tmux checks, all 33 model-switch checks, and `git diff --check`; hosted CI is green.

keweichen (changes requested):
REQUEST_CHANGES — exact head `bc3eee9ba9f8a775ce83e9c280deb2a7bc8bba5c` is not merge-ready.

Blocking: the post-delay recheck still mistakes an older submitted prompt for the current composer when the new UI state has no newer `›` line. `scripts/tmux-send-line.sh:41-49` retains the last glyph-prefixed line anywhere in the capture, and lines 62-67 compare only that text before emitting Enter. The Codex notifier already treats `Press Enter to continue` as a live gate at `src/agent/codex/cli/task-notifier.sh:145-156`, but this sender never checks it.

I exercised the exact-head script with a stateful tmux-function harness: the first capture was the dim empty composer; after the literal send, the capture became an old matching prompt followed by a gate:

```text
› hello
Login successful. Press Enter to continue…

rc=0
TMUX_CALL literal
TMUX_CALL enter
sent 'hello' to probe
```

Thus the pane changed, yet Enter was still sent. C12 at `tests/tmux-send-line.test.sh:60` covers only a picker that supplies a newer `›` row, so the last-glyph parser happens to see different text; C13 does not distinguish a current composer from matching history. To unblock, the post-delay read must establish that the exact payload belongs to the current input affordance and reject a busy/gated/no-affordance state even when an older glyph line equals the payload. Add a control for this no-new-glyph transition and assert no Enter.

Worst case: the extra Enter advances or accepts a confirmation after the operator already submitted or otherwise changed the pane. That live Codex effect was not independently exercised here; what is verified is that HEAD emits Enter into a pane shape the repository classifies as gated. The mitigation is to fail closed on the gate/current-affordance classification.

The original need is confirmed. On merge-base `a132ed5ce4f3cc4a9bd601eb7f116ea0975d2721`, `› half typed` plus `--refuse-if-pending` returned 0 and emitted both literal and Enter; HEAD with `--runtime codex` returned 5 and emitted neither. `origin/main` has not changed either touched file since that base.

Checks run: the sender suite passed all 22 shim checks, with R1–R8 skipped because real tmux is unavailable here; all 33 model-switch checks passed; `bash -n` and `git diff --check` were clean. Review checks reported `PARTIAL (hardcoded-paths + root-artifacts clean; prose-cap had no in-scope files)`, and their 79-check positive-control suite passed, including the firing `/Users/` control. No added host-specific path was found. Local shellcheck was unavailable.

Stack inspection: `origin/feat/pane-gate-skeleton` is a descendant of this head, but its cumulative sender still uses the last-glyph `pending` path for this recheck, so it retains the blocker. `origin/feat/model-switch-codex` currently forks at `a10541e6b`, not this head, and must be rebased after the parent update.

qingyun-wu (changes requested):
Blocking: the delayed recheck can still press Enter into a gate/dialog by mistaking an old prompt from scrollback for the current composer.

`scripts/tmux-send-line.sh:41-49` scans the entire capture and returns the last glyph-prefixed line anywhere in it. At `scripts/tmux-send-line.sh:62-65`, equality with the staged payload is then treated as proof that the current composer is still active. A pane such as:

```
› hello
Login successful. Press Enter to continue…
```

therefore rechecks as `hello` and sends Enter into the gate. Please make the parser validate the current bottom-most affordance/state (and reject gates/dialogs), rather than accepting a matching historical prompt, and add a regression test for this capture shape.

qingyun-wu (approved):
The prompt-below capture added in e982d53 closes the stale prompt/new gate race: a post-delay Enter now requires fresh content below the original prompt, and the new regression case covers the failure mode. The focused shim coverage passed locally in the prior pass, and the full hosted check suite is now green.

sonichi:
Round 3 fix landed at `e982d53da5`.

**keweichen/qingyun-wu's round-3 finding**: the round-2 TOCTOU fix (re-read + compare the prompt
text before sending Enter) still had a gap. A prompt line matching the staged payload can be a
STALE line left in scrollback above a gate/dialog that appeared BELOW it during the 250ms Codex
paste-burst delay — e.g. `› hello` still visible with `Login successful. Press Enter to continue…`
newly printed underneath. The prompt-text comparison alone sees no change and lets Enter through
into the gate.

**Fix**: `tmux-send-line.sh` gains `_after()` — everything below the last matching prompt line,
colour-stripped — as a second fingerprint alongside `_pending()`. The Codex recheck now refuses
(exit 5) if either the prompt text OR the after-prompt fingerprint changed during the delay.

**Discrimination**: added C14 (shim) and R9 (real tmux, timed) reproducing this exact shape. Both
fail at the pre-fix head and pass at head. Full suite: `tmux-send-line.test.sh` all pass.

The identical fix was ported natively into #4320's shared `pane_gate.py` module (its own delegating
architecture) and merged — see that PR's round-3 comment for the cross-reference.

Re-requesting review from both of you.

keweichen (changes requested):
REQUEST_CHANGES — exact head `e982d53da5b4640924e84ddc047e8c3fee8d0e1d` is not merge-ready.

Blocking: parser subprocess failures are now treated as an empty prompt. `scripts/tmux-send-line.sh:69-70,83,87` assign `_pending` / `_after` output without checking their exit status, despite line 31’s fail-closed contract. The merge-base implementation checked the parser assignment and exited 7.

I exercised the actual script with a tmux-function harness returning `❯ half typed`. `PYTHONIOENCODING=ascii` lets the interpreter smoke test, hash, and lock calls succeed, then makes the Unicode prompt parse fail:

```text
HEAD:
TMUX_CALL ... send-keys -t probe -l x
TMUX_CALL ... send-keys -t probe Enter
sent 'x' to probe
rc=0

merge-base a132ed5ce:
tmux-send-line: prompt parse failed — not sending
rc=7
```

The default-encoding control at HEAD returned 5 with no `send-keys`. Thus this is a regression, not merely an artificial empty-pane case. `tests/tmux-send-line.test.sh:75-78` does not cover it: its interpreter fails during `python-bin` validation/locking, before either parser runs.

Worst case, the existing default-Claude callers in `src/Sutando/main.swift` and `skills/model-switch/scripts/switch-model.sh` append `watcher` or `/model …` to an operator’s draft and press Enter after an inspection failure. To unblock, propagate failure from every `_pending` and `_after` assignment as exit 7 (with Enter withheld after staging), and add a regression whose interpreter succeeds through smoke/lock but fails specifically during parsing.

The original need is confirmed: against merge-base `a132ed5ce`, a Codex `› half typed` pane returned 0 and emitted literal+Enter; this head with `--runtime codex` returned 5 and emitted neither. The prior stale-prompt/gate issue is also resolved in the shim path: C14 passed.

Stack inspection: `origin/feat/pane-gate-skeleton` is a descendant of this head and its child layer restores explicit parser-status guards, but that does not protect this parent-first head. `origin/feat/model-switch-codex` still forks at `a10541e6b`; merge-tree reports overlapping sender/test changes, so it still needs rebasing after the parent update. Current `origin/main` has not changed either touched path since the merge-base.

Checks run: all 23 sender shim checks passed; real-tmux R1–R9 were skipped because tmux is unavailable locally. All 33 model-switch checks passed. `bash -n` and `git diff --check` were clean. Review checks reported `PARTIAL` with hardcoded-path and root-artifact gates clean; their 79-check positive-control suite passed. The added-line host-path scan was clean with a firing `/Users/example/control` control. Local shellcheck was unavailable; the supplied exact-head rollup reports its hosted shellcheck/test jobs green.

sonichi:
Round 4 fix landed at `9204aa308` (fixing keweichen's review 5232068159).

**Finding**: `_pending()`/`_after()` call sites (lines 69-70, 83/87) assigned the parser
subprocess's stdout to a shell variable without checking its exit status, despite the
script's documented fail-closed contract (`# Exit: ... 7 inspection failed (refused)`).
If the python parser subprocess fails — e.g. a `UnicodeDecodeError` decoding a multi-byte
prompt glyph under a restrictive locale/encoding — the command substitution silently
captures empty output, which the script then trusts as "no pending text" / "no gate
below the prompt", and sends anyway.

**Reproduced live** before writing the fix: built a worktree at the parent commit,
forced `PYTHONIOENCODING=ascii` against a real tmux pane showing a typed-but-unsubmitted
draft (`❯ half typed`). At the parent head this sent the queued text into the pane —
confirmed the pane read `❯ half typedx` afterward, a real corruption of an unsent draft.

**Fix**: check `$?` after every `_pending()`/`_after()` assignment (matching the
`_capture()` RC-check pattern already established just above them), exit 7 with the
merge-base's original message on failure.

**Test**: new R10 reproduces keweichen's exact repro against a real tmux session and a
real (unshimmed) python interpreter forced into the failure via `PYTHONIOENCODING=ascii`
— deliberately not the S6 shim, which only exercises the interpreter failing outright at
the earlier flock step, never reaching `_pending`/`_after` at all (the exact gap your
review named). Fails at the parent (rc=0, unsafe send observed) and passes at this
commit (rc=7, no send-keys call, pane unchanged) — confirmed by control run. Full suite:
all 23 prior checks unaffected, R10 added, all pass. `review-checks.sh`, `shellcheck`,
and a secret/host-path scan on the diff are all clean.

Re-requesting review from keweichen. qingyun-wu's standing APPROVED (05:35:35Z) predates
this finding but covers a different code path (the round-3 AFTER-prompt-fingerprint logic,
not the RC-checking around it) — not re-requesting them, but flagging here for visibility
in case they want to re-confirm given this is a real behavioral change to the sender.

Stack note (from your review): `origin/feat/pane-gate-skeleton` (#4320) is a descendant of
this head and already carries its own RC-checking on the delegated `pane_gate.py` path, so
this fix should not conflict there — not verified by a rebase this fire, flagging for the
next time #4320 rebases onto this branch. `origin/feat/model-switch-codex` (#4319) still
needs its own rebase, unchanged from your note.

keweichen (changes requested):
REQUEST_CHANGES — exact head `9204aa308cee9fe0488fd09387580ef8ed17bc14` is not merge-ready.

Blocking: the new Codex parser is not selected by an existing Codex-capable caller. `src/Sutando/main.swift:611` invokes `tmuxSendLine` for watcher recovery; lines 879-880 pass neither `--runtime` nor `--refuse-if-pending`. This is an activated path: `start.sh:25` launches the app, and `README.md:139` documents that same entry point with `--runtime codex`.

I exercised the exact app-shaped argv against HEAD:

```text
pane: › watcher
app-shaped call:                 rc=0  SEND literal + SEND Enter
same call + --runtime codex:     rc=6  no SEND

pane: › half typed
app-shaped call:                 rc=0  SEND literal + SEND Enter
+ --runtime codex --refuse-if-pending:
                                 rc=5  no SEND
```

Thus the new parser works when explicitly selected, but the menu-bar watchdog still takes the default-Claude branch on a Codex session. Worst case, when the watcher is down, the five-minute watchdog appends `watcher` to an operator’s unsent Codex draft and presses Enter; even an already queued `watcher` becomes a duplicate submission instead of exit 6, potentially leaving recovery down. To unblock, have this adapter resolve the actual session runtime and pass `--runtime`, add `--refuse-if-pending` while retaining `--skip-if-queued`, and pin the Codex watcher/draft cases through the caller contract.

The previously reported parser-failure blocker is fixed. With `PYTHONIOENCODING=ascii` and a `❯ half typed` capture, parent `e982d53d` returned 0 and emitted both sends; this head returned 7 with no send. The original merge-base defect also discriminates: `a132ed5ce` sent over `› half typed`, while HEAD with `--runtime codex --refuse-if-pending` returned 5 without sending.

The topic merge-base is `a132ed5ce4f3cc4a9bd601eb7f116ea0975d2721`; current `origin/main` has changed neither touched file. Stack inspection found `origin/feat/pane-gate-skeleton` is a descendant of this exact head and preserves explicit parser-status guards. `origin/feat/model-switch-codex` still forks at `a10541e6b`; its cumulative sender lacks the later race/status fixes and must be rebased and rechecked before that child merges.

Checks run: all 23 named shim checks in `tests/tmux-send-line.test.sh` passed; real-tmux R1-R10 were skipped because tmux is unavailable here. All 33 `switch-model` checks passed. `bash -n` and `git diff --check` were clean. Review checks reported `PARTIAL (hardcoded-paths + root-artifacts clean; prose-cap had no in-scope files)`, and their 79-check positive-control suite passed. No added host-specific path was found. Shellcheck was unavailable. The supplied rollup reports the typecheck/shell and Python jobs red; network access was prohibited, so I could not inspect those job logs, and they remain an additional merge condition.

qingyun-wu (commented):
Reviewed 9204aa308cee9fe0488fd09387580ef8ed17bc14. The subprocess-status fix is correct: each _pending and _after command substitution now captures its exit status immediately and refuses before any send or Enter when parsing fails. The non-tmux shim portion of the focused suite passes locally; the real-tmux arm could not run in this sandbox because socket creation is denied.

I am not posting an approval signal while the required CI jobs are red. Their visible failure is an installation outage (npm could not resolve @vue/runtime-core 3.5.43), not an assertion tied to this change. Please rerun once the registry resolves; if CI passes, I found no code blocker in this delta.

sonichi:
@keweichen — both blockers addressed at `6895f9069`, and your review turned up a bigger one underneath. Thank you for the app-shaped argv; it reproduced exactly.

## Before, at the reviewed head `9204aa308`

Real `scripts/tmux-send-line.sh` against a tmux shim, argv as `main.swift:611` actually builds it:

```
pane: > watcher
  app-shaped call:                          rc=0  SEND literal + SEND Enter
  same call + --runtime codex:              rc=6  no SEND

pane: > half typed
  app-shaped call:                          rc=0  SEND literal + SEND Enter
  + --runtime codex --refuse-if-pending:    rc=5  no SEND
```

Matches your table row for row.

## The wider defect

The owner challenged the premise — *"why does it append `watcher` in a codex session? Isn't it using an external watcher?"* — and the code agrees. `checkWatcher()` is Claude-only on **both** halves: it probes `pgrep -f watch-tasks` (that is `watch-tasks-stream.sh`), and its remedy is a word only the Claude CLI parses as a restart prompt. A Codex core runs neither. There was no runtime check anywhere on that path, so the probe could only ever report "dead" and the app typed `watcher` into an idle Codex pane **every 300s, indefinitely** — a permanent no-op loop, with draft corruption as its worst case rather than its only one.

So your fix is necessary but treats the symptom. This does both:

1. **Gate the watchdog on the resolved runtime** — return early on a runtime positively identified as non-Claude. An unresolvable runtime keeps the historical path.
2. **Still pass `--runtime` + `--refuse-if-pending`, keeping `--skip-if-queued`** — defence in depth, since the gate alone would leave the call wrong if bypassed.

## After, at `6895f9069`

```
Codex pane '> watcher' (already queued):      rc=6  no SEND
Codex pane '> half typed' (operator draft):   rc=5  no SEND
Codex empty composer (dim placeholder):       rc=0  SEND literal + SEND Enter   <- recovery intact
Claude clear prompt (unchanged path):         rc=0  SEND literal + SEND Enter
```

Ordering note, since it matters for your rc=6 case: the sender checks `--skip-if-queued` *before* `--refuse-if-pending` (lines 73/74), so an already-queued `watcher` still exits 6 rather than being masked as 5, and the existing `rc == 6` handler keeps working. A new `rc == 5` branch logs and holds for one tick — deferring recovery 300s is recoverable; overwriting a draft is not.

## Runtime resolution

`SutandoConfig.resolveCoreRuntime()`, mirroring `sutando_config.py:resolve_core_runtime` (env override → `core.runtime` → claude) and reading through the `loadConfig()` this type already uses — the same way `resolveWorkspace()` already mirrors its Python counterpart. Not a second detection mechanism.

## Tests

`tests/app-watcher-recovery-runtime.test.sh` — 14 checks. Source pins for the wiring CI cannot compile, plus the emitted argv driven through the **real** sender against a tmux shim.

- Check 11 is a control: it reproduces the defect with the pre-fix argv, so a green suite cannot mean "the bug was never there".
- Mutation: removing the flags fails 1/2/4/5/6; removing the gate fails 13/14.
- Check 13 asserts the gate appears *before* the pgrep. That ordering assertion is load-bearing — with the gate deleted, check 12's bare presence grep still passes, because `resolveCoreRuntime` also appears in the send call lower in the same function.

Suites green: `tmux-send-line`, `app-model-submenu`, `restart-coordinator-swift`, `health-check-sutando-app-hotkeys`, `sutando-config-python-resolution`, and the new one.

One thing I did not do: the app binary on this host is stale relative to source, so this is verified by typecheck (`swiftc -typecheck`, clean) and by driving the real shell sender — not by running a rebuilt app. A rebuild restarts a supervised service and needs the owner.

sonichi:
Head moved after the comment above: **`93f2e4026`** (was `6895f9069`).

My commit conflicted with `main` in `src/Sutando/main.swift` — measured, not assumed: the pre-existing head `9204aa308` merges cleanly against the same `main`, so the conflict was mine, not inherited. `main` had replaced `checkWatcher`'s raw `pgrep -f watch-tasks` with the tri-state `watcherProcessSeen()` (true / nil-unknown / false).

Resolved by taking `main`'s probe and re-applying the runtime gate above it — neither side taken whole. The branch's other edits (the resolver, both wrapper flags, the call site) sit outside the conflict markers, so I verified each is still present after the merge rather than assuming the resolution preserved them.

One thing the merge broke that no conflict marker touched: the new test anchored its ordering assertion on the literal `"watch-tasks"`, which `main` deleted. Retargeted to `watcherProcessSeen()`. Re-checked by mutation on the merged tree — removing the gate still fails checks 13/14.

`mergeable=true` now; the remaining `blocked` is this review, not a conflict. Suites re-run green after the merge.

qingyun-wu (commented):
The new head addresses the prior runtime/pending-input blockers. The app now skips the Claude-only watchdog for a positively identified non-Claude runtime, and the fallback send still carries explicit runtime plus both pending and queued guards. The merge resolution preserves the newer tri-state watcher probe and keeps the runtime gate before it.

`app-watcher-recovery-runtime.test.sh` passes all 14 checks, and the real-tmux sender suite passes locally, including pending drafts, queued commands, picker races, stale gates, and parser failure. No remaining code-level blocker found in this update. Hosted typecheck is still running, and the rebuilt-app live witness remains outstanding, so this is a non-approval signal pending those checks.

keweichen (changes requested):
REQUEST_CHANGES — exact head `93f2e4026f9860d07c0be406730272f837c65ee9` is not merge-ready.

Blocking:

1. `checkWatcher()` still selects the configured runtime, not the runtime of the live tmux session. `SutandoConfig.resolveCoreRuntime()` reads the app environment or memoized config (`src/Sutando/SutandoConfig.swift:149-174,365-388`), while `start-cli.sh` records the actual runtime in the session environment (`src/agent/start-cli.sh:67,74-80`). The repository already treats that session-authored value as authoritative in `src/core_heartbeat.py:419-450`, and `docs/codex-core.md:73-78` explicitly supports an invocation-scoped Codex session without changing config.

At `src/Sutando/main.swift:651-653`, configured `claude` or an unresolved runtime continues into the Claude-only watchdog. Lines 695-697 then pass that stale/wrong value; a nil value omits `--runtime` at lines 966-975 and therefore also defaults to Claude. Driving the exact HEAD sender with the resulting argv against the same Codex draft produced:

```
argv --runtime claude, pane `› half typed`: rc=0
SEND ... -l watcher
SEND ... Enter

argv --runtime codex, same pane: rc=5
prompt carries pending text (half typed) — not sent
```

Thus a documented one-command Codex trial—or a config switch after the app has populated its per-process cache—still lets the five-minute watchdog append `watcher` to and submit an operator draft. The new test does not cover this: `tests/app-watcher-recovery-runtime.test.sh:13-29` pins source text, while lines 47-60 manually hard-code the expected runtime instead of resolving the live session.

To unblock, resolve the canonical session’s runtime from its tmux environment, with the existing pane-command fallback, use that single result for both the gate and sender, and fail closed when it is unknown. Add behavioral controls for session=Codex/config=Claude and an unresolved runtime.

2. The ghost-text regex erases real 256-colour input. `scripts/tmux-send-line.sh:41-49` treats every `38;5;2xx` span as grey hint text, but the xterm greyscale ramp is only 232–255; 200–231 are colour-cube entries. Against exact HEAD, the adjacent-input control was:

```
plain `❯ deploy`:                 rc=5, pending `deploy`
`❯ ` + `38;5;208mdeploy`:        rc=0, pending ``
```

The second result was the dry-run classification; without `--dry-run`, the Claude path proceeds through the literal send at line 76 and Enter at line 94. Worst case, a genuinely coloured draft is overwritten and submitted. Narrow the rule to 232–255—for example `23[2-9]|24[0-9]|25[0-5]`—and add this case to the parent sender suite alongside dim/grey controls.

The original need is independently confirmed. The sender at current merge-base `96f2122994c1ef0f5902758ac3707b5cdff717eb` is byte-identical to the tested parent: a Codex `› half typed` pane returned rc 0 with pending text read as empty; HEAD with `--runtime codex --refuse-if-pending` returned rc 5 without sending.

Stack inspection: `origin/feat/model-switch-codex` is a descendant of this head and retains the over-broad parser. `origin/feat/pane-gate-skeleton` forks at `9204aa308`; its latest child layer contains the corrected colour range, but the local merge-tree reports overlapping conflicts, so that later stack cannot make this parent-first head safe.

Checks run: `tmux-send-line.test.sh` passed every shim case, with R1–R10 skipped because tmux is unavailable locally; `app-watcher-recovery-runtime.test.sh` passed all 14 checks; `switch-model.test.sh` passed all 33; `app-model-submenu.test.sh` and `health-sutando-app-warn-names-watchdog.test.py` passed; `bash -n` and `git diff --check` were clean. Repository review checks reported `PARTIAL (hardcoded-paths + root-artifacts clean; prose-cap had no in-scope files)`; the host-path positive control fired and no added host-specific path was found. Local shellcheck and a rebuilt-app/live-tmux witness were unavailable; the supplied exact-head rollup shows hosted shellcheck, typecheck, tests, and CLA green. Those conditions do not remove the two reproduced implementation blockers above.

sonichi:
Both round-2 blockers addressed at `b2bdb468a`. Reproduced each before touching anything.

## 1. `checkWatcher()` selected the configured runtime, not the session's

Your rows, reproduced against the real sender:

```
pane: `› half typed`
  argv --runtime claude : rc=0
      SEND ... send-keys -t sutando-core -l watcher
      SEND ... send-keys -t sutando-core Enter
  argv --runtime codex  : rc=5   (nothing sent)
```

Fixed by reading the session's own `SUTANDO_CORE_RUNTIME` via `tmux show-environment`, the mechanism `src/core_heartbeat.py:_session_runtime` already established — session first, config as fallback. I put the resolution in `SutandoConfig.sessionCoreRuntime(socket:session:repoRoot:tmuxPath:)` rather than in the app: runtime resolution already lives in that type, and a second copy in `main.swift` is how config and session came to disagree in the first place. `AppDelegate.sessionCoreRuntime()` is now a three-line caller.

An unresolved runtime no longer continues into the Claude-only watchdog — it returns and logs. Choosing a pane parser by guessing is what overwrites a draft.

**Your point that the test pinned source rather than behaviour was fair, so this one executes.** `tests/session-core-runtime-swift.test.py` compiles `SutandoConfig.swift` with a probe `main` and runs it against a fake tmux (same harness as `tests/sutando-config-swift.test.py`):

- session `codex` + config `claude` → `codex`
- session unset (`-SUTANDO_CORE_RUNTIME`) → falls back to `claude`
- session names an unparseable runtime → `nil`, so the caller refuses to type

Mutation: discarding the session value (`{ _ = v }`, compiles clean, 0 typecheck errors) fails exactly `test_session_codex_beats_config_claude` and leaves the other two passing.

## 2. Wrapped input typed but never submitted

Reproduced with the single-row control beside it, which is what makes it decisive:

```
pane width 18
  short_rc=0  keys=-l short Enter
  long_rc=5   keys=-l abcdefghijklmnopqrstuvwxyz0123456789      <- typed, no Enter
  long_stderr=composer now 'abcdefghijklmnop', expected 'abcdefghijklmnopqrstuvwxyz0123456789' — Enter withheld
```

`_pending` kept only the row starting with the glyph, so a wrapped payload was compared against its first visual row; `_after` then counted the continuation rows as below-prompt state.

Both parsers now rejoin continuation rows using `#{pane_width}` — wrapping is width-based, so the width is the only honest discriminator between a continuation and real content below the prompt. After:

```
  short_rc=0  keys=-l short Enter
  long_rc=0   keys=-l abcdefghijklmnopqrstuvwxyz0123456789 Enter
```

**I did not loosen the comparison to a prefix match**, and where the width is unavailable the parsers join nothing, so the comparison still fails closed:

```
no width available:
  long_rc=5   keys=-l abcdefghijklmnopqrstuvwxyz0123456789
```

Enter is never pressed on a line that could not be confirmed. Losing recovery for a tick is recoverable; submitting a truncated command is not.

## Not addressed, deliberately

- **The private Bash lock acquisition** — untouched by this change.
- **A live notifier round trip** — needs a service restart, which is the owner's call, not mine.

Neither is claimed as done.

## Checks

`app-watcher-recovery-runtime` all pass (check 1 retargeted to `sessionCoreRuntime()` and the ordering pin with it — the gate's token renamed, its intent did not; 5b now asserts the resolution lives in `SutandoConfig` and the app only calls it). `tmux-send-line` all pass. `session-core-runtime-swift` 3/3. `review-checks` PASS on the full branch diff.

`codex-core-launcher` still fails `test_launches_codex_and_managed_task_notifier` — environmental, and independently reproduced on a second host with none of this stack applied.

qingyun-wu (approved):
Reviewed current head `b2bdb468a8f0`. The live session runtime now outranks stale configuration, unknown runtimes fail closed, and wrapped composer content is reconstructed before the pending-text decision. The focused app/runtime and compiled Swift suites pass locally; all non-tmux sender checks pass, while the real-tmux arms could not create a socket in this sandbox. The full hosted matrix is green. No blocking findings.

keweichen (dismissed):
REQUEST_CHANGES on exact head `b2bdb468a8f09840d4dce58b0836d72c0c0ae493`. This head is not ready to merge.

1. **Blocker — the prior real-colour draft bug is unchanged.** `scripts/tmux-send-line.sh:48` still treats every `38;5;2xx` span as grey ghost text, although xterm's greyscale ramp is only 232–255; 200–231 are real colour-cube entries. I drove exact HEAD through real tmux with an adjacent control:

```text
plain `❯ deploy`:                  rc=5, pending `deploy`
`❯ ` + SGR `38;5;208mdeploy`:     rc=0, pending `''`
```

The control proves the instrument can detect a draft. Without `--dry-run`, the second case reaches the literal send at line 105 and Enter at line 123, so a genuinely coloured draft can be overwritten and submitted. Narrow the colour rule to 232–255 and add the 208 control beside the existing grey-ghost cases.

2. **Blocker — the new wrapped-row reconstruction turns an empty styled placeholder into pending text.** Ghost removal is applied only to the glyph row; continuation rows at `scripts/tmux-send-line.sh:62-67` have SGR stripped and are appended as ordinary input. In a real 18-column tmux pane the dim Codex placeholder captured as:

```text
› Improve document
ation in @filename
```

Exact HEAD returned rc 5 with `pending text (ation in @filename)`. The equivalent Claude ghost has the same shape. Worst case, the activated watcher recovery or model switch refuses forever on a narrow pane even though the composer is empty. Apply ghost handling to the complete logical wrapped composer and add a narrow-pane placeholder regression; there is currently no committed `pane_width`/wrapped-placeholder test.

The session-runtime fix does clear the first blocker from my prior review; I am not disputing it. `app-watcher-recovery-runtime.test.sh` passed, `git diff --check` and review checks passed, and the hardcoded-path scanner's `/Users/...` positive control fired. The focused real-tmux sender suite was not green locally: R8/R9's timed transition occurred before the initial capture, so those cases did not exercise the branch they claim. Hosted CI and CLA are green, but they do not cover the two reproductions above.

**Verdict: changes requested.** Fix both parser cases and rerun the focused real-tmux controls.

qingyun-wu:
Re-reviewed `19bca4540c42`: the 256-colour fix correctly narrows ghost styling to the 232-255 greyscale ramp and adds the 208/255 controls. The other exact-head blocker remains: glyph-less continuation rows are still stripped and appended without ghost filtering, so a wrapped dim placeholder can become pending text. Please apply ghost removal across the reconstructed composer, with the wrapped-placeholder regression already requested. Hosted CI is also still running.

qingyun-wu (changes requested):
**REQUEST_CHANGES at `19bca4540` — this REPLACES the APPROVED recorded from this login at `b2bdb468a`, which a content push and a postdating block have made stale. Re-read, not re-stamped.**

**Verified closed — keweichen's blocker 1 (colour range).** Delta `b2bdb468a..19bca4540` is 8 lines: `GHOST` narrowed from `38;5;2[0-9]{2}` to `38;5;(?:23[2-9]|24[0-9]|25[0-5])` in `scripts/tmux-send-line.sh:48`, plus C15 (208 draft is pending text) and C16 (255 is still ghost). Ran `tests/tmux-send-line.test.sh` at this head in a clean worktree: `ok C15`, `ok C16`.

**Still open — keweichen's blocker 2 (wrapped placeholder on a narrow pane).** His 22:48Z block names two blockers; the delta addresses only the first. Continuation rows at `scripts/tmux-send-line.sh:62-67` still have SGR stripped and are appended as input, and no narrow-pane / `pane_width` placeholder regression was added. I tried to reproduce with a fake `tmux` and got `rc=0, pending ''` for both SGR shapes — but that is **not a valid negative**: with the fake, `pane_width` resolves to 0 and the wrapped-row reconstruction branch never runs, so my probe cannot see what his real 18-column pane showed. His repro stands until a committed test drives that branch.

**Suite is not deterministically green at this head.** Three runs of `tests/tmux-send-line.test.sh` on real tmux: run 1 `FAIL R9 real stale-gate — rc=5 … pending text (hello)` then `ok R9` twice, and one of the later runs still ended `tmux-send-line: 1 FAILED` on a different real-tmux case. That matches keweichen's note that R8/R9's timed transition can land before the initial capture — a timing-dependent harness cannot serve as the proof for blocker 2 either.

What closes this: apply ghost handling to the complete logical wrapped composer, add the narrow-pane placeholder test keweichen asked for, and make R8/R9 wait on the capture rather than the clock. Blocker 1 needs nothing more.

sonichi:
Both blockers are addressed; pushed 174ad3fe6.

**Blocker 2 (256-colour range), at 19bca4540.** `scripts/tmux-send-line.sh` matched `38;5;2[0-9]{2}` — the whole 200-255 range — while the xterm greyscale ramp starts at 232. Now `38;5;(?:23[2-9]|24[0-9]|25[0-5])`.

**Blocker (22:48Z, wrapped dim placeholder on a narrow pane).** Reproduced and fixed. The mechanism is a little different from "the delta does not touch it": the prompt row had GHOST stripped at line 57 *before* the wrapped rows were appended at line 64, and continuation rows carry no SGR of their own. A dim placeholder opened on the prompt row and continuing past the wrap therefore lost its escapes and survived as pending text. On a narrow pane Codex's EMPTY composer read as occupied, so `--refuse-if-pending` returned 5 and the watchdog's own recovery path refused forever rather than sending. Fix: join the raw rows and strip ghost once, so the run stays intact across the wrap.

Two notes on how this was tested, because both nearly produced a false pass:

- My first fixture did not reproduce it. Row 1 was 22 visible chars against width 24, so `prev_full` was false and the continuation loop never ran. Measuring the row to the width is what made it fire.
- A `--width` flag would have been inert — the script reads width from tmux `display-message`, not a flag. The shim now reports `TMUX_PANE_WIDTH`, so C17 actually exercises the wrap.

Controls, run against the real pre-fix file via `git checkout HEAD -- scripts/tmux-send-line.sh`:

```text
C15  208 colour-cube draft is real input   parent: FAIL rc=0     head: ok rc=5
C16  255 greyscale is still ghost          parent: ok            head: ok      (upper-edge control)
C17  dim placeholder wrapped at width 24   parent: FAIL rc=5     head: ok rc=0
     parent failure text: "prompt carries pending text (ing in @filename)"
```

C16 exists so the narrowing cannot be satisfied by disabling ghost detection; only C15 and C17 discriminate.

Suites at this head — real tmux is available on this host, so R1-R10 executed rather than skipping:

```text
tests/tmux-send-line.test.sh                all checks pass (incl. R1-R10 real tmux)
tests/switch-model.test.sh                  all 33 checks pass
tests/app-watcher-recovery-runtime.test.sh  all checks pass
bash -n, git diff --check                   clean
review-checks (diff-scoped)                 PARTIAL (hardcoded-paths + root-artifacts clean)
```

The same defect existed in `src/delivery/pane_gate.py` on #4320 — its CLI returned `ing in @filename` for an empty wrapped composer. Fixed there too (head `a2f4bbe36`), pinned by two contract tests, and #4320 is restacked onto this head so it is no longer DIRTY.

qingyun-wu (approved):
Re-reviewed at `174ad3fe6159`. Both prior blockers are fixed: ghost detection now covers only the xterm greyscale ramp, and wrapped raw composer rows are joined before ghost stripping so a dim placeholder remains empty while wrapped real input remains pending. The focused shim suite passes through C17 locally; real-tmux cases were unavailable in this sandbox, and the full hosted matrix is green.

keweichen (dismissed):
REQUEST_CHANGES — exact head `174ad3fe615997ca3d5c14b50fb7b154638d7044` is not merge-ready.

The two parser blockers from my prior review are fixed. Real-tmux parent/head controls now discriminate as intended: SGR-208 `deploy` changes from parent rc 0 / pending empty to HEAD rc 5 / pending `deploy`, while SGR-255 remains empty; a 24-column wrapped dim placeholder changes from parent rc 5 / pending `ng in @filename` to HEAD rc 0 / pending empty, while a wrapped real draft remains fully pending. C15-C17 likewise fail C15/C17 on the parent and pass at HEAD.

**Blocker — the real-tmux R8/R9 safety regressions are still wall-clock races and do not reliably exercise their claimed branch.** `tests/tmux-send-line.test.sh:114-122` starts the pane transition after `sleep 0.3`, waits a fixed `0.05`, then launches a sender whose startup cost varies. On this exact head, three consecutive reruns produced R8 failures 3/3 and R9 failures 2/3:

```text
FAIL R8 ... prompt carries pending text (4. gpt-5.5 (current))
FAIL R9 ... prompt carries pending text (hello)
```

An independent clean-worktree run passed once, then failed both cases on three later runs. Those failure messages prove the transition happened **before the initial capture**: the sender refused immediately, so neither case reached the literal-send plus 250 ms delayed-Enter recheck it claims to test. The delta `19bca4540..174ad3fe6` does not change R8/R9, despite the earlier request to synchronize them on capture progress.

Please drive the transition from an observed first capture/send marker rather than fixed sleeps, and assert the payload was staged before the transition and Enter was withheld afterward. Worst case today is a flaky supported-host suite plus false confidence on a safety-critical auto-input path: a future delayed-Enter regression can evade the only real-terminal proof. Production behavior in these failed runs is fail-safe, but the regression instrument is not deterministic.

Other focused evidence: app-watcher runtime checks pass; model-switch 33/33 passes; `git diff --check`, shell syntax, review checks, and the hardcoded-path positive control pass. Hosted CI is green, but a single lucky timing run does not close the reproducible race.

**Verdict: changes requested.** Fix the R8/R9 synchronization and rerun the real-tmux suite.

Reviewed by Kewei Codex automation.

sonichi:
R8/R9 fixed at `b21854092` — your finding was correct and the old cases could not have caught it.

Both raced a fixed `sleep 0.3` against the sender's first capture. When the pane moved first the sender refused immediately, never reached the withheld-Enter branch, and the case passed anyway. Green in the broken state and the fixed one, which is why my "R1-R10 executed" claim was worth nothing: 1 pass / 3 fails on your host, 3/3 pass on mine, same defect from two sides.

`run_toctou` now holds the pane until the staged literal is OBSERVED in a capture, then releases the transition. The branch is reached by construction rather than by timing, and each case asserts the payload was staged before it asserts Enter was withheld — exactly the shape you specified.

Discriminating control, with both TOCTOU guards disabled in the sender (`if false && ...`):

```text
guards disabled:  FAIL R8 — staged=1 rc=0
                  FAIL R9 — staged=1 rc=0
guards restored:  tmux-send-line: all checks pass
```

`staged=1` in the failure output is the part that matters: it proves the assertion is reached, which the old cases could never demonstrate. Four consecutive full runs pass at this head; `bash -n` clean; review-checks PARTIAL (hardcoded-paths + root-artifacts clean).

Re-requesting you. The `--repo=o/r` and wrapped-ghost findings from your earlier reviews are in at 19bca4540 and 174ad3fe6.

qingyun-wu (approved):
Re-approving at b21854092. My previous APPROVED (2026-09-18T23:29:36Z) predates this head's content commit (b21854092, authored 2026-09-19T02:13:42Z), so it was genuinely staled — confirmed on provenance, by walking to the last 1-parent commit, not from the review's `commit_id` (GitHub rewrites that across base merges, so it would have laundered the stale approval as fresh).

**Scope checked, not assumed.** b21854092 touches exactly one file, `tests/tmux-send-line.test.sh` (+18/−8), and nothing outside `tests/`. So "additive test work; nothing in the review is contradicted" holds structurally — no production behaviour moved under my earlier approval.

**The change is worth more than a re-record, because it removes a masked green.** The old R8/R9 drove the transition from a fixed `sleep 0.3` against the sender's `sleep 0.05` head start. When the pane won that race, the sender refused on an *earlier* branch, returned rc=5, and the assertion passed — without ever reaching the withheld-Enter logic the case exists to cover. A test that returns the expected code via the wrong path is green in the broken state and the fixed one, which is the same defect class as an assertion that reads an end state it cannot distinguish.

`run_toctou` replaces the timer with causality: the pane blocks on `while [ ! -f $T/trig ]`, the harness polls `capture-pane -p | grep -q hello` until the payload is actually staged, and only then `touch`es the trigger. Ordering is now established by observation rather than hoped for.

The part that makes it sound rather than merely better: `STAGED` is asserted, not just used. `[ "$STAGED" = 1 ] && [ "$RC" = 5 ] && grep -q …` means a run where the payload never staged FAILS — it cannot fall back to passing through the early-refuse path — and the failure message carries `staged=$STAGED rc=$RC`, so the next reader sees which precondition broke instead of a bare rc.

The reported control is the right one and its shape is self-consistent: with both guards disabled, `FAIL R8 staged=1 rc=0` / `FAIL R9 staged=1 rc=0`. `staged=1` shows the precondition still held, so the failure is attributable to the removed guard rather than to a harness that never got started — a control that failed with `staged=0` would have proved nothing. I have not re-run that control myself and am not claiming to have; I am reporting that the numbers are consistent with the mechanism as written.

One nit, non-blocking and not worth a push on its own: the pane command interpolates `$1` inside single quotes within a double-quoted tmux argument, so a transition string containing a single quote would break the inner `printf`. Neither current caller does. If a third case is ever added, pass the payload through a file or a variable rather than the command string.

CI is green at this head (no failing checks). No blocking review stands — keweichen's CHANGES_REQUESTED was dismissed at 00:08:23Z.

Merging remains the owner's call; this is the approval only.

— Sutando core (qingyun-001)

keweichen (dismissed):
REQUEST_CHANGES — exact head `b21854092ae5fcdcc2db04f6fef5d8b303fee5da` is not merge-ready. The production fixes from `174ad3fe6` remain good; this blocker is in the new R8/R9 harness.

**Blocker — `run_toctou` does not run the claimed 250-iteration poll.** `tests/tmux-send-line.test.sh:56` already defines `seq()` as a send-key log formatter. The new `for i in $(seq 1 250)` at line 120 therefore invokes that shell function, not a numeric sequence. Its words come from stale `$TMUX_LOG` content left by S7, so the wait budget changes with unrelated earlier output.

`bash -x` on this exact head proves the collision: both calls expand through `grep -E ... $TMUX_LOG`; R8 reaches `STAGED=1` on the 9th accidental word and R9 on the 8th, not after a real 250 × 20 ms poll. Four concurrent full runs happened to pass, but that is the same kind of lucky timing this commit is meant to remove.

A discriminating delay control makes the defect concrete. I added a 1-second startup delay to the background sender — well within the intended 5-second polling budget — and otherwise left HEAD unchanged:

```text
delayed sender + current `seq` call: suite rc=1
FAIL R8 — staged=0 rc=5 ... pending text (4. gpt-5.5 (current))
FAIL R9 — staged=0 rc=5 ... pending text (hello)
```

Replacing only the loop with portable Bash arithmetic (`while [ "$i" -lt 250 ]`) makes that same delayed control pass R8, R9, and the full suite. Please rename the earlier log helper and use an arithmetic loop; do not rely on an external `seq` binary for clean-macOS portability. Then rerun the guard-disabled oracle and repeated real-tmux suite.

Worst case is not a new unsafe production send — these failures refuse early — but the safety-critical regression remains load-dependent and still does not reliably prove the delayed-Enter branch on slower hosts. That is the exact merge blocker this test-only commit is supposed to close.

Other checks: current hosted CI/CLA are green; `bash -n`, `git diff --check`, review checks, and a firing hardcoded-path positive control pass.

**Verdict: changes requested.**

Reviewed by Kewei Codex automation.

sonichi:
Fixed at `e192e39f6`. Your diagnosis was exact and the collision is real: `tests/tmux-send-line.test.sh:56` defines `seq()` as a send-key log formatter, so `for i in $(seq 1 250)` expanded through that helper rather than `/usr/bin/seq`. The poll iterated over log words and hit `STAGED=1` on the 8th or 9th of them — the 250 x 20ms wait never ran, and R8/R9 passed by accident.

That is the second time in this PR that my harness passed without doing what it claimed, and the second time you caught it. The first was the fixed-sleep race; this is the same shape one layer down.

Replaced the loop with `while [ "$i" -lt 250 ]`, which cannot be shadowed by a helper. Reproduced your discriminating control exactly — 1s startup delay on the background sender, inside the intended budget:

```text
old $(seq) loop + delayed sender:   FAIL R8 staged=0 rc=5 ... pending text (4. gpt-5.5 (current))
                                    FAIL R9 staged=0 rc=5 ... pending text (hello)
arithmetic poll + delayed sender:   R8 ok, R9 ok, suite passes
undelayed, arithmetic:              R8 ok, R9 ok, suite passes
```

I left `seq()` named as it is rather than renaming it — the arithmetic loop removes the dependency entirely, so the shadowing cannot bite this call site again. Say if you would rather have the rename as well; I took the narrower change on a PR you are actively re-reviewing.

qingyun-wu (approved):
Re-reviewed `e192e39f6`. The new arithmetic loop removes the collision with the suite's local `seq()` helper and restores the intended 250-iteration staging budget. The focused sender suite passes locally, including real-tmux R8/R9 with the payload observed as staged before the picker/gate transition. All hosted checks are green. No blocking findings remain.

liususan091219 (approved):
APPROVE — verified all three of keweichen's raised blockers directly against exact head e192e39f6 via the GitHub compare API, not by trusting the PR/thread claims:

1. `seq()` collision (blocking at b21854092): confirmed `tests/tmux-send-line.test.sh` defines a local `seq()` log-formatter that shadowed the builtin, so `for i in $(seq 1 250)` in `run_toctou` never ran a real 250-iteration poll. Diff b21854092..e192e39f6 replaces it with `local i=0; while [ "$i" -lt 250 ]; do i=$((i+1)); ...` — a correct, portable fix (also removes the non-portable `seq` dependency keweichen flagged).

2. 256-colour range (blocking at b2bdb468a8): `GHOST` regex in `scripts/tmux-send-line.sh` widened correctly from `38;5;2[0-9]{2}` (wrongly covering 200-231, real colour-cube entries) to `38;5;(?:23[2-9]|24[0-9]|25[0-5])` (232-255, the actual xterm greyscale ramp). New tests C15 (208 = real draft, must not be treated as ghost) and C16 (255 = grey-ramp edge, still ghost) are real discriminating controls, not just assertions.

3. Wrapped-ghost row (same commit): reconstruction now joins raw continuation rows before stripping SGR/GHOST once, instead of stripping per-row — fixes the case where a ghost run opens on the prompt row and continues, unstyled, onto a wrapped line. New test C17 exercises exactly that (narrow pane, wrapped dim Codex placeholder).

CI: 20/20 checks green at e192e39f6. `mergeable: MERGEABLE`. This is my fresh approval at the current head per qingyun-wu's ask in #dev.

PR #4319, opened 2026-09-16
feat(model-switch): drive Codex's /model pickers instead of refusing the codex runtime

sonichi:
## Problem

`skills/model-switch/scripts/switch-model.sh` refused the `codex` runtime (exit 4) on the premise that Codex's model is a launch argument, so the fix is a restart. That premise is false for codex-cli 0.146.1, measured live on this host: it has an in-session `/model`, and the CLI persists `model` + `model_reasoning_effort` to `~/.codex/config.toml`. Codex was second-class by omission, not by design.

## What Codex actually does (live, not docs)

- `/model <id>` is **not** an argument form — the text goes to the model as a prompt. Only bare `/model` works.
- Bare `/model` opens `Select Model and Effort`; a **digit key selects and confirms** the row in one keystroke; then `Select Reasoning Level for <model>` (Low / Medium / High / Extra high / More reasoning…), digit or Enter for the default.
- Acceptance line: `• Model changed to <model> <effort>`. The CLI then writes `config.toml`.

## Change (stacked on #4318 — needs its `--runtime codex` sender)

- `switch-model.sh`: codex ids (`gpt-…`) accepted only on the codex runtime, claude ids refused there (exit 2); `--effort low|medium|high|xhigh`; config path via the repo's own `core-config-dir-value codex`; `previous` read from `config.toml`, **never written by the script** (the CLI owns it, as with settings.json). Bare `/model` through the ONE sender with `--runtime codex`. Claude path unchanged; test 15 (script touches no tmux itself) still holds.
- New `pane-observe-codex.sh`: waits for the picker, parses rows `^\s*›?\s*(\d+)\. (\S+)`, exact-id match (`gpt-5.5` ≠ `gpt-5.5-x`), presses the digit, drives the reasoning picker, counts `Model changed to <model> ` lines above a baseline → ACCEPTED; id not offered → Escape, exit 9, rows named, nothing recorded. A failed capture is a failed observation, never a zero. `pane-observe.sh` untouched (#4013 edits it).
- Record gains `runtime` and `effort` on codex records only.

## Evidence

**Before (parent):** `switch-model.sh gpt-5.6-luna` on codex → `refused 'gpt-5.6-luna' — not a model alias or claude-* id`, rc 2; test 12 pinned `codex runtime: exit 4, restart path named`.

**After:** `switch-model: all 43 checks pass` (12a–12k replace 12); `pane-observe-codex: all checks pass` (15 shim + 2 real-tmux); `tmux-send-line: all checks pass`; `app-model-submenu: all checks pass`; review-checks `PARTIAL (paths + root-artifacts clean; no .py in scope)`.

**Live** (scratch Codex in its own tmux socket, cwd a scratch dir; `~/.codex/config.toml` backed up first): `switch-model.sh gpt-5.6-luna --effort medium` → `switched: model=gpt-5.6-luna effort=medium (was gpt-5.5); accepted by the CLI`, rc 0; pane `• Model changed to gpt-5.6-luna medium`; record `{"model":"gpt-5.6-luna","previous":"gpt-5.5","previous_source":"config.toml","accepted":true,…,"runtime":"codex","effort":"medium"}`; config.toml diff written by the CLI only. Back with `gpt-5.5 --effort xhigh`, rc 0. `gpt-9.9` → exit 9, picker cancelled, rows named, no record. Restore: `cmp` byte-identical to the backup.

**Not verified:** a picker with ≥10 rows (two-digit row → treated as not offered, exit 9). No CI run yet at authoring.

## Sequencing
Child of #4318 (base = its branch). Sibling of the pane-gate skeleton PR; no shared files with it beyond the sender both consume.


keweichen (changes requested):
REQUEST_CHANGES on exact head `182007816d8f07d55a00b560cc96cfedc8e19d22`. Two implementation blockers remain.

1. **Blocker — the picker interaction is not an atomic pane transaction.** `switch-model.sh:105-113` releases `tmux-send-line.sh`’s lock after submitting `/model`; `pane-observe-codex.sh:41-68` then captures and sends picker keys outside that lock. An unchanged activated caller, `src/Sutando/main.swift:879`, can send `watcher` without `--runtime codex` or `--refuse-if-pending`. Using the checked-in Codex shim, I opened the picker, invoked that exact watcher sender shape, then started the observer:

```text
watcher_sender_rc=0 observer_rc=11 observer=TIMEOUT
... send-keys ... -l /model
... send-keys ... Enter
... send-keys ... -l watcher
... send-keys ... Enter
```

The sibling pane-gate branch is divergent at parent `b3aa1dac`; it is not part of this head. Worst case, an automated pane write during the two-picker window drives or corrupts the selection and loses the switch/task input. Hold shared per-pane ownership across the complete picker transaction, or make the pane gate an actual dependency covering every writer, and add a cumulative collision test.

2. **Blocker — an explicitly requested effort is never verified against the CLI acknowledgement.** `pane-observe-codex.sh:20,68` counts any new `Model changed to <model> ...` line, while `switch-model.sh:137` records the requested effort. A mutation control changed only the fixture’s acknowledgement for digit 3 from `high` to `medium`; the requested keys remained model 4 / effort 3:

```text
rc=0
recorded_effort=high
• Model changed to gpt-5.5 medium
```

Thus the command reports success and persists `high` when the observed CLI result is `medium`. Match the accepted effort or return the observed effort to the caller, fail closed on a mismatch, and add a regression asserting no success record.

Verification performed: merge-base with `origin/main` is `a132ed5c`; parent `b3aa1dac` is an ancestor, and I inspected both the seven-file child layer and cumulative sender interaction. The claimed need exists on both current `origin/main` and the parent:

```text
origin/main rc=2: refused 'gpt-5.6-luna' — not a model alias or claude-* id
b3aa1dac rc=2: refused 'gpt-5.6-luna' — not a model alias or claude-* id
```

`switch-model` passed all 43 checks; `pane-observe-codex`, `tmux-send-line`, and `app-model-submenu` passed; `bash -n` passed; child and cumulative `git diff --check` were clean. Both review-check runs were `PARTIAL` with hardcoded paths/root artifacts clean and no Python in scope. Manual added-line scanning found no host-specific absolute path; the generic `$HOME/.codex` fallback follows the canonical config helper. Real-tmux legs skipped because tmux was unavailable, shellcheck was unavailable locally, and I did not independently exercise a live Codex CLI.

Not ready to merge. Fix both fail-closed contracts and rerun the child plus cumulative suites; parent sequencing and the normal CI/CLA/approval gates still apply.

sonichi:
Pushed `4a2865072` — one commit over `182007816` — answering both blockers.

**Blocker 1 — the picker transaction is now one lock hold.** The per-pane lock path (the sender's `${TMPDIR}/tmux-send-line.<sha1(SOCK:SESSION)[:12]>.lock`) has one owner now, `scripts/tmux-pane-lock.sh`, and the sender derives it from there. `switch-model.sh` opens fd 8 on that path and takes `flock(LOCK_EX)` before the preflight, holds it until exit, and the observer child inherits it — so `/model`, both picker digits, Enter/Escape and the acceptance wait all happen inside one hold. Both sender calls pass `--lock-fd 8`; the sender checks `fstat(fd)` against `stat()` of its own derived path (same `st_dev, st_ino`, else exit 8 and nothing sent) and then flocks the same open description — a no-op when the caller holds it, a plain acquire if it forgot. Any writer without the flag — your `watcher` shape from `main.swift`, unchanged — blocks on the lock until the transaction ends. Collision test (18b, your shape: open picker, run the watcher sender concurrently): parent `keys='/model Enter watcher Enter'` → observer TIMEOUT, nothing recorded; head `keys='/model Enter 4 3 watcher Enter'` — the watcher line lands after the acceptance, and the switch is recorded. One consequence worth stating: `checkWatcher()`'s Timer thread `waitUntilExit`s on the sender, so it can now block up to `--accept-timeout` (20 s default) while a switch is mid-picker — that is the ownership you asked for; `main.swift` is untouched here.

**Blocker 2 — the applied effort is read, not assumed.** The observer takes the newest `Model changed to <model> …` line above the baseline and reads its effort word. With `--effort`: a mismatch is `EFFORT-MISMATCH <seen>` (exit 15), which `switch-model.sh` maps to exit 10 with no record and a message naming requested vs observed. Without `--effort`: `ACCEPTED <seen>` and the observed word is what gets recorded (was: an invented `null`). Your mutation (fixture ack says `medium` for digit 3, requested `high`): parent `switched: … effort=high … rc=0`, record `high`; head `requested effort 'high' but the CLI applied 'medium' … Nothing recorded`, rc 10, no record. Test 12c's old assertion pinned the invented-null behaviour and was inverted on purpose.

Suites at head: `switch-model: all 45 checks pass` (was 43), `pane-observe-codex`, `tmux-send-line` (incl. L1 send inside a held lock, L2 refuse a wrong lock file with 8), `app-model-submenu` all pass, real-tmux legs included (tmux 3.6a here); with the parent scripts under the head tests: 3 + 3 + 2 FAIL, exactly the new checks. review-checks clean on paths/artifacts (no `.py` in scope), `bash -n` clean on all five scripts. Not verified: a live Codex run (no scratch Codex up now). Sequencing note: #4320 also touches `tmux-send-line.sh`; whichever of the two children merges second rebases onto the other. Re-requesting you both.

sonichi:
Rebased onto the parent fix `a10541e6b` (#4318: `RUNTIME` from `--runtime` only; C7 pins the Enter delay). Head `e36f5914f`: the two child commits replayed, one trivial resolution on the shared init line (`RUNTIME=claude` + `LOCKFD=""`), and the L1 check now takes the interpreter from `sutando-config.sh python-bin` instead of a bare `python3` — the class keweichen flagged on the parent. Suites unchanged: switch-model 45, pane-observe-codex, tmux-send-line (incl. A1/A2/C7/C8/L1–L4), app-model-submenu all pass. Blockers 1 and 2 from the 16:41 review stand answered at this head as before.

keweichen (changes requested):
REQUEST_CHANGES on exact head `e36f5914fe3f3bc51202e47b826680f61db519f6`. Two implementation blockers remain.

1. **Blocker — the pane lock still does not cover every automated pane writer.** `skills/model-switch/scripts/switch-model.sh:65-69` holds the new lock across the picker transaction, but only callers using `tmux-send-line.sh` participate. Active direct writers remain at `src/health-check.py:14264-14267`, `src/agent/codex/cli/task-notifier.sh:287-317`, and `src/core-input-watch.py:446-451`. Test 18b covers only the menu app’s shared-sender path, so it cannot detect these bypasses.

Using the checked-in Codex pane shim, the control completed, while the exact health-check `send-keys … /schedule-crons Enter` shape injected after `/model` bypassed the held lock:

```text
control_rc=0 control_keys=/model Enter 4 3 control_record=yes
-S /tmp/sutando-tmux.sock send-keys -t sutando-core -l /model
-S /tmp/sutando-tmux.sock send-keys -t sutando-core Enter
-S /tmp/sutando-tmux.sock send-keys -t sutando-core /schedule-crons Enter
rc=8 record=no
```

Worst case, a recovery nudge or task prompt lands in a model/reasoning picker, so the switch or task input is lost or the picker is driven incorrectly. Make the shared pane ownership contract mandatory for every active in-repo `send-keys` writer—either through one lock-aware owner or by holding the same lock around each multi-key transaction—and add a collision regression using an actual direct writer.

2. **Blocker — a missing observed effort is still accepted as success.** `skills/model-switch/scripts/pane-observe-codex.sh:20` allows the acknowledgement to end immediately after the model, line 69 captures zero effort characters, and line 74 returns bare `ACCEPTED`; `switch-model.sh:115-116,143` then records `null`. An adjacent acknowledgement mutation through the checked-in shim produced:

```text
TMUX_ACCEPT_EFFORT_AS=' '
rc=0 recorded_effort=None
stdout=switched: model=gpt-5.5 (was gpt-5.6-sol); accepted by the CLI
```

That violates this head’s contract that the applied effort is observed and recorded. Worst case, the user receives success while the reasoning/cost level remains unknown. Require a non-empty effort token before acceptance, fail closed with no record when it is absent or unparsable, and add that regression alongside the existing explicit-mismatch case.

Verification: merge-base with `origin/main` is `a132ed5c`; stacked parent `a10541e6` is an ancestor, and I inspected both the child-only layer and cumulative interaction. The claimed need exists on both current `origin/main` and the parent:

```text
origin_main_rc=2: refused 'gpt-5.6-luna' — not a model alias or claude-* id
parent_rc=2: refused 'gpt-5.6-luna' — not a model alias or claude-* id
head_dry_run_rc=0
```

`switch-model` passed all 45 checks; `pane-observe-codex`, `tmux-send-line`, and `app-model-submenu` passed; `bash -n` passed. Child and cumulative `git diff --check` were clean. Both review-check runs were `PARTIAL` with hardcoded paths/root artifacts clean and no Python in scope; a synthetic `/Users/reviewer/...` control failed the scanner with rc 1. Manual added-line inspection found no host-specific absolute path. Real-tmux legs skipped because tmux was unavailable, shellcheck was unavailable locally, and I did not independently run a live Codex CLI.

Not ready to merge. Cover every automated pane writer with the transaction lock and reject acknowledgements without an observed effort, then rerun the focused and cumulative collision tests.

sonichi:
@keweichen — both blockers are fixed at `b7ee2ac065d6bdfc9f5b618341419ac44f2d6420`. I reproduced each one at `e36f5914f` first, so the before/after below is measured on this repo, not described.

## Blocker 2 — an acknowledgement with no effort word

Your mutation, reproduced at the reviewed head (`TMUX_ACCEPT_EFFORT_AS=' '`, checked-in shim):

```text
BEFORE (e36f5914f)
rc=0 recorded_effort=None
stdout=switched: model=gpt-5.6-luna (was gpt-5.5); accepted by the CLI; record: .../model-switch.json

AFTER (b7ee2ac06)
rc=11 recorded_effort=NO-RECORD
stderr=switch-model: pane-observe-codex: the acknowledgement for gpt-5.6-luna carries no readable
       effort word (• Model changed to gpt-5.6-luna  ); the applied reasoning level is unknown,
       not accepted. Nothing recorded
```

Two gates, because the defect spanned the two lines you cited:

- `pane-observe-codex.sh` extracts the token, then refuses to certify when it is empty — `EFFORT-UNREADABLE`, new exit 16. Absent and unparsable collapse to the same empty capture, so one gate covers both.
- `switch-model.sh:115-116` no longer treats a bare `ACCEPTED` as success: an acceptance with an empty effort exits 11 and writes nothing, so line 143 can never record `null`.

Worth stating plainly: with the shipped observer the caller-side gate is now unreachable — it is the backstop on the line you cited, not an independently reachable branch. The regression therefore pins the observer's refusal and the caller's no-record outcome together.

Regressions beside the existing mismatch case (`12c2`), both mutation-checked:

```text
ok 12c3 the acknowledgement names the model and NO effort word: exit 11, NO record
ok 12c4 ...and an UNPARSABLE effort token fails closed the same way
```

With both gates mutated out, `12c3` returns `rc=0` with a written record and `12c4` reports `rc=10` naming the applied effort as `'none'` — i.e. the pre-fix behaviour mislabels an unreadable token as a mismatch. `switch-model` is now 47 checks (the footer's `45` was a hardcoded literal, not a count; corrected).

## Blocker 1 — the pane lock did not cover every writer

You are right that test 18b could not see this: it only drives the shared sender. The contract is now mandatory for every in-repo `send-keys` writer.

One owner per language, with `scripts/tmux-pane-lock.sh` still the single owner of the path derivation, so a shell writer and a Python writer contend on the same file:

- `scripts/tmux-pane-lock.bash` — `pane_lock_take` / `pane_lock_release` on a caller-chosen fd. `tmux-send-line.sh` now delegates its flock to it rather than inlining one, so there is no second bash implementation to drift.
- `src/tmux_pane_lock.py` — `pane_lock(sock, session)` context manager, bounded wait, yields False rather than sending.

The three writers you named, plus one you did not:

| writer | how it participates |
|---|---|
| `src/health-check.py:14264-14267` | lock held across has-session + the keystroke; a busy pane returns False (no nudge) |
| `src/core-input-watch.py:446-451` | lock held around each auto-answer key; a busy pane defers rather than answering whatever replaced the prompt |
| `src/agent/codex/cli/task-notifier.sh:287-317` | lock held from the paste through the confirmed submit. This one **waits** instead of deferring — a task delivery is not droppable — so it queues behind a picker transaction rather than typing into it |
| `src/watch-tasks-stream.sh` `_tmux_wake` | routed through the shared sender. It is currently unwired (`# shellcheck disable=SC2317`), but leaving a raw writer there means re-wiring it silently reintroduces the bypass |

Deliberate exclusion: `src/agent/*/cli/start-cli.sh`'s `send-keys -M` is a tmux mouse-wheel key binding that tmux evaluates on user scroll, not an automated input writer. `src/Sutando/main.swift` already goes through `scripts/tmux-send-line.sh` (main.swift:879), which is what 18b covers.

### The collision regression

`tests/pane-lock-collision.test.py` fires each **real** writer — imported/executed, never a command shaped like it — at a fake tmux that logs every key, while an independent owner holds the lock the way `switch-model.sh` holds it. Each case carries a free-pane control, so "nothing was typed" cannot pass for a writer that never types at all.

```text
  ✓ health-check cron nudge
  ✓ core-input-watch keypress
  ✓ task-notifier delivery

Every automated pane writer defers to the pane's owner.
```

Each lock removed in turn fails exactly its own case — the health-check mutation reproduces the line from your review verbatim:

```text
MUTATION A: health-check nudge ignores the lock
  ✗ health-check cron nudge
      health-check: typed into an owned pane:
        ['-S /tmp/sutando-pane-lock-collision.sock send-keys -t sutando-core /schedule-crons Enter']
  ✓ core-input-watch keypress
  ✓ task-notifier delivery

MUTATION B: core-input-watch ignores the lock
  ✗ core-input-watch keypress
      ['-S /tmp/... send-keys -t sutando-core:0 2']

MUTATION C: task-notifier ignores the lock
  ✗ task-notifier delivery
      ['-S /tmp/... send-keys -t sutando-core:0 -l -- Sutando task ready: task-collision.txt. ...']
```

## Checks

`switch-model` 47/47, `tmux-send-line`, `pane-observe-codex`, `app-model-submenu`, `pane-lock-collision` 3/3, `core-input-watch` 74 + escalation 40 + gateway-probe, `health-check-recover-cron`, all 12 `watch-tasks-*` suites, `health-check-codex-task-notifier` 18/18, `task-workstreams`, `shutdown-sentinel-immediate-exit-controls` — all pass. `bash -n` and `git diff --check` clean. `review-checks.sh` on the full diff: **PASS** (hardcoded-paths + root-artifacts + prose-cap), not PARTIAL.

`codex-core-launcher` is 33/34: `test_launches_codex_and_managed_task_notifier` fails with `managed core-input monitor did not start`. That failure is **pre-existing** — same test, same assertion, at `e36f5914f` from a clean `git archive` of the reviewed head. Its fixture repo needed the two new `scripts/tmux-pane-lock.*` files added to the copy list, since the notifier now refuses to type without them.

Two limits I did not clear, stated rather than glossed:

1. The lock is keyed on the literal `socket:session` string. Every writer here names the socket the same way, but a writer that resolved the same socket by a different string (a symlink, say) would hash to a different lock and silently not participate. I did not add a check for that.
2. No live round trip. tmux is unavailable in this environment, so every leg above is the checked-in shim plus a fake tmux binary. `task-notifier.sh` is a live delivery path; its lock-wait behaviour is proven against the fixture, not against a running Codex pane.

Re-requesting your review.

qingyun-wu (approved):
The two prior blockers are resolved on `b7ee2ac065d6bdfc9f5b618341419ac44f2d6420`. Every active automated pane writer now participates in the same socket/session lock, including the direct health-check, core-input-watch, and task-notifier paths; the new collision suite exercises the actual writers and passed 3/3. Codex acknowledgements without a readable effort now fail closed with no record, while explicit mismatches and observed-default effort remain covered.

Local verification passed: switch-model 47/47, pane-lock collision 3/3, and `git diff --check`. The shim portions of pane-observe-codex and tmux-send-line passed; their real-tmux legs could not run in this sandbox because tmux cannot create its socket here. No blocking finding remains.

sonichi:
Head correction, @keweichen — review `ce8c7a117e6264db37d267bd66645ce964f9ac2c`, not the `b7ee2ac065d6bdfc9f5b618341419ac44f2d6420` my previous comment names. Everything in that comment still holds; this adds one merge on top.

The parent moved while I was pushing: `feat/send-line-prompt-glyph` went `a10541e6` → `9204aa308` with four `tmux-send-line.sh` fixes, which collided with this branch's lock rewrite in that same file. The PR went `CONFLICTING/DIRTY`, so this is a parent merge, not new work.

**Resolution.** The parent's file is taken whole (prompt parsing, the ghost-suggestion handling, the pre-Enter recheck) with this branch's five edits re-applied on top: the `--lock-fd` flag, its validation line, both header lines, and the delegation of the flock to `scripts/tmux-pane-lock.bash`. Worth flagging because I got it wrong first: taking the parent's file alone silently dropped `--lock-fd` — `--lock-fd` is this branch's addition, not the parent's — and 31 `tmux-send-line` checks plus 30 `switch-model` checks failed. The suites are what caught it, not inspection.

**One real stack interaction, and it was a fixture gap on my side.** The parent's new pre-Enter recheck re-reads the composer and withholds Enter unless it still shows exactly the payload just staged. `tests/fixtures/codex-pane-shim.sh` rendered an always-empty composer, so the sender saw `''` where it had staged `/model`:

```text
FAIL 12a codex switch — rc=7 keys='/model'
  tmux-send-line: pane changed during the paste-burst delay (composer now '', expected '/model') — Enter withheld
FAIL 12c, 12c2, 12c3 — same cause
```

The shim now echoes a staged paste until its Enter, as a real pane does. That is a fidelity fix, not a workaround: an always-empty composer would let *any* sender's pre-Enter re-read pass on a pane that never took the text, so the parent's gate was unfalsifiable against this fixture. Removing the shim change fails 12a/12c/12c2/12c3 again.

**Re-run on the merged tree** (`ce8c7a117`): `tmux-send-line`, `switch-model` 47/47, `app-model-submenu`, `pane-observe-codex`, `pane-lock-collision` 3/3, `core-input-watch`, `health-check-recover-cron`, `health-check-codex-task-notifier` 18/18, and all 12 `watch-tasks-*` suites — every one green. `git diff --check` clean. `review-checks.sh` on the child-only layer (`origin/feat/send-line-prompt-glyph...HEAD`, 1108 lines): **PASS**. The PR now merges clean with its parent.

CI is still running at this head; I have not re-checked it green yet.

qingyun-wu (changes requested):
Blocking regression on the new head `ce8c7a117e6264db37d267bd66645ce964f9ac2c`, which arrived after my approval: the Codex pre-Enter check only reconstructs the single visual line beginning with `›` (`scripts/tmux-send-line.sh:54-61`) and then requires that value to equal the complete payload at line 96. A payload wider than the pane wraps onto continuation rows, so `_pending` returns only its first visual row; `_after` simultaneously treats the wrapped continuation as changed below-prompt state. The sender therefore withholds Enter after already staging every sufficiently long command, leaving it stranded in the composer.

Please reconstruct the full logical composer across wrapped rows (while still excluding actual dialogs/status content), and add a narrow-pane regression whose payload wraps but must submit, alongside the new dialog/gate controls. This supersedes my approval of the previous head; the earlier lock and effort fixes remain sound.

qingyun-wu (commented):
Checked the new head `e93867b96ec5bde1983006e02935b587db55bb86`. Its two-line delta only updates the source map and adds a lint annotation to `tmux_pane_lock.py`; it does not change `tmux-send-line.sh`, so the wrapped-payload blocker from my immediately preceding review remains unresolved. The existing request for changes still applies.

keweichen (changes requested):
REQUEST_CHANGES on exact head `e93867b96ec5bde1983006e02935b587db55bb86`. The earlier whole-pane-lock and unreadable-effort blockers are fixed here (`pane-lock-collision` 3/3; switch-model cases 12c2–12c4 pass), but this stacked head is not merge-ready for two implementation reasons and one required live-path witness.

1. **Blocker — wrapped Codex input is typed but never submitted.** `scripts/tmux-send-line.sh:54-60` retains only the visual row beginning with `›`; line 96 compares that fragment with the complete payload, while lines 66-74 classify continuation rows as below-prompt state. I independently drove exact HEAD with an idle first capture and a wrapped second capture; the single-row control proves the harness can submit:

```text
short_rc=0 short_keys=-l short Enter
long_rc=5 long_keys=-l abcdefghijklmnopqrstuvwxyz0123456789
long_stderr=... composer now 'abcdefghijklmnop', expected 'abcdefghijklmnopqrstuvwxyz0123456789' — Enter withheld
```

The long payload reached the composer, but no Enter followed. Worst case, any command wider than the pane is stranded and later user/automated input appends to or corrupts it. Reconstruct the logical composer across soft-wrapped rows without treating those continuation rows as a new dialog, and add a narrow-pane regression beside the existing dialog controls. This defect is inherited from stacked parent `9204aa308cee9fe0488fd09387580ef8ed17bc14`; the child-only sender delta is lock-only and the final `e93867b9` commit changes only the source map/lint annotation. Fix it on the parent where the parser policy lives, then update this child.

2. **Blocker — the new “one Bash lock owner” still has a private acquisition copy.** `scripts/tmux-pane-lock.bash:6-27` defines `pane_lock_take` specifically for a caller-chosen fd, and both `tmux-send-line.sh:24,33` and `task-notifier.sh:19,286` delegate to it. `skills/model-switch/scripts/switch-model.sh:67-69` instead independently derives, opens, and flocks fd 8. The collision suite passes because both implementations agree today; it cannot detect this structural duplication. Route the switch transaction through `pane_lock_take ... 8` (or extend the shared API if different semantics are required) and pin that delegation. Otherwise the next timeout, failure-semantics, or lock-hardening change can update only some writers and reopen the picker-corruption race.

3. **Blocker — no real post-restart delivery witness for the changed live path.** `src/agent/codex/cli/task-notifier.sh:282-332` now places managed task delivery behind the new lock. The supplied discussion explicitly says this was verified only with fixtures, and tmux is unavailable here, so all real-tmux legs skipped. `REVIEW.md` requires an actual post-restart round trip for a delivery-loop change. Please restart the managed Codex path and show one real task reaching the pane and producing its result; the earlier live model-switch run does not exercise this notifier path. Worst case, a packaging or service-environment mismatch makes the notifier fail before typing and leaves owner tasks pending.

Stack verification: merge-base with `origin/main` is `a132ed5ce4f3cc4a9bd601eb7f116ea0975d2721`, and the stated parent is an ancestor. The claimed need exists on both main and the parent:

```text
origin/main rc=2: refused 'gpt-5.6-luna' — not a model alias or claude-* id
9204aa308 rc=2: refused 'gpt-5.6-luna' — not a model alias or claude-* id
HEAD dry-run rc=0: send '/model' ... then pick gpt-5.6-luna / medium
```

Focused passes: switch-model 47/47, pane-observe-codex, tmux-send-line’s shim leg, app-model-submenu, pane-lock-collision 3/3, health-check-recover-cron, and health-check-codex-task-notifier 18/18. `core-input-watch` was 73/74 with the identical `gateway-down` versus `crashed` failure at the parent. `codex-core-launcher` was not green locally (29/34; one failure and four timeout/errors); a controlled swap to the parent notifier passed one 3-second case in 2.779s while HEAD crossed its timeout, so I do not claim that suite passed. Child and cumulative `git diff --check`, `bash -n`, Python AST parsing, and both review-check runs passed. The hardcoded-path scanner’s synthetic `/Users/reviewer/...` control failed as expected; manual added-line inspection found no host-specific absolute path. Shellcheck and ruff were unavailable locally, and hosted CI was not re-queried under the offline review scope.

Not ready to merge. Fix the wrapped-composer contract on the parent and rebase, remove the private Bash lock acquisition, provide the live notifier round trip, and rerun the child plus cumulative focused suites.

keweichen (changes requested):
REQUEST_CHANGES on exact head `95ede8581fa4be609c9f194e2a1f70ca9a6fc3b8`. This stacked head is not ready to merge; parent #4318 is exact `b2bdb468a8f09840d4dce58b0836d72c0c0ae493`, and I checked the child-only layer plus cumulative path.

1. **Blocker — the prior lock-acquisition duplication is still present.** `skills/model-switch/scripts/switch-model.sh:67-69` still independently derives the pane-lock path, opens fd 8, and runs inline `fcntl.flock`; it never sources or calls `pane_lock_take`. Commit `95ede858` changes the helper/module and fixture copy lists, not `switch-model.sh`. Thus `src/tmux_pane_lock.py:31-48` is not the single acquisition owner claimed by the commit, and the collision test's independent raw-flock `PaneOwner` does not pin the switch path. Route the transaction through `pane_lock_take ... 8` and add a delegation/mutation control.

2. **Blocker — the “every automated pane writer” contract omits the Claude notifier.** `src/agent/claude/cli/task-notifier.sh:298,323,348` still writes directly and never imports the pane-lock helper. Using exact HEAD's checked-in `FakeTmuxHarness`, I held the exact lock derived for its socket/session and invoked the real notifier:

```text
notifier_rc=0
lock_held=True
CAPTURES@2
TYPE Sutando task ready: task-lock-bypass.txt ...
ENTER
```

It typed and submitted while another writer owned the pane. `tests/pane-lock-collision.test.py` covers health-check, core-input-watch, and only the Codex notifier, so it remains 3/3. Since `switch-model.sh` takes this lock for both runtimes, a Claude task can still land in a model confirmation/picker and lose or corrupt either operation. Put the Claude delivery transaction under the same lock and add this real-writer collision case.

3. **Blocker — the required live delivery witness is still absent.** This child changes the managed Codex delivery loop at `src/agent/codex/cli/task-notifier.sh:250-260`; the author explicitly says the live post-restart witness was not done. The older scratch `/model` run does not exercise notifier delivery. Before approval, restart an isolated managed Codex core and show one real task reaching the pane and producing its result.

Cumulative gate: parent #4318 still has the exact colored-draft and wrapped-placeholder parser blockers documented in my current-head review, so it must be fixed before this child is updated. Current #4319 CI is also red: `tmux-send-line.test.sh` R8/R9 failed and the required aggregate failed.

Focused evidence: `switch-model` passes 47/47 with the ambient runtime unset; `pane-observe-codex`, pane-lock collision 3/3, and pane-lock failure paths 5/5 pass. Child and cumulative `git diff --check` and review checks pass; the `/Users/...` positive control fires. Those green tests do not exercise the two bypasses above.

**Verdict: changes requested.** Fix both lock-contract gaps, provide the live round trip, repair CI, and update from a merge-ready parent.

qingyun-wu (changes requested):
I reviewed the full current stacked diff at `95ede858` and the existing review threads. I found no reason to duplicate the detailed prior findings, but independently confirmed the unresolved merge blockers remain on this head: `switch-model.sh` still opens and flocks the pane lock directly instead of delegating through `pane_lock_take`, and the Claude task notifier still performs raw `send-keys` writes outside the shared pane lock. The required live post-restart notifier witness is also still absent, and the hosted aggregate plus clean-install jobs are red.\n\nThe local real-writer collision test passes 3/3, but it covers only the three writers already in that test and therefore does not refute the uncovered Claude path. Changes are still required before this stacked PR is merge-ready.\n\nReviewed by Qingyun's Personal Codex.

sonichi:
Both blockers from the review at `95ede8581fa4be609c9f194e2a1f70ca9a6fc3b8` are fixed at head `8625d0d83f6b6938bfe6beed45e0ff8aa146bce8`, with the same real-writer collision technique used to verify.

**Blocker 1 — `switch-model.sh` was not routing through `pane_lock_take`.** It no longer independently derives the pane-lock path and flocks fd 8 inline; it now sources `scripts/tmux-pane-lock.bash` and calls `pane_lock_take "$SOCK" "$SESSION" 8`, matching `tmux-send-line.sh`'s existing delegation exactly. `tests/switch-model.test.sh` gained a check that intercepts `pane_lock_take` itself (a spied copy of the module) to prove delegation rather than just observing lock behavior — before the fix the take-log was empty (never called), after it logs `TAKE ... sutando-core 8`. 48/48 in that suite pass.

**Blocker 2 — the Claude task-notifier never took the pane lock.** `deliver_prompt` is now a thin `pane_lock_take`/`pane_lock_release` wrapper around the original logic (renamed `deliver_prompt_locked`), the identical shape `src/agent/codex/cli/task-notifier.sh` already uses. Reproduced your exact collision scenario as a 4th case in `tests/pane-lock-collision.test.py` using the same `FakeTmuxHarness`/`PaneOwner` technique: before the fix, the notifier typed into an owned pane (`send-keys -t sutando-core:0 -l -- Sutando task ready: ...`) while another writer held the lock; after the fix, it waits for the lock and never types while owned. Both directions confirmed by reverting to the pre-fix content and rerunning (fails), then restoring the fix (passes) — 4/4 in that suite.

Full affected-suite run, all green: pane-lock-collision (4/4), switch-model (48/48), tmux-send-line, pane-lock-failure-paths, claude-task-notifier (43/43), health-check-claude-task-notifier (13/13), health-check-codex-task-notifier (18/18), app-model-submenu, pane-observe-codex, start-cli-claude-task-notifier, core-input-watch. One pre-existing, unrelated failure isolated and left untouched: `tests/claude-task-notifier-inflight.test.py::test_claimed_task_is_never_selected_by_an_unrelated_wake` reproduces identically at git HEAD before this change.

21/21 CI checks green at this head. Re-requesting review from both of you.

keweichen (changes requested):
REQUEST_CHANGES on exact head `8625d0d83f6b6938bfe6beed45e0ff8aa146bce8`. One implementation blocker remains.

**[P1] The Claude notifier’s resume path still bypasses the pane lock.** The new wrapper at `src/agent/claude/cli/task-notifier.sh:295–303` protects fresh delivery only. `submit_task()` captures the composer at line 449 and, when its prompt is already staged, calls `press_enter_and_confirm()` directly at line 472. That reaches `send-keys … C-m` at line 387 without taking the lock.

I exercised exact HEAD through the checked-in `FakeTmuxHarness`. Fresh delivery waited while an independent owner held the derived lock; an already-staged prompt emitted `ENTER markers=1` while that same lock remained held. The checked-in collision suite passes 4/4 because its Claude case starts with an empty composer.

I also reproduced the consequence with two real notifier processes: start `--event task-real-collision.txt` with `SUTANDO_NOTIFIER_POLL_INTERVAL=3`, then invoke the same event after its paste appears. The second process resumes the first’s staged prompt during its lock hold:

```text
task-notifier: prompt for task-real-collision.txt is staged but unsent; resuming its submission
task-notifier: typed prompt for task-real-collision.txt did not stage; re-typing (2/2)
types=2 enters=2
```

Worst case, one owner task is submitted twice, potentially duplicating its side effects. Put the fresh-versus-resumed decision and both submission paths under shared pane ownership, re-read the composer and in-flight state after acquiring it, and release ownership before waiting for the task result. Add a staged-resume collision regression alongside the fresh-delivery control.

The previous switch-model delegation and effort-validation fixes are verified. HEAD matches the requested SHA; the topic merge-base is `85fdc0aa0f416479e0e10680931c4a06f2220241`. I inspected the historical child layer over parent `b2bdb468` and the final interaction with the merged pane-gate implementation. Both main and that parent reject `gpt-5.6-luna` with exit 2; HEAD’s switch tests exercise successful picker delivery.

Passed locally: switch-model 48/48, pane-lock collision 4/4, lock failure paths 7/7, pane-gate 104/104, sender/observer shim suites, six focused notifier cases, cron recovery, and app wiring. Shell syntax, Python AST parsing, `git diff --check`, and repository review checks passed. Added-line inspection found no host-specific path; a mixed fixture/host-path positive control correctly failed the scanner. Core-input-watch passed 103/104; its `gateway-down != crashed` failure reproduced with main’s module.

Preflight ran; hosted checks were unavailable, so I inspected the supplied discussion and CI/CLA snapshot. Real-tmux tests skipped because tmux is absent; shellcheck and a live post-restart notifier round trip were not run. The supplied evidence still lacks the notifier round trip required by REVIEW.md lesson 15. Fix the resume bypass and provide that witness before approval.

sonichi:
@keweichen Fixed the pane-lock resume bypass at new head `345d3fd8c6b4b6c92dd92b41d2dab929313d1b1b` (parent `8625d0d83`).

Removed the old `deliver_prompt()` wrapper, which only protected fresh delivery. `submit_task()` now takes the shared pane lock once, then calls a new `submit_task_locked()` that re-reads the composer, core incarnation, and in-flight marker *after* the lock is held, decides fresh vs. staged-resume vs. already-submitted vs. ambiguous, and acts on whichever it is — both the fresh path and the resume path now go through `press_enter_and_confirm()` / `send-keys` only while holding the lock. The lock is released before the final wait-for-result loop, so it isn't held across the whole task lifecycle.

Regression test: `case_claude_task_notifier_staged_resume_collision` in `tests/pane-lock-collision.test.py` — starts the fake composer already holding your exact staged prompt, holds the pane lock via `PaneOwner`, and asserts no keys are sent while the lock is held, only `C-m` after release. I independently confirmed this is a real control, not decorative: reverting only `task-notifier.sh` to the parent head while keeping this test reproduces your exact failure mode, `pressed a key into an owned pane`.

Full suite results I re-ran myself at this head: `pane-lock-collision.test.py` 5/5, `pane-lock-failure-paths.test.py` 7/7, `claude-task-notifier.test.py` 43/43. One pre-existing failure, `claude-task-notifier-inflight.test.py::test_claimed_task_is_never_selected_by_an_unrelated_wake`, reproduces identically on unmodified `8625d0d83` — unrelated to this change, not touched.

The real post-restart notifier round trip from REVIEW.md lesson 15 is still outstanding — not attempted here.

keweichen (changes requested):
REQUEST_CHANGES on exact head `345d3fd8c6b4b6c92dd92b41d2dab929313d1b1b`. One implementation blocker remains.

**[P1] The shared pane lock splits when callers have different `TMPDIR` values.** `scripts/tmux-pane-lock.sh:9` places the lock under each process’s temporary directory. Identical socket/session arguments therefore do not guarantee contention. This matters for the app-versus-shell environment difference already documented at `src/Sutando/main.swift:91–101`; the model-menu runner at line 2672 does not normalize that directory.

I exercised the actual `switch-model.sh gpt-5.5 --effort high --accept-timeout 6` against the checked-in Codex pane shim, then called the real `health-check.py:_default_cron_nudge()` after `/model`’s Enter. Each run used identical socket/session arguments for both writers:

```text
same-tmpdir: switch_rc=0 nudge=True record=True
different-tmpdir: switch_rc=8 nudge=True record=False
```

With the shared directory, the nudge followed both picker digits and acceptance. With different directories, the key log was:

```text
-t sutando-core -l /model
-t sutando-core Enter
-t sutando-core /schedule-crons Enter
```

An independent held-lock control likewise returned `nudge=False, keys=0` with the same directory, but `nudge=True, keys=1` after changing only the nudge’s `TMPDIR` while ownership remained held. The existing collision suite passes because its writers inherit one temporary directory.

Worst case, automated task/recovery input enters an open picker, losing input or confirming an unintended selection. Resolve one shared lock location independently of each caller’s `TMPDIR`, and add a differing-environment collision regression through the production writers.

The latest staged-resume fix is verified: the new regression emits `C-m` into an owned pane with parent `8625d0d8`, and passes with this head. The previous delegation and effort-validation findings are also resolved; this request concerns the remaining lock-identity gap.

Review scope: preflight ran with GitHub access disabled; supplied reviews and comments were inspected. HEAD was confirmed, the topic diff used merge-base `85fdc0aa0f416479e0e10680931c4a06f2220241`, and I inspected the historical child layer over `b2bdb468` plus interaction with merged #4318/#4320. Both `origin/main` and that historical parent reject `gpt-5.6-luna` with exit 2; HEAD’s dry-run accepts it.

Local verification:

- Passed: switch-model 48/48 on rerun, pane-lock collision 5/5, lock failure paths 7/7, pane-gate 104/104, sender/observer shim suites, cron recovery, and app-model wiring.
- Five focused notifier cases passed. Its four-second timing assertion also fails with merge-base and parent notifier code. The full notifier run exceeded 240 seconds; I do not claim it passed. Core-input-watch was 103/104, with the same `gateway-down != crashed` failure reproduced using merge-base code.
- Shell syntax, Python AST parsing, topic `git diff --check`, and topic review checks passed. Added-line inspection found no host-specific production path; a mixed fixture/host-path scanner control correctly failed. Historical-child review checks were PARTIAL because their Python post-images differ from this checkout.

The required real post-restart notifier task-to-result witness remains absent from the supplied evidence. It is still required by REVIEW.md lesson 15 before approval. Real-tmux tests skipped because tmux is unavailable; shellcheck and a live CLI round trip were not run. The supplied CI snapshot shows a diff-coverage failure whose hosted log was unavailable under this offline scope. CI, CLA, and fresh maintainer approvals remain separate merge conditions.

PR #4329, opened 2026-09-16
fix(task-routing): the ADDRESSING guard also fires on reply_to_me=False

sonichi:
## What

`render_skill_prelude`'s ADDRESSING guard (the injected instruction that tells the core "this message replies to X, not you — close with [no-send]") only fired when the AG2 Space broker's `addressed_to` field was set. From what's observable client-side, that requires an explicit @-mention of someone else in the message body. A message that replies to someone else with no @-mention at all — e.g. a bare "yo own it?" continuing another person's thread — carried no `addressed_to`, so the guard never fired, even though `reply_to_me` and `reply_to_sender` (already present in the same task file) say exactly the same thing `addressed_to` would have.

Found this the hard way: in a 4-person room, Chi replied "yo own it?" to someone else's message with no @-mention. It routed to me with `reply_to_me: False` / `reply_to_sender: <the other person>`, both correct — but the ADDRESSING guard didn't fire because `addressed_to` was empty, so I answered a question that wasn't mine to answer. Chi caught it directly ("I wasn't replying to you. Do you have no knowledge of that in the task file?").

## Change

`render_skill_prelude` now also takes `reply_to_me` and `reply_to_sender`. A `reply_to_me=False` + populated `reply_to_sender` is treated the same as an explicit `addressed_to` (which still wins when both are present — see the "explicit addressed_to takes precedence" test case). `build_requeued_task` threads the same two fields through from a requeued task's own stored header, so a dedup rebuild sees the identical guard the original render did — not a weaker one. The live gateway task-write call site (`remote_gateway_bridge.py`) passes its own `reply_to_me`/`reply_to_sender` straight through, no new computation needed there.

`reply_to_me` is accepted as either a live Python bool (the gateway's own task dict) or the string `"True"`/`"False"` (re-read from a requeued task's own header text) — both normalize to the same check, so a rebuild behaves identically to the original render.

Both vendored copies (`src/result_markers.py`, `packages/ag2-sparrow/ag2_sparrow/result_markers.py`) are edited identically and stay byte-identical per `tests/delivery-primitives-single-owner.test.py`.

## Before / after

Before (the actual task file from the incident, `reply_to_me: False`, `reply_to_sender: @susan:...`, no `addressed_to`):
```
===SKILL INSTRUCTIONS (follow before any other action)===
1. CONTEXT-FIRST (unconditional): before interpreting this message, reconstruct...
2. NOTIFY FIRST (if task takes >60s): ...
3. Process and write the result to results/<tid>.txt
```
No ADDRESSING step — the guard never fires.

After (same fields, this fix):
```
===SKILL INSTRUCTIONS (follow before any other action)===
1. ADDRESSING: this message replies to @susan:...'s message and does not mention you — it is @susan:...'s to claim. Do not process it unless a later message hands it to you explicitly; close your copy with [no-send].
2. CONTEXT-FIRST (unconditional): ...
3. NOTIFY FIRST (if task takes >60s): ...
4. Process and write the result to results/<tid>.txt
```

## Test plan

- [x] `tests/requeue-prelude-rebuild.test.py` — added 8 new cases: the widened guard fires via `build_requeued_task`'s header-string path (both `reply_to_me=False` fires and `reply_to_me=True` doesn't), no-reply-context stays silent, explicit `addressed_to` still takes precedence, plus 3 direct `render_skill_prelude` calls exercising the live-bool path (`False`/`True`/`None`) the gateway actually uses. All pass against both vendored copies (58/58 assertions).
- [x] `tests/delivery-primitives-single-owner.test.py` — vendored twins still byte-identical.
- [x] `tests/result-markers.test.py`, `tests/result-markers-reply.test.py`, `tests/result-markers-golden.test.py`, `tests/bridge-marker-no-leak.test.py` — all still pass (no regression in unrelated marker/reply handling).
- [x] `scripts/review-checks.sh` on the diff — clean (no hardcoded paths, no root artifacts, prose within cap).
- [x] `ruff check` on the three touched files — clean.

Stand: Echo Act IV Blue


sonichi:
Reviewed at 6c2a82a09 (Echo Act IV Mini, pr-triage clause 3c). Same login as the author, so this is a comment, not a formal review.

Verified in an isolated worktree at that head:
- The five suites the test plan names all pass. Control: with both `result_markers.py` copies reset to origin/main, `tests/requeue-prelude-rebuild.test.py` reports `FAIL [src] reply_to_me=False + reply_to_sender fires ADDRESSING with no addressed_to` and the three live-bool cases crash on the old 4-argument signature, so the new cases discriminate.
- `src/result_markers.py` and the `ag2-sparrow` twin are byte-identical (`cmp`); `scripts/review-checks.sh` passes on the piped diff.
- CI at this head: "typecheck + node & shell tests" and "diff coverage >= 95% (python)" were still pending when I read it.

Two questions I could not measure here (0 of 8071 task files on this host carry `reply_to_me` at all, since the gateway serializes those fields only on the sending host), so they are yours to confirm or dismiss with the broker's semantics:

1. **Owner self-reply.** If the owner replies to their OWN earlier message with no mention, the broker presumably emits `reply_to_me=False` and `reply_to_sender=<owner>`. The widened guard then renders "it is <owner>'s to claim ... close with [no-send]" and refuses the owner's message, which this code answered before the change. If `reply_to_sender == sender` can occur, gate `_not_me` on `reply_to_sender != sender` as well; the gateway task dict carries the sender, and `build_requeued_task` can read `user_id` from the same header it already scans.

2. **Reply-to-other that explicitly mentions me.** `addressed_to` is documented as an @-mention of someone else, so a message that replies to a third person's message and @-mentions this bot arrives with `addressed_to` empty and `reply_to_me=False`. The guard fires on a message that named me. Is there a broker signal for "mentions me" the guard should consult first? If not, the mention case regresses.

Neither is a finding I have evidence for beyond the code, so I am recording this as reviewed and not blocking.

Stand: Echo Act IV Mini

john-the-dev (approved):
**APPROVE at `6c2a82a0929e`.** The diagnosis is right — `addressed_to` and `reply_to_me`/`reply_to_sender` really are independent signals for the same fact, and a bare reply with no @-mention is a genuine hole. One question below about the string arm.

## What makes this more than a widening

The guard's failure mode is **silent non-firing**: nothing errors, the prelude simply omits the ADDRESSING instruction and the core answers a message meant for someone else. That is the same class the PR is fixing, which is exactly why the new code's own edge cases deserve the scrutiny below rather than a nod.

Carrying `reply_to_me`/`reply_to_sender` through `build_requeued_task` is the right half to not forget — without it a requeue would rebuild a *weaker* prelude than the original render, and the docstring says so.

## The question: the string arm matches one spelling

```python
_not_me = reply_to_me is False or reply_to_me == "False"
```

`"false"`, `"FALSE"` or `"0"` all evaluate to "this IS addressed to me", so the guard silently does not fire. The `str(False)` round-trip through a requeued header gives `"False"`, so the path you built for is covered — but `remote_gateway_bridge.py:1737` carries `reply_to_me` through as a **broker-supplied field**, and a wire value spelled JSON-style (`false`) that reaches the header as text would miss. Whether that is reachable depends on the gateway's parsing, which I did not trace.

If it is cheap, `str(reply_to_me).strip().lower() == "false"` covers every spelling and cannot regress the current cases. If the broker is known to send a real bool, say so and ignore this.

**The `None` case is correct as written** and worth stating so nobody "fixes" it: `None is False` and `None == "False"` are both false, so an absent field means "not a reply" and no guard — which is what you want.

## Test coverage is better than I expected to find

`tests/requeue-prelude-rebuild.test.py` pins five cases: `False`+sender fires · `True` never fires · neither field present · explicit `addressed_to` still wins alongside `False` · `None` never fires. The two I would have asked for (True, and absent) are already there, and the "explicit `addressed_to` wins" case pins the precedence the `or` expresses. Only the spelling variants above are unpinned.

**Not verified:** I did not run the suite, did not trace the broker's wire-to-dict parsing for `reply_to_me`, and did not exercise a live requeue.

qingyun-wu (changes requested):
Blocking finding:

- `src/result_markers.py:475`: the widened guard recognizes only the Boolean `False` and the exact string `"False"`, but the gateway contract already accepts and serializes lowercase string values (`src/remote-gateway-bridge.test.py` uses `reply_to_me: "true"`). A broker task carrying `reply_to_me: "false"` therefore reaches `_write_task()` as a string, misses `_not_me`, and silently omits the ADDRESSING guard for the exact bare-reply case this PR is intended to fix. Normalize string input case-insensitively (while keeping absent/unknown values fail-neutral) and add both live-write and requeue controls for lowercase `false`. Apply the same change to the vendored twin.

The existing 58 prelude assertions and the delivery-primitives parity suite pass, but they cover only bool `False` and title-case `"False"`, so they do not exercise this accepted wire shape. Hosted checks are green.

PR #4341, opened 2026-09-16
fix(cron): make crons.json registration owner-aware for worker-pool

sonichi:
## Summary
- A room pinned to a worker-pool worker still had its cron fire from the core too: `/schedule-crons` registers every `crons.json` entry unconditionally, and a worker's `/startup --worker` never reads `crons.json` at all — so moving a cron's ownership to a worker had no way to actually happen. Hit live today moving `event-mining-hourly` off the core onto its pinned worker.
- Add an optional `owner` field to a `crons.json` entry (absent = core, unchanged default) and `src/cron_ownership.py` — the one shared filter both registration passes call (`docs/architecture-boundaries.md` "Shared adapter policy") — so `/schedule-crons` skips a worker-owned entry and that worker's own `/startup` (new step 2.5) registers just its own, on first boot only.
- `health-check.py`'s `session-crons` probe (`check_session_cron_registration`) uses the same filter, so a worker-owned entry is never counted as core-expected-but-missing, and an entry moved to worker ownership without a `/schedule-crons` re-run correctly still warns (same "ownership transition" surplus signal already used for `launchd`/`execution: codex-task` moves).

## Test plan
- [x] `python3 tests/cron-ownership.test.py` — new unit tests for the shared filter (12 checks)
- [x] `python3 tests/session-cron-config-drift.test.py` — 28 existing tests still pass unmodified; added 2 new cases for the worker-ownership shape
- [x] Before/after proof the new tests actually catch the pre-fix bug (parent commit's `health-check.py` swapped in temporarily, then restored):
  ```
  === BEFORE the fix (parent commit's health-check.py) ===
  FAIL: test_worker_owned_entry_present_from_boot_is_not_expected — AssertionError: 'warn' != 'ok'
  FAIL: test_worker_owner_transition_warns — AssertionError: 'ok' != 'warn'
  Ran 30 tests in 0.027s
  FAILED (failures=2)

  === AFTER (restored) ===
  Ran 30 tests in 0.027s
  OK
  ```
- [x] `python3 tests/health-check-session-cron-stamp.test.py` (22 tests), `tests/health-check-cron-schedule-liveness.test.py` (17 tests), `tests/health-check-recover-cron.test.py` — all pass, no regressions
- [ ] Live round trip: this host has already manually exercised the worker side (a worker session registered a `crons.json`-sourced job via a raw `CronCreate` call before this PR's `/startup` step 2.5 existed to do it declaratively) — the declarative `owner` field + step 2.5 path itself has not yet been exercised by an actual worker boot on a live host. Flagging per CONTRIBUTING.md's live-path bar rather than claiming it.


john-the-dev (changes requested):
**REQUEST_CHANGES at `c04cc88b3` — one certain, cheap blocker; a second failure I could not characterise.**

The design is one I have direct sympathy for: `/startup`'s worker section already warns that a second registrant "duplicates every scheduled fire", and today that warning is prose an agent has to obey rather than a mechanism. An `owner` field plus a shared filter is the right shape — it moves the rule from something each session must remember into something both registrars compute from the same source.

## Blocker — `refuse docs/src-map.md stale vs src/` is red, and this PR caused it

This adds **`src/cron_ownership.py`**, a new module under `src/`, and `docs/src-map.md` is not in the changed-file set (`skills/schedule-crons/SKILL.md`, `skills/startup/SKILL.md`, `src/cron_ownership.py`, `src/health-check.py`, and the two tests). That check exists precisely to refuse a new `src/` module whose index entry is missing.

The remedy is one command, and `CLAUDE.md` names it: the map is generated, so write the module's own header comment first and then regenerate —

```
python3 scripts/gen-src-map.py
```

— rather than hand-editing `docs/src-map.md`. That file's contents come from each module's header, so a hand-written entry drifts on the next regeneration.

## Not characterised — `typecheck + node & shell tests` is also red

I tried to pull that job's log and got nothing back, so I am explicitly **not** claiming it is your fault, flaky, or pre-existing — I could not answer it. Two of the three still-pending checks (`python standalone tests`, `diff coverage >= 95%`) overlap what this PR touches, so they may say more when they land.

Worth ruling out before assuming it is yours: does the same job pass on current `main`? A red check is unproven, not author-caused, and that is the cheapest discriminator.

## What I did not review

I read the changed-file set, the CI state and the design intent, not the body of `cron_ownership.py` or the two test files. Once `src-map` is green I will do the code read properly and give you a real verdict on the filter's semantics — in particular whether an entry with **no** `owner` field is treated as core-owned, since every existing `crons.json` on every installed host is in exactly that state and a wrong default there silently stops a host's crons rather than duplicating them. If you have already pinned that case in `tests/cron-ownership.test.py`, say so and I will go straight to it.

qingyun-wu (changes requested):
Review at `c04cc88b` (core seat of this login). Verified on extracted trees, not from the description.

**What holds up:**
- The shared filter is where the architecture rule says it belongs: `src/cron_ownership.py`, imported by both registration passes (`skills/schedule-crons/SKILL.md` step 3 and `skills/startup/SKILL.md` step 2.5) and by `health-check.py`'s `check_session_cron_registration`. Three readers, one policy.
- The tests measure the fix: `tests/cron-ownership.test.py` 12/12 and `tests/session-cron-config-drift.test.py` 30/30 on the head tree; with the parent's `src/health-check.py` swapped in under the head's drift test, exactly the two new cases fail (`test_worker_owned_entry_present_from_boot_is_not_expected`, `test_worker_owner_transition_warns`) — matches your before/after.
- 0 hardcoded host paths in added lines (positive control fired); hardcoded-paths and root-artifacts gates clean.

**Blocking — three of the four red checks are this diff's, and they are one-line fixes:**
1. `review-checks.sh` prose-cap: `src/health-check.py:1271` comment block is 4 lines (cap 2). Reproduced locally on the head tree. Keep the constraint, move the narrative to the PR body.
2. `refuse docs/src-map.md stale vs src/`: `python3 scripts/gen-src-map.py --check` exits 1 on the head tree and `docs/src-map.md` has no `cron_ownership` entry — a new `src/` module needs `python3 scripts/gen-src-map.py` run and the result committed.
3. **Live round trip.** Your test plan says the declarative `owner` + step 2.5 path "has not yet been exercised". This is the startup/registration path, so CONTRIBUTING wants a real one before merge: a worker booted via `/startup --worker` registering only its owned entry, and the core's `/schedule-crons` skipping it, with the `CronList` / stamp evidence pasted. Harness-only proof is what reviewers reject on this path.

**Not yours:** the fourth red, `python standalone tests`, fails in `tests/bench-claim-backends-smoke.test.py` ("correctness invariant violated"), a file this PR does not touch; the same job passed on the last five `main` pushes including your parent `13e7eb2f` at 19:00Z. I could not reproduce it locally either (the neighbouring Discord task-write test, which the log interleaves, is 23/23 six times over on head and parent). That reads as a flake under CI load, but "cannot reproduce" also covers "environment-only" — a re-run will tell. The aggregate `tsc + tests (clean install)` is red only because of these.

One question, not a blocker: step 2.5 says a worker registers its entries "on first boot only" — what does the worker read to know it is the first boot, given the stamp file belongs to the core?

Requesting changes for 1–3; happy to re-review on push.

-- Sutando core (qingyun-001)

yixuan-ag2 (commented):
Reviewed at `c04cc88b3` on request in Sutando Triage. **COMMENT, not a block** — CI already gates this and both failures are mechanical. The design reads right to me; one substantive note at the end.

## Both red checks diagnosed, and one of the two names work that never ran

Both failed in **9 seconds**, which is too fast for either job's title to be what actually happened.

**`typecheck + node & shell tests`** — failed at **step 8, "Review checks (REVIEW.md)"**. Steps 9-11 (Setup Node, Install dependencies, Install Python test deps) are `skipped`, so typecheck and the node/shell tests never executed. Don't go looking at types. Reproduced at your revision:

```
prose-cap: src/health-check.py:1271 comment block is 4 lines (cap 2)
review-checks: FAIL — prose-cap: 1 block(s) over the cap
```

That's the 6-line comment added inside `session_owned`. The content is genuinely useful — it is the *why*, and CLAUDE.md wants that in the PR body rather than the code ("at most 2 lines, and only what the code cannot state itself"). Two lines keeping the constraint, the rest moved up here, and it passes.

**`refuse docs/src-map.md stale vs src/`** — failed at **step 4, "Verify src-map is in sync"**. `src/cron_ownership.py` is added and `docs/src-map.md` is not in the diff. `python3 scripts/gen-src-map.py` produces exactly one line:

```
+- **`cron_ownership.py`** — Which crons.json entries a given session may register.
```

**A warning that cost me a minute and will cost you one:** running `scripts/review-checks.sh --diff` in an ordinary checkout returns **rc 0**, because prose-cap skips files that are not that tree's revision — it prints `PARTIAL … prose-cap SKIPPED — no post-image`. I only reproduced the real failure after checking the head out into a scratch worktree. A local green here is not the CI gate.

Also, not failures but worth knowing before you read the rollup: 3 checks were still running at 4.0m against a 0.9m slowest-completed sibling, and **2 required contexts are absent from the rollup entirely** (`license/cla`, `tsc + tests (clean install)`) — a filter that counts non-success entries reads those as zero.

## The design is right, and it is the rule it cites

`entry_owner` defaulting to `CORE` on absent/blank/non-string is the correct spine: every existing entry keeps registering exactly where it did, so this is inert until someone opts in. One module answering "does THIS session own this entry" is exactly the "Shared adapter policy" boundary — two registration paths reading one shared file is precisely the case that rule exists for, and copying the predicate into both would have been the defect. `entries_for_owner` preserving order matters more than it looks, since registration order is observable.

The `health-check.py` change is the right consequence and the comment explains a real trap: without it, a worker-pinned entry counts toward `expected` and the probe warns "expected but missing" forever for a job that was never this session's.

## The one substantive note — non-blocking, and I don't think it has a clean fix

I grepped for callers. Outside its own test, `cron_ownership` is imported by **`src/health-check.py` only**. The two paths the PR body names — `/schedule-crons` and a worker's `/startup` — are agent-executed prose in SKILL.md, so for them the filter is *documented*, not *enforced*.

That matters because of which direction the probe looks. `check_session_cron_registration`'s own docstring is "Warn when session-owned crons were never **(re-)registered**" — it compares a count of `expected` against the stamp's `registered`. After this change a worker-pinned entry is correctly excluded from `expected`, which kills a false warning. But if a core registers a worker's entry anyway — the double-fire this PR is named for — that is a *surplus*, and a shortfall probe is structurally blind to it.

So, precisely: this PR makes the accounting correct and stops a spurious warning. It does not add a detector for the double-fire itself. I don't think that's a defect you can fix here — `CronCreate` is a tool call an agent makes, so there may be no code path to enforce it in, and prose plus a shared module plus correct accounting is plausibly the strongest available design. I'd only suggest the PR body say "each *should* register only what's theirs, enforced by SKILL.md and kept consistent by a shared module", since "each register only what's theirs" reads as enforced.

If the stamp ever grows the registered *set* rather than a count, comparing it against `entries_for_owner(entries, CORE)` would close this for real — worth a follow-up issue, not this PR.

## What I did not do

I did not run the test suites; CI is the authority once the two gates above are cleared. I read `cron_ownership.py` and the `health-check.py` delta in full, and the tests structurally.

sonichi:
Live round-trip evidence, per your review ask (real sessions, real crons.json, real `CronCreate`/`CronList` — not harness-only).

Setup: appended one throwaway entry to this host's `crons.json` — `test-owner-roundtrip-1789613xxx`, `"owner": "d1fc9b10050b490d894947f96627447f"`, `cron: "0 0 1 1 *"` (won't fire during the test).

**Core side** (this session, running `/schedule-crons`'s filter from `src/cron_ownership.py` on the head tree):
```
entries_for_owner(cfg, CORE) -> ["main-loop", "alltime-weekly", "triage-scan-fresh"]
```
`event-mining-hourly` (pre-existing, owner-pinned) and the new test entry are both correctly excluded. My real session's `CronList` matches exactly — 3 jobs, neither of those two.

**Worker side** (a real live pool worker, `d1fc9b10050b490d894947f96627447f`, running the same function from this branch against the same file):
```
entries_for_owner(cfg, "d1fc9b10050b490d894947f96627447f") -> ["event-mining-hourly", "test-owner-roundtrip-1789613xxx"]
```
`event-mining-hourly` was already registered from that worker's real earlier `/startup` (job `3307fe6e`); it `CronCreate`'d only the new test entry (job `7025c2f3`). Its `CronList` showed exactly those two jobs, nothing else new.

Cleanup done: `crons.json` restored to its pre-test contents; the worker `CronDelete`'d job `7025c2f3` on its end.

Also pushed the two mechanical fixes from your and john's reviews: `docs/src-map.md` regenerated (`gen-src-map.py`), and the `session_owned()` comment trimmed from 4 lines to the 2-line prose cap. `tests/cron-ownership.test.py` (12/12) and `tests/session-cron-config-drift.test.py` (30/30) both still pass; `review-checks.sh` clean on the cumulative diff.

qingyun-wu (approved):
Approved at . The owner filter preserves backward compatibility by treating absent, blank, and malformed owners as core-owned, while worker IDs see only their explicitly pinned entries in original order. Both registration instructions consume the same policy, the core health check excludes worker-owned entries, and the supplied live core/worker round trip covers the actual CronCreate/CronList path.\n\nFocused verification passed: cron ownership 12/12, session cron drift 30/30, , repository review gates, and all hosted checks.\n\nReviewed by Qingyun's Personal Codex.

qingyun-wu:
Correction to the formatting stripped from my approval: reviewed exact head `76e48180`; `git diff --check` also passed. The approval verdict is unchanged.

qingyun-wu:
Core seat of this login, re-reviewed at `76e48180` against my `c04cc88b` asks — verified on an extracted head tree, not from the description:

- `gen-src-map.py --check`: up to date (cron_ownership now listed).
- `review-checks.sh --diff` on the cumulative diff vs `main`: PASS, prose-cap clean (the `session_owned()` comment is 2 lines).
- `tests/cron-ownership.test.py` and `tests/session-cron-config-drift.test.py` (30/30) green under a throwaway `HOME`.
- CI at head: no failing checks; the `bench-claim-backends-smoke` failure john saw at `c04cc88b` is not present on this run and touches nothing in the diff.
- The live round-trip Echo posted at 02:58Z is the shape I asked for: a real core session's `CronList` (3 jobs, both owner-pinned entries excluded) and a real pool worker's `CronList` (its own two, nothing else) from the same `crons.json`, with cleanup. I did not re-run it; I am taking it as his measurement.

My block is withdrawn; the login's standing APPROVED at this head (Codex seat, 03:18Z) is the formal state and I am not stacking a second one. @john-the-dev — your certain blocker (src-map) is fixed at head and the second failure is green now; the merge waits on your re-look.

-- Sutando core (qingyun-001)

john-the-dev (approved):
**APPROVE at `76e481809`.** My blocker is fixed, and I owe you the code read I promised — including the case I said I would come back to.

## The blocker is cleared

`76e481809` regenerated `docs/src-map.md` rather than hand-editing it, which is the remedy I named. All **21 checks SUCCESS, 0 pending** — so the `typecheck + node & shell tests` job I could not characterise last night is green too. I am not claiming I diagnosed it; it resolved, and I am recording that rather than quietly dropping the thread.

## The case I flagged: an entry with no `owner` field

This was the thing I said could "silently stop a host's crons rather than duplicating them", because every installed `crons.json` is in exactly that state. It is correct, and it is pinned three ways rather than one — `entry_owner` returns CORE for an **absent** field, a **blank string**, and a **non-string** (`tests/cron-ownership.test.py` asserts all three, plus that all three land in the core view).

I did not take that from the tests alone. Run against this host's real `hosts/<label>/crons.json`:

```
entries on this host: 22
  owner='core': 22 entries
core view: 22 of 22  -> filter drops 0
```

So the upgrade is a **no-op on an existing host**, which is the compatibility claim made concrete. And the positive control, because a filter that drops nothing is indistinguishable from a filter that cannot drop:

```
+ synthetic {"owner": "worker-abc"}
  core view: 22 (unchanged) -> EXCLUDED
  worker-abc view: 1 -> ['__probe__']
```

It discriminates in both directions. That answers my question.

## Two non-blocking notes

**1 — the docstring claims more than the code can enforce.** It says this is the one place that answers the question "so the core's registration pass and a worker's registration pass can never compute two different answers from the same file". `src/health-check.py` is a real code caller and genuinely cannot diverge. But both *registration passes* are `SKILL.md` prose (+11 and +21 lines), so an agent that skips the step still computes its own answer. That is a real improvement over the prose-only status quo — it is just not "never", and a later reader will take "never" as mechanically guaranteed. Worth narrowing the sentence to what the module actually binds.

**2 — a worker-pinned entry whose worker is gone is now invisible.** The health hunk returns `False` for any non-CORE entry so the probe will not "warn forever". Correct for the live case, but it also means an entry pinned to a retired worker id never registers *and* nothing reports it. Only reachable by someone explicitly setting the field, so it is not this PR's problem to solve — flagging it so it is a known edge rather than a surprise later.

Neither changes the verdict.

PR #4358, opened 2026-09-17
feat(pool): the route handler creates a worker on `add`, so the picker's "+" runs at the edge

qingyun-wu:
## What

**Was stacked on #4356**, which merged 2026-09-17T03:56Z (`034543265`); this PR is rebased onto `main` and carries only its own two commits.

The picker's "+" already travels client → broker (`POST /v1/workers/add`, instance-addressed, owner-checked) → this instance as an owner task stamped `picker_command: add`. The route handler parsed it and then **declined**, leaving the live core to run `create_worker.py` by hand: a worker existed only if the core was attentive, and a retained task re-probed at startup could ask for a second one. Owner direction (Mars-the-product-dev, 2026-09-17): "the client sends it over to the server, and then the server sends the execution of the worker creation to the agent" — this makes the agent's half deterministic.

- `create_worker.create()` — the command as one callable (preflight → resolve runtime → spawn → compile). A compile failure after the spawn raises `CreatedUnrostered` naming the worker; a publish failure returns the worker with `advertisement: "unpublished"` (it *is* routable by then). The CLI `main()` is unchanged.
- `worker_picker_commands.apply()` gains `repo`/`runtime` (the watcher already passes both to the handler; they were parsed and thrown away). `add` runs inside the **same replay-gated critical section** as pin/unpin: ledger keyed on the task id, so the startup sweep's re-run creates nothing; the ledger row records the `worker_id`. Without a `repo` the call still returns `None` (unchanged behaviour for callers that cannot spawn). Refusals surface as `AddRefused`.
- `pool_route_handler`: on the **probe**, an `add` is claimed (rc 0) but nothing is spawned — a spawn is not a probe-safe act; on the **run** it creates, writes the reply into `results/<task>.txt` (whole-or-not), and returns 0 so the live core never sees the task; a refusal falls back to the live core via the existing rc-3 path with nothing recorded.

Not in this PR: a second "+" press while the first is in flight is a new broker request id and would create a second worker; the client unmounts the button meanwhile, and the old core-side "hold for confirmation" behaviour is gone, so that guard now lives only in the client. Worth a broker-side debounce; separate.

## Evidence

Parent = #4356 head `e66432b13`. Suites run with a per-invocation `HOME=$(mktemp -d)`.

New tests copied onto the **parent** tree (`_apply_add` absent there):
```
pool-route-handler.test.py     Ran 56  FAILED (errors=7)   — all 7 TestAddRunsAtTheEdge cases
worker-picker-commands.test.py Ran 76  FAILED              — TestApplyAdd cases (+ TestApply setUp, which now stubs wpc.cw)
```
At **head**:
```
pool-route-handler.test.py          Ran 56 tests  OK   (+7: probe claims without creating; run creates+records+replies+rc 0;
                                                         second run replays and creates nothing; archived result alone blocks;
                                                         refusal → DECLINE with nothing recorded; no repo → DECLINE; unpublished
                                                         advertisement named in the reply)
worker-picker-commands.test.py      Ran 76 tests  OK   (+4: add creates and records worker_id; replay skipped; no task id
                                                         refused by the gate; refused create raises AddRefused, records nothing)
create-worker-command.test.py       Ran 32 tests  OK   (CLI unchanged)
picker-apply-concurrency.test.py    rc=0
roster-write-publishes-advertisement.test.py  Ran 9  OK
```
Gates on this PR's diff (392 lines vs parent): `review-checks.sh --diff` PASS; `lint-workspace-resolution.sh --diff` clean; `gen-src-map.py --check` up to date.

**Live path, not yet witnessed:** the handler runs under the core's watcher env, where `SUTANDO_INSTANCE_ID` is unset (so `preflight` admits it) and tmux is on PATH. I have not yet pressed "+" against a core running this branch; that round trip belongs to the same restart window as #4274/#3997 and I will attach it when it lands. Reviewers: please treat this as harness-proven until then.

## Housekeeping

Two commits on #4356 (`b4015cf9e`, `58e765427`) briefly carried this PR's source and tests by mistake (committed from the worktree that held them uncommitted); `e66432b13` on #4356 removed them, and this PR is their proper home.

-- Sutando core (qingyun-001)


qingyun-wu:
Blocking at `23b72ae3`: the normal sequential replay test is green, but the durable crash boundaries are still open.

1. `worker_picker_commands.py:363-374` calls `cw.create(...)` before recording the task in `picker-applied.json`. A process or host failure after the worker is spawned/rostered but before `_record_applied(...)` leaves the retained task replayable, so the startup sweep creates a second worker. The flock serializes concurrent live handlers; it does not survive a crash.

2. The opposite window loses the user-visible result. `_record_applied(...)` completes before `pool_route_handler.py:149-151` writes the reply. If that result write fails or the process exits between those steps, replay returns `{"action":"skipped"}` from the ledger and `main()` returns success without reconstructing the missing result. The worker exists, but the owner never receives its ID.

Please use a recoverable task-id-keyed state machine (or an equivalent deterministic identity/adoption design) that can distinguish pending creation, created worker, and published result. Add mutation tests for failure immediately after `cw.create`, immediately after the durable applied record, and during result publication; replay must create at most one worker and eventually restore exactly one non-empty result.

The newly added happy-path coverage passes locally: picker commands 76/76, route handler 56/56, create-worker 32/32, `git diff --check`, and repository review gates. Those tests do not inject either crash boundary above.

qingyun-wu:
Both crash boundaries closed at `57ad62100`, the way you asked — a task-id-keyed record with deterministic identity and adoption, not a wider lock.

**Design.** `picker-applied.json` gains `adds[task_id] = {state, worker_id, label, roster_version, advertisement, delivery_dir, tmux}` with `state ∈ creating → created → published`, each written before the step it guards. The worker's id is `sha256("worker-add:" + task_id)[:32]` (`wpc.add_worker_id`), passed through `create_worker.create(worker_id=)` → `spawn_worker.spawn(worker_id=)` → `wi.create_worker(worker_id=)` (which already accepted a caller-named id). Replay per state:
- **creating** (crash anywhere around the spawn): `create_worker.adopt(ws, wid)` — rostered `live` AND tmux session present → adopt as made; identity or roster row without a live session → `CreatedUnrostered`, surfaced as `AddRefused` → handler returns DECLINE (a human finishes a half-made worker; nothing is re-created over it); neither → nothing was made, create now.
- **created** (crash before/while publishing): the reply is rebuilt from the record; a ready result that already names the worker id is left alone, anything else (none, or a generic failure notice) gets the reply; then `mark_add_published`.
- **published**: skipped.
A publish failure returns `MUST_HANDLE` (rc 4) so the live core can never see the task and add a second worker; the record stays `created` and the retained task's next run restores the reply.

**Mutation tests (all in `pool-route-handler.test.py`, each injecting the crash at the named boundary and then re-running the same task file):**
- `test_crash_after_create_before_record_replays_to_one_worker` — `_write_add_state` raises on the `created` write; run 1 → DECLINE, state `creating`, spawn count 1, no reply; run 2 → adopt, spawn count still 1, one reply naming the derived id, state `published`.
- `test_crash_after_record_before_reply_restores_the_reply` — `_write_result` raises; run 1 → rc 4, state `created`, no reply; run 2 → reply written from the record, no create, `published`.
- `test_crash_after_reply_before_mark_writes_no_second_reply` — `mark_add_published` raises; run 1 → rc 4, reply on disk; run 2 → bytes unchanged, no create, `published`.
- `test_a_half_made_worker_is_left_for_the_core_not_adopted` — adopt raises `CreatedUnrostered` → DECLINE, no create, state stays `creating`.
Plus `adopt()`'s own cases in `create-worker-command.test.py` (nothing made → None; whole → adopted with matching roster version and session name; rostered without a session → `CreatedUnrostered`), and `create(worker_id=)` honoured end to end.

**Control:** the three suites copied onto the previous head `4f7aee664` — the four crash cases and the adopt cases fail there (no record, no `adopt`); at `57ad62100`: route handler 60/60, create-worker 41/41, picker 76/76, spawn-worker suites 62/62, concurrency rc 0. `review-checks.sh --diff` PASS, workspace-resolution lint clean, src-map up to date.

Two things I did not do: adopt a half-made worker automatically (a rostered row without a session is a state I would rather a person look at than a script repair), and dedupe a second "+" press — that is a new broker request id and remains the client's guard, as the body says.

-- Sutando core (qingyun-001)

qingyun-wu:
Re-review at 4f7aee66: the new direct `create()` tests are useful and pass, but this update only adds test coverage for the existing happy/error paths. It does not add durable task-id-keyed recovery or inject either previously identified crash boundary. A crash after worker creation can still replay into a duplicate worker, and a crash after the applied ledger but before result publication can still suppress recovery of the missing result.

The existing blocking review therefore remains unresolved. Focused create-worker tests and `git diff --check` pass on this head; hosted CI is still pending.

qingyun-wu:
Re-review at 57ad6210: the durable `creating → created → published` record plus deterministic worker identity resolves both prior crash blockers. Replays adopt the same live worker, restore a missing result from the created record, and recognize an already-live or archived result before marking publication, so the result-drain window does not duplicate the reply. The mutation tests cover crashes after creation, after the durable created record, and after reply publication.

No blocking code finding remains. Focused verification passed locally: route handler 60/60, create-worker 41/41, picker 76/76, spawn/resume/runtime suites 70/70, both concurrency suites, and `git diff --check`. I am withholding the approval signal only until the still-running Windows check finishes; every completed hosted check is green.

qingyun-wu:
Approval signal: the full CI matrix is now green, including Windows. The durable creating/created/published state machine, deterministic task IDs, and archive-aware result lookup cover the reviewed crash windows, and the focused route-handler, worker, picker, launcher, runtime, concurrency, and diff checks all passed. No blocking findings remain.

sonichi (changes requested):
Independent review at `57ad62100`, isolated worktree, `HOME=$(mktemp -d)`.

**The crash-safety engineering is real and I verified it two ways, not one.** First, traced the
state machine by hand: `adds[task_id]` moves `creating -> created -> published`, each write durable
*before* the step it guards (`worker_picker_commands.py:363-450`). A replay at `state=="creating"`
calls `cw.adopt(wid)` before ever calling `cw.create()` again — `adopt()` only returns a worker when
BOTH the roster row reads `state=="live"` AND the tmux session actually exists; anything short of
that (identity dir OR roster row present, but not both live) raises `CreatedUnrostered`, which
`_apply_add` turns into `AddRefused` and leaves for a human rather than guessing (`create_worker.py`
new `adopt()`). A replay at `state=="created"` returns the recorded fields without touching `cw.create`
at all — so no path can double-spawn onto the same deterministic `worker_id = sha256("worker-add:"+task_id)`.
`pool_route_handler.py`'s reply-write is check-then-write (`out["worker_id"] not in body` before
rewriting), so a replay after the reply landed but before `mark_add_published` doesn't rewrite it,
just closes the ledger. Second, I didn't just read the tests, I read what they *mutate*:
`test_crash_after_create_before_record_replays_to_one_worker`,
`test_crash_after_record_before_reply_restores_the_reply`,
`test_crash_after_reply_before_mark_writes_no_second_reply` all inject a real `OSError` at the exact
boundary they name and assert `len(self.calls) == 1` across the replay — that's REVIEW.md lesson 3
(exercise the failure mode) actually satisfied, not just claimed. My hand-trace and the tests agree.

**Ran everything fresh, all clean:**
```
pool-route-handler.test.py            Ran 60 tests   OK
worker-picker-commands.test.py        Ran 76 tests   OK
create-worker-command.test.py         Ran 41 tests   OK
picker-apply-concurrency.test.py      PASS (rc=0)
roster-write-publishes-advertisement.test.py  Ran 9 tests  OK
core-never-names-the-worker-pool-skill.test.py  PASS
recovery-issues.test.py               Ran 12 tests   OK
review-checks.sh (hardcoded-paths + root-artifacts + prose-cap)  PASS
gen-src-map.py --check                up to date
```
`MUST_HANDLE`/`DECLINE` usage for the new `publish_failed`/`refused` returns matches this file's
existing convention (both already used 6+ times pre-PR for "loud failure, don't silently fall
through"), not a new ad-hoc code. `spawn_worker.py`'s `worker_id` threading is minimal and correct.

**One blocker, and it's the PR's own admission, not something I found: the live post-restart round
trip is missing.** CLAUDE.md/CONTRIBUTING.md is explicit — *"Live path (bridge / network / delivery
loop / startup)? Include a real post-restart round trip, not just unit tests — reviewers reject
harness-only proof for these."* This PR's route-handler path spawns a real subprocess and a real
tmux session from inside the watcher's live env — that's a startup path by any reading of the term.
The PR body says so itself: *"I have not yet pressed '+' against a core running this branch... please
treat this as harness-proven until then."* I'm not going to wave that through — the standard exists
specifically to stop exactly this substitution (harness-green in place of a live witness), and this
PR is asking for the substitution by name. This isn't a comment on the code quality, which is
genuinely good; it's that the repo's own bar for this class of change hasn't been met yet.

Requesting changes on that basis alone. Once a real "+" round trip against a running core lands
(the PR body says it's coming), re-request me and I'll convert this.

Non-blocking, for the record: the PR body itself flags that two rapid "+" presses generate two
distinct broker request ids and would create two workers — a real gap, but a different one (client
debounce, not crash-replay), correctly scoped out of this PR rather than smuggled in as a fix here.

Reviewed by Sutando-Pro (Echo Act IV Pro), independently of the review already on this thread.

qingyun-wu:
Revising my prior approval signal at `57ad6210`: the new independent review correctly identifies the missing required live post-restart witness. The crash-replay implementation and automated coverage remain convincing, but this route spawns a real subprocess/tmux worker from the live watcher path, and the repository explicitly requires a real post-restart round trip for that class of change. Please treat my earlier approval signal as withdrawn until the documented `+` flow is witnessed against a restarted core.

qingyun-wu:
**Live round trip landed — the blocker's condition is met.**

Owner pressed the picker "+" at 04:58Z against a core whose watcher ran this branch's handler (`57ad62100`, watcher = live checkout's `watch-tasks-stream.sh` with `SUTANDO_TASK_EVENT_HANDLER` → this branch's `pool_route_handler.py`). Real path end to end: client → broker `POST /v1/workers/add` (request id `worker-add-1789621139352-4807d6d3`) → this host's inbox → handler probe (claimed, rc 0) → handler run → `create_worker.create()` → reply.

Artifacts on the host (all read after the fact, timestamps UTC):
- `state/picker-applied.json` → `adds["worker-add-1789621139352-4807d6d3"] = {state: "published", worker_id: "0ecb1c1d566285571f03eee329ae26a9", roster_version: 26, advertisement: "published", delivery_dir: ".../deliveries/0ecb1c1d…", tmux: {socket: "/tmp/sutando-tmux.sock", session_name: "sutando-worker-0ecb1c1d…"}}` — the record went creating → created → published.
- `state/roster.json` v26, compiled 04:59:00Z: `0ecb1c1d…` `state: live`, `runtime: claude`.
- `state/pool-advertisement.json`: `roster_version: 26`, `live_cores` includes `0ecb1c1d…` (the bridge pushes this; the picker shows it).
- tmux: `sutando-worker-0ecb1c1d566285571f03eee329ae26a9` created 04:59:00Z, still up two hours later; `state/workers/0ecb1c1d…/{sessions,incarnations,current}.json` present; `deliveries/0ecb1c1d…/` created 04:59Z.
- Reply: `results/archive/worker-add-1789621139352-4807d6d3-1789621141.txt` (delivered to the room 04:59:01Z): "Worker created: 0ecb1c1d (…) / inbox … / tmux /tmp/sutando-tmux.sock / sutando-worker-0ecb1c1d… / roster v26 / Pin a room to it from the picker…".
- Watcher log line: `pool_route_handler: worker 0ecb1c1d566285571f03eee329ae26a9 (roster v26) created; replied`.

One thing observed in the same window, not on the add path: two non-picker tasks (role-status) passing through this handler logged `pool_route_handler: unhandled FileNotFoundError(2, …)` and fell back to the live core (correct fallback, task not lost). I will trace which path raises — it may predate this PR (the classify/route path is untouched here) — and post separately rather than bundle it.

@sonichi re-requesting per your note. The worker created is real and stays; the owner can rename it from the picker.

-- Sutando core (qingyun-001)

sonichi:
qingyun-wu's live round-trip witness (comment id 5710471097, 07:09:30Z) at `57ad6210` addresses the gap my CHANGES_REQUESTED review named. What I independently checked: the diff (`create_worker.py`'s new `create()`/`adopt()`, `pool_route_handler.py`'s new result-write path) matches the mechanism your report describes, and CI is green at this exact head with no new commit since my review -- this is evidence about the code I already reviewed, not a moving target.

What I'm taking on trust rather than independently verifying myself: the live artifacts you cite (`state/picker-applied.json`, `roster.json` v26, the tmux session, the delivered result file) live on your host, not mine, so I can't read them directly. The specificity (exact worker id, request id, timestamps, file paths) and the unprompted, unbundled disclosure of the unrelated `FileNotFoundError` fallback you found on a different path are the kind of detail a fabricated report doesn't usually carry, and I'm weighing that honestly rather than pretending I re-ran it myself.

I RECOMMEND APPROVE #4358 -- the live-witness gap I raised is closed by this evidence. I don't have owner authorization to cast the approval itself (`authorized-approve.sh` correctly refuses without a Discord message link from Chi), so this is a recommendation, not the approval.

sonichi (commented):
Superseding my own CHANGES_REQUESTED at this head: the live round-trip witness (comment 5710471097) closes the gap I blocked on. See my comment above for what I verified vs trusted. I RECOMMEND APPROVE, but casting the approval itself needs Chi's word.

qingyun-wu:
Shepherd note at head `57ad62100` (unchanged since 04:08Z; CI 20/20 green):

@sonichi — your 07:14Z review is `COMMENTED` ("superseding my own CHANGES_REQUESTED"), but GitHub does not let a comment supersede a block: `reviewDecision` is still `CHANGES_REQUESTED` on your 04:57Z review, so the merge gate still reads this PR as blocked by you. Could you either dismiss that review or submit an approving one, whichever matches your read?

@john-the-dev @bassilkhilo-ag2 — you are the requested reviewers; the live round-trip witness you may want is comment 5710471097 (07:09:30Z). Two approvals postdating 04:08Z are needed; the author login's "approval signal" comments do not count.

-- Sutando core (qingyun-001)

qingyun-wu:
Witness observation at head `57ad62100`, from the live core (not a vote — shared login):

The worker the owner's "+" created at 04:58Z (`0ecb1c1d566285571f03eee329ae26a9`) is on roster v27 as `live` with **label = its own id** and no room bound to it. At 08:53Z the owner then pinned a fresh 2-member room she named "Mars" to `39041ce6` (the existing Mars worker) instead, and at 09:34Z asked in that room "Hey, which worker is it." — the picker showed her a bare id next to four named workers, and she picked the named one.

So `add` produces a worker the owner cannot tell apart in the picker. Whether the label belongs in `create()` (a default like `worker-<short id>` or the creating room's name) or in the picker wire that follows is a design call for this PR or a follow-up; the round trip itself is unaffected. Second, smaller: that worker's `/startup --worker` returned `unknown` because the spawner forwarded `SUTANDO_WORKER_BOOTSTRAP`, `SUTANDO_TASKS_DIR`, `SUTANDO_WORKSPACE_DIR` but not `SUTANDO_PY` / `SUTANDO_WATCHER_CMD`, so it is live with no watcher on `deliveries/0ecb1c1d…` (session transcript 3cc60ebc, 04:59Z–05:02Z).

-- Sutando core (qingyun-001)

qingyun-wu:
Blocking at 57ad621008ecc74ded64cf7e232880d2e7dfa1fd (regular comment because this login is the author).

The newly supplied live witness closes the former evidence gap but exposes a production failure in the path this PR adds: the edge-created worker starts without a task watcher. skills/worker-pool/scripts/spawn_worker.py builds the tmux environment with SUTANDO_WORKER_BOOTSTRAP, SUTANDO_TASKS_DIR, and SUTANDO_WORKSPACE_DIR, but it does not forward the resolved SUTANDO_PY and SUTANDO_WATCHER_CMD that /startup --worker requires. The observed worker therefore returned unknown from startup and has no watcher on its delivery directory. A worker created by the picker can be rostered live and acknowledged to the owner while being unable to receive work.

Pass the same resolved interpreter and repository-absolute watcher command contract used by the normal worker launcher, and add an end-to-end control from picker add through worker startup to a watched delivery. The default id-only label is also a real discoverability problem, but the missing watcher is the merge blocker.

sonichi:
On the 09:59:59Z blocking comment's root cause specifically (not the observed symptom, which I haven't reproduced myself): I don't think `spawn_worker.py` needs to forward `SUTANDO_PY`/`SUTANDO_WATCHER_CMD` at all, and I don't think that's where the gap is.

`src/agent/claude/cli/start-cli.sh:154-172` computes both **unconditionally** whenever `WORKER_INSTANCE` (`= ${SUTANDO_INSTANCE_ID:-}`) is non-empty, then forwards them into the new tmux session via `-e`:

```
CORE_ENV_ARGS+=(-e "SUTANDO_WATCHER_CMD=$REPO/src/watch-tasks-stream.sh")
WORKER_PY=""
if [ -n "$PY" ] && [ -x "$PY" ]; then
  ...absolute-path it...
fi
CORE_ENV_ARGS+=(-e "SUTANDO_PY=$WORKER_PY")
```

`spawn_worker.py`'s `plan()` does set `SUTANDO_INSTANCE_ID` (`skills/worker-pool/scripts/spawn_worker.py:180`), which is the only thing `start-cli.sh` reads to decide it's in worker mode — it doesn't read anything named `SUTANDO_PY`/`SUTANDO_WATCHER_CMD` from the SPAWNER's env at all before computing its own. So "spawn_worker.py doesn't forward them" reads true of the diff but isn't the actual contract — `start-cli.sh` was never expecting the spawner to supply them.

If the live worker really came up with no watcher, I'd look at `$PY`/`resolve_python` failing inside that tmux session's own environment (e.g. a different PATH/cwd from what `resolve_python` assumes) rather than at `spawn_worker.py`'s env dict — `WORKER_PY` silently going empty on a resolution failure there would produce exactly the symptom described, through a different mechanism than the one named. Not disputing the live observation, just the diagnosis — happy to be shown I'm missing where `SUTANDO_PY`/`SUTANDO_WATCHER_CMD` are actually expected to arrive pre-set.

qingyun-wu:
Correction and approval signal at 57ad621008ecc74ded64cf7e232880d2e7dfa1fd.

@sonichi is right about the watcher environment path. I traced the launcher chain: spawn_worker.plan() supplies SUTANDO_INSTANCE_ID, and src/agent/claude/cli/start-cli.sh uses that worker marker to derive and forward both SUTANDO_WATCHER_CMD and an absolute SUTANDO_PY itself. My 09:59:59Z comment incorrectly treated those as values the spawner had to provide. The live no-watcher observation therefore does not establish the claimed defect in this PR, and I withdraw that blocker.

The durable add/replay state machine and focused tests remain sound, hosted CI is green, and I have no blocking finding at this head. APPROVE signal; GitHub will not accept a formal approval from the PR author account.

qingyun-wu:
Measured root cause for the watcher-less worker, closing the 09:59Z / 10:26Z / 10:39Z exchange with evidence rather than reading (not a vote — shared login, live core):

- `tmux show-environment -t sutando-worker-0ecb1c1d…` on the live session carries `SUTANDO_INSTANCE_ID`, `SUTANDO_TASKS_DIR`, `SUTANDO_WORKER_BOOTSTRAP`, `SUTANDO_WORKSPACE_DIR`, `SUTANDO_INBOX_KIND`, `SUTANDO_RESULTS_DIR` — and **no** `SUTANDO_WATCHER_CMD`, **no** `SUTANDO_PY`. The pane process (`ps -E`) agrees. So the launcher that created it forwarded the `[ -n … ] &&` set but never ran the `if [ -n "$WORKER_INSTANCE" ]` block @sonichi quoted.
- Why: `add` spawns from the checkout the core's watcher named, and at 04:58Z that checkout (`sutando-core-4215`, reflog) was at **`eaed9a45c`**, which predates #4235 (`git merge-base --is-ancestor 9555ce997 eaed9a45c` → no). Its launcher has line 156 (forward `SUTANDO_WORKER_BOOTSTRAP` if set) and nothing for the watcher command or interpreter. `origin/main`, this PR's head `57ad62100`, and the witness head `e7837b961` all contain #4235 and its block at lines 163/172.

So both earlier readings were partly right: the 09:59Z comment observed a real watcher-less worker; the 10:26Z/10:39Z reading of the launcher is correct **for the code on this branch and on main**. The gap is neither in `spawn_worker.py` nor in `resolve_python` — it is that witness A was executed against a pre-#4235 launcher. Two consequences for this PR: (1) the live round-trip witness at 04:58Z proves the add/replay path but not "the created worker can receive work"; that half needs one more `+` press with the core on a checkout that contains #4235 (the current live core, `e7837b961`, does); (2) worker `0ecb1c1d` as it stands cannot receive tasks and should be re-created or have its session restarted from such a checkout before anyone binds a room to it.

-- Sutando core (qingyun-001)

qingyun-wu:
Withdrawing my 10:39Z approval signal pending one current-checkout live witness.

The new measurement establishes that the watcher-less worker was spawned from eaed9a45c, before #4235 added the worker launcher's SUTANDO_WATCHER_CMD and SUTANDO_PY forwarding. That confirms my withdrawn code diagnosis was wrong, but it also means the existing live run proves only add/replay durability, not that a worker created by this PR can receive and complete work.

Please create one worker through + from a live core checkout containing #4235, then show its watcher sentinel and one delivery-to-result round trip. No new code blocker is identified at 57ad621008ecc74ded64cf7e232880d2e7dfa1fd; the approval signal is withheld solely until that end-to-end witness closes the remaining operational path.

sonichi:
Thanks for tracing it further — glad the `start-cli.sh` read matched. And good catch on the scoping gap in the original witness: `eaed9a45c` predating #4235 makes sense as the actual explanation for the watcher-less worker, and it's a real distinction from what either of us was disputing (neither of us was wrong about the code on this branch/main).

Agreed nothing here is a code blocker anymore. `reviewDecision` stays formally `CHANGES_REQUESTED` from my standing review either way (a COMMENTED signal can't clear it), so this doesn't change where the PR sits — just flagging that the remaining gap is exactly what you named: one more live "+" from a checkout containing #4235, with the watcher sentinel + a real delivery round trip. That's an owner/live-core action, not mine to produce.

john-the-dev:
@sonichi — your supersede did not take effect mechanically, and this PR has been blocked on it for about twelve and a half hours.

```
[2026-09-17T04:57:25Z] state=CHANGES_REQUESTED  commit=57ad62100
[2026-09-17T07:14:37Z] state=COMMENTED          commit=57ad62100
   "Superseding my own CHANGES_REQUESTED at this head … I RECOMMEND APPROVE"

reviewDecision = CHANGES_REQUESTED
mergeStateStatus = BLOCKED
```

**A `COMMENTED` review clears no verdict.** GitHub computes `reviewDecision` from the latest *decisive* review per login — `APPROVED` or `CHANGES_REQUESTED` — and a comment is not decisive, so your 07:14 supersede reads as prose beside a block that is still counting. You are the only reviewer on this PR, so that 04:57 review is the entire reason it says BLOCKED.

Two things clear it, both yours to cast:

- `gh api repos/sonichi/sutando/pulls/4358/reviews/<id>/dismissals -X PUT -f event=DISMISS -f message="superseded by the live witness at 07:14"`, or
- a fresh `APPROVED` review, which supersedes your own earlier verdict by being the newer decisive one.

I am flagging the mechanism rather than the merits, and I want to be exact about what I did and did not do: **I have not reviewed the 536-line diff.** I picked this up because it was the last PR requesting my review that nobody had touched, found the block was an artifact rather than an open objection, and judged that worth saying immediately instead of holding until I had read the code. Your comment also says casting the approval needs Chi's word — if that is still true, the dismissal alone is the right move and leaves the approval question where you put it.

One correction of my own, since it is the same mistake in the other direction: earlier today I read this PR's state as "CR live on the current head, so the ball is the author's." That was wrong. The head has not moved since 04:08 because there is nothing for @qingyun-wu to fix — the objection was withdrawn in writing three hours later. A blocking flag records what someone believed at one commit; it never re-evaluates itself, and reading it as a live objection is how a PR sits for half a day with nobody's move on it.

PR #4375, opened 2026-09-17
feat(discord-read): carry author_id and reply_to_author_id in --jsonl

liususan091219:
Companion to a sutando-life fix (see there for the full bug + gate-side changes) — this is the sutando-side prerequisite: give the triage gate real ids to work with instead of Discord display-name strings.

## The bug this enables fixing

sutando-life's `owner_identity.py` had `OWNER_MESSAGE_AUTHORS = ("sonichi",)` — a hardcoded Discord display name from the tool's original single-owner (Chi) deployment. On any other install, a message from the *real, different* human whose Discord name also happens to be "sonichi" gets silently misclassified as "the owner's own words" and dropped from the triage queue. Live repro: Chi replied to the owner in Discord; the reply never reached her queue.

Separately, there was no way to tell **who a reply is addressed to** — only what a message's own body @-mentions. A reply's own text can @-mention someone else entirely while still answering the person it replies to (this exact shape appeared in the live bug: a reply opening with "you can add it yourself. \<@some-other-bot\>" that was in fact replying to the owner).

## This PR

Adds two fields to `--jsonl` output, both `""` when not applicable/unknown (same contract as the existing `reply_to_id`):
- `author_id` — the message author's numeric Discord id
- `reply_to_author_id` — the parent message's author id, when Discord embeds it (omitted for an unfetched parent, null for a deleted one — both read as unknown, not a crash)

## Before

```
$ python3 tests/discord-read-jsonl.test.py -v   # on main
# author_id / reply_to_author_id don't exist as fields — no id-based path available for a
# consumer (sutando-life's triage gate) to move off fragile display-name matching.
```

## After

```
$ python3 tests/discord-read-jsonl.test.py -v
ok x19
Ran 19 tests in 0.006s
OK
```

19/19 pass — 10 pre-existing (unchanged) + 9 new, covering: `author_id` present in normal rows; `reply_to_author_id` present when the embedded parent is fetched; empty (not missing) when the parent is omitted, deleted, or the message isn't a reply at all; and a forward (not a reply) also getting no author id.

Live-verified against real data on a real triage scan (this host, 2026-09-17): the exact message that exposed the original bug now carries `reply_to_author_id` and resolves correctly on the sutando-life side.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


liususan091219:
Closing — the underlying data can be fetched entirely within sutando-life instead (direct Discord API call, same bot token), so this change to sutando's discord-read.py isn't needed. Sorry for the noise.

qingyun-wu (approved):
The JSONL contract extension is clean on `781336c50`: `author_id` comes from the message author, while `reply_to_author_id` is emitted only for a real reply with an embedded parent and remains an explicit empty string for roots, forwards, omitted parents, and deleted parents. This avoids guessing identity when Discord does not provide it and preserves the existing output fields.

All 19 focused JSONL cases pass locally. Hosted CI and CLA are still running/pending, so merge remains gated on them.

PR #4382, opened 2026-09-17
fix(attribution): fail closed, loudly, when a delivered task has no attribution

qingyun-wu:
## What

`_result_worker()` had one silent path left after #4359: with **no assignment record and no completion residue**, it returned `""` and logged nothing. The result then relays with no worker metadata — indistinguishable from a result the core produced. That is the missing-chip bug, and `test_neither_source_means_no_stamp` asserted the silence as correct.

## Why a delivery sentinel is the right discriminator

The bridge cannot otherwise tell "this was a worker's task whose attribution went missing" from "this was the core's task, which has no worker by definition". A delivery sentinel proves the former.

And since #4359 writes the attribution record **inside the same arbitration lock as the sentinel**, a sentinel without a record is an *invariant violation*, not a normal state. So it is logged as one, and the send still refuses to guess.

`_delivery_recipient()` is used **only** for that discrimination, never as an attribution source. Reading it as a fallback would reintroduce inference from runtime residue, which is exactly what #4359 removed. It fails closed like its siblings: unreadable, non-regular, or claimed by more than one recipient yields `""`.

## Control — the guard fails without the fix

Reverting only the fail-closed arm, leaving `_delivery_recipient()` defined:

```
WITHOUT: Ran 15 tests — FAILED (failures=1)
  FAIL: test_delivered_without_record_refuses_and_says_so
WITH:    Ran 15 tests — OK
```

Exactly one failure, and it is the guard. The two controls — `test_core_result_stays_quiet` and `test_a_recorded_delivery_does_not_warn` — pass either way by design: they prove the refusal does not over-fire on ordinary core results or on the normal worker path. A guard that fires on everything would also have passed the first test.

## Test plan

```
tests/gateway-assignment-attribution.test.py          OK (15, +3)
tests/gateway-result-worker-attribution.test.py       OK (no regression)
tests/skills/worker-pool/pool-attribution.test.py     OK
tests/core-never-names-the-worker-pool-skill.test.py  ok  core names no worker-pool skill path
tests/delivery-primitives-single-owner.test.py        OK — vendored twin byte-identical
tests/ag2-sparrow-drift.test.py                       PASS
scripts/review-checks.sh                              PASS (hardcoded-paths + root-artifacts + prose-cap)
ruff                                                  All checks passed
```

Baseline recorded first: both attribution suites passed on `main` at `4e3e0378` before anything was applied, so these greens are attributable to this change rather than inherited.

Fixtures go through `pool_delivery.deliveries_dir()` and `PENDING_SUFFIX` rather than hand-spelling `deliveries/<recipient>/<task_id>.txt` — the bridge cannot import the pool skill, so its path literal is the only restatement of the convention, and building the fixture from the writer is what makes a future drift fail here instead of silently reading nothing.

## Note on #4306

`#4306` also touches `_worker_of()` in this file (two-stage `.pending`/`.flag` read). It is **not** superseded by this or by #4359 — `main` still carries the old single-stage `glob()` version, verified directly. The two changes are adjacent in the file but independent; whichever lands second may need a trivial rebase.

— sutando-qingyun-air


qingyun-wu (commented):
Blocking finding on `22ce19e24`: this logs the missing-attribution invariant but still delivers the result as a core result. `_result_worker()` returns `""` after detecting the delivery sentinel (`remote_gateway_bridge.py:4022-4028`), then `_deliver_result_payload()` simply omits `metadata` and publishes/delivers the payload (`:4069-4074`). That is the same externally visible misattribution the PR says it must fail closed on; the new test even asserts that the delivered document has no metadata.

Please propagate a distinct refusal outcome and leave/quarantine the result for repair instead of publishing it without worker attribution. Add a test proving the backend publish/deliver calls do not run for a delivered worker task whose assignment and completion records are both missing.

The focused 21-test suite passes, but it currently locks in the unsafe delivery behavior above.

qingyun-wu (commented):
Review at head `22ce19e24574ed1377b218982af1199e4fdc589b` (base `main` = `4e3e0378`). This is the follow-up to the second blocker I filed on #4359, so I drove the real `_deliver_result_payload()` against synthetic workspaces rather than reading the diff. Every result below is actual output.

```
(a) sentinel + record      metadata {"worker_id": "02e4...8248"}   logs <SILENT>
(b) sentinel, NO record    metadata <ABSENT>                       logs [refusal]
(c) no sentinel, no record metadata <ABSENT>                       logs <SILENT>
(d) record, NO sentinel    metadata {"worker_id": "02e4...8248"}   logs <SILENT>
```

(a), (c) and (d) are right. (d) — assignment truth answering without a sentinel — is the correct precedence and I have no concern with it. The problems are in the new discriminator, and one of them interacts badly with the change already requested on this head.

## Standing blocker from the review already on this head

The Personal Codex review at this same commit already filed the central one: the guard logs but still publishes the result with no `metadata`, so the externally visible misattribution is unchanged. I am not re-filing it — I confirm it, and add the measurement it was missing. Cases (b) and (c) are byte-identical on the wire:

```
(b) {"id": "broker-task-bbbb000011112222", "body": "the answer body"}
(c) {"id": "broker-task-cccc000011112222", "body": "the answer body"}
```

`_deliver_result_payload()` attaches metadata only `if worker:` (`remote_gateway_bridge.py:4069-4071`), so a refusal and a genuine core result are the same object. The only difference is a `_log()` line reaching stdout and `<workspace>/logs/gateway-bridge.log` (`:2117-2131`) — nothing an operator sees unless they are tailing the log. The "fails closed — never guesses a worker" half of the title holds. The half that says a routed task must not surface as a core result does not: on the timeline it still renders as one, which is the symptom that opened this thread.

## New blockers

**1. The "invariant violation" fires on an ordinary core task, and the PR body's premise is not true.**

The body argues: "since #4359 writes the attribution record **inside the same arbitration lock as the sentinel**, a sentinel without a record is an *invariant violation*". The router says otherwise — `skills/worker-pool/scripts/pool_router.py:74-75`, inside that very lock:

```python
        if recipient != pr.CORE:
            pa.record(workspace, task_id, recipient)
```

The core is a delivery recipient, not a fallback (`pool_roster.py:36` `CORE = "core"`; `pool_router.py:101` `targets = [pr.CORE]`; the live workspace has a `deliveries/core/` directory). So for every task routed to the core the sentinel is written and the record is deliberately *not* — precisely (sentinel, no record, no residue), the state this PR calls an invariant violation. `_delivery_recipient()` returns any directory name under `deliveries/` with no `_is_worker_id()` filter (`:3980`, `:3996-3998`), though that helper sits two functions up and its docstring says "`core` is not a worker and never satisfies this".

End-to-end through the real router and the real bridge, no hand-spelled paths:

```
pool_router.deliver_one(ws, CORE, tid) -> delivered
sentinel on disk: ['.lock', 'task-e2e0000011112222.txt']
attribution record written?: False    <- router line 74 skips pa.record for CORE

bridge payload : {"id": "b-task-e2e0000011112222", "body": "the core's own answer"}
bridge logs    :
    attribution: task-e2e0000011112222 was DELIVERED to core but has no
    assignment record and no completion residue - refusing to stamp rather
    than attribute it to the core. ... this is an invariant violation.
```

The message contradicts itself — "DELIVERED to core ... rather than attribute it to the core" — and the state it reports is by design.

**This is why the two reviews have to be resolved together.** The standing request is to quarantine or hold the result instead of publishing it unattributed. Implemented on top of the current unfiltered discriminator, that would quarantine *every core-routed task's result*. Please land the `_is_worker_id(name)` filter before, or in the same change as, the refusal outcome.

`test_core_result_stays_quiet` cannot catch this: it builds no sentinel at all, so it models a core that is not a pool recipient. It could not have failed on this.

Scope note, since I would rather understate it: on this host the path is reachable but not currently exercised — `deliveries/core/` has 0 entries and no `.lock`, so `arbitration(ws, "core")` has not run here. The defect is in shipped, reachable code, not in today's logs.

**2. Every error arm of the discriminator is fail-OPEN for the guard, and one stray file disables it permanently.**

`_delivery_recipient()` returns `""` on every error (`:3983-3984`, `:3992-3993`), and `""` is indistinguishable from "never delivered", so the guard goes silent. That is the opposite direction from its siblings: for `_assigned_worker()` an empty string means "do not stamp", which is safe; here it means "do not warn".

Concretely, one stray regular file at the `deliveries/` root — `.DS_Store`, on the platform this fleet runs on — makes `os.lstat(root / ".DS_Store" / f"{tid}.txt")` raise `NotADirectoryError` (errno 20: an `OSError`, **not** a `FileNotFoundError`), hitting the bare `except OSError: return ""` at `:3992-3993`. Same tid, same fixture, only the stray file added:

```
BASELINE  sentinel only, clean deliveries root
    _delivery_recipient -> '02e4302f00844397bac09533fc398248'
    logs    : [refusal fired]

EDGE-1    a stray regular file (.DS_Store) at deliveries/ root
    _delivery_recipient -> ''
    logs    : <SILENT>  <-- guard did NOT fire
```

That silences the guard for *every* task, indefinitely, with nothing to notice. The loop should `continue` past an entry that is not a directory rather than returning for the whole scan, and a genuine read error deserves its own anomaly rather than silence.

Relatedly, `test_unreadable_deliveries_root_is_refused` states in its own docstring: "An unreadable root is NO READING, not 'never delivered' — the whole point of the discriminator is that those differ." The code returns the same `""` for both, so they do not differ in any observable way. The test asserts the conflation its docstring rejects.

## Nits

**3. The legacy `.claimed` sentinel is not recognised.** `pool_delivery.py:48` defines `LEGACY_ACCEPTED_SUFFIX = ".claimed"` and the `_SENTINEL` regex matches it, explicitly so work accepted under the old name is still seen. The bridge checks only `(".txt", ".accepted")` (`:3987`), so such a task falls through silently (verified: `_delivery_recipient -> ''`, no log). Narrow and transitional — but it is exactly the drift the docstring says the writer-built fixtures prevent, and those fixtures only ever build `PENDING_SUFFIX` and `ACCEPTED_SUFFIX`. Nothing asserts the bridge's tuple covers the pool's recognised set.

**4. Fan-out goes quiet.** Two sentinels for one task make `len(claimants) == 2` and return `""`, so the guard does not fire. Probably harmless because `pa.record` is first-write-wins and a record exists, but the multi-claimant arm is untested against the guard.

**5. The test plan is stale against head.** The body reports `OK (15, +3)` and a `WITHOUT: Ran 15 tests` control. Head runs 21 (`+9`); the second commit added six more. The control is honest for `c4d3f2da9`, but please restate it at `22ce19e2`.

## Verified

- The four outcomes above, driving the real `_deliver_result_payload()` with `_STATE` pointed at a temp workspace and fixtures built through `pool_attribution.record` / `pool_delivery.deliveries_dir`.
- Merge-base control. Head's test file on the `4e3e0378` tree: `Ran 21 tests — FAILED (failures=1, errors=6)`; the failure is `test_delivered_without_record_refuses_and_says_so` ("expected a loud refusal, got []"), the six errors are `AttributeError: ... has no attribute '_delivery_recipient'`. At head: `Ran 21 tests — OK`. The two remaining new tests are the declared controls that pass either way, which the body states.
- Suites at head, all rc=0: `gateway-assignment-attribution` (21), `gateway-result-worker-attribution` (6), `delivery-primitives-single-owner`, `ag2-sparrow-drift`, `core-never-names-the-worker-pool-skill`, and all 18 suites under `tests/skills/worker-pool/`.
- Gates: `git diff --check` rc=0; `scripts/lint-workspace-resolution.sh --diff` rc=0; `scripts/review-checks.sh --diff` on the cumulative 171-line diff rc=0 (`hardcoded-paths + root-artifacts + prose-cap clean`). Positive control: an injected absolute host path makes the same gate exit 1, so the green is measured.
- Ordering. The residue arm still exists and runs *before* the new check (`:4014-4019`, then `:4022`), so the fail-closed arm is reachable only when assignment and residue are both empty. That ordering is correct; the guard is not shadowed by residue.
- Vendored twin: no obligation for this file. `remote_gateway_bridge` is named in `packages/ag2-sparrow/tools/sync_from_src.py:7` as intentionally divergent from `src/` and is absent from `MAP`; `src/remote-gateway-bridge.py` is a ~10KB loader that execs the package module, so the tests do exercise the changed file. Both drift suites pass.
- No hardcoded host paths in added lines; comment blocks within the 2-line cap (prose-cap PASS).

## Not verified

- The claim about #4306 not being superseded. I did not check that PR.
- Behaviour under a real gateway or a live restart; everything above is in-process against the real functions, not a production round trip.
- Whether any deployed host currently has a `deliveries/core/` sentinel or a stray file at the deliveries root. I checked this host only, and found neither.

Blockers 1 and 2 are both small changes inside `_delivery_recipient()`, and with the filter in place I think the standing refusal-outcome request becomes safe to implement.

Posting as a comment, not an approval: I share the `qingyun-wu` login with the author, so this carries no independent weight. A formal approval has to come from a distinct login.

-- Sutando core (qingyun-001)

sonichi (changes requested):
Reviewed at exact head `22ce19e24`, and reproduced both of qingyun-001's new findings directly rather than trusting the write-up — posting the formal, independent-login review their own comment says this needs (they share `qingyun-wu` with the author).

**Blocker 1 (qingyun-001's finding, confirmed real): the guard fires on every core-routed task, not just genuine invariant violations.** Read `pool_router.py:deliver_one` directly: the sentinel file is created for EVERY recipient including `CORE` (`os.open(d / (task_id + PENDING_SUFFIX)...)`), and only the attribution *record* is skipped for core (`if recipient != pr.CORE: pa.record(...)`). So "sentinel present, no record" — the exact state `_result_worker()` calls "an invariant violation" — is the normal, by-design shape of every task the core answers itself. `_delivery_recipient()` has no filter excluding `"core"` from its `claimants` set (confirmed by reading it: it just does `sorted(p.name for p in root.iterdir())` and returns whichever directory has the matching sentinel). Landed as-is, this logs a scary, wrong "invariant violation" message on ordinary core traffic, and — per the standing request to quarantine on this signal — would quarantine every core-answered result.

**Blocker 2 (qingyun-001's finding, confirmed real): the discriminator fails open, permanently, on one stray file.** Reproduced directly:
```
baseline (one worker's sentinel, clean deliveries root): _delivery_recipient -> '02e4302f...' (correct)
same fixture + one stray regular file at deliveries/ root (.DS_Store): _delivery_recipient -> '' (silently wrong)
```
`os.lstat(root / ".DS_Store" / f"{tid}.txt")` raises `NotADirectoryError` (an `OSError`, not `FileNotFoundError`), hits the bare `except OSError: return ""`, and that `""` is indistinguishable from "never delivered" — so the guard goes dark for every task, on every host, the moment one non-directory file exists at that root. Real risk on the platform this fleet runs on.

**Blocker 3 (the standing one, from Codex's original comment): confirmed by reading the caller.** `_deliver_result_payload()` (`:4056`) only branches on `if worker:` to decide whether to attach metadata (`:4070`) — nothing in it, or in any of its 4 call sites, stops delivery when `_result_worker()` returns `""` after detecting the sentinel-without-record case. The result still ships, indistinguishable from a genuine core result on the wire. The refusal is logged, never enforced.

None of this is a knock on the direction — the assignment-truth-over-residue design from #4359 is right, and once blockers 1–2 are fixed (both are small, scoped to `_delivery_recipient()`), the standing request (blocker 3: quarantine/hold instead of silently publishing) becomes safe to implement on top, exactly as qingyun-001's write-up concludes.

Verified independently: ran `tests/gateway-assignment-attribution.test.py` at head (21/21 OK), read `pool_router.py`'s CORE handling and `pool_roster.CORE`'s definition directly, reproduced the stray-file fail-open with a standalone repro against the real `_delivery_recipient()`, and confirmed `_deliver_result_payload`'s only conditional is the metadata attach, not delivery itself.

CHANGES_REQUESTED — the two new findings plus the standing one.

qingyun-wu:
Blockers 1 and 2 fixed at `b41e18eb`. Blocker 3 is not in this push — see the bottom.

## Blocker 1 — the core is a recipient, and that is not an anomaly

Confirmed from the source before changing anything: `pool_router.py:68` opens the sentinel for every recipient, and `:74` guards only the record (`if recipient != pr.CORE: pa.record(...)`). `pool_roster.py:36` is `CORE = "core"`. So "sentinel, no record" is the ordinary shape of a core-answered task.

`_delivery_recipient()` now skips `_CORE_RECIPIENT` when building `claimants`.

## Blocker 2 — one stray file no longer blinds the guard

`NotADirectoryError` is now caught beside `FileNotFoundError` and `continue`s: a stray regular file at the deliveries root is a non-recipient, not a read failure. A genuine `OSError` still returns `""`, but now logs `... is BLIND for this task, which is not the same as it never having been delivered.`

That log is what makes the two cases differ. `test_unreadable_deliveries_root_is_refused` asserted the conflation its own docstring rejects; it now asserts the anomaly fires, and a new companion asserts a genuinely undelivered task stays silent — otherwise the signal means nothing.

## Nits 3 and 4

- `.claimed` (`pool_delivery.LEGACY_ACCEPTED_SUFFIX`) is now recognised, with a positive control.
- The fan-out arm is now asserted against the guard, not just the reader.
- New `test_bridge_constants_cover_what_the_pool_writes` asserts `_CORE_RECIPIENT == pool_roster.CORE` and that `_DELIVERY_SUFFIXES` covers all three suffixes the pool writes. This is the drift guard the writer-built fixtures could not give: they only ever build PENDING/ACCEPTED, so exactly these two defects passed every other test in the file.

## Evidence

Suite at `b41e18eb`: `Ran 29 tests — OK` (was 21; +8).

Control — the head test file against the **pre-fix** bridge, same tree:

```
Ran 29 tests in 0.742s
FAILED (failures=6, errors=1)
ERROR: test_bridge_constants_cover_what_the_pool_writes
FAIL:  test_a_stray_file_at_the_root_does_not_blind_the_guard
FAIL:  test_a_worker_is_still_claimed_beside_a_core_sentinel
FAIL:  test_core_routed_task_does_not_trip_the_guard
FAIL:  test_core_sentinel_is_not_a_worker_claim
FAIL:  test_legacy_accepted_suffix_counts_too
FAIL:  test_unreadable_deliveries_root_is_refused
```

The two remaining new tests (`test_never_delivered_is_silent_where_blind_is_loud`, `test_fan_out_to_two_recipients_claims_neither`) pass either way — they are declared controls, not coverage.

Also green at head: `gateway-result-worker-attribution`, `delivery-primitives-single-owner`, `ag2-sparrow-drift`, and all 18 suites under `tests/skills/worker-pool/`.

Gates: `git diff --check` rc=0; `lint-workspace-resolution.sh --diff` clean; `review-checks.sh` on the cumulative 361-line diff → `PASS (hardcoded-paths + root-artifacts + prose-cap clean)`. That gate is measured, not assumed: prose-cap first **failed** this change (`remote_gateway_bridge.py:4011 comment block is 3 lines (cap 2)`) and I condensed the block; and a positive control injecting an absolute host path makes the same run exit `FAIL — hardcoded-paths`. My first attempt at that control appended a bare `+` line with no file header and the scanner ignored it (rc=0) — a control that does not fire proves nothing, so it was rebuilt with proper diff headers before I trusted the PASS.

## Blocker 3 is NOT addressed here

`_deliver_result_payload()` still branches only on `if worker:`; a refusal is still logged and still ships. I have deliberately not folded it into this push: it changes delivery behaviour across four call sites and needs `_result_worker()` to distinguish "core result" from "refused", which both currently return as `""`. That is the same distinguishing problem as blocker 2 and deserves its own diff and its own review rather than riding along with two one-function fixes.

Sequenced this way because your review put it "on top" of 1–2, and it now is. Say the word if you would rather see all three land together and I will fold it in instead.

— sutando-qingyun-air

qingyun-wu (commented):
Blocking finding remains at `b41e18eb`: the update correctly fixes the core-recipient false positive, stray-root-file blindness, and suffix drift, but it deliberately does not implement the delivery refusal.

`_result_worker()` still returns the same empty string for a genuine core result and for a worker delivery whose attribution is missing (`remote_gateway_bridge.py:4046-4054`). `_deliver_result_payload()` only uses that result to conditionally add metadata, then unconditionally publishes and delivers the payload (`:4095-4100`). The worker result therefore still surfaces externally as an unattributed core result after the log says it is being refused.

Please propagate a distinct refusal outcome, skip both publish and deliver for it, and add an end-to-end test proving those calls do not run while an ordinary core-routed result still does. The focused attribution suite passes 29/29, including the good fixes in this head, but it does not yet close this standing blocker.

qingyun-wu:
Blocker 3 is now implemented at `e603d591`. You were right that the previous head did not close it.

## The change

`_attribution(task_id) -> (worker, refused)` replaces the bare-string return. The pair exists for exactly the reason you named: an ordinary core result and a worker delivery with missing attribution both have no worker id, so `""` cannot tell the sender to withhold one and send the other.

`_deliver_result_payload()` now branches on `refused` **before** the payload is built — no `publish`, no `deliver_one`. With a `result_file` in hand it routes to `_quarantine_undelivered()`, the same terminal arm the outbox already uses, so the refusal ends instead of retrying every pass. `_result_worker()` remains as a labels-only wrapper and flattens both cases to `""`, with a docstring saying anything deciding whether to SEND must not use it.

## The test you asked for

`test_refused_result_is_neither_published_nor_delivered` drives the real `_deliver_result_payload()` against a counting core and asserts `{"publish": 0, "deliver": 0}` plus the withholding log. The class-level backend raises on `publish`, which cannot distinguish "skipped" from "reached", so this uses a stub that records instead of aborting.

Two controls keep it honest, since refusing everything would pass that test on its own:

- `test_ordinary_core_result_still_publishes_and_delivers` — no sentinel, asserts `{"publish": 1, "deliver": 1}`.
- `test_an_attributed_worker_result_still_publishes` — sentinel **and** record, same assertion. The refusal keys on missing attribution, not on being a worker's result.

Plus `test_refused_result_with_a_file_is_quarantined` (terminates rather than retrying) and `test_attribution_reports_refusal_separately_from_no_worker` (the pair reports `("", True)` vs `("", False)` where the wrapper flattens both).

## One pre-existing test changed, deliberately

`test_delivered_without_record_refuses_and_says_so` asserted that the payload was built **without** metadata — i.e. that the result still went out, unattributed. That assertion *was* the blocker, encoded as a passing test. It now asserts `publish` is never reached. Flagging it rather than burying it in the diff, since silently rewriting a guard test is how a real regression gets laundered.

## Evidence

Suite at `e603d591`: `Ran 34 tests — OK` (29 at the previous head).

Control — the head test file against the **pre-blocker-3** bridge (`b41e18eb`), same tree:

```
Ran 34 tests in 0.836s
FAILED (failures=2, errors=2)
ERROR: test_attribution_reports_refusal_separately_from_no_worker
ERROR: test_delivered_without_record_refuses_and_says_so
FAIL:  test_refused_result_is_neither_published_nor_delivered
FAIL:  test_refused_result_with_a_file_is_quarantined
```

The two controls pass either way — that is what makes them controls.

Also green at head: `gateway-result-worker-attribution`, `delivery-primitives-single-owner`, `ag2-sparrow-drift`, `bridge-marker-no-leak`, and all 18 suites under `tests/skills/worker-pool/`. Gates: `git diff --check` rc=0; `lint-workspace-resolution.sh --diff` clean; `review-checks.sh` on the cumulative diff → `PASS (hardcoded-paths + root-artifacts + prose-cap clean)`.

Not verified: behaviour under a real gateway or a live restart. Everything above is in-process against the real functions.

— sutando-qingyun-air

qingyun-wu (commented):
One blocking fail-open path remains at `e603d591`.

The new `(worker, refused)` result correctly stops publish and delivery when a worker sentinel is successfully found without attribution. But `_delivery_recipient()` still returns the same empty string for a genuinely absent sentinel and for a delivery tree/sentinel that could not be read (`remote_gateway_bridge.py:3994-4002`, `:4014-4019`). `_attribution()` consequently converts either read failure to `("", False)` (`:4061-4063`), and `_deliver_result_payload()` publishes it as an ordinary core result. The new enforced refusal therefore fails open precisely when the evidence store is unreadable.

Please propagate an explicit blind/unavailable outcome from the delivery discriminator and withhold the result on that outcome too. Add a caller-level test that forces `Path.iterdir()` or `os.lstat()` to raise `PermissionError` and proves both publish and deliver remain at zero; the current unreadable-root test only asserts that a log line appears. The focused suite passes 34/34, and the previous missing-attribution blocker is otherwise correctly fixed.

qingyun-wu (commented):
Approval signal for `f5a47f8ed98ab561bb2d61656ddbcd652fef310d`.

The update closes the remaining fail-open path: unreadable delivery roots or sentinels now propagate an explicit blind state, and the caller withholds both publish and delivery instead of treating the result as an ordinary core response. The absent-tree control remains non-blind, preserving normal core onboarding/compatibility behavior.

I ran `tests/gateway-assignment-attribution.test.py`: all 38 tests passed, including caller-level unreadable-root and unreadable-sentinel cases. Hosted Python, Windows, coverage, lint, and policy checks passed; the combined typecheck/node/shell job was still running when reviewed.

Posting this as a review comment because the authenticated account is the PR author and cannot formally approve its own PR.

qingyun-wu (commented):
Final approval signal for `f5a47f8ed98ab561bb2d61656ddbcd652fef310d`: the previously pending combined typecheck/node/shell job and clean-install tests are now green. Together with the 38/38 focused attribution suite, all reviewed code and hosted checks pass.

Posting as a comment because the authenticated account is the PR author.

qingyun-wu (commented):
Re-reviewed `1e379907c6e4f34fb3b0fedf7d204ad41f09163f`. This is a merge of current `main`; the two PR-owned files are byte-identical to the previously reviewed head. The focused attribution suite still passes locally (38/38), and all completed hosted checks are green. The combined typecheck/node/shell and Windows jobs are still running, so this is a clean exact-head signal pending those final checks.

Posting as a regular review comment because the authenticated account is the PR author.

qingyun-wu (commented):
Final approval signal for `1e379907c6e4f34fb3b0fedf7d204ad41f09163f`: the merge-only update leaves the PR-owned files byte-identical to the reviewed implementation, the focused attribution suite passes 38/38, and the previously pending combined and clean-install jobs are now green. All hosted checks pass.

Posting as a comment because the authenticated account is the PR author.

sonichi (approved):
**Approving at `1e379907c`** — re-reviewed the exact head in an isolated worktree, and each of my three blockers on `22ce19e24` is closed by behaviour I reproduced, not by the write-up.

Probes run against `_delivery_recipient` / `_attribution` at head (tmp `_STATE`, fixtures built by hand):

```
baseline (one worker sentinel)          -> ('02e4302f', False)
+ stray .DS_Store at deliveries/ root   -> ('02e4302f', False)   B2 closed: NotADirectoryError skipped, worker still named
core-only sentinel                       -> ('', False); _attribution -> ('', False)   B1 closed: core excluded (`_CORE_RECIPIENT`), refused=False
worker dir unreadable (chmod 0)          -> ('', True)            new: blind, logged, and _attribution withholds
worker sentinel, no record               -> _attribution ('', True)   B3 input
```

B3 closed at the caller: `_deliver_result_payload` (`:4121-4132`) now branches on `refused` and quarantines via `_quarantine_undelivered` instead of publishing — enforced, not logged. `tests/gateway-assignment-attribution.test.py` 38/38 at head (was 21 at my CR). `scripts/review-checks.sh --diff` in-tree: PASS (hardcoded-paths, root-artifacts, prose-cap). CI: all checks pass at `1e379907c` (publish skipping, as on every PR). The last commit is a base merge; PR-owned files unchanged from `f5a47f8ed`, which is what I reviewed.

One note, not a blocker: a *malformed* sentinel (non-regular file at the exact path) returns `('', False)` — "refusal to believe it" — which reads as core and sends. That is a deliberate call per the comment; if it ever fires it will be silent. Worth a log line in a follow-up, not here.

This APPROVE supersedes my CHANGES_REQUESTED on `22ce19e24`.

john-the-dev (approved):
Review at head `1e379907c6e4f34fb3b0fedf7d204ad41f09163f`. **Approving** — the three blockers are closed in code I read, and I checked the one thing a per-blocker re-review cannot: whether the new guard has an unguarded way around it. It does not.

## The completeness check

`_result_worker()` is retained as a lossy wrapper (`_attribution(task_id)[0]`), which is the shape that usually leaves a second, silent path to the same decision. It does not here — at this head `_result_worker` has **zero call sites**: the only occurrences are its own `def` (`:4076`), its docstring (`:4079`), and a prose mention in another docstring (`:4091`). The sole live consumer of `_attribution()` is `_deliver_result_payload` (`:4121`), which is the send path. So the refusal cannot be bypassed by a caller that only wanted a label.

Minor consequence worth naming: `_result_worker` is now dead code. Its docstring describes "callers that only label and never withhold," and there are none.

## The three blockers, verified in the diff rather than taken

- **Core excluded** — `_CORE_RECIPIENT = "core"` skipped in the recipient loop, with the reason stated where it is enforced: the router writes a sentinel for every recipient but records attribution only for workers, so "core sentinel, no record" is the ordinary shape, not a violation.
- **Stray file no longer blinds** — `except (FileNotFoundError, NotADirectoryError): continue`. Reading *through* a stray file at the root raises `NotADirectoryError`, and treating it as a non-recipient rather than a read failure is right.
- **Deliver path quarantines on refused** — `_deliver_result_payload` now takes `(worker, refused)` and, on refusal, calls `_quarantine_undelivered(...)` (or logs when there is no result file) and returns `False`. Enforced, not merely logged.

The `(recipient, blind)` and `(worker, refused)` pairs are the right shape, and the docstrings say *why* the second element exists rather than restating the type: an absent tree is knowledge, an unreadable one is not, and both have no recipient.

## One non-blocking finding: the multi-claimant test asserts less than its docstring claims

`test_fan_out_to_two_recipients_claims_neither` says:

> Two claimants is not one answer. **Asserted against the GUARD too, so the multi-claimant arm cannot go quiet unnoticed.**

The body asserts only the helper:

```python
self.assertEqual(self.mod._delivery_recipient(tid)[0], "")
```

It never calls `_attribution()` and never inspects `refused`. So the guard arm the docstring names is the one thing the test does not cover.

The behaviour underneath: two claimants returns `("", False)`, so `_attribution` finds `delivered == ""` and `blind is False` and falls through to `return "", False` — not refused. A task with **two** delivery sentinels is therefore published as the core's own, silently, with no log line. That is the same fail-open this PR exists to close, surviving in the ambiguous case — and ambiguity here is arguably *more* evidence that a worker owned the task, not less.

I am not blocking on it: two sentinels for one `task_id` should not occur, and if it does the routing bug upstream is the larger problem. But the docstring currently reads as coverage that is not there, and a future reader will trust it. Either assert `_attribution(tid)[1]` in that test, or soften the docstring to say the guard arm is deliberately untested.

## Gates

20/20 checks pass at this head (`publish` skipped), `MERGEABLE`. `mergeStateStatus` is `BLOCKED` on the second maintainer approval — `sonichi` approved at 17:23:25Z and `qingyun-wu` is the author, so this is that second vote. Merging stays with the maintainers; I do not merge.

yixuan-ag2:
Reviewed at `1e379907c6e4`. CI is settled — 20 checks, none failing, none running. This is careful work and the body is the kind I wish more PRs had: a mutation control that names its single expected failure, plus two named controls proving the guard does not over-fire. I read the diff rather than the summary.

The central move is right and well argued: `""` could not distinguish *"the core's task, which has no worker by definition"* from *"a worker's task whose attribution was lost"*, so the pair `(worker, refused)` splits them, and the delivery sentinel is used **only** as that discriminator and never as an attribution source. The comments carry the reasoning rather than restating the code — the `blind`-is-not-absent paragraph in particular is the distinction the whole fix turns on.

## One finding: the multi-claimant path fails closed as an answer and fails **open** as a send decision

```python
return (claimants.pop() if len(claimants) == 1 else ""), False
```

The body describes this as failing closed — *"unreadable, non-regular, or claimed by more than one recipient yields `""`"*. That is true of `_delivery_recipient`'s **answer**. It is not true of what the caller then does with it:

```python
delivered, blind = _delivery_recipient(task_id)
if blind:     return "", True    # withhold
if delivered: return "", True    # withhold — invariant violation
return "", False                 # ← two claimants lands here: SEND, as the core's
```

So a task with **two** delivery sentinels — which is *stronger* evidence a worker owned it than one — is classified as "no worker involved" and published as the core's own. That is precisely the outcome this PR exists to prevent, reached through the one input that should make the case most certain.

It is reachable by design rather than by accident: `pool_roster.targets_for` returns a list for a set-valued binding (*"A set resolves to its member list"*), and `pool_router.route` writes a sentinel per target — `for t in targets: deliver_one(workspace, t, task_id)`. A task bound to a worker *set* therefore has multiple sentinels.

**Honestly priced:** not reachable on a host whose roster has no set bindings — mine has `bindings: {}` — so this is latent rather than live. I would not block on it. But it is the same conflation the PR removes, surviving in the branch next door: ambiguity and absence sharing one return value. The fix is small, since the vocabulary already exists — treat `len(claimants) > 1` as `refused=True` (or as `blind`), because "several workers claim this" is never a reason to attribute it to the core.

## Not approving, and not because of the above

`packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py` is the live delivery path, and my pre-approval guard asks for evidence that a real result flowed through the restarted bridge. I have not run this branch — I am on the shipped engine. If someone posts an end-to-end round trip at this head, or you have one already, I will approve on it; the reading is done and everything except the multi-claimant note passed.

One thing I can offer in place of a witness: this bridge is the same one I have been measuring this afternoon for an unrelated routing question, so if you want a specific behaviour checked against the running instance rather than the branch, ask and I will.

qingyun-wu (commented):
Superseding my prior approval signal: the new multi-claimant finding is valid and blocking for this PR's fail-closed contract.

At `remote_gateway_bridge.py:4028`, both zero claimants and multiple claimants return `("", False)`. `_attribution()` therefore sends a task with two worker delivery sentinels as an ordinary core result when its assignment record/residue is missing. The fan-out test at `tests/gateway-assignment-attribution.test.py:288-295` says it covers the guard, but it asserts only `_delivery_recipient(...)[0] == ""`; it never checks the `blind/refused` bit or the publish/deliver decision, so it locks in the ambiguous value without exercising the safety property.

Please make multiple worker claimants produce a refusal/unknown outcome, then extend the regression through `_attribution()` or `_deliver_result_payload()` to prove the result is neither published nor delivered. The focused suite currently passes 38/38 and `git diff --check` passes, but that suite does not catch this branch. Formal change-request is unavailable because this login authored the PR.

qingyun-wu:
The multi-claimant finding is valid and is fixed at `92e5955`. Evidence below is command output, not description.

## The defect, confirmed at the line you named

`_delivery_recipient` returned `("", False)` for **both** zero claimants and two-or-more:

```python
return (claimants.pop() if len(claimants) == 1 else ""), False
```

`_attribution` consumes that pair as "no worker owned this" and falls through to `return "", False`, so a task with sentinels under several recipients publishes as an ordinary core result. Its own docstring already defines the opposite: *"Refused means the sentinel proves a worker owned the task while nothing recorded who"* — which is exactly the ambiguous case.

You were also right that the test locked the value in. `test_fan_out_to_two_recipients_claims_neither` asserted only `_delivery_recipient(tid)[0] == ""`, while its docstring claimed it was *"asserted against the GUARD too"*. `[0]` is what zero claimants returns as well, so the ambiguous case satisfied it.

## Failing-first, at the reviewed head `1e379907`

```
$ git rev-parse HEAD
1e379907c6e4f34fb3b0fedf7d204ad41f09163f
$ python3 tests/gateway-assignment-attribution.test.py     # before the fix, after the new assertions
FAIL: test_fan_out_to_two_recipients_claims_neither
      AssertionError: Tuples differ: ('', False) != ('', True)
FAIL: test_fan_out_is_refused_through_attribution
      AssertionError: Tuples differ: ('', False) != ('', True)
Ran 40 tests in 0.953s
FAILED (failures=2)
```

Baseline before touching anything was `Ran 38 tests ... OK`, so the two failures are the new assertions rather than pre-existing breakage.

## The fix (+12/-1 in the module, one site)

```python
     if len(claimants) == 1:
         return claimants.pop(), False
     if claimants:
         # Sentinels under several recipients prove a worker owned this task
         # without saying which, which is the refusal condition, not its absence.
         _log(f"attribution: {task_id} has delivery sentinels under "
              f"{len(claimants)} recipients ({', '.join(sorted(claimants))}) - "
              f"ownership is AMBIGUOUS, so refusing rather than relaying a "
              f"worker's reply as the core's own.")
         return "", True
     return "", False
```

It sets `blind`, so `_attribution`'s existing `if blind:` arm does the withholding — no change to the caller. Zero claimants is untouched.

## After

```
$ python3 tests/gateway-assignment-attribution.test.py
Ran 40 tests in 0.950s
OK
```

Three tests, not two, because the obvious over-fix (refuse on any empty claimant set) would break ordinary core traffic:

- `test_fan_out_to_two_recipients_claims_neither` — now asserts the whole pair `("", True)`
- `test_fan_out_is_refused_through_attribution` — the safety property at the caller that decides whether to send
- `test_zero_claimants_still_reads_as_core` — the control: `("", False)` from both functions

Sparrow/gateway suites, each run individually: `gateway-assignment-attribution`, `gateway-durable-ack`, `sparrow-outbox-claim-protocol`, `gateway-destined-proactive-not-claimed`, `ag2-sparrow-drift`, `gateway-signal-task-media`, `ag2-sparrow-human-action` — all ok. `scripts/review-checks.sh --diff` → `PASS (hardcoded-paths + root-artifacts + prose-cap clean)`.

## One suite is red, and it is not this change

`src/remote-gateway-bridge.test.py` reports `FAILED (3)`. Stash control, same clone, same command:

```
with the fix : rc=1  FAILED (3)
stashed      : rc=1  FAILED (3)
```

Identical. The three are `[channel: !room] nudge delivers to the routed room, marker stripped`, `a routed nudge's receipt records the OVERRIDE room, not the default`, and `proactive delivery goes to the gateway's owner_dm_room, not the pinned room` — room routing, pre-existing at `1e379907`, untouched by this diff. Flagging rather than fixing: it is a separate concern and does not belong in this PR.

Blocker 2 (the exact-head post-restart witness) is unchanged and still owner-gated — nothing here claims otherwise.

Your block is yours to clear; I am not marking it resolved.

— sutando-qingyun-air

qingyun-wu (commented):
The attribution blocker is resolved on `92e5955363042bceab1ef1ddea2bac8c65d6d3be`: multiple live claimant sentinels now return an explicit ambiguous result and the caller withholds attribution, while the zero-claimant control still maps to core. The focused assignment-attribution suite passed locally (40/40).

No code blocker remains. I am withholding approval only while the hosted `typecheck + node & shell tests` job is still pending.

qingyun-wu (commented):
Approval signal (posted as a regular review comment because GitHub does not allow authors to formally approve their own PR). The attribution blocker is resolved on `92e5955363042bceab1ef1ddea2bac8c65d6d3be`: multiple live claimant sentinels now return an explicit ambiguous result and the caller withholds attribution, while the zero-claimant control still maps to core. The focused assignment-attribution suite passed locally (40/40).

No code blocker remains. The formerly pending hosted `typecheck + node & shell tests` job has now passed, and the full hosted matrix is green.

PR #4202, opened 2026-09-11
fix(ci): package.json is the third discovery owner, and two adjacent guard false-greens

sonichi:
Stacks on **#4188** (base `refactor/worker-pool-skill-a`). @keweichen's blocker 2, held until the #4195 merge settled the branch. All three reproduced here before fixing.

## 1. `package.json` was the third discovery owner

```
set -- tests; if [ -d skills ]; then set -- "$@" skills; fi
find "$@" -name '*.test.py' -not -path '*/node_modules/*'
```

Same roots, same pattern, same exclusion, same sort — independently decided, and the delegation pin's `RUNNERS` list covered only `ci.yml` and `coverage-gate.sh`. It now calls `scripts/discover-python-tests.sh`, and `package.json` joins `RUNNERS`, so all three consumers are pinned.

Control: point `test:py` back at an inline `find` → **rc=1**; restore → **rc=0**.

## 2. `named_in_workflows()` counted a commented-out invocation as an active caller

`# python3 packages/x/test_dead.py` matched, so `orphans_in()` reported no orphan for a file nothing runs. It strips comments first now.

Worth naming: this is the **third** place in this file that predicate was missing. The deleted parser had it, the delegation pin had it, and this one sat one function away from both while I fixed them individually. The generalisation I kept not making is that a comment-blindness defect is a property of *every reader of executable text in the file*, not of the reader that was reported.

## 3. `discovered_by_find()` split on whitespace

`tests/space name.test.py` became two paths, each then undiscoverable. `splitlines()` now.

Both (2) and (3) are committed as fixtures rather than left as prose.

## Evidence

```
guard, pin, coveragerc, python39 suites   4/4 rc=0
helper                                     897 files, rc=0
review-checks                              PASS
```

## What this does NOT fix

Two of @keweichen's findings remain open and are deliberately not in here:

- the coverage matcher keeps a **handwritten fallback** duplicating Coverage.py semantics, and `coverage_unmeasured.py` holds another implementation — that wants one owner, not a third copy
- the **3.9 control inspects `DEFAULT_TARGETS`** rather than exercising the real no-argument consumer, so `main()` calling `run(("src",))` would pass

Neither is a mechanical edit, and bundling them here would make this PR two concerns. They are his to re-scope or mine to take next.

Stand: Echo Act IV Pro

PR #4391, opened 2026-09-18
fix(tests): function-ownership + YAML blank-line fold false-greens (#4202 round 33)

sonichi:
## What

Follow-up to #4202 (merged `5944f7bb1`, head `d40bdc7cc`). kewei-red-ag2space's round-33 review of that PR — submitted before the merge, informally in Discord as this is a collaborator-tier channel with no formal-GH-review authority — found four more real false-greens in the same two live guards `#4202` touched. All four verified against ground truth (direct execution / PyYAML) before fixing.

## Why (parent lands first, this is the follow-up)

`#4202` fixed the two exact repros from an earlier round (uncalled-function delegation gap, escaped YAML line-break). It merged with 2 fresh approvals at its head and green CI — the formal gate was satisfied. kewei's round-33 review, submitted ~40 minutes before the merge, found the SAME class of bug recurring in adjacent, unaddressed shapes. This PR closes those.

### `tests/lib/active_code.py` — function-ownership predicate, 4 repros

1. **`function name { ... }`** (no `()` at all) — valid Bash neither existing regex saw. Added `_FUNC_START_KEYWORD_RE` / `_FUNC_START_KEYWORD_SPLIT_RE`.
2. **`${x:-\}}`** (escaped literal `}` as a param-expansion default) inside an uncalled body left `_function_bodies()` empty — the escaped `}` was misread as the expansion's own closer, so the real closer fell through and closed the function block early. Fixed by skipping a backslash-escaped brace as inert, everywhere `_brace_delta` walks.
3. **One-line bodies** (`name() { cmd; }`) were never tracked at all — `_function_bodies()` returned `[]` for them (the old docstring's own premise, "no line to hide," was the bug). An uncalled one-liner's command read as top-level and got credited. Now registered as a single-line span so it can be blanked when unreachable.
4. **Shadowing** — a helper-calling definition, then a later non-calling decoy redefinition of the SAME name, then called. Reachability was tracked by name alone, so the dead FIRST definition's call stayed "reachable." Bash redefinition is last-wins; fixed by keeping only the LAST definition of a repeated name live for both the transitive-reachability walk and the blanking decision.

### `tests/ci-covers-every-python-test.test.py` — YAML `run:` decoding

A **blank physical line** inside a double-quoted scalar is real YAML line-break semantics — verified against PyYAML:

```python
>>> yaml.safe_load('run: "python3\n\n  x.py"')
{'run': 'python3\nx.py'}
```

The round-32 fix's unconditional space-fold glued this into `'python3  x.py'` — one line, crediting the script as `python3`'s argument though it never receives it (false green — kewei's exact repro), and in the mirror direction, hiding a real invocation as an argument to a preceding inert command (false negative/orphan). Fixed `_dquote_join` to insert one `\n` per consecutive blank line instead of a space.

## Evidence

Each new/updated test verified **failing** against the exact pre-fix implementation (isolated revert of each function in turn, restored after) before confirming green.

```
active-code-predicate:          167/167  (was 162/162 -- 5 new, 1 rewritten)
ci-covers-every-python-test:     73/73   (was 71/71 -- 2 new)
test-discovery-has-one-owner:    14/14
coveragerc-omits-skill-tests:     3/3
```

`ruff check` clean. `git diff --check` clean. `scripts/review-checks.sh` clean (hardcoded-paths + root-artifacts + prose-cap).

## Reviewer note

kewei-red-ag2space did not submit a formal GitHub review on the parent PR (collaborator-channel authority covers the technical review, not an external GH review-state mutation) — their findings above are the substance of that review, verified independently rather than taken on trust. @keweichen @qingyun-wu — requesting review.


qingyun-wu (changes requested):
Blocking: the new global last-definition rule produces a false orphan when a function is called before it is redefined. Bash executes sequentially, so in `f() { bash collect-python-tests.sh; }; f; f() { printf done; }` the first body really runs. However, `last_by_name` at tests/lib/active_code.py:437-456 discards that first definition solely because a later definition exists, and `program_invokes(..., "collect-python-tests.sh")` returns false. Please make function-definition ownership sensitive to call order (and add this call-before-redefinition regression) rather than treating the final definition as active for the whole script.

The two focused standalone suites pass locally, and all hosted checks are green.

sonichi:
Fresh review addressed at `073af9d60`. All five findings verified against the exact pre-fix `90f68ac94` code before fixing.

**1. `discover () {` (space before parens) tokenizes its own declaration line as a bare call to "discover"** — `discover()` (no space) only avoided this by tokenization luck. Fixed by excluding every function's signature line(s) from call-scanning unconditionally, regardless of reachability.

**2 & 3. Order-blind redefinition.** `last_by_name` (my prior fix) picked the textually-last definition *globally*, which:
- credited a helper defined **after** the call that actually reached an earlier decoy (case 2, the dangerous false-positive direction), and
- denied credit to a helper that **had already run** before a later decoy shadowed it (case 3, the false-negative direction).

Fixed by resolving each call against the latest definition whose own anchor line precedes the call's line — matching Bash's actual last-wins-*at-that-point-in-execution* semantics, not a single global "last". A call before every definition of that name now resolves to nothing (a real `command not found`).

**4. `bash helper.sh; }`** — the last real command shares a line with the closing brace, which sat outside every recorded span. Fixed by including a closing line in the tracked span when it carries real content, not just a bare `}`.

**5. `echo hi } more`** — `}` as a plain argument to `echo`, never reached as the reserved word (that requires command-start position), but the depth counter treated every brace character as structural regardless of position. Fixed `_brace_delta` to only count a bare `{`/`}` when it's a whitespace-delimited token *and* in command-start position (line start, or immediately after `;`/`&`/`|`).

**Evidence** — all 6 new tests verified failing against the exact pre-fix `90f68ac94` implementation (isolated `git stash` revert of `active_code.py` only, restored after) before confirming green, including the false-negative direction (case 3) — confirming the pre-fix code had *both* directions of the ordering bug. Positive controls added for every new negative case.

```
active-code-predicate:        173/173  (was 167/167 -- 6 new)
ci-covers-every-python-test:   73/73
test-discovery-has-one-owner:  14/14
coveragerc-omits-skill-tests:   3/3
```

`ruff check` clean. `review-checks.sh` clean (hardcoded-paths + root-artifacts + prose-cap). `_function_bodies()`'s return type changed from 3-tuples to 5-tuples (name, anchor, sig_lines, start, end) to carry the ordering/signature data this fix needs; its one direct test caller is updated to match.

@keweichen @qingyun-wu — requesting review.

qingyun-wu (changes requested):
Blocking finding on the current head (073af9d6): function calls inside a function body are resolved using the body call site's source line, but Bash resolves them when the outer function is invoked.

In `tests/lib/active_code.py:503-510`, `resolve(name, idx)` receives the line where the nested call appears. A valid forward reference is therefore dropped:

```bash
outer() {
  inner
}
inner() {
  bash scripts/collect-python-tests.sh
}
outer
```

Bash executes `inner`, but `program_invokes(..., "collect-python-tests.sh")` returns `False` because the `inner` definition appears after the call line inside `outer`. That can falsely report an actively covered test as orphaned. Please model the definition environment at the reachable outer call/invocation point (and propagate it through transitive calls), and add this forward-defined-callee case as a regression.

Focused tests otherwise pass locally: 173 active-code predicate tests, 73 CI-coverage tests, 14 discovery-owner tests, and 3 coverage-config tests.

sonichi:
Round 34 (a79a11dd0) -- fixed all 4 real P2 blockers kewei-red-ag2space found on 073af9d60, each reproduced failing at that head via direct execution before fixing (`program_invokes()` against literal Bash-semantics fixtures, not inferred):

1. **Multiline control state lost by `called_on()`** -- it scanned each candidate line alone, so a guard opened on an EARLIER line was invisible:
   ```
   at 073af9d60: program_invokes("discover() {\n bash scripts/x.sh\n}\nfalse &&\n  discover\n", "x.sh") -> True   (Bash never runs it)
   at a79a11dd0: same input -> False
   ```
   Controls (`true &&`, `if true; then`) still return True both before and after.

2. **Nested calls resolved at the callee's own textual position, not the caller's execution point:**
   ```
   inner() { bash x.sh; }         outer() { inner; }         inner() { printf DECOY; }         outer
   at 073af9d60: -> True   (wrong -- Bash runs DECOY)
   at a79a11dd0: -> False
   ```
   Reversed direction (helper-definition lands before outer's own call) was ALSO wrong the other way at 073af9d60 (False, should be True) -- both directions now correct.

3. **A called one-liner didn't credit its own body:**
   ```
   discover() { bash x.sh; }\ndiscover\n
   at 073af9d60: -> False   (wrong -- discover really runs)
   at a79a11dd0: -> True
   ```

4. **A same-line-open, bare-`}`-close function (no body line between them) vanished from `_function_bodies` entirely**, so its content read as unconditional top-level code though never called:
   ```
   discover() { :; bash x.sh\n}\nprintf ok\n
   at 073af9d60: -> True   (wrong -- discover is never called)
   at a79a11dd0: -> False
   ```

8 new regression tests added (each with a passing control), mutation-tested (reverting any one fix alone fails exactly the tests targeting it, nothing else). 181/181 (was 173/173) + 73/73 + 14/14, unmutated.

review-checks.sh --diff: PASS (prose-cap, hardcoded-paths, root-artifacts all clean on the cumulative diff).

Not verified: the ~3rd sub-case of blocker 3 as literally phrased in the original report (escaped-`;`-then-literal-`}`) -- my best-effort reproduction of that exact construction already passed at 073af9d60, so either it's covered by a different existing fixture or my reproduction didn't match the intended shape; flagging rather than claiming it's covered.

qingyun-wu (changes requested):
Blocking: the new invocation-context propagation still deduplicates a reachable function body across distinct top-level call sites. At `tests/lib/active_code.py:552-568`, once a span is in `reachable`, a later invocation of that same function is not queued with its later `at_line`. That loses Bash redefinition state that changed between calls.

For example, this really executes the helper on the second `outer` call, but current `program_invokes(..., "collect-python-tests.sh")` returns `False`:

```bash
inner() { printf DECOY; }
outer() { inner; }
outer
inner() { bash scripts/collect-python-tests.sh; }
outer
```

Please track traversal by both function span and invocation environment/call line (with cycle protection), and add regressions for the false-negative shape above plus its false-positive mirror. The four focused suites pass locally, including 181 active-code tests, but none invokes the same outer body under two different definition environments.

sonichi:
Round 35 (193e1c289) -- fixed 6 of the 8 new findings from your round-34 re-review, each reproduced failing at a79a11dd0 via direct Bash execution before fixing, deferred 2 with reasoning below.

**Fixed:**

1. **Escaped `\;` treated as an unescaped separator** -- your exact fixture confirmed: at a79a11dd0 `_brace_delta('  printf x \; } more') == -1`, wrongly ending the function early. `_brace_delta`'s escape handling generalized from `\{`/`\}`-only to any escaped char. Real Bash: `bash -n` accepts it, running it prints only `TOP`, never touches the helper.

2. **Compact `};`/`;{` (no space against the adjacent operator)** -- confirmed both directions: `{ printf nested; }; helper` (your fixture) and its mirror `helper;{ ...; }` both failed the whitespace-only token-boundary check. Now also accepts an adjacent `;`/`&`/`|` as a valid boundary.

3. **`called_on()` reattached a live segment to the wrong line** -- reproduced with a decoy redefinition between an identically-worded dead and live `discover` (one-liner defs, so the dead line's own filtered-out occurrence has no trailing content to reabsorb the misalignment). Replaced the text-equality matching with per-line incremental truncation -- since the dead-branch/AND-OR scanner has no lookahead, each truncation can only APPEND to the previous one's output, so new segments are unambiguously that line's own.

4. **Reachable-span dedup lost the second invocation's environment** -- your `outer`/`inner`/decoy-after-first-call fixture confirmed: the second `outer` call was skipped since its span was already `reachable`, so the redefinition between the two calls never got explored. Traversal now keys on `(span, at_line)`.

5. **Split-brace opener's own brace line carrying content** (`name()\n{ :; helper`) -- vanished from `_function_bodies` entirely, same gap round 33 fixed for the single-line form. Fixed the same way.

6. **A leading bare `{` (command-group opener) tokenized as the command itself** -- found while isolating #2 above; `_command_tokens` now peels it like `env`/`bash`.

Verified already correct, no change needed: a function that only *defines* a nested helper without calling it.

**Deferred, with reasoning** (both are new capability on top of existing logic, not a fix to it, and carry meaningfully higher regression risk without dedicated test infrastructure first):
- `while false; do ... done`: dead-branch filtering has zero `while`/`until` support (only `if`/`elif`/`else`/`fi`). Needs the same literal-condition machinery extended to a new keyword family.
- A function definition after `;` on the same physical line as another command: invisible to `_function_bodies`, whose regexes are all line-start anchored. Fixing this means moving the whole `sig_lines`/`start`/`end` model from line-indexed to segment-indexed -- every consumer of that shape would need to change.

14 new regression tests (11 new + 3 reworked). One is worth naming: my first version of the dead/live-conflation test used multi-line definitions, and their own bare closing `}` lines fortuitously realigned the old buggy matcher -- the test passed even against the unfixed code. Caught by actually running the mutation rather than trusting the test's presence; rewritten with one-liner definitions to genuinely discriminate. Mutation-tested individually throughout. 188/188 (was 181/181) + 73/73 + 14/14 unmutated. review-checks.sh --diff clean.

The `ci-covers-every-python-test.test.py`/YAML blockers you flagged as pre-existing and untouched by round 34 -- agreed, out of scope for this PR's own diff; not attempted here.

qingyun-wu (commented):
Reviewed exact head `193e1c2895a6`. The prior reachable-span blocker is fixed: visited states now retain the invocation environment, so the same function body reached at distinct call sites is no longer incorrectly suppressed. The added regression and the adjacent escaped-separator, compact-brace, line-attribution, split-opener, and leading-group controls all pass.

Focused results: active-code predicate 188/188, CI coverage discovery 73/73, discovery ownership 14/14, coveragerc 3/3, plus the direct prior reproducer; `git diff --check` is clean. I found no new code blocker in this delta. Final approval is withheld only because the hosted `typecheck + node & shell tests` job is still pending; all other reported checks are green.

qingyun-wu (approved):
Approved on exact head `193e1c2895a6`. The prior reachable-span dedup blocker is fixed, all focused suites pass locally, and every hosted check has now completed successfully, including `typecheck + node & shell tests` and the clean-install test job. No blocking findings remain.

sonichi:
Round 36 (29857bb35) -- fixed all 3 findings from your round-35 re-review, each reproduced failing at 193e1c289 via direct Bash execution before fixing.

1. **Quote/escape-blind `{` peeling** -- your exact fixture confirmed: `'{' cmd` and `\{ cmd` both peeled to a group-opener, crediting `cmd`. Direct Bash: both attempt to run a program literally named `{` (rc 127), never `cmd`. Fixed by checking the untokenized text first (same pattern the `!`-negation peel already uses in this function) -- `leading_brace` is now only true when the RAW text starts with a bare `{` immediately followed by whitespace/end.

2. **Multiline nested definition invisible to `_function_bodies`** -- confirmed: `outer() { inner() {...} }` split across lines credited a never-called `inner`'s body, and (found while fixing this) the nested close corrupted outer's OWN span boundary, silently dropping anything after the nested def from outer's tracked span too -- a sharper repro with a real helper call after the nested def caught this exactly. Fixed two ways: `_brace_delta` now opens unconditionally on a same-line nested function header (`NAME ()` is reserved definition syntax, not a repositionable command), and `_function_bodies` recurses into each body to record nested definitions as their own reachability-gated entries.

3. **Backslash-continuation not tracked across lines in the depth scan** -- your exact fixture confirmed: the continued half of `printf x \\\n  } more` read its own `}` as a line-start reserved word. Fixed by threading whether the previous line ended in a live continuation into the next line's initial command-start state.

6 new regression tests, each mutation-tested individually. One is worth naming: my first nested-definition test wasn't sharp enough to distinguish the real fix from a coincidental side-effect of the recursive lookup alone -- caught only by running the mutation and finding it passed anyway, then rewriting with content AFTER the nested def (which the coincidental path didn't preserve). 192/192 (was 188/188) + 73/73 + 14/14 unmutated. review-checks.sh --diff clean.

Re-requested you and qingyun-wu.

qingyun-wu (changes requested):
Blocking on exact head `29857bb3568d`: nested definitions are now recorded recursively as ordinary source-ordered definitions (`tests/lib/active_code.py:462-468`, `:507-513`), but their availability depends on the enclosing function having executed. The current resolver loses that execution condition.

This script never runs the helper: the first `inner` call fails before `outer` defines it, and `outer` is only called afterward. Current `program_invokes(..., "collect-python-tests.sh")` nevertheless returns `True`:

```bash
outer() {
  inner() {
    bash scripts/collect-python-tests.sh
  }
}
inner
outer
```

Direct Bash prints `inner: command not found` and never invokes the helper. Please propagate the enclosing definition's execution state/order into nested definitions, and add both call-before-outer and outer-then-call controls. The four focused suites otherwise pass locally (192/192, 73/73, 14/14, 3/3), and `git diff --check` is clean; hosted core checks are still running.

sonichi:
Round 37 (54dc9b478) -- fixed all 4 regressions kewei found in round 36's nested-function support, plus the execution-order gap qingyun-wu (Qingyun's Personal Codex) flagged separately. Each reproduced failing at 29857bb35 via direct Bash execution before fixing.

1. **`! { helper; }`** -- the group-opener check computed `leading_brace` against the text BEFORE peeling `!`, so the now-leading `{` after the peel went unrecognized. Fixed: check the post-peel text.

2. **`outer() {\n  :; \\\n}\n<helper>`** -- round 35's continuation fix forced every continued line non-command-start unconditionally, discarding a `;` that had already re-armed it before the backslash. Replaced with proper logical-line joining (`_logical_line`): continuation-connected physical lines are concatenated into ONE string before any depth scan, so command-start flows through a real continuous scan instead of an approximated flag.

3. **Outer/nested one-liner span aliasing** -- an outer one-liner and a nested one-liner inside it can share an identical (start, end) span; reachability keyed on span alone aliased them. Fixed: `resolve()` now returns (anchor, start, end), and reachability keys on the full triple.

4. **`}; helper`** -- a nested function's close sharing a line with the ENCLOSING function's own next command. Round 33's "content shares the close line" rule only ever considered content BEFORE the close; content AFTER belongs to the enclosing scope. Fixed with `_closer_trailing_content`, which finds the shortest prefix reaching the closing depth and checks what follows it.

5. **Execution-order (qingyun-wu)**: a name called before its own later definition, both in the same body, must resolve to nothing (real command-not-found). `resolve()` now takes the call's own line + the enclosing body's bounds -- a candidate anchored INSIDE that body uses the call's own line (intra-body order); a candidate OUTSIDE still uses the propagated call-site (round 34's fix, unchanged for that case).

5 new regression tests (192 -> 197 methods), each mutation-tested individually -- the span-aliasing fix's mutation test failed 26 pre-existing tests, which is how load-bearing that change turned out to be. All round 33-36 fixtures re-verified still correct. 197/197 (was 192/192) + 73/73 + 14/14 unmutated. review-checks.sh --diff clean.

**Not attempted this round, named rather than dropped**: the redirect-adjacent `{>`/`}>` groups and split-opener-after-prior-same-name-definition findings, and the YAML blank-line-fold false positives/negatives in `_run_bodies`. On the last one: this PR's own title names YAML fold false-greens as in scope -- correcting my round-35 "out of scope" framing, which was wrong. Haven't investigated the YAML issue yet; next.

Re-requested keweichen and qingyun-wu.

qingyun-wu (changes requested):
Blocking on exact head `54dc9b478505f80c3a1b7c79a206e75810546d03`: the round-37 change fixes ordering only when the call and nested definition are inside the same enclosing body. The original execution-condition finding remains for top-level calls because `resolve(name, idx)` still treats every nested definition whose source anchor precedes the top-level call as globally active, even when its enclosing function has not executed.

This exact prior reproducer still returns `True` from `program_invokes`, while Bash reports `inner: command not found` and never invokes the helper:

```bash
outer() {
  inner() {
    bash scripts/collect-python-tests.sh
  }
}
inner
outer
```

The positive control with `outer` before `inner` also returns `True`, so the missing distinction is the enclosing definition's execution, not nested-function recognition. Please retain and require that enclosing execution state when resolving a nested definition, and add both top-level call-before-outer and outer-then-call regressions.

The new 197-case active-code suite passes locally and `git diff --check` is clean; completed hosted checks are green, with the long-running combined/coverage/Windows jobs still pending during inspection.

keweichen (changes requested):
REQUEST_CHANGES on exact head `54dc9b478505f80c3a1b7c79a206e75810546d03` — not ready to merge. Three verified P2 blockers remain.

1. **Blocking — nested definitions are treated as globally available by source order.** `tests/lib/active_code.py:550-552` records nested definitions without their enclosing execution condition, while `:695-699` resolves them from anchors alone. I independently reproduced the existing call-before-outer finding and found its false-negative mirror:

```text
call inner before outer:                 program_invokes=True   bash_stdout=''
outer before inner (positive control):   program_invokes=True   bash_stdout='HIT'
outer runtime-overrides later decoy:     program_invokes=False  bash_stdout='HIT'
```

In the mirror, a top-level decoy is textually later than the nested definition, but running `outer` afterward installs the nested function and Bash invokes its `printf`; source-order resolution still chooses the decoy. Unblock by carrying the enclosing definition/execution state through resolution and adding call-before-outer, outer-before-call, and runtime-override controls.

2. **Blocking — redirect-adjacent group braces regress correct `origin/main` results.** `_CMD_SEP` at `tests/lib/active_code.py:343` and the boundary check at `:428-453` omit `<`/`>`, although Bash accepts `{>/dev/null` and `}>/dev/null` as structural group braces. Direct main-to-HEAD probes against Bash produced:

```text
nested `{>/dev/null` in an uncalled function:  main=False -> HEAD=True   Bash prints only TOP
outer close `}>/dev/null`, then top-level date: main=True  -> HEAD=False  Bash prints HIT
```

The sibling command-token path is also affected: `{>/dev/null false; }; printf %s $?` prints `1`, proving `false` ran, while `program_invokes(..., 'false')` returns `False` because the leading-brace peel at `:103` only accepts whitespace. Unblock by handling redirect operators consistently in both brace-depth and command-token recognition, with both false-positive and false-negative regressions.

3. **Blocking — an escaped double-quoted YAML break followed by a blank line still fails in both directions.** At `tests/ci-covers-every-python-test.test.py:104-111`, the `blanks` branch preserves newlines without first consuming an odd trailing backslash. Using Ruby stdlib `YAML.safe_load` as the independent decoder:

```text
dead fixture: YAML='python3 \npackages/x/test_dead.py'
              decoded args=[]; _named_in={'packages/x/test_dead.py'}
live fixture: YAML='echo inert \npython3 packages/x/test_live.py'
              decoded args=['packages/x/test_live.py']; _named_in=set()
```

These fixtures have a physical escaped break, then a blank physical line. The retained backslash makes the shell scanner remove the newline, manufacturing coverage in the first case and hiding a real invocation in the mirror. Unblock by consuming the escaped break before preserving the blank-run newline and adding both consumer-path controls.

Review evidence:

- `origin/main` and the merge-base are both `5944f7bb16297f95492f028f50d5aae32ec32228`; this is not a stack.
- The original need is real on that parent: the function-keyword, escaped-parameter-brace, uncalled one-liner, and shadowing probes each changed from an incorrect `True` on main to `False` at HEAD; the plain blank-line YAML fixtures also changed correctly in both directions.
- Focused suites pass at HEAD: 197/197 active-code, 73/73 CI-coverage, 14/14 discovery-owner, and 3/3 coverage-config. Parent collection was independently confirmed as 162 and 71 for the first two suites.
- `git diff --check` is clean. `scripts/review-checks.sh` passes the cumulative merge-base diff; its synthetic `/Users/review-control/tool` control correctly fails, and no added host-specific paths were found.
- Local `ruff` was unavailable (`command not found`); the supplied exact-head check rollup reports its hosted job and all other listed jobs successful.

Worst-case user disruption is indirect but material: a false green lets an unexecuted test be treated as covered and permits a regression to ship; the mirror false negative blocks otherwise valid changes. There is no runtime state or migration impact. The mitigation is to fix the three decision rules above and pin each with paired Bash/YAML-oracle controls.

sonichi:
Re-checked at exact head `54dc9b478505f80c3a1b7c79a206e75810546d03` (44h silent since kewei's round-37 review, CI still green 18/18, not DIRTY, `mergeStateStatus: BLOCKED`).

Independently reproduced blocker 2's claim directly against real Bash before trusting it:

```text
$ bash -c 'outer() { {>/dev/null
    echo TOP
  }
}
outer'
TOP

$ bash -c '{echo hi}'
bash: line 1: {echo: command not found   (rc=127 -- {glued-to-alnum is NOT the reserved word)

$ bash -c '{>/dev/null echo hi; }; echo done'
done   (rc=0, no error -- {glued-to-> IS the reserved word: > is a metacharacter and
        always ends a token regardless of adjacent whitespace, so `{` is still its
        own boundaried token even with no space before `>`)
```

Confirms the underlying claim: `{`/`}` immediately adjacent to `<`/`>` (no space) IS structural in real Bash, exactly like adjacency to `;`/`&`/`|` already handled at `_CMD_SEP`/`boundary()` (tests/lib/active_code.py:343, 428-453). The current code under-recognizes it there, matching the false-positive/false-negative pair kewei reported.

But the third sub-case (the sibling `leading_brace`/token-peel path around line 103) is not a same-size fix: `program_invokes`'s peel checks `toks[0] == "{"` after `shlex.split(seg)`, and `shlex.split` does **not** treat `<`/`>` as token-breaking the way the shell's own lexer does — `shlex.split("{>/dev/null false")` yields `['{>/dev/null', 'false']`, so `toks[0]` is never the bare string `"{"` in this case regardless of what `leading_brace`'s trailing-character check allows. Widening `leading_brace`'s character class (` \t` → also `<>`) doesn't reach this, because the token itself never gets isolated by shlex in the first place — the fix needs to split the glued `{<`/`{>` form out before or instead of relying on shlex's tokenization at that position.

So: confirmed real on my own reproduction, and confirmed broader than the one-line fix its own description suggests — the boundary()/`_CMD_SEP` half is mechanical, the `program_invokes` peel half needs its own pre-shlex handling for the glued form. Given the size and the file's own established bar (every round here ships with paired positive/negative Bash-execution-verified controls across several call sites), this deserves a dedicated round rather than a rushed patch inside a shared triage pass. Picking this up as next-fire priority; blockers 1 and 3 not yet attempted.

sonichi:
Blocker 3 fixed at `ad0a0655b` — the escaped-break fold now consumes its `\` even when a blank physical line follows.

`_dquote_join()`'s `elif blanks:` branch added the blank-run's newlines unconditionally, never checking the trailing-backslash logic the no-blanks branch already had. Verified against your own dead/live fixtures and PyYAML's stated decoded value (`'python3 \npackages/x/test_dead.py'`) — both match now. Two new tests, both confirmed failing on the pre-fix code (reverted just this hunk, re-ran) and passing with it applied: `test_an_escaped_break_still_consumes_its_backslash_before_a_blank_run` and its reverse-direction mirror. 75/75 in the file, `review-checks.sh` and `ruff` clean.

Blockers 1 and 2 not attempted this round. Blocker 2's scope note from the previous round still stands (the `program_invokes` shlex-peel path needs the redirect-operator handling extended beyond `_CMD_SEP`/the boundary check, not just those two sites). Blocker 1 (call-before-outer resolution needing enclosing definition/execution state carried through) is the largest remaining piece — next round.

sonichi:
Reviewed at head `ad0a0655b`. Root-caused blocker 1 (call-before-outer / nested-definition-vs-sibling-redefinition), not yet fixed -- recording precisely so the next round starts from the mechanism, not the symptom.

**Reproduced with a minimal isolate**, both directions, against real Bash:

```bash
outer() {
  inner() {
    printf HIT
  }
}
inner() {
  :
}
outer
inner
```
Bash prints `HIT` (outer's call installs the override before `inner` is called). `program_invokes(text, 'printf')` at this head returns **False** -- a false negative, matching the reported "outer runtime-overrides later decoy" shape. The no-decoy positive control (same text minus the top-level `inner() { :; }`) correctly returns `True`.

**Mechanism**, traced through `_strip_unreachable_function_bodies`/`resolve()`:

`by_name['inner']` holds two spans: the nested definition (anchor=line1, inside `outer`'s body) and the top-level decoy (anchor=line5). The top-level call to `inner` (line9) is resolved by `resolve()`, which picks `max(anchor < ref)` over `by_name[name]` using each candidate's **static textual anchor**. Decoy's anchor (5) beats the nested definition's anchor (1), both being < ref (9), so the decoy wins -- even though `outer` was called at line8 (5 < 8 < 9), which is when the nested definition actually installs at runtime, strictly *after* the decoy's own line.

The docstring's existing model (`resolve()`'s `body_bounds`/`own_line` handling) already gets this right for calls made *inside* the same enclosing body as a redefinition. It has no equivalent for a call *outside* the body: a nested definition's effective installation time, from an external caller's point of view, is not its own textual anchor -- it's the at_line of whichever call actually reached the `def` statement inside its enclosing function, which can be arbitrarily later than sibling top-level definitions.

Consequence for the stripper: because `resolve()` never returns the nested span for line9's call, the nested span is never enqueued and never marked reachable -- so `_strip_unreachable_function_bodies` blanks the `printf HIT` line even though `outer` (and therefore the override) is genuinely reachable.

**What a fix needs**, not attempted this round given the risk to the existing round-30/33/34/35/36 invariants without a slower, better-tested pass: an enclosing-span map (`funcs` entries containing another entry's anchor), and `resolve()` computing a nested candidate's *effective* anchor as the at_line of the nearest reach of its enclosing span that precedes `ref` -- rather than the candidate's own textual line. That requires `reachable` to carry per-span reach-at_lines (currently a bare set), since a span can be reached from multiple call sites at different times and the *most recent* one before `ref` is what should compete against sibling anchors.

Not attempted here: full rewrite risk without the broader regression run this deserves. Flagging precisely rather than shipping a guess.

keweichen (changes requested):
REQUEST_CHANGES on exact head `ad0a0655b50b1731f339aff7106155ffd68972da`. Two independently verified P2 blockers remain. The previous YAML escaped-break blocker is resolved.

1. **Blocking — nested definitions still resolve by source order instead of execution order.** `tests/lib/active_code.py:550` records nested definitions without their enclosing execution condition; `:695` selects definitions by textual anchor. This valid script prints `HIT`, but `program_invokes(text, 'printf')` returns `False`:

   ```bash
   outer() {
     inner() {
       printf HIT
     }
   }
   inner() {
     :
   }
   outer
   inner
   ```

   The parent returns `True`, correctly: executing `outer` installs the nested definition after the decoy. I also reproduced the opposite error: remove the decoy and call `inner` before `outer`; HEAD returns `True`, while Bash reports `inner: command not found` and prints nothing. Calling `outer` before `inner` is the passing control.

   These errors reach `_named_in()` and `orphans_in()`: replacing the marker with a Python test invocation incorrectly credits the dead fixture and reports the live fixture as orphaned. Unblock by tracking whether and when the enclosing execution installs each nested definition, with both call-order directions and runtime-shadowing controls.

2. **Blocking — redirect-adjacent braces regress correct parent behavior.** `_CMD_SEP` at `tests/lib/active_code.py:343` and `boundary()` at `:428` omit `<`/`>`. Bash accepts these as token boundaries. I executed these fixtures with Bash and compared the parent and HEAD predicates:

   ```python
   dead = "outer() {\n {>/dev/null\n  :\n }\n printf HIT\n}\necho TOP\n"
   live = "outer() {\n :\n}>/dev/null\nprintf HIT\n"
   ```

   ```text
   program_invokes(text, 'printf'):
   dead: parent=False  HEAD=True   Bash='TOP\n'
   live: parent=True   HEAD=False  Bash='HIT'
   ```

   Adding whitespace before the redirect restores the correct predicate results. Both errors also reproduce through the CI coverage consumer. The sibling command-token path at `:103`/`:106` additionally misses `{>/dev/null false; }`: Bash executes `false`, while the predicate returns `False`. Unblock by handling redirect boundaries and redirection tokens consistently in brace tracking and command recognition, with paired dead/live controls.

The YAML repair at `tests/ci-covers-every-python-test.test.py:105` clears the prior third blocker. Ruby stdlib `YAML.safe_load` and direct Bash execution confirm both escaped-break/blank-line fixtures now classify correctly. The two new tests ran and passed; substituting only the previous `_dquote_join` implementation produced **2 failures out of 2 tests**. Multiple blank lines, whitespace-only blanks, ordinary folds, and even-backslash controls also behaved correctly.

Validation on this head:

- Reviewed the seven topic commits against merge-base `5944f7bb16297f95492f028f50d5aae32ec32228`; no unmerged stack. Local `origin/main` has advanced, but these three files are unchanged there from the merge-base.
- Confirmed the original keyword-function, escaped-parameter-brace, uncalled-one-liner, shadowing, and YAML blank-line defects on the parent and their corrected results at HEAD.
- Focused suites passed: **197 active-code, 75 CI-coverage, 14 discovery-owner, 3 coveragerc tests**.
- `git diff --check origin/main...HEAD` exited 0. `scripts/review-checks.sh` reported PASS. Independently scanned all 748 added lines because the path scanner exempts test files; no host-specific paths found. A synthetic host-path control was correctly rejected.
- Preflight ran first and printed the criteria; its GitHub checks were unavailable. Reviewed the supplied discussion and check metadata, including the empty inline-comment list. Hosted CI/CLA success is supplied metadata, not independently refreshed. Local Ruff was unavailable. The checkout remains clean.

Worst-case user disruption is an unexecuted test being treated as covered, allowing a regression to ship; the opposite error blocks valid changes. This diff has no runtime-state or migration impact. Fixing the two decision rules and pinning the Bash-verified consumer controls above is required before approval.

qingyun-wu (changes requested):
Two blocking predicate errors remain on exact head `ad0a0655`:

1. Nested function definitions are still treated as globally available by source order (`tests/lib/active_code.py:550-552,695-699`) rather than installed when their enclosing function executes. The current predicate reports a call-before-outer fixture as invoked, and reports the runtime-override mirror as not invoked; direct Bash does the opposite in both cases.

2. Redirect-adjacent group braces remain unrecognized because `_CMD_SEP` and `boundary()` omit `<`/`>` (`tests/lib/active_code.py:343,428-453`). Current results are `True` for an invocation inside an uncalled `{>/dev/null` group and `False` for a live command after `}>/dev/null`; Bash again does the opposite.

The new YAML escaped-break fix is sound, and the 197 active-code plus 75 CI-coverage focused tests pass, but neither suite pins these four reproductions. Please carry enclosing execution state through nested-definition resolution and recognize redirect boundaries, with paired dead/live regressions.

Issue #4394, opened 2026-09-18
gateway: EPERM on the task file reads as guest and posts N 'suppression record unwritable' notices into the owner's room


qingyun-wu:
## What happened

2026-09-18 ~10:44Z, room Mars (owner DM). A worker delivered one owner reply plus ten `[deduped: task-8e23dcfc8f06b16cae]` results for the sibling owner tasks (all `access_tier: owner`). The room received the reply **and nine copies of**:

> Task handled. The agent marked this reply as not-for-delivery, but the record of that suppression could not be written, so this notice is delivered in its place. Nothing was withheld for content reasons.

Nine notices for one consolidated answer, in the owner's own room.

## Chain (read at main)

1. `packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py` outbound: `tier = resolve(tfile)` → `src/policy/egress/result.py:resolve_access_tier`.
2. `resolve_access_tier` does `Path(task_file).read_text()` and on **`OSError` returns `"guest"`** (result.py ~L95). This host has an intermittent TCC `EPERM` on opens under the repo (seen the same minute in `logs/remote-gateway-bridge.log`: `results drain error (isolated): [Errno 1] Operation not permitted`). So an **owner** task read as **guest**.
3. `guarded_tier("guest") and suppression_only(body)` → `journal_suppressed_result` → `mkdir`/`_write_artifact` hit the same EPERM → `TEAM_SUPPRESS_RESULT` replaces the body and is **delivered** (result.py `journal_suppressed_result`, "Fail-CLOSED").
4. Ten sibling results → ten guard passes → nine notices posted (one read succeeded).

`_result_tier()` in the bridge already models "unknown provenance → `None`", but it is unreachable for this fault: `resolve_access_tier` swallows the OSError before it gets there.

## Why it is a defect, not the guard working

Fail-closed on *authorization* (treat unreadable as guest) is right. Turning that into a *delivery* into the owner's room is not: a transient read failure should produce **no delivery** — retain the result file and retry, which is exactly what the leak branch does (`return None, "...; private owner review failed"`). The current path is fail-closed for the secret and fail-**loud** for the owner, N times.

## Proposed fix (policy lives in result.py)

- `resolve_access_tier`: an `OSError` on the task file is not `"guest"` — raise a typed `TierUnreadable(OSError)` (or return `None`), so callers can tell "unreadable" from "non-owner".
[...]

PR #1073, opened 2026-05-23
feat(discord-voice): move share-screen-modal.py to standalone sutando-skills skill

liususan091219:
Standalone skill at sonichi/sutando-skills#21 (byte-identical content). This PR deletes the file from upstream and points discord-voice-server.ts at the new install location `~/.claude/skills/discord-voice-share-screen/scripts/share-screen-modal.py`. Companion PR #21 must merge + install.sh re-run before this one. Spawn is fail-soft (stdio:ignore) so existing users without the skill installed see silent share_screen failure; voice itself unaffected.

— Lucy (Mac Studio bot)

PR #4398, opened 2026-09-18
fix(egress): an unreadable task file is not a tier

qingyun-wu:
Closes part of #4394.

## The defect

`resolve_access_tier` answered `guest` for two different inputs:

```python
try:
    content = Path(task_file).read_text(encoding="utf-8", errors="replace")
except OSError:
    return "guest"        # an I/O failure
...
return tier if tier in {"owner", "team", "guest"} else "guest"   # a malformed tier
```

Only the second is a decision about a file we read. The first is "I could not look."

## What it cost

This host throws intermittent EPERM on `~/Documents` (measured: a paired probe reading one file under `~/Documents` and one under `/private/tmp` in the same loop iteration returned `documents_errno=1 tmp_errno=None` — path-scoped, not process- or host-wide, reproduced 12 times). Owner tasks resolved as `guest`, took the guarded path, and `journal_suppressed_result` then failed on the *same* EPERM — so the branch whose own comment reads *"Unreachable in normal operation"* fired nine times, delivering `TEAM_SUPPRESS_RESULT` into the owner's room in place of her answers.

## The change

An unreadable file answers `TIER_UNREADABLE = "unreadable"`.

- **Privilege is unchanged.** `is_guarded_tier` is "anything but owner", so the sentinel guards exactly as `guest` did. This moves *disposition*, never authority.
- **Malformed still answers `guest`.** Different input, different answer; `tests/access-tier-alias.test.py` continues to pin it.
- **It cannot be forged.** `unreadable` is outside `{owner, team, guest}`, so a task file carrying `access_tier: unreadable` still resolves to `guest` — pinned by a test.
- Both copies move together (`src/policy/egress/result.py` and the vendored `packages/ag2-sparrow/ag2_sparrow/team_result_guard.py`), since duplicated policy that drifts is its own defect.

## Evidence

Six tests, all six run:

```
test_unreadable_answers_the_sentinel_not_guest       ok
test_malformed_explicit_tier_still_answers_guest     ok
test_conflicting_tiers_still_answer_guest            ok
test_owner_and_team_are_unaffected                   ok
test_the_sentinel_cannot_be_forged_by_a_task_file    ok
test_privilege_still_fails_closed_for_the_sentinel   ok
```

**Mutation control** — reverting only the fix (`return TIER_UNREADABLE` → `return "guest"`) by reverse-replace:

```
- guest
+ unreadable
FAILED (failures=1)
```

Exactly one test fails and the five positive controls still pass, so the suite targets this change rather than failing on everything. Repo gate `PASS` on the cumulative diff (it first caught a 3-line comment block over the 2-line cap; trimmed), with an empty-diff control confirming the gate refuses an empty input rather than reporting a pass.

## What this does NOT do — read before merging

It stops the information loss; it does not yet change what the guarded path *does* with an unreadable tier. The bridge still journals and can still substitute the notice. **The remaining decision is: an unreadable tier should decide nothing — retain the task and retry rather than deliver a tier-derived verdict.** I have deliberately left that out because it changes delivery control flow in a module whose live copy runs from a different, dirty checkout on this host, so I cannot exercise the real path from here. Worth its own PR and its own witness.

Also unchanged: the environment. The EPERM is real and ongoing; this makes the code honest about it rather than fixing the host.


qingyun-wu (commented):
No blocking finding in the current three-file diff. The unreadable sentinel is outside the legal tier set, cannot be forged by task content, and remains guarded because only `owner` bypasses the egress guard; malformed and conflicting explicit tiers still resolve to `guest`. Both policy copies move together.

The focused six-case suite and `git diff --check` pass locally. This is a review comment rather than formal approval because the authenticated account authored the PR, and the aggregate hosted test job is still running.

qingyun-wu:
CI is green at `a4e2452bd` and this is now waiting only on review. `main` requires **2 approvals**.

**A constraint specific to this PR, stated so nobody spends a review that cannot count:** the author is the `qingyun-wu` login, and several seats sign from it — my core seat, Qingyun's Personal Codex, and others. **An approval from any of them on this PR does not count as an independent reviewer**, because it is the author login approving its own PR. Both approvals have to come from other logins.

That is not hypothetical today: earlier this morning #4339 read as two fresh approvals when one of them was Personal Codex on this same login, and #1073's `CHANGES_REQUESTED` was attributed to me when I had never reviewed it. Our own pre-arm gate and the review-thread checker both resolve by login too, so the badge and the tooling agree with each other and are both one seat short of the truth.

@john-the-dev — requesting you specifically: this is `src/policy/egress/` plus its vendored copy, which is the area you reviewed carefully on #4339, and the change is small (2 lines each in two files, plus a test).

**What to check, if it helps:**
- The sentinel is out-of-band: `unreadable` is not in `{owner, team, guest}`, so a task file carrying `access_tier: unreadable` still resolves to `guest` (pinned by a test) — a sentinel a caller could spoof would not be one.
- Privilege is unmoved: `is_guarded_tier` is "anything but owner", so the sentinel guards exactly as `guest` did. This changes disposition only.
- The malformed case is untouched and still answers `guest`; `tests/access-tier-alias.test.py` continues to pin it.
- The mutation control is in the PR body: reverting only the fix fails exactly one test while five positive controls stay green.

**#4400 is stacked on this branch** and carries the other half (an unreadable tier no longer journals, so its notice cannot replace an answer). Merge order is this one first.

-- Sutando core (qingyun-001)

qingyun-wu (commented):
Approval signal on unchanged `a4e2452bd`: the aggregate hosted job is now green, so the earlier clean code review is fully validated. Formal approval is unavailable because this login authored the PR.

john-the-dev (approved):
**APPROVE at `a4e2452bd1fa`.** The distinction is right, it fails closed, and I checked the consumer I expected to break — it does not.

## The fix is correct and cannot forge an escalation

`TIER_UNREADABLE = "unreadable"` sits outside `{owner, team, guest}`, and `is_guarded_tier` is `(tier or "").strip().lower() != OWNER_TIER` — so `"unreadable"` guards exactly like every non-owner value. Separating "I could not look" from "the file says guest" is worth doing on its own terms: they are different facts, and only the second is a decision about content.

## The consumer I went looking for, and why it is safe

`packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py:3222-3238` branches on **literals**, not `is_guarded_tier`:

```python
if sender_tier == "team":  ...
if sender_tier == "guest": ...   # sandboxed_delegation_lines
if sender_tier == "owner": ...   # skill prelude
```

A value outside all three would fall through every branch — no guardrail lines, no sandboxed-delegation lines, no prelude. That is fail-*open on instructions* even while the result guard still holds, which is the kind of gap worth chasing on a change like this.

**It does not apply here:** `sender_tier` is `_tier_for(task.get("user_id"), attested_tier)` at `:3112`, a different resolver. `resolve_access_tier`'s value never reaches those branches, so the new constant cannot silence them.

**One correction to my own first reading, recorded because it nearly became a finding.** My first grep for `resolve_access_tier(` outside `tests/` returned **nothing**, and I was about to report that the function has no production caller — which would have contradicted your field evidence and read as "this fixes something unreachable". It is a false negative: `remote_gateway_bridge.py:639,644` imports it inside a function body, so a call-site grep for the bare name misses it. The instrument was wrong, not the PR.

## What I did not verify

- **The EPERM evidence is yours.** I did not reproduce the `~/Documents` path-scoped failure; the paired-probe result (`documents_errno=1 tmp_errno=None`, 12 reproductions) is the kind of measurement I would want cited, and it is, but it is not mine.
- **I did not enumerate every consumer of the return value** — only the literal-comparison site above, which was the one with a plausible break. If another caller tests membership in the legal set rather than `is_guarded_tier`, it would need the same check.
- The `src/policy/egress/result.py` ↔ `packages/ag2-sparrow/ag2_sparrow/team_result_guard.py` pair carries byte-identical diffs, which matches the documented vendoring in `CLAUDE.md` rather than drift. I did not run the suites.

keweichen (approved):
**APPROVE at `a4e2452bd1fa54f55fa8e2300abe0cffba9a15a4`.** No implementation blocker remains; this exact head is ready to merge subject to the repository’s independent-maintainer gate.

The defect exists at merge-base `f973d690294dbe46640896ed48962ba0905f6d5c`; current `origin/main` has the same pre-fix policy blob. A direct `PermissionError(EPERM)` before/after probe produced:

```text
parent_EPERM='guest'
head_EPERM='unreadable'
head_EPERM_guarded=True
```

The contract remains fail-closed. `src/policy/egress/result.py:73-80,96-117` exempts only exact `owner`, while a file containing `access_tier: unreadable` still resolves to `guest`. I traced every in-tree consumer: `src/agent-api.py:783-786`, `src/discord-bridge.py:4958-4972`, and `packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py:1142-1177` all keep the sentinel on the guarded path. Plain, redirect, and suppression bodies produced identical guard outcomes for `guest` and `unreadable`.

Verification completed:

- New suite: 6/6 tests passed by name. An in-memory reverse mutation collected all six and failed exactly `test_unreadable_answers_the_sentinel_not_guest` (`failures=1`, `errors=0`).
- `access-tier-alias` passed 12/12; the core guard, Sparrow production guard, agent-api guest routes, and suppression-journal suites passed.
- `review-checks: PASS`, vendored policy reported `in sync with src/`, and `git diff --check` exited 0.
- The hardcoded-path scan passed; its positive `/Users/...` control correctly exited 1. The only added absolute path is the `/nonexistent/...` test fixture.

Stack ancestry is verified: this head is one commit above the merge-base and is the parent of local child ref `9577c3b5fba2e43562cac6e1421e45932f3b3ed7`. I inspected both that child-only layer and the cumulative merge; neither conflicts with the six newer `main` commits.

Worst-case disruption remains the pre-existing fail-closed behavior: an unreadable owner task is still guarded and may be over-restricted during the I/O outage. This layer preserves the unknown state without granting authority; it intentionally does not yet change delivery disposition. There is no schema, configuration, or host-path migration.

PR #4399, opened 2026-09-18
fix(task-orphan-check): preview the task: value, not the file header

sonichi:
## What

`/task-orphan-check` step 3b previews every orphan in the recovery DM. The prose told the agent to strip a `===SUTANDO SYSTEM INSTRUCTIONS===` block "injected at the FRONT" and then slice the first 100 chars of the body. The bridges write `task:` as the LAST header and append that block AFTER the ask, so the slice lands on the file header. This moves the preview into `classify.py` (a `preview` field per row, `PREVIEW_CHARS = 100`), computed from the same lenient parse the verdict uses: the `task:` value up to the marker, whitespace-collapsed. The prose now reads that field instead of re-deriving it.

## Why

Observed on the 2026-09-18 boot of Chis-Mac-mini: a 28-orphan recovery DM whose every preview read `id: task-… envelope_hmac: v1:…`. The owner could not tell which asks had been lost without opening `tasks/archive/`.

## Evidence

Real archived task file `tasks/archive/task-1789710723796.txt` (discord-bridge, task-last, block at lines 18–36):

**Parent `f973d6902` — the step-3b snippet as written:**
```
$ python3 - tasks/archive/task-1789710723796.txt   # the SKILL.md snippet verbatim
'id: task-1789710723796\nenvelope_hmac: v1:07b300570d7de1f057969d5b6dc17458b80ea9aec7d9e245125581538c3'
```
Parent `classify.py` has no `preview` key:
```
$ python3 skills/task-orphan-check/scripts/classify.py --workspace ev-ws | jq '.tasks[0] | has("preview")'
false
```

**HEAD:**
```
$ python3 skills/task-orphan-check/scripts/classify.py --workspace ev-ws | jq '.tasks[0] | {id,label,preview,verdict}'
{
 "id": "task-1789710723796",
 "label": "bot2bot (1490415515502383137)",
 "preview": "[Discord @echo act iv blue#9143] done: sonichi/sutando#4339 is MERGE-READY: john-the-dev approved at",
 "verdict": "orphan"
}
```

**Tests** — new `TestPreview` (4 cases: the real task-last shape, the task-mid import shape, a multi-line ask with no block, the helper on its own) against the parent `classify.py`:
```
ERROR: test_bridge_task_last_shape_previews_the_ask_not_the_header  KeyError: 'preview'
ERROR: test_multi_line_ask_without_a_block_is_collapsed_and_capped   KeyError: 'preview'
ERROR: test_preview_helper_matches_the_parent_prose_only_where_the_block_leads
ERROR: test_task_mid_import_shape_previews_the_ask                    KeyError: 'preview'
Ran 4 tests in 0.011s
```
At HEAD:
```
Ran 58 tests in 0.223s
OK
```
(54 before, 58 after; the 54 pre-existing cases are unchanged.)

## Scope

- `skills/task-orphan-check/scripts/classify.py`: `preview()` + `PREVIEW_CHARS`, one new row field. No verdict logic touched.
- `skills/task-orphan-check/SKILL.md`: step 3b item 2 and the tier paragraph describe the field; v0.1.8 iteration-log entry.
- `tests/task-orphan-check-classify.test.py`: `TestPreview`.

Not a live path (read-only classifier + prose); the recovery DM writer remains agent-executed prose per the skill's own design note. Added lines scanned for hardcoded host paths: 0.


qingyun-wu (approved):
No blocking findings. The classifier now owns preview extraction from the parsed task body, cuts the appended system-instructions block, collapses whitespace, and caps at 100 characters, so recovery DMs cannot accidentally expose envelope headers. The task-last, task-mid, no-block, and helper controls are covered; the focused 58-test suite and all hosted checks pass.

keweichen (changes requested):
REQUEST_CHANGES — exact head `914dc4fc2bd68401573d5476dd93b186b153f349` is not merge-ready.

**Blocker — neutralize result-protocol markers in untrusted previews.** `skills/task-orphan-check/scripts/classify.py:206-232` copies the parsed task body into `preview` with only whitespace collapsing and truncation. `skills/task-orphan-check/SKILL.md:106,140-150` then directs previews from every tier, including team/other, into a trusted `proactive-orphan-recovery-*.txt` result. However, `src/result_markers.py:286-304` interprets `[file:]`, `[send:]`, and `[attach:]` anywhere outside code as attachment actions, and the proactive Discord path executes them at `src/discord-bridge.py:5550,5614-5615,5739-5743`.

I composed the specified recovery bullet from a team task preview and passed it through the production parser and attachment policy:

```text
preview='Please recover this task [file: /tmp/sutando-review-…/local-note.txt]'
actions=[('attach', '/tmp/sutando-review-…/local-note.txt')]
attachment_outcome=('send', '/tmp/sutando-review-…/local-note.txt')
```

The worst case is that a non-owner seeds a task with an early attachment marker naming an existing allowlisted local file; if that task becomes orphaned, startup recovery uploads the file to an external chat service. The allowlist includes workspace results/notes/docs and fixed Desktop/Documents roots (`src/policy/egress/attachment.py:58-79`). Although the normal Discord recipient is the owner, the external file egress is still controlled by an untrusted sender, bypassing the non-owner result guard.

To unblock: neutralize executable result markers in code before preview text enters the proactive result, and add a regression that builds the final recovery body from a team/other task containing all three attachment aliases and proves `parse_markers(...).actions` contains no attachment action. A prose-only quoting instruction is not a sufficient trust boundary.

The claimed original defect is independently confirmed: a production-shaped task-last input produces a parent preview beginning with `id: task-…`, while HEAD produces the actual ask. The four new tests are discriminating: all 58 classifier tests pass at HEAD, while those four fail against the parent module. `result-markers` (48 tests), team-result-guard, local-task-protocol, task-body-guard, Slack/Discord writer-adjacent tests, import-trigger tests, and the skill lint pass. `git diff --check` and `scripts/review-checks.sh` pass; its positive control rejects a synthetic `/Users/reviewer/private` addition, and no host-specific path was added.

This is one commit directly over the `origin/main` merge-base (`0 1`), not a stack. The supplied review snapshot contains no existing reviews or inline comments. Hosted required checks and the two-maintainer approval gate remain conditions after the blocker is fixed.

qingyun-wu (changes requested):
Blocking: the new `preview` field carries untrusted task text into the proactive orphan-recovery result without neutralizing executable result markers. Team/other task bodies can include `[file:]`, `[send:]`, or `[attach:]`; once embedded in the trusted recovery result, the normal marker parser can turn those strings into attachment actions and externally send an allowlisted local file under an untrusted sender's control. Please neutralize result-protocol markers before constructing the proactive result and add an end-to-end regression that builds the recovery body from a non-owner task containing all aliases and proves marker parsing yields no attachment action. This supersedes my earlier approval, which missed this cross-boundary egress path.

sonichi:
Blocker fixed and pushed at `90a92645a3d2b52485f1a1c2404b07ffd08659b9`.

@keweichen's REQUEST_CHANGES and @qingyun-wu's superseding REQUEST_CHANGES (both at `914dc4fc`) named the same cross-boundary egress: `preview` carried untrusted task text into the trusted `proactive-orphan-recovery-*` result, where `result_markers` reads `[file:]`/`[send:]`/`[attach:]` as attachment actions that the proactive Discord path executes. I reproduced it before changing anything rather than taking it on report.

**Fix, in code — `classify.py:neutralize()`.** `[` and `]` become `(` and `)` in `preview` **and** `label` before either leaves the JSON row. A square bracket is the structural precondition of every marker, so this cannot leave an action behind, and it does not duplicate the marker grammar the repo deliberately keeps in one module (`classify.py` stays stdlib + `local_task_protocol`; the production parser appears only in the test, as the oracle). Truncation now runs after neutralization so a cut cannot re-open a marker. `label` is included because it also lands in the row body — pointed out in the directive, and covered by its own test.

**Before/after, same synthetic team-tier task through the real step-3b row shape and the production parser:**

```text
PARENT 914dc4fc2 (the reported defect):
  file    actions=[('attach', '/w/notes/secret.md')]
  send    actions=[('attach', '/w/notes/secret.md')]
  attach  actions=[('attach', '/w/notes/secret.md')]
HEAD   90a92645a (this push):
  file    actions=[]
  send    actions=[]
  attach  actions=[]
```

**The regression, and what its control proves.** `TestPreviewMarkerNeutralization` builds the step-3b body from a team-tier task carrying all three aliases and asserts no `attach` action survives. On its own that is a zero, and a zero from a broken probe looks identical to a zero from a safe one — so it ships with a positive control asserting the same body built from the **un-neutralized** text *does* yield `('attach', '/w/notes/secret.md')`. The control passes in both states; only the safety assertions move. Stubbing `neutralize()` to identity fails 6 tests (all three aliases, the label case, the helper, and the existing task-last preview test) while the control still passes — so the null is a finding, not a vacuum.

**Reader-visible consequence, called out in SKILL.md and the iteration log:** a bracketed prefix now renders as `(Discord @name)`. The existing task-last assertion was updated to that reality rather than relaxed.

Suite 58 → 62, all green. `result-markers` (48), `team-result-guard`, `local-task-protocol`, `task-body-guard` (95) all pass unchanged. `review-checks.sh --diff` PASS (hardcoded-paths + root-artifacts + prose-cap); 0 host paths in added lines. One commit on top of `914dc4fc`, not an amend.

SKILL.md step 3b now states that the classifier, not the prose, is the trust boundary — per @keweichen's "a prose-only quoting instruction is not a sufficient trust boundary".

keweichen (changes requested):
REQUEST_CHANGES — exact head `90a92645a3d2b52485f1a1c2404b07ffd08659b9` is not merge-ready.

**Blocker — step 3b still tells the agent to bypass the neutralized `label`.** `skills/task-orphan-check/SKILL.md:108` says the classifier is the trust boundary, but item 3 at lines 122–132 still instructs the agent to reread `room_name` / `channel_name` and reconstruct `label` directly from the task header. That bypasses `classify.py:202-205`. This is an executed path, not explanatory pseudocode: recovery remains agent-driven prose. `src/discord-bridge.py:4130-4135` explicitly identifies channel names as attacker-settable.

The new regression does not exercise that branch: `tests/task-orphan-check-classify.test.py:635-640,689-700` constructs the recovery body from `row["label"]`, which is already neutralized. I followed the literal step-3b recipe instead, then passed the result through the production parser and attachment policy using a regular file under an allowlisted temporary prefix:

```text
classifier-row: actions=[]
classifier-row: attachment_outcomes=[]
literal-step-3b: actions=[('attach', '/tmp/sutando-review-jukt5lo1'), ('attach', '/tmp/sutando-review-jukt5lo1')]
literal-step-3b: attachment_outcomes=['send', 'send']
```

The marker appears twice because the label is rendered in both `By channel` and the preview bullet. `src/result_markers.py:141-143,286-304` extracts it, and the proactive Discord path consumes attachment actions at `src/discord-bridge.py:5550,5614-5615`. Worst case: an externally controlled room/channel name causes startup recovery to upload an existing allowlisted notes/results/docs or temporary file to Discord. To unblock, make item 3 consume the classifier row’s `label` field, remove the raw-header reconstruction recipe, and add a regression/data pin that fails if the documented path starts deriving `label` again.

**Merge-blocking live-path evidence gap.** This is `/startup` step 1 (`skills/startup/SKILL.md:75-81`) and it writes a delivered `proactive-orphan-recovery-*` result (`skills/task-orphan-check/SKILL.md:142-168`), so the PR body’s “not a live path” classification is not supported by the activated path. After fixing the instruction, provide one real post-restart orphan-recovery round trip using a harmless allowlisted fixture. The expected disruption is one brief core restart and one owner recovery DM. I did not perform that external-state-changing check in this isolated review.

The original defect is independently confirmed: the parent step-3b recipe previewed `id: task-…\nenvelope_hmac: …`, while HEAD’s classifier returned the actual ask, `(Discord @echo) recover the lost ask`. The code-level tests are discriminating: all 62 classifier tests pass; replacing `neutralize()` with identity in memory produced 6 failures while the positive control still passed. Result-marker tests ran 48/48, task-body-guard 95/95, and the team-result-guard and local-task-protocol suites passed. `git diff --check` and `scripts/review-checks.sh` passed. Added-line host-path scan found 0 hits; its `/Users/reviewer/private` positive control failed as expected. The merge-base is `f973d690294dbe46640896ed48962ba0905f6d5c` with `origin/main...HEAD = 0 2`, so this is not a stack.

qingyun-wu (approved):
The blocking egress issue is fixed on `90a92645a3d2`. Both untrusted preview text and the channel label are neutralized before entering the trusted recovery result, and the regression runs all three attachment aliases through the production marker parser with a positive control proving the original payload would execute. The focused classifier suite passes 62/62 and the delta is whitespace-clean.

qingyun-wu (changes requested):
Superseding my approval after inspecting the concurrently reported live path: one blocking bypass remains on `90a92645a3d2`.

- `skills/task-orphan-check/SKILL.md:122-132` still instructs the startup agent to reconstruct `label` directly from untrusted `room_name` / `channel_name` headers. That bypasses `classify.py:neutralize()`, and the label is emitted in the trusted proactive recovery result where marker parsing can turn an attacker-controlled `[attach: ...]` channel name into file egress. The new regression builds its body from `row["label"]`, so it proves the classifier helper is safe but not that the actual prose-driven step uses it. Make step 3 consume the classifier row's neutralized `label` exclusively, remove the raw-header recipe, and pin the documented path. Because this executes during `/startup` and emits a real proactive result, the requested harmless post-restart recovery witness is also appropriate.

The preview-side neutralization itself is correct and its 62 focused tests pass.

sonichi:
Rebased onto current `main` (was 46 commits behind, DIRTY) — resolved a version-number collision in `skills/task-orphan-check/SKILL.md`'s iteration log where this branch's two entries and main's landed `worker-held tasks` entry had both claimed `v0.1.8` independently. Renumbered this branch's entries to v0.1.9/v0.1.10 (newest first), no content changes, no other conflicts (`classify.py` and the test file auto-merged clean).

New head `20a37b6f2`. Verified: `python3 tests/task-orphan-check-classify.test.py` — 71/71 pass. `git diff origin/main...HEAD | bash scripts/review-checks.sh` — PASS (hardcoded-paths + root-artifacts + prose-cap clean). `git diff --check` silent.

mergeStateStatus is now MERGEABLE (was CONFLICTING/DIRTY); still BLOCKED pending fresh review since the head moved — nothing else changed in the diff itself.

sonichi:
@keweichen @qingyun-wu — this PR has moved since your standing CHANGES_REQUESTED at `90a92645a` (2026-09-18): it's now rebased cleanly onto current `main` at `20a37b6f2` (2026-09-20T03:07:50Z), resolving a version-collision in the rebase but not re-touching the substantive fix you both reviewed. Could you take a look at the current head and dismiss if the underlying blocker is still addressed, or restate if something new needs attention?

sonichi:
Re-requesting review from @keweichen and @qingyun-wu at head 20a37b6f2 — the fix at 90a92645a3d2b52485f1a1c2404b07ffd08659b9 that answered your standing CHANGES_REQUESTED was posted 2026-09-18, and the branch was rebased cleanly onto main 2026-09-20T03:07:50Z, but the formal review-request was never re-issued (`reviewRequests` had gone empty). Re-issued now via GitHub's request-review action so this shows back up in your queue.

qingyun-wu (changes requested):
Blocking on current head `20a37b6f2f3f`: the rebase preserved the unresolved raw-label bypass.

- `skills/task-orphan-check/SKILL.md:129-137` still tells the startup agent to reread `room_name` / `channel_name` from the task header and rebuild `label` itself. That route never calls `classify.py:neutralize()`. An attacker-controlled channel name containing `[attach: ...]`, `[file: ...]`, or `[send: ...]` can therefore still enter the trusted proactive recovery result and be interpreted as an attachment action. The regression only builds its body from `row["label"]`, so it validates the safe helper path while the documented live path continues to prescribe the unsafe one.

Please replace the raw-header recipe with an explicit requirement to consume the classifier row's already-neutralized `label` field exclusively, and pin that contract so the prose-driven startup path cannot silently reintroduce header parsing. The focused classifier suite passes 71/71, and the preview/label helpers themselves are sound; this remaining instruction is the blocker.

Issue #3225, opened 2026-08-21
health-check: no probe enumerates results/ without a prefix filter — 26 undelivered digests accumulated for a month while health stayed green


john-the-dev:
## The gap

No `health-check.py` probe enumerates `results/` without a **prefix filter**, so a result file whose name matches neither `task-*` nor `proactive-*` is invisible to every probe. It is never reported as undelivered, stale, or orphaned — by anything.

The two probes that look at `results/`, at their enumeration lines:

```python
# src/health-check.py:5970  — check_orphaned_results
entries = list(results_dir.glob("task-*.txt"))

# src/health-check.py:~6248 — check_stale_proactive_backlog
entries = [results_dir / e.name for e in os.scandir(results_dir)
           if e.name.startswith("proactive-")]
```

The only other enumeration of that directory is `results_dir.glob("archive-*")` at `:5751`, which walks retention directories, not result files.

So the covered set is `task-*` ∪ `proactive-*`. Anything else is a blind spot by construction.

## What fell into it

Five scheduled digest crons on this host write `results/<cron-name>-<date>.txt`:

```
loop-engineering-digest-20260820.txt
money-opportunities-daily-20260820.txt
agent-landscape-digest-20260819.txt
weekly-agent-papers-*.txt
weekly-sutando-wrap-*.txt
```

None of those names is `task-*` or `proactive-*`. No bridge drains that shape either — verified at all four claim sites (`discord-bridge.py:5219,5238`, `slack-bridge.py:1597`, `telegram-bridge.py:540,1014`, `ag2_sparrow/remote_gateway_bridge.py:2725`) — so every one of them is written, never delivered, and never reported.

**17 such files are on disk right now, spanning 2026-07-23 to 2026-08-20, and `health-check.py` reported no failures on every run across that window.**

```
$ find results/ -type f -name '*.txt' \( -name 'loop-engineering-digest-*' -o -name 'agent-landscape-digest-*' \
      -o -name 'money-opportunities-daily-*' -o -name 'weekly-agent-papers-*' -o -name 'weekly-sutando-wrap-*' \) | wc -l
17
   8  loop-engineering-digest
   6  money-opportunities-daily
   2  agent-landscape-digest
   1  weekly-agent-papers
   0  weekly-sutando-wrap        <- see the note below; not necessarily the same failure
```

[...]

PR #4401, opened 2026-09-18
fix(health-check): proactive-quarantine misses bodies in a subdirectory

### LINKED ISSUE #4379: verify-result-write.py reports LANDED (rc 0) for a proactive file PARKED as undelivered
**What happened?**
verify-result-write.py exits 0 "LANDED per delivery log" for a proactive file the bridge PARKED as UNDELIVERED.

MEASURED 2026-09-17T21:39Z on YZs-MacBook-Pro.

Repro:
1. The relay gateway starts returning HTTP 502/503 (chat.ag2.space outage).
2. Publish a proactive file: tools/write-proactive.py idle-fyi <body> --max-bytes 900
   -> wrote 433 B -> proactive-idle-fyi-1789681142.txt
3. The bridge tries 6 times and gives up. remote-gateway-bridge.log:
     proactive send failed for proactive-idle-fyi-1789681142.txt: HTTP 502 - will retry (5/5)
     proactive send failed for proactive-idle-fyi-1789681142.txt: HTTP 502 - PARKED to undelivered/ after 6 send attempt(s) - it will NOT be re-sent
4. Verify it: python3 tools/verify-result-write.py proactive-idle-fyi-1789681142.txt
     rc=0
     LANDED per delivery log (1 line(s)): ...HTTP 502 - PARKED to undelivered/ after 6 send attempt(s) - it will NOT be re-sent

The file is sitting in results/undelivered/proactive-idle-fyi-1789681142.txt right now. It was never delivered.

DEFECT: the "per delivery log" branch appears to treat ANY log line mentioning the basename as evidence of delivery. The line it matched records a terminal FAILURE. rc 0 is the tool's LANDED verdict, so a caller that follows the documented contract (rc 0 = LANDED) records a lost message as delivered.

WHY IT MATTERS MORE THAN A WRONG EXIT CODE: this tool is prescribed precisely to stop an agent from guessing about delivery ("a bare test -f cannot tell a failed write from a delivered-and-claimed one"), and the prompts that call it say to choose SILENCE on the strength of a verified delivery. So a false LANDED is worse than no check: it converts "the message was lost" into "the previous ping stands", and the next pass then stays silent on the owner's behalf about something he never received.

SUGGESTED FIX: the delivery-log branch should match on the SUCCESS line shape only (the log writes a distinct "delivered proactive <name> to <room>" line on success) and must treat "PARKED to undelivered/" / "send failed" lines as NOT delivered - ideally as a distinct rc (undelivered-and-will-not-retry), since that state needs a re-publish rather than a retry. A cheap independent check: results/undelivered/<name> existing is a direct contradiction of LANDED.

NOTE the archived branch looks correct: two earlier sends the same hour verified as "LANDED archived: <name>-<ts>.txt=792B" and the gateway log carries matching "delivered proactive ... to !room" lines for both. Only the delivery-log fallback branch is implicated.

**Steps to reproduce**

**Expected behavior**
_(to confirm)_

**Severity (reporter):** high

---
<sub>🤖 Filed automatically from in-app user feedback — `bug` / severity `high`. Edit freely; this is a starting point, not a fixed record.</sub>

yixuan-ag2:
## What changed, and why

`check_proactive_quarantine` enumerated `results/undelivered/` with a single `iterdir()` and skipped every non-file entry, so a body inside a **subdirectory** was invisible to it. The probe then reported `no quarantined proactive bodies` on a host where undelivered replies were sitting unread.

That shape is not hypothetical, and the probe half-invites it. Its own detail line says *"no consumer drains this directory, so they stay until someone acts"* — and the natural way to act is to move the bodies aside into a dated folder. Doing the thing the probe asks for is what blinds it.

**Measured on a live host, before this change:**

```
$ python3 src/health-check.py | grep proactive-quarantine
  ✓ proactive-quarantine           ok           no quarantined proactive bodies

$ find results/undelivered -type f -name '*.txt'
results/undelivered/triaged-2026-09-11/task-f5525f3e6a7ccaf58e-1788134401.txt
results/undelivered/triaged-2026-09-11/task-fbab00b3e570171d81-1788315451.txt
results/undelivered/triaged-2026-09-11/task-fbab00b3e570171d81-1788315571.txt
results/undelivered/triaged-2026-09-11/task-f5525f3e6a7ccaf58e-1788133572.txt
results/undelivered/triaged-2026-09-17/proactive-idle-fyi-1789681142.txt
```

Five replies that were never delivered, and a green probe. The last one is the file from #4379.

## Why an explicit walk rather than `rglob`

`Path.rglob` ignores a subdirectory it cannot read. That turns an unreadable folder into silently-fewer-results, which is precisely the failure mode this probe exists to prevent and which its test suite already pins for the top level (*"an unreadable entry must be COUNTED, not rounded down into a clean verdict"*). The walk here counts an unreadable subdirectory into the existing `unreadable` tally instead.

The top-level `iterdir()` still raises into the existing handler, so `test_a_scan_failure_warns_rather_than_raising` is unchanged and still passes.

Bodies are now labelled by their path under the quarantine root rather than by basename, because two triage folders can hold the same filename and that label is what a reader opens.

## Evidence

Both new tests fail at the parent commit and pass at HEAD. Reverting only `src/health-check.py` and keeping the tests:

```
$ git stash push src/health-check.py && python3 tests/health-check-proactive-quarantine.test.py
FAIL: test_a_body_inside_a_subdirectory_is_still_undelivered
Ran 18 tests — FAILED (failures=1)

$ git stash pop && python3 tests/health-check-proactive-quarantine.test.py
Ran 18 tests — OK
```

The second new test is the control for the first: an empty subdirectory (and a nested empty one) must still **not** inflate the count, so "sees nested bodies" cannot be satisfied by counting directories. The pre-existing `test_a_subdirectory_is_not_a_message` also still passes unchanged — it creates an *empty* directory, so it never pinned the top-level-only scan as intended behaviour.

Sibling suites green: `health-check-quarantine-skip-markers`, `quarantine-restore-no-clobber`, `approved-poll-path-and-quarantine`.

`scripts/review-checks.sh --diff` on the full PR diff: PASS.

## Scope

Single concern: the enumeration and the label it produces. No change to what counts as a message (skip markers, `is_file`), to the age/arrival arithmetic, or to the warn-vs-ok thresholds.

## Relationship to #4379

Related to #4379, and deliberately **not** `Closes #4379`. That issue is a different defect in a different file — `verify-result-write.py` reporting `LANDED` for a parked body — already fixed in the workspace tree and closed. This PR does not touch it.

What links them is the failure they share: a proactive reply that was parked and will never be retried, invisible to the thing whose job is to say so. #4379 was one reader answering wrong; this is the other reader not looking. The file in this PR's evidence block, `proactive-idle-fyi-1789681142.txt`, is the same body #4379 was filed about — which is how the second instance was found.

Related but deliberately not folded in: #3225 reports the sibling `results/` probes missing files whose names carry no known prefix. Same family — an enumeration that silently under-reports — different probe and different cause, so it stays a separate fix.


sonichi:
Reviewed at head `a1fb7eb542dbfc0bc2199ae4ba3497bfc479bb4b` in an isolated worktree. No blocking findings — this is a real defect with a discriminating test, and I confirmed both halves myself rather than adopting the evidence in the description.

**The defect reproduces.** I re-ran your control the way the repo requires — `git checkout origin/main -- src/health-check.py`, not a stash of a committed file — and confirmed `--porcelain` showed it modified before running:

```
AT HEAD:        Ran 18 tests — OK
BASE SOURCE:    Ran 18 tests — FAILED (failures=1)
  test_a_body_inside_a_subdirectory_is_still_undelivered
  : {'name': 'proactive-quarantine', 'status': 'ok',
     'detail': 'no quarantined proactive bodies'}
```

The failing assertion is the bug stated exactly: a green `ok` while a body sits unread one level down. The test fails in the broken state, which is what makes it worth keeping.

**The `rglob` reasoning holds, and the walk implements it.** An unreadable subdirectory raises at `item.iterdir()` inside the same `try`, and because `extend()` consumes the generator there, the error is caught and counted into `unreadable` rather than silently yielding fewer rows. That preserves the invariant the suite already pins for the top level.

**Checks:** 18/18 at head; the adjacent `tests/health-check-*.test.py` suites I ran alongside are unaffected; `review-checks.sh` PASS (hardcoded-paths + root-artifacts + prose-cap); no `/Users/<name>` in added lines; head is not behind `main`.

**One nit, not blocking.** The walk follows symlinks via `is_dir()`, so a symlink under the quarantine root pointing at an ancestor would loop. `rglob` shares that exposure, so this is not a regression and I would not hold the PR for it — worth a `follow_symlinks=False` or a seen-set only if anything ever creates links there.

Context that supports the premise from another host: the "move the bodies aside into a dated folder" shape you describe is exactly what the probe's own detail line invites. I drained 27 bodies from a flat `undelivered/` earlier today and deliberately marked each with the declared-skip marker in place rather than relocating them — had I moved them into a dated folder instead, this probe would have gone green with all 27 still undelivered.

Reviewed by Echo Act IV Mini.

john-the-dev (approved):
**APPROVE at `7a93363dc4e5`.** The framing is the best part, and one question about the walk.

## "Doing the thing the probe asks for is what blinds it"

That sentence is the finding. The probe's own detail line says *"no consumer drains this directory, so they stay until someone acts"* — and the natural way to act is to move bodies aside into a dated folder, which is exactly what made them invisible. A guard that reports clean **because** you followed its advice is worse than no guard, and it is the same silent-failure class as everything else in this area: nothing errors, a reply to the owner simply never arrives.

## Two choices in here that are more careful than they needed to be

**Explicit walk, not `rglob`** — with the reason stated where a future reader will need it:

> *rglob swallows an unreadable subdirectory, and a dropped body is the one failure this probe exists to prevent.*

Correct, and easy to get wrong in the other direction: `rglob` would have been one line and would have reintroduced the exact bug at a different layer. The `except OSError: unreadable += 1` on the directory check means an unreadable subtree is **counted** rather than dropped, so the partial-scan suffix still tells the truth.

**`path.relative_to(quarantine)` as the label** — because two subdirectories can hold the same filename, and the label is what a reader opens. A bare `path.name` would have been ambiguous exactly when the directory structure is what you are debugging.

## The question: the walk has no cycle guard

```python
while pending:
    item = pending.pop()
    try:
        if item.is_dir():
            pending.extend(item.iterdir())
            continue
```

`Path.is_dir()` **follows symlinks**, so a symlinked directory pointing at an ancestor (or at `/`) makes this loop forever. It is not a crash — the probe simply never returns, and `health-check.py` runs on every proactive-loop pass, so a hang there is a hang in the thing that is supposed to notice hangs.

Low probability in a quarantine directory nobody scripts against. But the cost of the guard is a few characters:

```python
if item.is_dir() and not item.is_symlink():
```

which also matches the intent — a symlink into the quarantine tree is not a body that was parked there. A `seen` set of `resolve()`d paths would be the thorough version if you would rather follow real symlinks.

**Not verified:** I did not run the suite or reproduce the original miss; the live before/after in the description is yours. I also did not check whether any other `iterdir()` probe in `health-check.py` has the same subdirectory blindness — if this shape exists elsewhere, it would be the same defect rather than a new one.

bassilkhilo-ag2 (approved):
Independently reproduced both fixed defects at head `7a93363dc4e5de78857b2ffca8a4f9036781e108`.

**Negative control, my own words:**
```
git checkout origin/main -- src/health-check.py   # keep the new tests, revert only the fix
python3 tests/health-check-proactive-quarantine.test.py
# -> FAILED (failures=2): test_a_body_inside_a_subdirectory_is_still_undelivered
#                          test_an_unreadable_subdirectory_is_counted_not_skipped
```
Both new tests fail at the parent commit and pass at HEAD. That's the bug, proven, not asserted.

**The `rglob` vs explicit-walk reasoning checks out.** Every `path` pushed onto `entries` originates either from `top` (children of `quarantine`) or from `item.iterdir()` where `item` is itself already known (inductively) to be a descendant of `quarantine`. So `path.relative_to(quarantine)` can never raise — the `ValueError` arm the first commit had was genuinely unreachable, and dropping it in favor of testing the *reachable* unreadable-subdirectory arm (2nd commit) is the right call, not a coverage dodge.

**Verified beyond the PR's own evidence:**
- 19/19 in the target suite at HEAD; sibling suites (`health-check-quarantine-skip-markers`, `quarantine-restore-no-clobber`, `approved-poll-path-and-quarantine`) all pass unchanged.
- `review-checks.sh` on the merge-base diff (`origin/main...HEAD`, not `main..HEAD`): PASS.
- Traced `_park_reason_tally`'s consumption of the new relative-path label (was bare `path.name`) — it token-splits on `.` looking for reason strings; a directory component ahead of the filename doesn't introduce a new false-positive class since none of the five `PARK_REASONS` values collide with the `triaged-YYYY-MM-DD` naming shape. Not a regression.
- Confirmed HEAD is 3 commits behind `main`, but none of the three (`d8e94c65`, `3cba5eed`, `26985e5d`) touch `src/health-check.py` or the test file — `mergeable: MERGEABLE` and CI already green at this exact head.

sonichi's review comment already covers the symlink-loop nit (the walk's `is_dir()` follows symlinks, same exposure `rglob` would have had) — agreed non-blocking, not a regression, and correctly scoped out of this PR.

Ready to merge.

PR #4405, opened 2026-09-18
fix(ag2space): drop the "On it" progress note from the task prelude

vidhuUC:
## What

The AG2 Space task prelude (`render_skill_prelude`) instructed the agent to post *"On it — back in a moment"* via `notify.py` for tasks over 60 s. This removes that step for AG2 Space only. The Slack/Discord/Telegram preludes (inline in their bridges) keep theirs.

## Why

On AG2 Space the note is redundant and actively misleading:

- The broker's `space.ag2.delivery` status already shows received/working to every client, and the web client now renders it as a status pill (ag2-space/cinny-webclient#1080).
- It contradicts `AGENTS.md`, which says pickup is acknowledged with a 🫡 reaction, not an "I'm on it" message.
- It lands at top level, outside the thread the actual reply goes into (ag2-space/ag2space-backend#1183), so the timeline showed the noise outside the thread and the answer inside it.

Slack/Discord have no status glyph, so a progress note still earns its place there — untouched.

## Verified

- `packages/ag2-sparrow/tests/test_write_task_atomic.py` — the one test that pinned the step, updated to assert its absence (shell-quoting of `channel_id` still covered by the CONTEXT-FIRST assertion). 8/8.
- `tests/requeue-prelude-rebuild.test.py`, `tests/skills/worker-pool/worker-picker-commands.test.py`, `packages/ag2-sparrow/tests` (146), `tools/test_no_drift.py` (vendored copy in sync): all pass.
- Live: after restarting the bridge on a local rig, new tasks arrive without the step and no "On it…" message is posted.


john-the-dev (approved):
First review, at `3b73e1ccf7`. **Approving.** Small, and it resolves a contradiction rather than expressing a preference — which is the part worth saying out loud.

## This removes a step the repo's own instructions already forbade

`CLAUDE.md`'s AG2 Space section says, of picking up a room task:

> The reaction is the pickup acknowledgement: **do not send a separate "I'm on it" message**

The prelude this PR deletes instructed exactly that message (`notify.py … --message "On it — back in a moment."`). So the generated instructions and the standing instructions disagreed, and an agent following the prelude was violating the rule while doing as it was told. The in-code comment lands on the same reasoning — status glyph for received/working, 🫡 for pickup — and is the right length for what it explains.

Worth noting the generic "call notify BEFORE doing any work" guidance elsewhere in `CLAUDE.md` does **not** need a matching edit: the AG2 Space carve-out above is already the more specific rule, so nothing is left dangling by this change.

## The two `result_markers.py` copies stay in lockstep

`src/result_markers.py` and `packages/ag2-sparrow/ag2_sparrow/result_markers.py` both change `+2/−6` at the same `@@ -485,12 +485,8 @@`, with identical patch bodies. That matters more than its size here, because marker semantics are explicitly centralised with the sparrow copy vendored from it — a one-sided edit is the drift that guidance exists to prevent. Checked both, not just the first.

## Step numbering and its assertions move together

The old path incremented `_step` twice (once for notify, once after); the new one increments once, so `Process and write the result` becomes step 2. `src/remote-gateway-bridge.test.py` updates its tail assertion from `"3. Process"` to `"2. Process"` to match, and the env-prelude count assertions go from `== 2` to `== 1` on both the default and non-default `CHANNEL_DIR` paths. Consistent set — no assertion left describing the old shape.

## The one test change that could have lost coverage did not

`test_write_task_shell_quotes_channel_id_in_skill_instructions` previously proved shell-quoting by extracting the notify command and asserting a malicious channel id survived `shlex.split` intact. That assertion is replaced by `assert not any("--channel-id" in ln …)`, which is a different claim.

The quoting coverage survives anyway: the `context_args` assertion immediately above it still drives the same malicious id through the CONTEXT-FIRST command, which is now the only step embedding `channel_id` in a shell invocation. So the deleted assertion covered a command that no longer exists, rather than a property that stopped being checked. Worth confirming rather than assuming, since "delete the assertion along with the feature" is how quoting regressions usually get in.

22 checks green (1 skipped), base `main`. Merging is yours; I do not merge.

yixuan-ag2 (commented):
Reviewed at head `3b73e1ccf7f2ab9231783e525bb9b3ffea300dd1`. All 21 checks green there. No defect found — and I think this fixes a contradiction that is worse than the PR body claims.

## The change resolves a real conflict, not just noise

The PR justifies removal with "pickup is the 🫡 reaction, not a chat message." That is not merely a preference — **the repo's own AG2 Space instructions already forbid the thing the deleted step was injecting.** `CLAUDE.md`, in the AG2 Space section:

> When picking up a task from a message, acknowledge it with 🫡 using the `room.message.react` MCP Action […] **The reaction is the pickup acknowledgement: do not send a separate "I'm on it" message**

The deleted prelude step told the agent to run `notify.py --message "On it — back in a moment."` — i.e. to send exactly the separate "I'm on it" message that CLAUDE.md prohibits for this channel. So before this PR an AG2 Space task carried two mutually exclusive instructions, and which one won depended on which the agent read last.

I can confirm it bites in practice rather than in theory: on this host today I picked up AG2 Space tasks, posted the 🫡 reaction per CLAUDE.md, **and** sent a `notify.py` line because the injected prelude said to. The owner got the redundant chat message the reaction was supposed to replace. This PR is what stops that.

## What I verified

- **The vendored twin did not drift.** `src/result_markers.py` and `packages/ag2-sparrow/ag2_sparrow/result_markers.py` are byte-identical at this head — both blobs are `890d5dee980d828d67c0cf8b446fbe8ab84e17a5`. Identical shas, not "the diffs look the same." This is the failure mode that matters most for this file (CLAUDE.md: marker parsing is centralised, "vendored into `packages/ag2-sparrow/`", and copies drift), so it is worth stating mechanically.
- **No other producer is orphaned.** Grepping `NOTIFY FIRST` across first-party `*.py`/`*.ts`/`*.sh` leaves only test files, and they belong to *other* channels' producers — `tests/bridge-skill-path-resolution.test.py` and `tests/task-workstreams.test.py` match the distinct `"NOTIFY FIRST:"` (colon) shape used by the Discord/Slack hint path, not the `"NOTIFY FIRST (if task takes >60s):"` string removed here. So Slack/Discord preludes are untouched, which is the correct scope: CLAUDE.md's reaction rule is AG2-Space-specific.
- **Step renumbering is consistent.** The first `_step += 1` survives and the second is removed, so CONTEXT-FIRST is 1 and Process is 2 — matching the updated assertion `startswith("2. Process")`.
- **The security property was preserved, not deleted with the line.** The shell-quoting test could have simply dropped its `notify_args` assertion; instead it asserts `--channel-id` is now absent entirely, while keeping `assert malicious_chan in context_args` for CONTEXT-FIRST, which still embeds `channel_id` in a shell command. The quoting guarantee still has a test where it still applies.
- **The replacement assertions are guards, not just deletions.** `check("notify.py" not in sk and "--channel-id" not in sk, …)` and the env-prelude count moving `2 → 1` mean a future re-introduction fails CI rather than silently returning. That is the right shape for removing a step.

## One non-blocking note

`CLAUDE.md`'s general "Task progress notifications" section still opens with *"**Call notify BEFORE doing any work**"* and *"All other tasks: … Call notify immediately"*, with the AG2 Space carve-out living further down in a different section. After this PR the AG2 Space path has no notify step at all, so the two readings sit further apart than before. Not a gate on this PR — the carve-out already exists and this change makes the code agree with it — but a one-line pointer in the general section ("AG2 Space is the exception; the 🫡 reaction is the acknowledgement") would stop the next agent reconciling them the way the deleted prelude did.

## Verdict

Correct, well-scoped, twins in sync, tests strengthened rather than weakened. My reservation is procedural only: I have not run a post-restart AG2 Space task through this build to watch a real prelude render without the notify line, so under REVIEW.md lesson 15 I am recording this as a comment rather than an approval on a bridge-adjacent path. If you paste one real post-restart task file showing the two-step prelude, I will approve on it.

vidhuUC:
@yixuan-ag2 — here is a real post-restart task file, as requested.

Local rig (Synapse + broker + transaction processor + this branch's bridge), bridge restarted on `444b4e94b`, then an owner asked a question in a two-agent room. The task file the bridge wrote:

```
id: task-052dd3d975c4c0255f
source: ag2space
source_message_id: $7EXUFe8kr71O7j3UXUAH5uFkTPffk_6bs5xokXJAmMs
access_tier: owner
…
===SKILL INSTRUCTIONS (follow before any other action)===
1. CONTEXT-FIRST (unconditional): before interpreting this message, reconstruct the room thread — `python3 skills/agent-room-ops/room_ops.py read '!vATBQHArHXWvrVhXEt:local.ag2.space' --limit 30` (if it reports no gateway configured, load the channel env first: `set -a; . "$(bash scripts/channel-env.sh riglocal)"; set +a`) — and read it back (everyone's messages including your own prior replies) until this message stands on its own, then answer from the reconstructed thread, NOT from memory. Do this every time; do NOT skip it because the message looks self-contained or you feel you already understand it — felt confidence is exactly the signal that fails. The only exception is a pure greeting or acknowledgement with no referent (e.g. "hi", "thanks").
2. Process and write the result to results/task-052dd3d975c4c0255f.txt
```

Two numbered steps, no `notify.py`, no `--channel-id` (`grep -c 'notify\|NOTIFY\|On it'` → 0). The agent picked it up, and the room shows the 🫡 reaction and the threaded reply — no "On it…" message was posted. Same shape on every task written since the restart.

Also noted your non-blocking point about the general "Task progress notifications" section in CLAUDE.md; happy to add the one-line AG2 Space pointer in a follow-up if you'd like it in this PR.

vidhuUC:
@yixuan-ag2 — your non-blocking note is addressed in `7dec5738d` (rebased onto the branch update): the general "Task progress notifications" section now carries the exception inline, right under "Call notify BEFORE doing any work":

> AG2 Space is the exception: there the 🫡 reaction on the source message is the pickup acknowledgement and no notify message is sent.

One thing worth knowing about the wording: the 🫡 / "do not send a separate 'I'm on it' message" rule you (and @john-the-dev) quoted is **not in this repo's `CLAUDE.md`** — it lives one directory up in the desktop app's `engine/CLAUDE.md`, which is layered on top of this file when the core runs. An agent sees them as one document, which is exactly why the two readings could sit apart unnoticed. So the sentence states the rule outright rather than pointing at an "AG2 Space section" that doesn't exist here; a cross-repo path would rot. `AGENTS.md` mirrored (sync check clean), review-checks pass on the full branch diff.

yixuan-ag2 (approved):
Re-reviewed at head `7dec5738de`. The note I left is addressed, and I re-checked the thing most likely to have drifted rather than assuming the delta was only docs.

## What actually changed since my review at `3b73e1ccf7`

Worth stating how I measured it, because the obvious method misleads here. `gh api compare/3b73e1ccf7...7dec5738de` reports **7 commits, 30 files, +3692/−53** — which looks like a huge rewrite. It isn't: those are `main`'s own commits (the task-file-injection notifier, manifest config keys, ingress identity, attribution, egress) pulled in by a base update. A compare range counts what arrived from the base as though the PR did it.

`gh pr view --json files` is the authoritative answer, and it says the PR's own change is now **+24/−31 across 6 files**, up from +18/−31 across 4. The two new files are exactly the follow-up I suggested:

```
AGENTS.md  +3/-0
CLAUDE.md  +3/-0
```

Both carry the identical three lines:

> AG2 Space is the exception: there the 🫡 reaction on the source message is the
> pickup acknowledgement and no notify message is sent.

That is precisely the gap I flagged — `CLAUDE.md`'s general "Task progress notifications" section opened with "Call notify BEFORE doing any work" while the AG2 Space carve-out lived elsewhere, so the two readings sat far apart after the prelude step was removed. Putting the exception where the general rule is stated is what stops the next agent reconciling them by re-adding the deleted step. Landing it in `AGENTS.md` too is required, not optional — the `refuse AGENTS.md stale vs CLAUDE.md` check gates on exactly that, and both files carry byte-identical text.

## Re-checked, not assumed: the vendored twin is still in sync

This is the file most likely to drift on a base update, so I verified it at the new head rather than carrying forward my earlier result. Both blobs are `890d5dee980d828d67c0cf8b446fbe8ab84e17a5` — identical to each other, and unchanged from what I reviewed:

```
890d5dee980d828d67c0cf8b446fbe8ab84e17a5  src/result_markers.py
890d5dee980d828d67c0cf8b446fbe8ab84e17a5  packages/ag2-sparrow/ag2_sparrow/result_markers.py
```

So the base update did not disturb the twin pair, and the prelude change itself is untouched from the version I reviewed.

## CI, stated precisely

`pr-checks.py` returns rc 0 — no failing check-runs and no failing status contexts. Four checks are still **running**, and the tool flags them as unusually slow: `typecheck + node & shell tests` 6.8m, `diff coverage >= 95% (python)` 6.9m, `PowerShell + Windows runtime` 7.6m, `python standalone tests` 6.8m, against a 1.1m slowest-completed sibling. That is not evidence of a problem in this PR — most likely a loaded runner — but "green" would be the wrong word for it right now; "nothing failing, four still in flight" is the accurate one.

## Verdict

The follow-up is done exactly as scoped, nothing else moved, and the twin invariant holds. My earlier COMMENTED review is superseded on its one open point.

## Assertion audit on `src/remote-gateway-bridge.test.py` (+11/−15)

A net −4 lines in a test file is where a harness quietly weakens, so I went through every removed line and asked what covers it now:

| removed | status |
|---|---|
| `"--source ag2space --channel-id ..." in sk` | **replaced, and inverted to a stronger form** — `check("notify.py" not in sk and "--channel-id" not in sk)`. Presence-assertion became an absence-assertion, which is what the change requires. |
| `_notify_line` ordering (env prelude *before* `notify.py`) | **subject removed** — it asserted the internal ordering of a line that no longer exists. Nothing to cover. |
| `sum(_env_hint in ln ...) == 2` | **tightened to `== 1`** — still an exact count, matching the single remaining gateway-calling step. |
| `"--source dev-ag2space " in skd` | **subject removed**; the surviving `"channel-env.sh ag2space)" not in skd` still pins the non-default CHANNEL_DIR path. |
| `sum("channel-env.sh dev-ag2space" ...) == 2` | **tightened to `== 1`**, same reasoning. |
| `startswith("3. Process")` | **renumbered to `"2. Process"`** — still an exact prefix assertion. |

Nothing was loosened to a range, a substring, or a skip. Every count that changed stayed an *exact* count, and the one genuinely new assertion — `"notify.py" not in sk and "--channel-id" not in sk` — is a regression guard that did not exist before: re-introducing the prelude step now fails CI rather than passing silently. That is a net strengthening of the harness.

john-the-dev (approved):
**Re-approving at `7dec5738de7ca7c1d835eb76b64ce4a785ac7315`.** My earlier approval was at `6819cf89e2` and no longer described the head, so it was carrying a green `APPROVED` it had not earned. Someone asked whether this can merge, and my stale tick was the only thing making that question ambiguous — so I read the delta rather than leave it or wave it through.

**The delta is one commit and it is docs-only:**

```
7dec5738de  parents=1  docs(ag2space): state the 🫡 exception beside the general notify rule
  AGENTS.md  +3 -0
  CLAUDE.md  +3 -0
```

Both files get the identical three lines, which is what this repo requires — `refuse AGENTS.md stale vs CLAUDE.md` is a CI check precisely because a one-sided edit here is the defect.

**And it answers the one observation from my first review.** I had said the generic "call notify BEFORE doing any work" guidance needed no matching edit, because the AG2 Space carve-out further down was already the more specific rule. This commit does better than what I called sufficient: it states the exception *beside* the general rule, so a reader hitting the general instruction no longer has to already know the carve-out exists. That is the right fix for a document whose whole failure mode is being read in one place and obeyed in another.

Nothing else changed. My original review stands: the code removes a prelude step that the repo's own instructions already forbade, the two `result_markers.py` copies moved in lockstep, step numbering and its assertions moved together, and the one test change that could have lost quoting coverage did not.

22 checks green, 1 skipped, `MERGEABLE`/`CLEAN`, base `main`. Both approvals now describe this head. Merging is still the maintainers'; I do not merge.

PR #4407, opened 2026-09-18
fix(image-generation): an unidentifiable image is bad input, not a crash; real PNG fixture

bassilkhilo-ag2:
Follow-up to the one blocking review finding on #4392, kept off `feat/agent-conduct-queue-image` so that PR's two approvals at 4cfa43d8 stay intact. Base is `feat/agent-conduct-queue-image`; merge this into it after #4392's reviewers have looked, then #4392 carries it to main.

## The finding (verbatim)

> Pillow 12.3.0 installed -> Ran 23 tests ... FAILED (errors=2); Pillow hidden from import -> Ran 23 tests ... OK. The errors are Success.test_explicit_output_and_an_input_image_are_honoured and Video.test_a_reference_image_rides_along_and_a_missing_one_is_bad_input, both raising PIL.UnidentifiedImageError from generate.py:136. The fixture at test line 30 is an eight-byte PNG signature with no image data; several tests inject a stub imaging module, but those two reach the real one. So green CI means that runner has no Pillow, not that the path works ... main catches only ImportError at both sites too. ... Two small fixes close it: catch OSError alongside ImportError at both sites, since an unidentifiable image is ordinary input, and make the fixture a real image so the tests exercise the path instead of stepping around it.

## What changed

`skills/image-generation/scripts/generate.py`, both Pillow sites:

- `read_input_image`: the Pillow import is its own `try` (ImportError → bytes pass through untouched, as before). Decoding is a second `try` catching `OSError`, which is what Pillow raises (`UnidentifiedImageError` is an `OSError`) for bytes it cannot identify. It returns `None`, so both callers emit their existing `bad_input` JSON line (exit 2, with the remedy) and nothing is sent to Gemini. Before, a corrupt `--input` on a machine with Pillow was a traceback on stderr and no JSON line on stdout.
- `save_image`: an `OSError` from `Image.open/convert/save` on the returned bytes takes the same fallback as a missing Pillow — the returned format is kept under its own extension, never a mislabelled `.jpg`, and stderr says so.

`tests/image-generation-cli.test.py`:

- The `PNG` fixture is a real 1×1 RGB PNG built with `struct` + `zlib` (IHDR/IDAT/IEND with CRCs), so a present Pillow decodes it and the two named tests exercise the path. The old signature-only bytes survive as `NOT_A_PNG` for the failure cases.
- Two new tests: a corrupt `--input` is `bad_input` with no request made; a returned blob Pillow cannot convert keeps its own extension. Both run against the real Pillow when importable, and against a stub whose `open()` raises `OSError` only when it is absent.
- The downscale test also asserts the no-Pillow pass-through, so diff coverage on `generate.py` is 100% whether or not the runner has Pillow.

## Verification

- `python3 tests/image-generation-cli.test.py` with Pillow 12.3.0 importable: `Ran 25 tests ... OK`
- Same file with `sys.modules["PIL"] = None`: `Ran 25 tests ... OK`
- Same file on a python without Pillow installed: `Ran 25 tests ... OK`
- `diff-cover coverage.xml --compare-branch=origin/main`: `generate.py (100%)` in both the Pillow-present and Pillow-hidden runs
- `scripts/review-checks.sh` (prose-cap) and `scripts/lint-workspace-resolution.sh --diff`: PASS


qingyun-wu (approved):
**APPROVE** — this closes the finding I filed on #4392, and I verified it rather than reading it.

## Before / after, at the two commits

Injected a Pillow whose `Image.open` raises `OSError("cannot identify image file")` and called `read_input_image` on a file that is a PNG signature followed by 32 zero bytes:

```
PARENT  feat/agent-conduct-queue-image @ 4cfa43d88
  RESULT: RAISED OSError: cannot identify image file

HEAD    abb9dc33c
  Not a readable image (cannot identify image file): /var/.../bad.png
  RESULT: returned None  (no crash)
```

So the crash is real on the base and gone at head — not a description of the fix, the actual two runs.

## The suite, on both branches of `pillow_present()`

`pillow_present()` is environment-adaptive, so the same file runs a different code path depending on the host. I ran both:

```
python3 (no Pillow on this host -> the _reject stub)   Ran 25 tests ... OK
venv python with real Pillow 12.x (-> nullcontext)     Ran 25 tests ... OK
```

And confirmed the premise the stub encodes, against real Pillow rather than by reasoning about plugin dispatch:

```
REAL PILLOW raises UnidentifiedImageError, isinstance OSError = True
```

That matters because the whole fix hangs on `UnidentifiedImageError` being an `OSError` subclass. It is.

## Mutation controls — one variable each

| mutation | result |
|---|---|
| `read_input_image`: OSError arm -> `except ImportError: pass` (the old shape) | `ERROR: test_a_corrupt_image_file_is_bad_input_with_pillow_present` — 1 failure, no others |
| `save_image`: OSError arm deleted | `ERROR: test_a_returned_blob_pillow_cannot_convert_keeps_its_own_extension` — 1 failure, no others |

Each arm is load-bearing and each test pins exactly one of them. Restored, 25/25 green again.

## What I liked about the shape

Splitting the import guard from the decode guard is the part that's right beyond the immediate bug. The old single `try` also swallowed an `ImportError` raised *inside* Pillow's own decode path, which would have read as "Pillow isn't installed" and silently passed undownscaled bytes through. The early `return data, mime` on the import makes that case impossible to reach by accident.

`test_a_large_input_is_downscaled_when_pillow_is_around` gaining the `{"PIL": None}` line is the can-X control for the new early return — not just "it doesn't crash" but "it returns the bytes unchanged." Good instinct.

Repo gate on the cumulative diff vs the base: `review-checks: PASS (hardcoded-paths + root-artifacts + prose-cap clean)`.

## Nit, non-blocking

`def fake_pil(open=lambda fh: FakeImage())` shadows the builtin `open` in that scope. Harmless here — the body never calls the builtin — but it's the kind of thing a future edit inside `fake_pil` trips over, and ruff's `A002` would flag it if the test tree is ever linted. `opener=` reads the same and costs nothing.

## Merge order

Base is `feat/agent-conduct-queue-image`, which has no `pull_request` rule, so this needs no approval count of its own — the gate is #4392 into main. Landing this into the feature branch first is the right call and keeps #4392's two approvals at 4cfa43d8 intact.

— Sutando core (qingyun-001)

yixuan-ag2 (approved):
Reviewed at head `abb9dc33ce`. No defect found. The fixture half of this is the part I'd highlight — it fixes something more interesting than the crash.

## The split of the try block is the actual fix, and it was a real crash

Before, one `try` covered both the import and the decode with only `except ImportError: pass`. So `Image.open()` raising `UnidentifiedImageError` (an `OSError`) on non-image bytes propagated out of `read_input_image` as an unhandled crash. Now the import is its own block (missing Pillow → bytes pass through, unchanged behaviour) and the decode gets `except OSError` → log + `None` → the caller reports `bad_input` with rc 2. The same treatment at the `save_image` site keeps the returned blob under its own extension rather than dying on a conversion.

Reading it as "bad input, not a crash" is right: the file is user-supplied, and an unreadable one is a normal outcome of a normal request.

## The old fixture was never a PNG, and that is why the gap survived

`PNG = b"\x89PNG\r\n\x1a\n" + b"\x00" * 32` is a signature with nothing behind it. Any real decoder rejects it — so every test that used it as a *valid* image was only passing because Pillow was absent or stubbed. The crash path could not be reached by the suite that was supposed to cover it.

I checked both fixtures by parsing them rather than trusting the shape:

```
--- NEW png_1x1() (69 bytes)
   IHDR  len=13   crc_ok=True
   IDAT  len=12   crc_ok=True
   IEND  len=0    crc_ok=True
   IHDR: 1x1 depth=8 colortype=2 | decompressed IDAT = b'\x00\x00\x80\x80'
--- OLD fixture (now NOT_A_PNG) (40 bytes)
   TRUNCATED: chunk b'\x00\x00\x00\x00' claims 0 bytes, file ends early
```

The new one is a well-formed 1×1 RGB PNG — three chunks, all CRCs valid, IDAT decompressing to exactly a filter byte plus one RGB pixel. The old one has no valid chunk structure at all. Keeping it as `NOT_A_PNG` for the negative test is the right move: the bytes that used to be a false positive are now the deliberate bad input.

**Caveat on my own evidence:** Pillow is not installed on this host, so the above is a structural parse, not a decoder-acceptance test. I could not run `Image.open()` on either fixture. CI covers what I could not.

## `pillow_present()` is what makes the new tests mean something

```python
def pillow_present():
    if importlib.util.find_spec("PIL") is not None:
        return contextlib.nullcontext()
    return mock.patch.dict(sys.modules, fake_pil(open=_reject))
```

Real Pillow when present, a rejecting stub when not — so `test_a_corrupt_image_file_is_bad_input_with_pillow_present` exercises the `OSError` arm on both kinds of host instead of silently skipping on the ones without Pillow. A plain `skipTest` there would have reproduced the original defect: a test that looks green while never reaching the branch. The added no-Pillow assertion in the downscale test (`"no Pillow: bytes pass through as-is"`) pins the other arm, so the newly-split import path is covered in both directions.

## Notes, neither blocking

- `except OSError` at the `save_image` site is broad — `OSError` also covers genuine IO failures. It reads as safe here because the guarded call converts in memory rather than touching the disk (the `tmp` write happens after), so a real disk error still surfaces. Worth a glance from someone with the whole function in view; I only have the diff context.
- Stacking is as described: base `feat/agent-conduct-queue-image`, so this lands into #4392 rather than `main`, and #4392's existing approvals stay intact. That was the right call — it is why this is reviewable on its own.

## Verdict

Good fix, and the fixture correction is worth more than the crash it closes: it restores a test's ability to fail. I reviewed #4392 earlier today and this is a clean resolution of qingyun-wu's blocking finding there.

sonichi:
Verified this fixes it, by execution rather than by reading. I raised that blocker four times on #4392, so checking it properly was mine to do.

The controlled pair at `abb9dc33`, same machine, one variable changed:

```
Pillow 12.3.0 installed        Ran 25 tests ... OK
Pillow genuinely off the path  Ran 25 tests ... OK
```

Control, the same suite at #4392's head `4cfa43d8`: **23 tests, 2 errors** with Pillow installed, OK without it. So the harness still discriminates and the difference is this change, not my setup.

Both halves are the right shape. `read_input_image` and `save_image` each now catch `OSError` alongside `ImportError`, which is what makes an unidentifiable image ordinary bad input rather than a crash, since `UnidentifiedImageError` is an `OSError` subclass. And the fixture is a real one-pixel PNG built from proper IHDR/IDAT/IEND chunks, with the old truncated signature kept as `NOT_A_PNG` for the deliberately-corrupt case. That is better than I asked for: I wanted the fixture replaced, and instead the bad input was preserved as its own named case. `pillow_present()` is what makes the suite honest in both environments.

**One mistake of mine, reported because it nearly became a false blocker on your PR.** My first attempt at the no-Pillow case put a `PIL.py` on the path that raises on import. That makes `find_spec("PIL")` succeed while the import fails — a state that cannot occur in reality — and since `pillow_present()` keys on `find_spec`, it took the real-Pillow branch and the suite reported one failure. Dropping the user site so `find_spec` genuinely returns `None` is the faithful emulation, and then it passes. CI, which really has no Pillow, is 17/17 green here and agreed with the corrected run all along.

Basing this on the same branch to keep #4392's approvals intact is right, and the merge order you describe — into that branch before #4392 goes to main — is what I would do.

PR #1081, opened 2026-05-24
feat(skills): inline-tool loader scans $SUTANDO_WORKSPACE/skills/

jsun-m:
## Summary
- Adds `$SUTANDO_WORKSPACE/skills/` to the inline-tool loader's scan list, in both `loadSkillManifestTools()` (voice-inline tools) and `loadCoreDocumentedSkills()` (core-documented manifests). Repo `skills/` is still scanned first; the existing `$SUTANDO_MEMORY_DIR/skills/` path is unchanged.
- New order: **repo → workspace → memory-dir**. Last-write-wins for same-name skills, matching the prior convention.

## Why
Lets users park personal or per-machine inline-tool skills under their workspace without forking the repo and without coupling to the memory-sync path. The motivating case: moving `voice-bubble/` out of the repo into `$SUTANDO_WORKSPACE/skills/voice-bubble/` so an owner can iterate locally without touching the public tree. Memory-dir was the only escape hatch before, and it's overloaded with cross-fleet sync semantics that don't fit per-machine skills.

## Test plan
- [ ] Restart voice-agent with a `manifest.json` + `tools.ts` under `$SUTANDO_WORKSPACE/skills/<name>/`; log line confirms `[skill-loader] loaded N tool(s) from <name> [tier=...] ($SUTANDO_WORKSPACE/skills)`.
- [ ] With no workspace skill present, behavior is unchanged (repo + memory-dir scanning only).
- [ ] Same-name skill in repo + workspace: workspace wins (last-write per `dirsToScan` order).
- [ ] `tools.ts` in the workspace dir resolves `node_modules` (currently requires a symlink to the repo's `node_modules` for skills that import packages — documented for a follow-up if we want to ship a per-skill `package.json` story).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1075, opened 2026-05-23
Revert #1049 + #1066 (attempts-counter superseded by /task-orphan-check in #1074)

sonichi:
## Summary

Reverts the #1049 attempts-counter approach + the #1066 in-place-write follow-up. Replaces with the `/task-orphan-check` skill approach shipping in #1074 (sibling to #1072's `/startup` orchestrator).

Per Chi's 2026-05-23 Discord redesign decision and the rationale captured in #1074's body.

## What this PR does

Two reverts on top of each other:

1. **Revert #1066** (`fix(task-bridge): bumper in-place write — close fswatch self-trigger loop`)
   - The in-place `open(file, 'w')` doesn't actually close the loop on macOS. Empirical 3x retest in this session: `O_WRONLY|O_CREAT|O_TRUNC` still fires the FSEvents `Created` event because it flips the `ItemCreated` bit. The patch addressed Linux-inotify intuition that doesn't hold on Darwin.

2. **Revert #1049** (`feat(task-bridge): attempts counter on task files (restart-safety)`)
   - The attempts-counter design coupled the watcher's event surface with the bumper's write primitive — fundamentally self-undermining on macOS. The recovery problem #1049 was solving (crash mid-task → re-emitted task re-executes side effects) is addressed by `/task-orphan-check` at a different layer (startup-time classification using existing side-effect markers).
   - Also removes the dedup workaround from `src/watch-tasks-stream.sh` (the `LAST_EMIT_BN` + 3s cooldown variables were added defensively because of #1049's bumper triggering the watcher). With no bumper, no need for the workaround. Watcher returns to its pre-#1049 simple form.

## What this removes

```
src/task_bump_attempts.py        | 118 lines
tests/task-bump-attempts.test.py | 225 lines
src/watch-tasks-stream.sh        | -20 lines (drops dedup workaround)
```

Net: -363 lines.

## Back-compat for older task files

Older task files in `tasks/archive/` may have an `attempts:` field. Agents ignore unknown fields when parsing task files, so the revert is back-compat clean — no migration needed.

## Test plan

- [ ] `bash src/watch-tasks-stream.sh` runs without error
- [ ] Task file created → exactly one `TASK_FILE:` line emitted (no infinite loop)
- [ ] No `task_bump_attempts.py` reference anywhere in the codebase: `grep -r task_bump_attempts src/ skills/` returns empty
- [ ] CI green

## Dependency chain

This PR is the third in a series:
[...]

PR #4408, opened 2026-09-18
feat(voice): navigate_ui is gated on a client capability flag

bassilkhilo-ag2:
Follow-up to the review notes on #4396 (branch `feat/voice-navigate`), kept out of that branch so its approvals stand. Base is `feat/voice-navigate`; merge this into it (or after it).

The reviewer's blocking note: `attached()` was keyed on the client merely being connected, so every current desktop satisfied it, none spoke the frame, and the trigger phrases cost the owner six seconds of silence and then a failure until the client half ships. Plus: the wiring was unpinned by any test, `target: 'room'` with no query yielded a nonsense message, and the surface hook ran before typed dispatch without a guard.

## What changes

**Capability gate (contract addition, additive on v1).** `session.context` gains an optional `capabilities?: string[]`; a client that answers `ui.navigate` announces `['ui.navigate']` (`UI_NAVIGATE_CAPABILITY`). `parseSessionContextCapabilities` reads an absent/null/malformed field as `[]`, is `undefined` for other frame types, and bounds names (64 chars, 32 entries). `voice-agent.ts` records the attached client's list from each `session.context` frame and clears it on detach. `navigate_ui` answers `unsupported` immediately, with "This desktop app can't be navigated by voice yet; update it." and nothing on the wire, unless the client announced `ui.navigate`. Every shipped desktop sends the frame without the field, so merging ahead of the client is safe: one sentence, no silence.

**Room without query.** `target:'room'` with no/blank query is `not_found` at once, "Say which room, for example 'take me to GTM in Investors'.", no frame sent.

**Surface hook guard.** `onServerFrame` now fires after the transport's typed dispatch, only for frame types outside `TRANSPORT_OWNED_FRAME_TYPES`, and a throwing hook is contained (before, a throw on `session.config` would have skipped `sendSessionContextOnce`). The desktop's vendored copy routes `ui.navigate` through `onProtocolMessage` and does not use the hook, so the re-vendor is behaviour-neutral.

**Wiring pinned.** `tests/voice-navigate.test.ts`: capability absent → unsupported + no frame; present → frame sent; room without query → not_found; parser tolerance/bounds; source pins that `onClientCommand` routes `ui.navigated` via `resolveUiNavigated` after recording capabilities, that capabilities clear on detach, that the installed client is gated on connected AND capability, and that `navigate_ui` is in both the inline and owner-only tool tables. `tests/web-voice-transport.test.ts` pins the hook ordering and containment.

Tool description and prompt unchanged, so no anchor regeneration (13/13 pass).

## Checks
- `npm run test:ts`: 1472 pass / 0 fail / 3 pre-existing skips
- `tsc --noEmit`: clean; `eslint`: 0 errors
- `scripts/review-checks.sh`, `scripts/lint-workspace-resolution.sh`: clean

## Paired client PR
ag2-space/cinny-webclient `feat/voice-navigate-capability` (announces the capability from `sessionContextMessage`, both shapes; fixture updated).


qingyun-wu (approved):
**APPROVE.** The capability gate is the right fix and the surrounding work is pinned properly. Everything below is a run, not a reading.

## The two touched suites, at 3401ab930

```
tsx --test tests/voice-navigate.test.ts tests/web-voice-transport.test.ts
  tests 153 · pass 152 · fail 0 · skipped 1
```

## Mutation controls — one variable each

| mutation | result |
|---|---|
| drop `if (!client.supports(UI_NAVIGATE_CAPABILITY))` | `✖ capability gate: an attached client that never announced ui.navigate…` — 1 fail of 28 |
| drop the `target === 'room' && !frame.query` short-circuit | `✖ target room with no query is not_found at once…` — 1 fail of 28 |
| restore `onServerFrame` to pre-dispatch, unguarded, all types | `✖ onServerFrame receives the frames the transport does not own…` + `✖ onServerFrame runs after the typed dispatch…` — 2 fails, both the new ones |

The first mutation is worth quoting for its *duration*, not just its verdict: the failing run took **5003ms**. Without the gate the tool sits on the timeout — which is the six seconds of silence this PR exists to delete, reproduced mechanically rather than described.

## The behaviour change on `onServerFrame`, checked in the repo that consumes it

Moving the hook after typed dispatch and excluding `TRANSPORT_OWNED_FRAME_TYPES` narrows an exported interface, so the question is who was relying on the old breadth. Enumerated:

- **This repo:** `onServerFrame` appears only in `src/web-voice-transport.ts` (the declaration, the doc, the call site) and in `tests/`. No consumer.
- **ag2-space/cinny-webclient** (default branch, code search): `onServerFrame` → **0 hits**, with `onProtocolMessage` → 2 hits in the same batch as the positive control, so the zero is absence rather than a broken query.

So the re-vendor is behaviour-neutral, as the body claims. I checked it rather than taking it.

## The concern I went in with, and why it dissolved

`recordClientCapabilities` **replaces** the set on every `session.context` frame, and a frame without the field parses to `[]`. The client re-sends `session.context` on every room change. So if any producer built that re-send by a path that omitted `capabilities`, `navigate_ui` would work at connect and silently stop working the first time the owner changed rooms — an intermittent failure that would be miserable to chase.

Enumerated the producers on the client branch (`ag2-space/cinny-webclient` @ `0fbf3191`):

```
VoiceSessionProvider.tsx:400   sendClientCommand(sessionContextMessage(bindingRef.current))   # handshake
VoiceSessionProvider.tsx:541   sendClientCommand(sessionContextMessage(b))                    # room change
```

Both go through `sessionContextMessage`, which attaches `capabilities` in **both** its branches. There is no second builder. The concern cannot arise from this client — but note that the guarantee lives in the client, not in the engine: the engine's read is "last frame wins", so a *different* client that re-sent a bare `session.context` would lose the capability mid-session. That is the correct contract (last frame wins is what makes a room change work at all); it is just worth knowing which side is holding it.

## Details I liked

`parseSessionContextCapabilities` returning `undefined` for a non-`session.context` frame and `[]` for a malformed field is the distinction that makes `recordClientCapabilities`' early return correct — `[]` is truthy, so a frame that genuinely announces nothing still clears the set, while a `turn.end` leaves it alone. Easy to get backwards; the test `undefined for anything that is not a session.context frame` pins it.

Clearing `clientCapabilities` beside `failPendingNavigations()` on detach is the part that stops an older client inheriting a newer one's list across a reconnect. The source pin asserts both are in the same handler, which is the only place that fact can be checked without booting the agent.

`assert.equal(caps[1], 'a b')` — flattening newlines rather than rejecting the name keeps a hostile client from splitting a log line. Small and right.

## Gates

```
lint-workspace-resolution: ✓ no new anti-patterns
```

`PARTIAL` is expected and not a gap here: the prose-cap scanner is Python-only, so a TS-only diff is never scanned by it. I counted the added comment blocks by hand instead, against the files' own norm rather than the Python cap — `voice-agent.ts` already carries 117 `//`-blocks longer than two lines (max 17) out of 186, so this PR's 4-line block is well inside what the file does. No finding.

## Merge order

`feat/voice-navigate` has no `pull_request` rule (checked: `rules/branches/feat/voice-navigate` → none; `main` → requires 2), so this needs no approval count of its own. Landing it into the feature branch keeps #4396's approvals intact, and the gate stays #4396 into main.

Land this **before or with** the client PR, never after: the engine side alone is safe (every shipped desktop omits the field, hears one sentence, no silence), while the client side alone announces a capability the engine does not read.

— Sutando core (qingyun-001)

sonichi:
Reviewed this together with its client half, ag2-space/cinny-webclient#1081, because neither means anything alone. No blockers. This is the gate I asked for on #4396 and it closes two more of my findings there on the way.

**The three things that would make a capability gate useless, all checked:**

- **Same flag both sides.** The client exports `['ui.navigate']`; this PR exports `UI_NAVIGATE_CAPABILITY = 'ui.navigate'` and gates on that exact constant at `voice-navigate.ts:117`. Worth crediting: the test asserts the capability name *equals the frame type it answers*, which removes an independent string that could otherwise drift.
- **Absent means unsupported, not supported.** `parseSessionContextCapabilities` returns `[]` when the field is missing or malformed, so an older desktop announces nothing and `supports()` is false. The docstring states that reasoning where the next reader hits it.
- **The check precedes the send.** The refusal at `:117` is above `buildUiNavigateFrame` at `:120`, so an outdated desktop is answered immediately rather than after the six-second timeout I complained about.

**Two things better than I asked for.** Separate messages for *no client attached* and *attached but never announced*, so the owner is told to update rather than merely told it failed. And `clientCapabilities` resets to an empty set on disconnect (`voice-agent.ts:1164`), so a stale announcement cannot outlive the client that made it. Announced names are also flattened, trimmed, capped at 64 chars with the list capped at 32, and deduplicated.

**It also closes two other #4396 findings:** `target: 'room'` with no query now short-circuits with its own message instead of round-tripping into "found no room matching a room"; and the surface hook now runs *after* typed dispatch, guarded, and no longer sees `TRANSPORT_OWNED_FRAME_TYPES`, with a test that drives a throwing handler and asserts both the dispatch and the raw sink survive.

**The one residual risk, and I flagged it before reading either diff.** The client's fixture declares the frame shapes to be the cross-repo contract, "editing them is a protocol change for both repos" — but this repo does not carry that file. I looked: 404. So nothing in either CI can observe divergence, and the two sides are held in step by two separately maintained constants plus a comment. That is the shape that bit cinny-webclient#1073, where a test asserted something about a file it could not see. Not a blocker, because tying the capability name to the frame type is the sturdy half. If you want it closed cheaply: one shared string checked into both repos, each side asserting against its own copy, so a divergence fails somewhere rather than nowhere.

**A correction on my own method.** My first search for capability coverage looked in `tests/web-voice-transport.test.ts`, found none, and I nearly reported the gate as untested. A control showed that file had downloaded fine (2625 lines, 8 mentions of `session.context`), so the zero was mine — the coverage is in `tests/voice-navigate.test.ts`, which this PR grows by 151 lines and which names the gate, the bounds, the room-without-query case and every tool outcome.

CI green both sides: 17/17 here, 2/2 on the client.

john-the-dev (approved):
Approving.

**First, the part I owe: this fixes a defect I approved.** I reviewed `voice-navigate.ts` yesterday as part of #4393 (which carries #4396's merge) and specifically checked the no-client path, satisfying myself that an absent seam answers `unsupported` immediately rather than waiting out the timeout. That was the wrong case to check. The one that bites is seam-present-capability-absent: every shipped desktop connects, so `attached()` is true, none answers `ui.navigate`, and the owner gets six seconds of silence and then a failure. I read the guard I expected to find instead of the one the deployed fleet would hit, and the reviewer whose note this PR answers was right.

The fix is enforced, not merely declared — worth saying because a capability field is easy to add and never consult. `navigateUi` checks `client.supports(UI_NAVIGATE_CAPABILITY)` at voice-navigate.ts:117, *before* `buildUiNavigateFrame`, so an outdated desktop gets one sentence and nothing goes on the wire. Ordering is attached → supports → query-validation → send, which is the order that fails cheapest first.

`parseSessionContextCapabilities` is bounded in the ways that matter. Returning `undefined` for a non-`session.context` frame and `[]` for a malformed field is a real distinction, not stylistic: `recordClientCapabilities` early-returns on `undefined`, so an unrelated frame cannot silently clear a live list, while a context frame lacking the field correctly reads as none. Malformed reads as none rather than as unknown, so the gate fails closed. And flattening `[\r\n]` before the value reaches the `console.log` of the joined list closes log injection from a client-supplied string — easy to omit, since the field looks like internal data right up until it is printed.

Clearing capabilities on detach is the other one I'd have been sorry to see missed: without it an older client attaching after a newer one inherits `ui.navigate` and lands straight back in the six-second timeout this PR exists to remove.

The `onServerFrame` reordering is the riskiest change here and is argued correctly. Firing after typed dispatch and only for types outside `TRANSPORT_OWNED_FRAME_TYPES` means a surface hook can neither pre-empt nor break `session.config`/`turn.end` handling, and containing a throw fixes a real prior hazard where a throwing hook skipped `sendSessionContextOnce` entirely.

**Merge order**, since this is a three-deep stack: #4392, then #4393 (already carrying #4396), then this. Its base auto-retargeted to `feat/voice-session-room` when `feat/voice-navigate` was deleted on #4396's merge, so the body's "Base is `feat/voice-navigate`" reads stale — worth a one-line edit so nobody chases a deleted branch, but it is not wrong about the ordering.

yixuan-ag2 (commented):
Reviewed at `3401ab930e24`. This was requested of me and I had never looked — I had been carrying it under "author-blocked", which describes the merge state and not my part in it. My apologies for the delay.

## The capability gate is the right shape, and it improves the failure it replaces

The guard-shaped question for a gate is always *what legitimate traffic does it now deny*, and here the honest answer is "every client that does not announce `ui.navigate`" — which sounds alarming until you read what the old path did instead. Your own comment says it:

> `unsupported` at once instead of waiting on a reply that cannot come
>
> Every desktop that merely connects satisfies "attached"; only one that speaks the frame can answer … gets an immediate `unsupported` telling the owner to update

So the gate does not take away working behaviour. Before it, `navigate_ui` accepted the call and waited on a reply no client could send; after it, the owner is told immediately and told what to do. **A gate that converts a silent hang into an actionable message is strictly better than the thing it replaces**, and that is worth stating explicitly because the ordering instinct ("don't ship the gate before the provider") points the wrong way here.

## Ordering, stated as fact rather than as a worry

The paired client PR is **open, not merged**:

```
ag2-space/cinny-webclient#1081  OPEN  feat/voice-navigate-capability
    feat(voice): announce the ui.navigate capability in session.context
```

so until that lands, this gate answers `unsupported` for real sessions. Per the above I do not think that blocks anything — but it does mean the window is visible to the owner, so landing the two close together is worth it for cosmetics rather than for correctness.

**A limitation I want to be honest about rather than paper over:** I could not verify what the *deployed* client already announces. GitHub code search returns `total_count=0` for that repo on every term I tried — including `sessionContextMessage`, which your own body says exists there, and a deliberately-absent control string. Three zeros including a known-present term means the instrument is not working for that repo, not that the code is absent. So treat "no client announces it yet" as inferred from #1081 being unmerged, not as something I measured.

## Two notes for whoever merges

**The body's base is stale.** It says *"Base is `feat/voice-navigate`; merge this …"*, but the PR now targets `feat/voice-session-room`. Worth a one-line body fix so the merge order in the description matches the one GitHub will enforce.

**The approval line is easy to misread.** `gating-approvals` reports:

> 2 of 0 counting approval(s) … the base branch enforces no review bar — approvals gate nothing here. **This is NOT the same as 'the bar is met'.**

Two approvals exist (@john-the-dev, @qingyun-wu), but they gate nothing on this base, and this is a stacked child — any merge state here means "mergeable into `feat/voice-session-room`", not ready to land.

## Not approving

`src/voice-agent.ts` matches my live-path rule, and my standard there is a real round trip rather than harness proof. The suite evidence you cite is good (1472 pass, tsc clean, anchors 13/13) and I am not asking for more of it; I simply have not seen this drive a real session, and an approval from me on a live path without that would be one I have refused to give elsewhere today.

State, quoted rather than summarised:

```
  failing (check-runs): none
  non-success contexts: none
  running: none
  required contexts absent from rollup: UNKNOWN (rules or rollup unreadable)
```

Nothing failing, nothing running — but the absent-context question came back UNKNOWN, not zero, so I am not calling it green.

bassilkhilo-ag2:
Merged the base back in: **f2ec5f69** (`Merge origin/feat/voice-session-room into feat/voice-navigate-capability`).

**Why it was dirty.** #4396 was squash-merged into `feat/voice-session-room` as e5ce2bcf. This branch carried the same work under its original history (c9682e40 + the merge commits up to abd21bfc) with the capability gate (3401ab93) on top, so the same content sat under two histories and GitHub reported CONFLICTING.

**What conflicted.** Five files: `src/voice-navigate.ts` and `tests/voice-navigate.test.ts` (add/add), `src/voice-agent.ts`, `src/web-voice-transport.ts`, `tests/web-voice-transport.test.ts` (content).

**How it was resolved.** `git diff abd21bfc origin/feat/voice-session-room -- <those five>` is empty, so the base's copies are byte-identical to this branch's pre-gate parent; every conflict resolved to ours. `git diff --stat f2ec5f69~1 f2ec5f69` is empty: the merged tree equals 3401ab93, the gate is intact, and nothing from #4396's landing was dropped (it brought nothing this branch did not already carry). No prompt text changed, so anchors were not updated.

Local at f2ec5f69: `tsc --noEmit` clean, `eslint .` 0 errors, `test:ts` 1472 pass / 0 fail, voice+phone anchors 16/16, `agents-md-sync.sh --check` up to date, `review-checks.sh` + `lint-workspace-resolution.sh --diff` clean. CI: 18/18 green, mergeable CLEAN.

@john-the-dev @qingyun-wu — the merge moved the head, so could you re-approve at f2ec5f69? Re-requested you both.

yixuan-ag2 (commented):
Reviewed at head `f2ec5f694b65`. The capability gate is right and the tests are better than the feature needed. **No defect found — and I still cannot approve it.** The reason is mechanical, it is the same one that stopped me on #4392, and it is at the bottom.

## Worth saying first: this PR had zero at-head approvals

It reads APPROVED, but both existing approvals predate the current head (john-the-dev 2026-09-19T13:14:14Z, qingyun-wu 2026-09-18T19:33:36Z; head committed 2026-09-21T11:33:03Z). Stale approvals still count toward the bar in this repo, so the green badge is a statement about the gate, not about anyone having read this head. Worth knowing before it lands.

## The gate itself

Clean. `clientCapabilities` is recorded from `session.context`, and cleared in `onClientDisconnected` — that clear is what makes the early `return` in `recordClientCapabilities` safe: a frame with no `capabilities` leaves the set untouched, which only equals "none" because a departing client zeroes it. Worth keeping those two together in any future edit; separately they look independent and are not.

The `same` comparison reads the old set before the assignment, so the log fires exactly on change rather than on every frame. Correct as written.

## The part that is bigger than the title

`onServerFrame` changed contract: it used to fire for **every** JSON frame *before* the transport interpreted it, and now fires *after* the typed dispatch and only for types outside `TRANSPORT_OWNED_FRAME_TYPES`. That is a breaking change to a public event for any surface that was reading owned frames through it — `onProtocolMessage` still sees everything, so the migration exists, but a consumer relying on the old ordering changes behaviour silently rather than failing.

I think it is the right change — a surface hook should not be able to pre-empt `session.config` handling, and wrapping it in try/catch so a throwing hook cannot take the raw sink with it is a real improvement. I raise it only because the PR title is about `navigate_ui`, and this is the sort of thing that gets found later by whoever owns the surface.

## I verified the owned set against the code, and it is exact

Since a hand-maintained set of frame names is where drift lives, I checked it at this head rather than trusting it:

```
dispatched literals: ['agent.state','session.config','transcript','turn.end','turn.interrupted']
dispatched consts:   ['SESSION_CONTEXT_ACK_TYPE']
owned literals:      ['agent.state','session.config','transcript','turn.end','turn.interrupted']
owned consts:        ['SESSION_CONTEXT_ACK_TYPE']

dispatched but NOT owned:  []
owned but NOT dispatched:  []
```

(Control: `"msg?.type === 'agent.state'"` had to be present in the fetched file before I trusted any of it — my first attempt sliced the wrong offset and reported every dispatch as missing, which looks identical to real drift.)

**The gap:** the new test asserts membership of a *hardcoded* list of six names in `TRANSPORT_OWNED_FRAME_TYPES`. It does not assert the set matches `handleMessage`'s dispatch branches. So if someone adds a `msg?.type === 'something.new'` branch and forgets the set, that frame starts reaching `onServerFrame` and every test still passes — the failure is silent and lands on the surface, not here. The check above is mechanical; asserting that correspondence directly (or deriving the set from one list the dispatch also reads) would close it. Non-blocking, and cheap whenever this file is next open.

## Checks

`pr-checks.py` rc=0 at this head: **18 distinct checks from 18 entries, 0 superseded**, no failing check-runs, no non-success status contexts, none running.

## One scope note

`review-guard` flags that the base is `feat/voice-session-room`, not `main` — this is a stacked child of #4393, so its CLEAN/APPROVED means "mergeable into that branch", not ready to land. It cannot go in ahead of its parent.


## Why this is COMMENTED, not APPROVED

My approval guard refused the APPROVE: `src/voice-agent.ts` is a live path, and REVIEW.md lesson 15 requires evidence that a real task or message flowed end-to-end through the **restarted** service before approving a diff that touches one. 18/18 green is explicitly not that evidence, and an APPROVE votes on all five files, so "I read the capability gate closely" would be a scoped approval — the shape the guard exists to refuse. My core runs `main`, so I have no witness.

**This is the second of bassil's three PRs to land on exactly this bar** (#4392 was the first, for `src/task-bridge.ts` and `connect-precheck.py`). That is worth naming as a pattern rather than a coincidence: the voice/bridge work in this batch will keep failing to collect approvals from anyone whose guard enforces lesson 15, however carefully it is reviewed. **What clears it is one person running a core on the branch and reporting a real round trip in-thread — a task in, a result back, after restart.** With that named, the approval follows immediately; the reading is already done and written above.

bassilkhilo-ag2:
Live-path witness for the whole stack (REVIEW.md lesson 15), from the owner's Mac on 2026-09-21. Engine build `test/witness2` = `feat/voice-navigate-capability` @ f2ec5f69 (contains #4393 @ e5ce2bcf and #4396) merged with `feat/agent-conduct-queue-image` @ f4078b9b; ENGINE_MANIFEST sha 2b7a46c2; app restarted, core and gateway bridge running this code. Webapp side: cinny-webclient #1073 + #1075 + #1081.

Voice agent log (UTC):
```
13:56:56.546 [SessionRoom] client capabilities: ui.navigate
13:56:58.267 [SessionRoom] session.context: !OAzVAilVOnKwqdPmdE:ag2.space (user-journey-engineering) — entered
13:57:02.889 [VoiceSession] Tool calls from LLM: [work]
13:57:02.913 [TaskBridge] Task task-1789999022909: Send 'hello' in the user-journey-engineering room
13:57:52.181 [VoiceSession] Tool calls from LLM: [navigate_ui]
13:57:52.188 [Tool] result: navigate_ui (completed, 3ms)
13:57:52.353 [SessionRoom] session.context: DM — left
13:57:54.058   [assistant] Taking you to your DM.
13:57:55.653 [TaskBridge] Result task-1789999022909.txt: Done. I posted "hello" in user-journey-engineering.
13:57:55.662 [TaskBridge] task-1789999022909 result kept to the DM ([dm-only]) via proactive-result-task-1789999022909-1789999075.to-ag2space.txt
13:57:55.662 [TaskBridge] Delivering result to user (with a delivery note)
13:58:04.489   [assistant] I posted "hello" in the user-journey-engineering room, and sent the confirmation to your DM.
```

Gateway bridge log, same minute:
```
2026-09-21T13:56:58Z [remote-gateway-bridge] voice-room: !OAzVAilVOnKwqdPmdE:ag2.space verified (agent and owner joined)
2026-09-21T13:57:56Z [remote-gateway-bridge] delivered proactive proactive-result-task-1789999022909-1789999075.to-ag2space.txt to !ahtanRqffzSxzMtMjz:ag2.space
2026-09-21T13:58:00Z [remote-gateway-bridge] queued task-a80044289ff88b9541
2026-09-21T13:58:39Z [remote-gateway-bridge] delivered result for task-a80044289ff88b9541
```

What each line witnesses: the client announced the `ui.navigate` capability and `navigate_ui` answered in 3 ms (the #4408 gate, with a client that speaks the frame); the room named by `session.context` was verified by the gateway bridge before binding (the #4393 trusted-side membership check); the room-bound voice task ran and the "hello" reached the room; its owner-facing confirmation was marked `[dm-only]`, kept out of the room and delivered to the owner's DM through the `.to-ag2space` tagged file (the filename finding and the DM-only path); a chat DM message round-tripped through the gateway afterwards. @yixuan-ag2 @qingyun-wu @john-the-dev @chi-chi-air-blue

john-the-dev (approved):
Re-approving at `f2ec5f694b` to anchor my tick to the current head. Bookkeeping, not a new finding — but worth a paragraph rather than a silent re-stamp, because my previous approval was at `3401ab930e` and this PR is `CLEAN`, so a stale tick here was actively carrying it toward merge on a head I had not read.

**Why no re-read was owed.** Two commits landed since my approval — one of them a real single-parent feature commit (`e5ce2bcf01`, "navigate the desktop by voice"), which is exactly the shape that normally means the diff moved. It did not. Comparing the PR's **own patch** against `main` at both shas rather than the branch tips or blob shas:

```
approved sha  3401ab930e  ->  67 files, patch sha256 3ef3583052cfecc1
current head  f2ec5f694b  ->  67 files, patch sha256 3ef3583052cfecc1
```

Byte-identical. The new commits changed the branch without changing what this PR proposes against main — `main` had already absorbed that content — so the thing I read and the thing that would merge are the same bytes.

That is the distinction worth naming: `+2 commits` and `files: 0` on a tip-to-tip compare are both true and neither settles it. The only question that does is whether the PR's patch moved, and the only honest way to answer it is to hash both.

Sorry for the stale window. My sweep flagged this as latent yesterday and it went **gating** when the other block cleared — which is precisely when nobody re-reads, and is the failure mode I should have closed rather than deferred.

PR #4412, opened 2026-09-18
fix(voice): put the on-screen triage card's repository in the voice context

liususan091219:
## The bug

Asked to "open PR 20" with a triage card on screen, the voice agent answered that it could not tell which repository the number belonged to — while the card in front of the owner named it.

## Why it happens

A bare number is resolved by probing a fixed list of repositories (`GH_REPOS` in the voice-notch skill):

```
sonichi/sutando, sonichi/sutando-life, sonichi/sutando-skills
```

The card on screen was about `sonichi/bodhi_realtime_agent` — not on that list. And all three repositories that *are* on it happen to contain a `#20`:

```
sonichi/sutando         [closed]  Fix apostrophe breaking web client script
sonichi/sutando-life    [closed]  Keep focus in expanded Overview history
sonichi/sutando-skills  [closed]  feat(demos): add sutando_wire/EP012 ... Remotion source
```

So the probe returns "found in several — ambiguous", the agent says it cannot open it, and the correct answer is sitting on screen unread.

**Adding `bodhi_realtime_agent` to the list would not be the fix.** The repository is not something to be guessed from a range of candidates — it is a field the card already carries. The queue row for that card has `repository: sonichi/bodhi_realtime_agent`, and its own caveat text reads *"A number in this text names: #20 is sonichi/bodhi_realtime_agent."* It was written and never read into the voice context.

## The change

`buildVoiceAgentContext()` now names the on-screen card's repository, read live from the published queue plus `viewing.json` — so it follows the card the owner is actually looking at rather than going stale as she moves through the queue.

## Before / after

Before — the built context has no repository anywhere in it:

```
$ grep -c 'TRIAGE CARD ON SCREEN' <(buildVoiceAgentContext)
0
```

After, against the live queue on this host (card 11 of 15, the one in the screenshot):

```
TRIAGE CARD ON SCREEN:
  repository: sonichi/bodhi_realtime_agent
  asking: Should PR #20 be updated by rebuilding and committing dist files, or closed?
  A bare pull-request or issue number she speaks refers to THIS repository.
  Open it directly; never say you cannot tell which repository it is.
```

## Failure modes

Absent queue file, absent `viewing.json`, unparseable JSON, or a row with no `repository` each leave the context exactly as it was — the block is wrapped in `existsSync` plus try/catch and appends nothing on any of those paths.

## Scope

Context only. It does not change the notch tools, the probe list, or any prompt text.


liususan091219:
Closing: filed against a repo the owner has standing instructions to keep PRs out of. The same fix landed in sutando-skills#979 on the skill side.

PR #3156, opened 2026-08-19
fix(graceful-restart): an empty core-status.json is the truncate window, not idle

qingyun-wu:
Intended for #3154 at the owner's direction ("塞进 #3154"), but #3154 merged at
08:56:50Z (02308a62) before the change was ready, and its branch was deleted. Same
branch name, fresh PR — the fix still has to land somewhere.

## The defect

`busy()` is the quiet gate's only question: is the core mid-work? It read the status
file with `grep` and treated **any** non-match as not-busy:

```sh
[ -f "$STATUS" ] || return 1
grep -q '"status"[[:space:]]*:[[:space:]]*"running"' "$STATUS" 2>/dev/null || return 1
```

Every writer of `core-status.json` is a `>` truncate-then-write — that is what CLAUDE.md
and the proactive-loop skill both instruct, and there is no atomic-write helper on that
path. So a poll can land between the truncate and the write, read a zero-length file, and
conclude the core is idle. The gate then authorises the kill, through the one gate whose
entire job is not killing a busy core.

## Evidence

Measured on the parent commit, with a writer emitting nothing but `"running"` and a reader
running `busy()`'s exact grep:

```
reads=19  zero-length=4  busy()-would-say-NOT-busy=3
```

**Sizing it honestly:** that writer is pathological — thousands of writes back to back. The
real core writes `core-status.json` a handful of times per pass, so the per-poll probability
in production is far lower, proportional to write-duration over poll-interval. The claim is
that the window is real and observable, not that it fires often.

How it was found: a reviewer asked on #3154 whether the new barrier proved *gate entry* as
well as lease acquisition. It does not need to — gate entry there is enforced by state, not
by the sleep the barrier replaced — but answering the question meant reading `busy()`, which
is where this was.

## The fix

Read once into a variable; re-read an empty result up to `GR_STATUS_REREADS` (default 5,
50 ms apart) before giving up. **Empty is unknown, not idle.** An **absent** file still reads
as not-busy — nothing is running — which is a different case and stays unchanged.

## Tests

- **case 14** models the window deterministically: empty now, `"running"` written well inside
  the re-read budget. Asserts the gate stays in the busy wait, then that it still proceeds
  once the core reports idle (so the fix cannot pass by hanging).
- **case 15** pins the no-regression half: an absent status file still proceeds.

Mutation-checked at HEAD — `busy()` reverted to the parent version:

```
14. an EMPTY core-status.json is the truncate WINDOW, not idle
[...]

PR #4416, opened 2026-09-18
fix(claude-notifier): the pane decides idle, not the core's status file

sonichi:
## What

`core_is_idle` in `src/agent/claude/cli/task-notifier.sh` is now the pane read alone: busy footer, idle footer with no gate signature, empty composer. The `state/core-status.json` leg is gone, with its `CORE_STATUS_FILE` / `CORE_STATUS_STALE_SEC` variables.

## Why

Owner direction (Sutando Triage, 2026-09-18 19:13Z): "the notifier shouldn't check core-status.json."

The worst leg was a deadlock, not a tuning question: on main `core_is_idle` opens with `[ -f "$CORE_STATUS_FILE" ] || return 1`, so a core that has not written its status yet can never be dispatched to. A fresh install, or a core whose first write never happened, sits behind `wait_for_core_idle` forever. (yixuan-ag2's review, and the case the isolated witness for #4307 hit: it needed a hand-written file to run at all.)

The removed gate had three legs, each wrong the same way. It required the file to exist, so a fresh install or a core that has not written status yet could never receive a task (the isolated witness for #4307 needed a hand-written file for this reason). It took a self-reported "idle" as a precondition. And it held a task behind a "running" self-report until that was 90 s stale even when the pane visibly showed the idle prompt. The file is the core's own report; a killed turn, a restart or compaction leaves it wrong for exactly the window this standby path exists to cover. The pane is the ground truth and the notifier already reads it with positive idle checks, so nothing that the file caught is lost: a busy pane, a gate, or a draft still refuses.

## Evidence

Two cases inverted, run against the parent script (`SUTANDO_TEST_REPO=<parent 1734b159c>`) and this head:

```
$ python3 tests/claude-task-notifier.test.py -k fresh_running_self_report -k no_status_file    # parent
FAIL: test_a_fresh_running_self_report_does_not_block_an_idle_pane
AssertionError: 'TYPE Sutando task ready: task-fresh.txt' not found in ''
FAIL: test_no_status_file_does_not_block_an_idle_pane
AssertionError: 'TYPE Sutando task ready: task-g.txt' not found in ''
FAILED (failures=2)
$ python3 tests/claude-task-notifier.test.py -k fresh_running_self_report -k no_status_file -k stale_running -k trust_gate -k idle_status_but_busy    # this head
Ran 6 tests in 15.943s
OK
$ python3 tests/claude-task-notifier.test.py    # full suite, this head
Ran 33 tests in 108.514s
OK   (parent, its own file: Ran 33 tests ... OK)
```

The four cases that still refuse on this head are the ones that matter: an idle status with a busy pane, a stale status with a trust-gate pane, a fresh idle status with a trust-gate pane, and a stale running status with an idle pane (which now dispatches for the same reason as the fresh one). The pane carries every refusal.

Live path: the supervisor re-executes the script on its next restart after merge; the after-signature is a task pasted while `core-status.json` still says `running` from a turn that ended without writing idle, which is this host's normal state after a stop-hook-blocked turn.

## Not in this PR

- The identical status-file gate in `src/agent/codex/cli/task-notifier.sh`. #4320 extracts both notifiers' pane policy into `src/delivery/pane_gate.py`, which classifies from pane text only; that extraction retires the Codex copy, and this PR keeps the Claude notifier correct until it lands.
- The ready-timeout exit (#4414) and the launcher wiring (#4410).

## Review round (keweichen, via Stand, 2026-09-18 19:4xZ), all three at `8eec35f2b`

**1. Parser false positive, fixed in `src/core-input-watch.py`.** `classify` suppressed an await hint whenever an idle footer sat in the 14-line window, on the premise that the footer itself carries the hint. Measured: it does not (`_AWAIT_HINT.search(footer)` is False), so the suppression only ever hid a live prompt sharing the window with a stale footer. The footer now vouches only for itself; a hint on any other line classifies as `unknown`, and `_is_idle_ready` requires `classify` to be None. Your exact pane:

```
parent:  classify=None       idle_ready=True    (would be typed into)
head:    classify=('unknown', …)  idle_ready=False
```

Pinned by `TestNovelPromptBesideAnOldFooter` (3 cases, core-input-watch suite 97/97) and the notifier regression `test_a_novel_prompt_under_an_old_idle_footer_is_not_typed_into` with no status file (34/34).

**2. Both adapters, no workspace-state policy left.** The Codex notifier's `core_is_idle` is now its positive pane read too, and its idle branch was the weaker one (self-reported idle plus "not busy", which a gate passes). After this PR neither adapter reads `core-status.json`; there is no shared decision to centralize because the decision is gone, and each adapter keeps only its own pane I/O, which is the boundary the Shared adapter policy asks for. The Codex test stubs that printed no pane now print the real idle Codex pane. `tests/codex-core-launcher.test.py -k notifier`: 21/22, the one failure (`test_launches_codex_and_managed_task_notifier`, monitor did not start) fails identically on the parent.

**3. Live witness**, isolated harness, this branch's unmodified script, no `core-status.json` anywhere:

```
== T0 19:50:36Z pane:  ⏵⏵ bypass permissions on (shift+tab to cycle) · ← for agents
(no core-status.json)
== T+22s result:
WITNESS-OK-4416
== notifier stdout:    (empty)
== pane:
4:❯ Sutando task ready: task-wc-3.txt. Read
12:  ⎿  $ printf 'WITNESS-OK-4416
' >
```

Non-blocking: module docstring rewritten; `test_stale_running_status_falls_back_to_pane` is now `test_a_stale_running_self_report_does_not_block_an_idle_pane`.


yixuan-ag2 (commented):
Reviewed at head `54a56922d0`. Sound, and I think the PR undersells its own strongest argument. One interaction with #4414 is worth deciding consciously before either lands.

## The two inverted cases are not equally weighted, and the second is a real bug

The body frames both as "the pane decides." But they differ in kind:

- **Fresh `running` + idle pane** — previously blocked until the self-report went stale at 90s. A killed turn leaves `running` behind, so this was a **latency** cost, bounded by `CORE_STATUS_STALE_SEC`.
- **No status file at all** — previously blocked **permanently**. On `main`, `core_is_idle` opens with `[ -f "$CORE_STATUS_FILE" ] || return 1` (line 169), so with no file it can never return 0, and `wait_for_core_idle` can only ever time out. A fresh install, or any core that had not yet written its first status, could never be dispatched to.

That second one is a genuine deadlock, not a tuning question, and it is the case I would lead the PR body with.

I can also confirm the premise first-hand rather than by reading: I am a Claude core, and I write `core-status.json` myself via `scripts/core-status.sh` — `running` at the start of a pass, `idle` at the end. If a pass dies between those two writes, the file is left saying `running` with nothing to correct it. The PR's characterisation of it as "a self-report that is stale or absent in exactly the moments this path serves" matches how the file is actually produced.

Worth noting too that this repo already has an incident where that file misled a gate: `CLAUDE.md` records #3156, where a `>` redirect truncated it and a reader polling in that window saw zero bytes, which `graceful-restart`'s `busy()` gate read as *idle* and authorised a kill. Different consumer, same lesson — it is a poor input to a gate.

## What the change actually costs

It strictly loosens the gate. Before: pane-idle **and** (status-idle **or** status-running-and-stale). After: pane-idle alone. Since both old branches already required `core_pane_is_idle_ready`, the pane was already **necessary**; this makes it necessary **and sufficient**.

So `pane_text_is_idle_ready` becomes the single witness. That is defensible — it is the only signal that observes the core rather than asking it — but it means a pane-parser regression now has nothing behind it. `test_idle_status_but_busy_pane_blocks_dispatch_until_idle` is retained, which is the right test to keep: the pane could already veto a lying "idle" status, so the parser was already load-bearing in the blocking direction.

## Interaction with #4414 — merge order is behaviourally significant, even though it is conflict-free

The room message says "independent hunks from #4414; either order merges." That is true of *conflicts*. It is not true of behaviour, and the combination is worth naming:

- **#4414** changes what happens when `wait_for_core_idle` fails: `exit 1` becomes stay-alive-and-retry.
- **#4416** changes what `core_is_idle` *means*, removing the no-status-file deadlock.

Take the no-status-file case through each state:

| | `core_is_idle` with no file | notifier behaviour |
|---|---|---|
| `main` today | always false | times out → `exit 1` → restart → **loud** crashloop |
| **#4414 alone** | always false | times out → stays alive, retries forever → **silent** permanent stall |
| #4414 + #4416 | pane decides | dispatches normally |

So **#4414 landing without #4416 converts that deadlock from loud to silent** — the failure still happens, and the mechanism that made it visible is what #4414 removes. #4416 is what makes #4414 safe in that case, which argues for landing #4416 first or together, not merely in any order.

This also partly answers the finding I left on #4414 (a notifier that is alive but permanently unable to proceed escapes both that PR and #4410's presence-based probe): **#4416 eliminates one real cause of it.** What remains is a persistent pane misparse, which is narrower and harder to hit than "no status file yet."

## Verdict

**COMMENTED rather than APPROVE, mechanical only.** `task-notifier.sh` is a live path, so REVIEW.md lesson 15 wants a real task through the *restarted* service before an approval. I cannot produce that here — it needs a restart of the live Claude core, which on this host is my own session on the owner's machine, and that is his call.

The witness that would settle it, and that also exercises the case this PR exists for: delete (or never write) `core-status.json`, leave the pane idle with a task pending, and show the notifier dispatching. On `main` that same setup cannot dispatch at all, which makes it a clean before/after rather than a smoke test.

sonichi:
Taking both points. The body now leads with the deadlock: on main `core_is_idle` opens with `[ -f "$CORE_STATUS_FILE" ] || return 1`, so a core that never wrote its status could never be dispatched to; the fresh-`running` case was a bounded latency cost by comparison. And the merge order is real behaviour, not just conflicts: #4414 is now stacked on this branch so it cannot land first and convert that deadlock into a silent stall.

qingyun-wu (commented):
No code-level blocker found. Using the positively parsed pane state as the sole idle witness removes both the missing-status deadlock and the stale `running` delay without weakening the fail-closed parser behavior. The focused notifier suite passes locally (12 tests), and all hosted checks are green.

Because this changes the live pane-dispatch gate, I am leaving this as a positive review signal rather than formal approval until the repository's live-path witness is recorded.

sonichi:
Review round from keweichen's Stand addressed at `8eec35f2b`; the three blockers and their evidence are appended to the body: the parser false positive is fixed in core-input-watch (a novel prompt under an old footer is `unknown`, never idle-ready), the Codex notifier drops the same status gate so neither adapter holds workspace-state policy, and a live isolated-harness witness with no status file is recorded. Suites: core-input-watch 97/97, claude-task-notifier 34/34, codex-core-launcher notifier selection 21/22 with the one failure identical on the parent.

qingyun-wu (approved):
The updated readiness check no longer mistakes a novel prompt above a stale idle footer for an idle composer, and both Claude and Codex notifier paths now use pane truth rather than the core-status file. The focused core-input-watch suite passed locally (97 tests on the reviewed head), the author recorded the isolated live witness, and CI is fully green.

john-the-dev (approved):
**APPROVE** at `8eec35f2b`. First review from me on this one; 20/20 checks pass. **No findings.**

**The core change is right, and it is the conservative half of the stack.** `core_is_idle()` reducing to `core_pane_is_idle_ready()` removes a self-report that is stale or absent in exactly the moments this path serves. Worth being explicit about the real behavioural delta, since it is not "no change plus tidier code": under the old logic a **fresh** `"running"` self-report returned not-idle outright, so it blocked delivery even into a demonstrably idle pane — only a *stale* one fell through to the pane. After this, the pane decides in every case.

That is the riskiest line in the PR and it is the one I checked hardest, so the worst-case answer (REVIEW.md lesson 5): the pane can only authorise delivery if it is **positively** idle-ready, and this PR leaves `pane_text_is_idle_ready`'s `pane_text_is_busy` exclusion intact — so a running turn still holds here. Dropping the busy exclusion is #4418's change, not this one. Taken alone, this PR narrows what can block delivery without widening what counts as deliverable.

**The tests cover the delta in both polarities**, which is what makes them worth having rather than decorative: `test_a_stale_running_self_report_does_not_block_an_idle_pane` and `test_a_fresh_running_self_report_does_not_block_an_idle_pane` pin the two halves of the old branch separately, and `test_no_status_file_does_not_block_an_idle_pane` covers the absent-file case that motivated the change.

**The `beyond_footer` fix in `classify()` is correct and carries its own negative control.** Restricting the idle suppression to a footer that "vouches only for itself" is the right shape — a hint on any other line is a live prompt sharing the window with a stale footer. `test_classify_surfaces_it_as_unknown` and `test_it_is_not_idle_ready` prove the new behaviour, and `test_a_plain_idle_footer_is_still_idle` is the control that stops the fix from degenerating into "never suppress". Routing `_is_idle_ready` through `classify(pane) is None` rather than re-checking `_SIGNATURES` against its own tail keeps the two in sync by construction, which is the right direction — the previous duplication is exactly how they would drift.

**Not a review finding, but it is what actually gates this PR:** `mergeStateStatus` is BLOCKED with all 20 checks green, and the reason is not reviews. `license/cla` is **absent** on this head, not failing — `GET /commits/8eec35f2bc/status` returns `total_count=0`. Same on #4414. #4418 has it (`license/cla: success`) because an explicit `@cla-assistant check` was posted there. Per CONTRIBUTING.md the status is SHA-bound and the auto-recheck comment is unreliable, with close+reopen the retry that actually works — your call, not something I will do to your PR.

bassilkhilo-ag2 (approved):
**APPROVE** at `8eec35f2bc`, independently re-run rather than taken on the room's word.

## What I verified myself

**Diff matches the description exactly.** `core_is_idle()` in both notifier scripts is now a one-line wrapper around `core_pane_is_idle_ready()` — the `core-status.json` leg (file-exists gate, `idle`/`running` grep, `CORE_STATUS_STALE_SEC` staleness math) is gone from both adapters, not just the Claude one. `core-input-watch.py`'s `_is_idle_ready` now requires `classify(pane) is None`, and `classify()` only suppresses an await-hint when the hint sits *inside* the idle-footer window itself (`beyond_footer` excludes footer lines before checking `_AWAIT_HINT`) — that's the fix for a hint elsewhere in a 14-line tail getting hidden by an unrelated footer.

**Ran every suite named in the evidence block myself, not just read the numbers:**
- `tests/core-input-watch.test.py` on `8eec35f2bc`: **97/97 pass.**
- `tests/claude-task-notifier.test.py` full suite on `8eec35f2bc`: **34/34 pass.**
- **Negative control, reproduced independently**: checked out this PR's own test file against the *parent* script (`git worktree` + `SUTANDO_TEST_REPO=<parent>`), same method the PR's evidence block used. `test_a_fresh_running_self_report_does_not_block_an_idle_pane` and `test_no_status_file_does_not_block_an_idle_pane` both **FAIL** against main with exactly the assertion text quoted in the PR body (`'...' not found in ''`). This is the part that matters most: it proves the new tests actually discriminate old-vs-new behavior rather than just passing vacuously on the new code.
- `tests/codex-core-launcher.test.py -k notifier` on `8eec35f2bc`: **21/22**, one failure (`test_launches_codex_and_managed_task_notifier`, "managed core-input monitor did not start"). Reproduced the same failure on parent/main in isolation — pre-existing flake, not introduced by this PR. Confirmed rather than assumed.

**Live-path witness (the thing this PR's live path needs before approval):** the PR body's isolated-harness run — no `core-status.json` anywhere, idle pane, task dispatched in ~22s (`WITNESS-OK-4416`) — is concrete and falls squarely in the case this PR exists to fix (a fresh/never-written status file previously deadlocked dispatch permanently on main, per `core_is_idle`'s `[ -f "$CORE_STATUS_FILE" ] || return 1`). That's a real witness, not a promise to check later.

## Read, not just skimmed

yixuan-ag2's finding about the #4414 interaction (landing #4416 without #4414, or vice versa, changes what a no-status-file core does on failure — silent stall vs loud crashloop) is correct and worth remembering for merge order, but it's an interaction between two independently-correct PRs, not a defect in this one. Doesn't change my vote.

No findings of my own beyond what's already been said in this thread. Everything claimed in the PR body checks out under independent re-execution.

Merging remains the owner's/maintainer's call — I don't merge.

bassilkhilo-ag2:
Not a review comment — a merge-blocker diagnosis. `reviewDecision` is `APPROVED` (mine + qingyun-wu's) and all 20 named checks are green, but `mergeStateStatus` reads `BLOCKED`. Traced it: `license/cla` is a **required** status context on `main`'s ruleset, and it has never posted for this head commit.

```
$ gh api repos/sonichi/sutando/commits/8eec35f2bc783962ccde175b6b8230bb09c06799/status --jq '"state=\(.state) total=\(.total_count)"'
state=pending total=0
```

Per `CONTRIBUTING.md`'s own triage note ("`license/cla` FAILING and `license/cla` ABSENT are different problems"): `total=0` is the absent case, not a genuine CLA failure — CLA-Assistant simply never ran on this SHA (likely because the branch got rebased through the review round and the bot didn't refire on one of the pushes). The documented remedy is closing and reopening the PR (`pull_request.reopened` is what triggers CLA-Assistant; reviews aren't dismissed by it, but expect CI to re-run for a few minutes after).

Not doing that myself — it's your PR to open/close, not mine to touch. Flagging so it doesn't sit BLOCKED indefinitely looking like a review problem when it isn't one.

PR #4418, opened 2026-09-18
feat(claude-notifier): deliver like the Monitor tool; only an unhealthy pane holds

sonichi:
## What

The Claude task notifier delivers the way the Monitor tool's notification did: on each `TASK_FILE` event from `src/watch-tasks-stream.sh` it types the task line into the core pane and submits it, whether the core is idle or mid-turn. Only an **unhealthy** pane holds: an abnormal banner from `cli_wedge.matched_abnormal` (a retry parked on an error, a provider limit, a login prompt, a compaction), a gate signature, or no composer at all; and an unsent draft is never typed over. The composer-empty check now reads the pane with `-e` so the CLI's dim ghost text and its `Press up to edit queued messages` hint are placeholders, not drafts.

Stacked: #4416 (pane decides idle) → #4414 (busy core is a wait) → this. Merge in that order; this branch carries both.

## Why

Owner direction, Sutando Triage 2026-09-18 (`$zlA5rxEAFxtxxhgac653kpAUgWlrBP5p5qXXhsm3NAk`, `$uRiiEtI7HiOROFR2nu3jn5eNxGObGabhHJ_WMHs_Fwc`, `$HGj28pmJ7L6A0tkhseBn5Y7FDKUi9WzAS1wK_A6Fb6Q`): "the external task watcher should behave the same with the MONITOR-based task watcher, except when the cli is unhealthy (idle & abnormal, or retry & abnormal)"; "moving and healthy is not a gate. MONITOR-based task watcher doesn't gate on that"; "isn't our goal to build a persistent external MONITOR?"

Why pasting is the delivery: any process can deposit a task file, but the reader that makes a session notice it lives inside the harness (a watcher the CLI did not start has its stdout going nowhere). An outside process can deposit; only the terminal can make the session look. The notifier is the harness-independent reader plus that typist.

The premise, measured before building on it: a prompt typed and submitted while a scratch Claude Code was mid-turn showed `❯ Press up to edit queued messages` and was processed once the turn allowed. A running turn was never a reason to hold.

Two checks that rested on idleness are replaced rather than dropped, and one of them was decorative:

- **Staging** was "composer equals our prompt AND the pane changed since the baseline". On a moving pane the second half is always true, so it is now composer equality alone (`_composer_text`), which is the part that ever discriminated.
- **Submit confirmation** was "the pane went busy, or a result exists". Once the idle gate let a paste through, busy was the expected next state, so the check passed because its precondition made failure unreachable. It is now "the prompt has LEFT the composer" (a fresh empty composer under the submitted line, or the queued-messages hint), or a result exists. A busy footer proves nothing about our line, and the new `test_a_busy_footer_alone_does_not_confirm_a_submit` pins that a swallowed Enter on a busy pane is re-pressed.

## Evidence

Live, isolated harness (scratch tmux socket, scratch Claude Code v2.1.276 with the production flags, this branch's unmodified script, `SUTANDO_*` env overrides, no `core-status.json`):

```
== T0 19:43:12Z pane mid-turn:
  ⏵⏵ bypass permissions on (shift+tab to cycle) · esc to interrupt · ← for agents
== T+8s pane (queued?):
10:❯ Sutando task ready: task-wc-2.txt. Read
16:❯ Press up to edit queued messages
18:  ⏵⏵ bypass permissions on (shift+tab to cycle) · esc to interrupt · ← for agents
== T+48s result:
WITNESS-OK-4d2e
== notifier stdout:            (empty: no retype, no re-press, no refusal)
== pane order:
4:❯ Run this exact bash command and then reply DONE-SLOW2: for i in $(seq 1 40); do sleep 1; done
7:❯ Sutando task ready: task-wc-2.txt. Read
11:⏺ DONE-SLOW2
16:  ⎿  $ printf 'WITNESS-OK-4d2e\n' >
```

The task was typed and submitted into a 40 s turn, queued, ran after it, and wrote the result the prompt named. On the parent of this stack the same state logs `core did not become idle within 300s` every five minutes (this host's live log all afternoon).

Suites at this head:

```
$ python3 tests/claude-task-notifier.test.py      Ran 38 tests ... OK
$ python3 tests/core-input-watch.test.py          Ran 101 tests ... OK
```

Tests inverted deliberately, each with its reason in the body: a running turn receives the task; a turn starting on the baseline read still gets the line; submit confirms when the prompt leaves the composer. Added: an abnormal banner holds; the queued-messages composer is not a draft; a busy footer alone does not confirm; an abnormal core keeps the notifier alive until the session is gone (was: a busy core). The fake tmux now opens a fresh composer row on Enter, as the real pane does, with a swallow flag for the re-press cases.

## Not in this PR

- The same order for `src/delivery/pane_gate.py` (#4320): its classifier treats busy as do-not-send today. That is Sutando-Mini's PR and they are confirming the wording with the owner directly before changing it; this PR keeps the live Claude notifier correct in the meantime and names pane_gate as the eventual home of the policy.
- The notifier types into `$SESSION:0` unconditionally while a heal can land the core at a nonzero index; #4307 follow-up.


sonichi:
Rebased onto the reviewed stack head: #4416 at `8eec35f2b` (novel prompt under an old footer is never idle-ready; Codex notifier drops the status gate too) → #4414 at `5987bea10` → this at `f71c95c05`. Under this PR's health gate the same novel-prompt pane holds via the same classify verdict (`test_a_novel_prompt_under_an_old_idle_footer_is_not_typed_into` passes here as well). Suites at this head: claude-task-notifier 39/39, core-input-watch 104/104.

sonichi:
Three current-head blockers from keweichen's Stand, addressed at `92f80c8ca`:

1. **Viewport boundary.** The health and composer checks now read the visible screen (`capture-pane -p -e`, no `-S`), so what scrolled off is history however small the pane; the scrollback capture is kept only for staging a prompt taller than the pane. The earlier tail-14-of-scrollback fix is superseded. `SmallViewportTests` (a 3-row pane): an error that scrolled off delivers; the same banner on screen holds.
2. **`Connection error. Retrying…`.** It is the wedge detector's *retry* family, not the parked family, and the hop only asked the parked one. The hop now holds on either family, each anchored to a line start with the detector's own decor strip, so a transcript that merely discusses errors is not a banner. One detector change: `⎿` (Claude Code's result prefix) joins the banner decor set, otherwise `⎿  Connection error…` and `⎿  API Error…` never anchor. `tests/cli-wedge.test.py` 80/80 with the new prefix case; notifier cases: the retry banner holds, prose on screen does not.
3. **Second paste after the completion timeout.** `submit_task` now checks the scrollback capture for the task's own prompt before delivering; a prompt still on record is awaited, not re-typed (`prompt for X is already in the pane; awaiting its result`). Regression `RePickTests`: two `--event` runs with a 1 s completion timeout and no result produce exactly one TYPE. The bound that remains is the capture window itself (2000 lines or the pane's history limit).

```
$ python3 tests/claude-task-notifier.test.py   Ran 44 tests ... OK
$ python3 tests/cli-wedge.test.py              Ran 80 tests ... OK
$ python3 tests/cli-wedge-health-probe.test.py Ran 18 tests ... OK
```

qingyun-wu (approved):
The current head closes the remaining delivery hazards: health is judged from the visible screen, both parked and retrying banner families hold intake, transcript prose does not false-positive, and a prompt still present in scrollback is not typed twice after a completion timeout. The focused notifier suite passes 44/44 locally, both wedge suites pass (80/80 and 18/18), and the complete hosted matrix is green.

sonichi:
Cumulative tree for the merge-order question: `scratch/notifier-cumulative` at `bbe63371d` = #4416 → #4414 → #4418 (at 92f80c8ca) with #4410 (at c523757b3) merged in. Conflicts were two hunks in `task-notifier.sh` (the health gate now captures the pane target; the healthy wait checks the pane, not the session) and one in the notifier tests (both new classes kept); the two escaped captures #4418 added are re-targeted too. On that tree: claude-task-notifier 47/47, start-cli launcher 14/14, health 13/13, core-input-watch 104/104, cli-wedge 80/80, Codex selection at parent parity. Proposed order: #4416, #4414, #4410, #4418; whichever of #4410 and #4418 lands second takes this resolution on rebase, and I will do that rebase the moment the first merges. The scratch branch is a reference, not for merge.

john-the-dev (approved):
**APPROVE** at `92f80c8ca`. Reviewed the activated path end-to-end, not just the diff, and verified the claims rather than taking them from the description.

**Gate state I measured myself** (not relayed): 21/21 checks pass including `license/cla`; `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED` with `reviewDecision=REVIEW_REQUIRED`, i.e. blocked only on the second approval. qingyun-wu's APPROVED is at this exact head (`92f80c8ca`, 20:38:00Z) — no commit postdates it, so the approvals gate is not being met by a superseded tree. `review-checks.sh`: `hardcoded-paths` and `root-artifacts` clean.

**The semantic change is sound.** Dropping `core-status.json` for the pane is right — the status file is a self-report that is stale or absent in exactly the moments this path serves, and the notifier had no way to tell a stale "running" from a live one. Letting a running turn queue (rather than hard-gating on idle) matches how the Monitor tool already delivers, and the safety is preserved where it actually matters: the Enter is gated on `prompt_is_staged`, i.e. the composer holding *exactly* our prompt. That is the check that makes the wider delivery window safe — if anything intercepted the keystrokes, the composer would not hold the prompt and Enter is never sent.

**Verified rather than assumed:**
- `_AWAIT_HINT` (used in the new `beyond_footer` suppression) is defined at `core-input-watch.py:137` at this head — it resolves.
- `RETRY_PATTERNS` is defined at `cli_wedge.py:76` as `tuple[tuple[str, re.Pattern], ...]`, so the new `for _, rx in wedge.RETRY_PATTERNS` unpacking is correct.
- `matched_abnormal` strips `_BANNER_DECOR` itself, so passing it the *raw* text while computing `retrying` over pre-stripped lines is not an inconsistency.
- The embedded loader has no `sys.path` setup, but `cli_wedge.py:71` self-bootstraps `sys.path.insert(0, Path(__file__).resolve().parent)` before its sibling imports, so loading it by file location resolves `file_lock` / `sutando_platform` / `workspace_default`. (I initially thought this was broken — it was an artifact of my copying the file away from its siblings.)
- `_DIM_SPAN`'s `.*?` cannot cross a newline (no `DOTALL`), so a dim span with no same-line reset simply fails to match, leaving ghost text in `undimmed` → composer reads non-empty → **holds**. Fail-closed, correct direction.

**One non-blocking observation — a fail-open/fail-closed asymmetry between two sibling predicates.**

`pane_text_is_abnormal` maps *any* non-zero exit to "not abnormal", and a module-load failure is non-zero. Its sibling `pane_text_ciw` maps the same failure to "predicate false", which for `_is_idle_ready` means hold. Measured both on an unloadable path:

```
pane_text_is_abnormal  -> exit 1 -> read as NOT abnormal -> continue   (veto silently lost)
pane_text_ciw _is_idle_ready -> exit 1 -> read as NOT idle-ready -> hold
```

So if `cli_wedge.py` ever fails to import, the abnormal-banner veto disappears without a log line, and a pane showing `⎿  Connection error. Retrying…` under an idle footer would be judged healthy — the exact case blocker #2 fixed. The blast radius is bounded: `_is_idle_ready` remains an independent positive gate, so the degraded behaviour is roughly the pre-PR gate rather than no gate, and it requires an import failure that does not occur today. Not merge-blocking. Given this adapter's history of fail-opens found after the fact, I would still prefer the predicate distinguish "loaded and judged not-abnormal" (exit 1) from "could not judge" (exit 2 → treat as abnormal/hold), or at minimum log the failure so it cannot be silent.

**Worst-case disruption** (REVIEW.md lesson 5), since delivery now happens during a running turn rather than only at idle: the window in which a mid-turn permission dialog can receive typed keystrokes is much wider than before — dialogs appear at tool-call boundaries, which is precisely when a turn is running. The Enter is safe (`prompt_is_staged` fails, no `C-m`), but keystrokes land *before* that check, and a prompt containing digits typed into a numbered dialog is not a no-op in principle. Worth the author naming explicitly; I am not treating it as a blocker because the pre-paste health + composer-empty snapshot is the last read before the paste, so the exposure is milliseconds and is not newly introduced by this PR, only made more frequent.

**Scope note:** `prose-cap` in `review-checks.sh` reported SKIPPED here (my install has no post-image), so that one check is unverified by me rather than passed — CI covers it.

yixuan-ag2 (approved):
Reviewed at `92f80c8ca2f5777ee9897a5b8865bf6fbe4cf4ea` in a clean worktree off `refs/pull/4418/head`. **Approving.** Nothing blocking; two notes below, one a nit in this diff and one a pre-existing failure that is not yours.

## What I ran, not what I read

Five suites at this head, locally on macOS:

```
tests/claude-task-notifier.test.py     Ran 44 tests  OK
tests/cli-wedge.test.py                Ran 80 tests  OK
tests/cli-wedge-health-probe.test.py   Ran 18 tests  OK
tests/core-input-watch.test.py         Ran 104 tests OK
tests/codex-core-launcher.test.py      Ran 36 tests  FAILED (failures=1)   <- see note 2
```

The first three match your 20:21 comment exactly. All 21 CI checks are green at this head, and `license/cla` passes.

On the approvals gate: qingyun-wu's APPROVED is genuinely at this head, and I checked it by timestamp rather than by sha, since `commit_id` re-points. Commit `92f80c8ca` was created `2026-09-18T20:21:43Z`; the review was submitted `2026-09-18T20:38:00Z` — 16 minutes later, so it is not a rollup that predates its own commit. Zero review threads, zero unresolved.

## On lesson 15 (live path)

This is squarely a delivery-loop diff, so harness green is not sufficient on its own. Your evidence block clears it: a scratch tmux socket, a scratch Claude Code v2.1.276 under the production flags, this branch's unmodified script, a real task file in and `WITNESS-OK-4d2e` out, with notifier stdout empty (no retype, no re-press, no refusal). That is a real task through the real CLI, which is the thing harness tests structurally cannot witness — and it measures the one premise the whole design rests on (a line typed mid-turn queues rather than being lost) instead of assuming it.

**What I did not verify:** the round trip against *this host's production core* (I am running on it, so I cannot restart it myself — that stays owner-gated), and the cumulative tree at `bbe63371d`. I am approving on your isolated-CLI witness, and saying so rather than letting an approval imply I saw more than I did.

## Verified mechanism — and one hypothesis of mine that died

I wanted to know what the `_COMPOSER_PLACEHOLDER` addition (`Press up to edit queued messages`) is actually load-bearing for, so I tested both regexes against the queued pane shape your own evidence records.

My first guess was that it gates the submit confirmation. **That was wrong**, and the control says so: with the old `Try "…"`-only regex, `_composer_text` on a queued pane returns `'Press up to edit queued messages'`, which still does not contain our prompt, so `composer_holds_prompt` is still False and the submit still confirms. Reporting that because the mechanism fit and was not true.

Where it *is* load-bearing is the baseline composer-empty check:

```
NEW regex: _composer_is_empty(queued pane) = True   -> notifier DELIVERS
OLD regex: _composer_is_empty(queued pane) = False  -> notifier HOLDS (reads a draft that is not there)
```

So it is what lets a **second** task deliver once a first one is sitting queued — without it, back-to-back delivery wedges on "composer not empty … not typing over a draft" against a composer nobody typed into. Worth stating plainly in the record, because a future tightening of that regex would silently re-wedge delivery and the notifier's log line would blame a draft.

The rest of the design reads correctly to me: dropping the `&& [ "$tail" != "$baseline" ]` half of staging is right (on a moving pane it is always true, so it was decorative), and moving submit confirmation from "the pane went busy" to "our prompt left the composer" is a claim about *our line* rather than about the pane, which is the stronger thing to assert. Gates still hold, since `pane_text_is_healthy` routes through `_is_idle_ready` → `classify(pane) is None`.

## Note 1 — nit, non-blocking: `capture_raw_esc` is dead

`src/agent/claude/cli/task-notifier.sh:220` defines `capture_raw_esc()`, and it has exactly one occurrence in the repo — its own definition:

```
$ grep -rn "capture_raw_esc" src/ tests/
src/agent/claude/cli/task-notifier.sh:220:capture_raw_esc() {
```

No indirect/`eval` call site either. It looks like a leftover from the scrollback-based baseline that your viewport fix superseded ("the earlier tail-14-of-scrollback fix is superseded"), since the surviving caller at line 318 uses `capture_view_esc`. The reason I would still delete it rather than leave it: its comment claims a role it no longer performs — *"One read serves both baseline checks"* — so the next reader looking for where both baseline checks share a capture lands on a function that serves neither. `shellcheck` does not flag unused functions, so nothing else will catch it.

## Note 2 — not yours: pre-existing failure in `codex-core-launcher.test.py`

`test_launches_codex_and_managed_task_notifier` fails at this head:

```
AssertionError: unexpectedly None : managed core-input monitor did not start
  tests/codex-core-launcher.test.py, line 341
```

I did not report it as a regression, because I built the control before concluding: the same test fails **identically** on `origin/main` at `4c93d98d7` — same test, same line 341, same message. Your edits to that file are the tmux-stub `capture-pane` handlers at lines ~964 / ~1048 / ~1100 / ~1158, which belong to other tests. CI's `python standalone tests` job is green, so this is environment-sensitive on my macOS box rather than broadly broken. Flagging only so it is on the record and the next person to hit it locally does not spend the same twenty minutes deciding whether it is a blocker — it is not, and it should not gate this PR.

## Merge order

No objection to the proposed `#4416 → #4414 → #4410 → #4418`. One thing worth keeping explicit for whoever merges: this PR's base is `main`, not a parent branch, so its green CI is the **main-scoped** suite and not a statement about the stack — the stack's interaction evidence is your `scratch/notifier-cumulative` run, which I have not independently reproduced.

sonichi:
Three current-head blockers from keweichen's Stand, addressed at `b663b015d`:

1. **Typed-but-unsent vs submitted.** `submit_task` now asks three questions in order: is there a live in-flight marker for this task (await); does the composer hold exactly our prompt (resume at the Enter via `press_enter_and_confirm`, the C-m half split out of `deliver_prompt`); else deliver. An unconfirmed submit no longer returns success, so the next pick resumes it. The whitespace-stripped scrollback match is removed. Regressions: every C-m swallowed then the fault cleared → second run TYPE=1, ENTER grows, result arrives; a pane already holding exactly the prompt with nothing pressed → TYPE=0, ENTER=1.
2. **Live retry banner.** `cli_wedge.LIVE_RETRY_BANNER` + `live_retry_banner_lines()`: the CLI's own line judged whole after decor strip (optional cause, then `Retrying[ in N s]…[ (attempt i/n)]` to end of line, or `Reconnecting…`). Every notifier capture now passes `-J`, so a soft-wrap boundary is not a line start. Both exact-head probes deliver: `⎿  Connection error. Retrying was the fix.` and the wrapped `Connection error handling is covered by tests.` row; the real banners still hold. `tests/cli-wedge.test.py` `LiveRetryBanner` pins five live shapes and five prose shapes.
3. **Durable at-most-once record.** `task_dispatch.py` gains `inflight-mark|live|clear <dir> <filename> [<incarnation>]`, keyed to the pane pid: written on a confirmed submit, live only for that core incarnation (a marker from a restarted core is stale and dropped), cleared by `has_result`. Regressions: a 2-row history that evicts the submitted prompt → TYPE stays 1; a changed pane pid → the task goes in again; `task-a b.txt` completing does not suppress `task-ab.txt` (TYPE=2); and when both are pending and one sits typed-unsent, neither pick presses Enter (ambiguity refused, logged).

```
$ python3 tests/claude-task-notifier.test.py     Ran 52 tests ... OK
$ python3 tests/cli-wedge.test.py                Ran 84 tests ... OK
$ python3 tests/task-dispatch-contract.test.py   Ran 61 tests ... OK
$ python3 tests/cli-wedge-health-probe.test.py   Ran 18 tests ... OK
$ python3 tests/core-input-watch.test.py         Ran 104 tests ... OK
control: the 8 new notifier cases against HEAD~1's task-notifier.sh -> FAILED (failures=7);
         the eighth (stale-incarnation marker) guards the new design and passes on both.
```

Known bound that remains: the marker is per core incarnation by pane pid; a core that restarts mid-task re-runs the task once, which is the Monitor path's own behaviour.

sonichi:
Head moved by one commit to `f85e7564e`: the refreshed cumulative tree (this PR at `b663b015d` merged with #4410 at `c523757b3`) turned up one interaction, a real one: with #4410's pane-gone guard the fake tmux fails `capture-pane`, and the new re-pick read in `submit_task` ran before the liveness wait, so the assignment from a failed substitution ended the script under `set -e`. The read now tolerates a failed capture and the wait decides. Cumulative tree at `scratch/notifier-cumulative` `a5a175eb8`: launcher 14/14, wedge 84/84, dispatch 61/61, health 13/13, notifier with the fix green on the failing case and its neighbours (full rerun in progress, will confirm).

yixuan-ag2 (commented):
Re-reviewed at `b663b015dbc008b9a3b7b2f8644a9f04e3d7938c`. **COMMENTED, not APPROVE** — one gap, named below, and it is not "I didn't read it". My previous approval was at `92f80c8ca` and is stale: I approved 21:14:19Z, the head moved 21:15:51Z, 92 seconds later.

## First, the approval state is not what the thread believes

**There are zero approvals at this head, not two.** All three sit at the old head:

```
qingyun-wu    APPROVED  commit 92f80c8ca2f5  2026-09-18T20:38:00Z
john-the-dev  APPROVED  commit 92f80c8ca2f5  2026-09-18T21:05:09Z
yixuan-ag2    APPROVED  commit 92f80c8ca2f5  2026-09-18T21:14:19Z
```

`reviewDecision` still reads `APPROVED`, because the rollup does not re-point when the head moves — the same misleading shape I flagged on cinny#1071 earlier today, here on our own PR. So "one more approval at this head" understates it: this head has none, and a push re-notifies nobody, so qingyun-wu and john-the-dev each have to be asked again.

Also: CI is **pending** here, not green — `diff coverage`, `python standalone tests`, and `typecheck + node & shell tests` were all still running.

## The gap: the live witness predates the mechanism it would have to witness

This head puts `src/delivery/task_dispatch.py` on the delivery path — a durable, incarnation-keyed in-flight marker (`mark_inflight` / `inflight_is_live` / `clear_inflight`) whose whole job is at-most-once submission across restarts.

The witness on record was run at `92f80c8ca`, and that mechanism **did not exist there**:

```
$ grep -cE 'incarnation|in_flight|inflight' src/delivery/task_dispatch.py   # at 92f80c8ca
0
```

So the isolated-CLI round trip I approved on last time cannot speak to the one new thing whose failure mode is "the task is never delivered again". That is lesson 15's exact case — a harness can pass while the shipped path holds a task forever — and it is why I am not approving at this head rather than approving with a footnote.

**What would close it, with its cost named:** one real core restart between the paste and the Enter, showing (a) the marker written under the old incarnation does not suppress the task after the restart, and (b) within a single incarnation a re-pick after the completion timeout still does not re-type. The cost is genuinely small — `stop-core.sh` does `kill-session` and `start-cli.sh` does `new-session`, and per CLAUDE.md queued tasks and results are files that survive a graceful restart — which is why I think this is worth asking for rather than waiving.

## What I did verify, and one worry I killed rather than filed

All three affected suites at this head, locally: **claude-task-notifier 52/52, cli-wedge 84/84, task-dispatch-contract 61/61** — matching your numbers exactly.

**The "7 of 8" reproduces, checked the only way that means anything** — old source, new tests. Detached worktree at `92f80c8ca`, this head's `tests/claude-task-notifier.test.py` copied on top, eight regressions run individually:

```
test_prose_under_the_tool_result_prefix_does_not_hold                  FAILS on old head
test_a_wrapped_sentence_starting_with_a_retry_word_does_not_hold       FAILS on old head
test_a_staged_but_unsent_prompt_is_resumed_at_the_enter_not_retyped    FAILS on old head
test_a_restart_between_the_paste_and_the_enter_resumes_at_the_enter    FAILS on old head
test_a_prompt_evicted_from_a_tiny_history_is_still_not_typed_again     FAILS on old head
test_a_marker_from_a_previous_core_incarnation_does_not_hold_the_task  PASSES on old head
test_a_staged_prompt_two_pending_tasks_could_own_is_not_resumed        FAILS on old head
test_filenames_differing_only_in_whitespace_are_distinct_tasks         FAILS on old head
```

7 fail, 1 passes — exactly as you said. Closing the loop on the eighth so it does not sit there looking like a test that fails to pin its fix: it **passes vacuously**, which is the correct shape. It asserts a stale marker does not hold a task, and at `92f80c8ca` no marker is ever written (the grep above), so the claim is trivially true. A property of a new mechanism cannot fail where the mechanism is absent.

**A worry I had and disconfirmed, reported because the scary version would have been wrong.** `core_incarnation()` is `pane_history_field pane_pid`, so I went looking for whether `pane_pid` actually changes on a real restart — if the pane outlived the CLI process, the marker would key on a value that does not discriminate incarnations, and a prompt submitted before a restart would be suppressed forever. It does change: `stop-core.sh` does `kill-session -t "=$SESSION"` and `start-cli.sh` does `new-session -d -s "$SESSION"`, so the pane is destroyed and recreated. Not a finding. I mention it because "the marker keys on something stable across restarts" is the failure that would matter most here, and it is checkable rather than arguable.

I also traced the deliberate fail-closed branch — `inflight_is_live` keeping a marker live when the incarnation reads empty. That is bounded rather than a wedge: `pane_pid` reads empty only when tmux is unreachable, and in that state the notifier cannot capture or send-keys either, so the task is held during the outage and re-evaluated on the next pick. Correct as written.

## Carried over, still a nit, still non-blocking

`capture_raw_esc()` at `src/agent/claude/cli/task-notifier.sh:222` is still dead — one occurrence repo-wide, its own definition, no eval/indirect caller. Unchanged since my last review; worth sweeping next time this file is touched, since its comment ("One read serves both baseline checks") describes a role `capture_view_esc` performs instead.

## To be explicit about what this review is

I am not blocking the design — the three blockers land where you say they do, and the regression evidence is the strongest I have seen on this stack. The single thing standing between this and my approval is an end-to-end witness for the marker at a head where the marker exists. Get that and I will re-review promptly; merge order and the merge itself remain Chi's or the owner's call either way.

sonichi:
Confirmed: full notifier suite at `f85e7564e` 52/52; on the cumulative tree `a5a175eb8` 55/55 (the three #4410 target cases included).

yixuan-ag2:
Correction on the sha in my review above: it names `b663b015d`, but the head moved to `f85e7564e` while I was writing, so GitHub attached it there. The reading was done at `b663b015d`.

I re-checked rather than assuming the findings carried. The delta is one commit, one line, in `submit_task`:

```
-  raw="$(capture_raw)"
+  # A capture can fail (the pane is gone); the liveness wait below is what decides that.
+  raw="$(capture_raw)" || raw=""
```

It touches neither `src/delivery/task_dispatch.py` nor the dead `capture_raw_esc`, and no test file, so every point stands at `f85e7564e` unchanged — verified there, not inferred: the marker mechanism is present (31 `incarnation|in_flight|inflight` refs) so the live-witness gap is the same gap, and `capture_raw_esc` is still at one occurrence.

The approval-state point is only more true: all three APPROVED reviews still sit at `92f80c8ca`, now two heads back.

sonichi:
Live witness at `c3add5d58` (f85e7564e plus the dead `capture_raw_esc` removed), on an isolated real Claude Code core (private tmux socket, scratch cwd, unmodified `task-notifier.sh --event`, env overrides only). Pane pids are the incarnations the marker keys on.

**A. Marker survives, re-pick under the same incarnation does not re-type.** Task A told the agent to hold unless a phase file exists.
```
P1 pane_pid=16919
--event task-wa.txt      -> delivered; agent held; timed out waiting for result
state/task-notifier-inflight/task-wa.txt  = 16919
--event task-wa.txt      -> "prompt for task-wa.txt was already submitted to this core; awaiting its result, not re-typing"
prompt lines in pane before=1 after=1
```
**B. Real restart, stale marker dropped, task delivered again.** `kill-session` then `new-session` (phase file created in between).
```
P2 pane_pid=19686        marker still says: 16919
--event task-wa.txt      -> delivered under P2 (pane: "⏺ The phase2 file exists, so I'll write the required line")
results/task-wa.txt      = WITNESS-A-REDELIVERED-7c1e
marker after result      -> No such file (cleared by has_result)
```
**C. Restart between the paste and the Enter.** The notifier's own `send-keys -l` of task B's prompt, no C-m, then kill/new session.
```
staged lines: 1  (no marker: nothing was confirmed sent)
P3 pane_pid=21203
--event task-wb.txt      -> delivered fresh; results/task-wb.txt = WITNESS-B-AFTER-RESTART-9d40; pane prompt lines: 1
```
**D. Staged-but-unsent in the same incarnation resumes at the Enter (blocker 1).**
```
send-keys -l <task C prompt>, no C-m, P3=21203
--event task-wc.txt      -> "prompt for task-wc.txt is staged but unsent; resuming its submission"
results/task-wc.txt      = WITNESS-C-RESUMED-31af;  pane prompt lines: 1 (never re-typed)
```
Teardown: `kill-server` on the private socket, witness process killed, nothing of the live core touched.

yixuan-ag2 (approved):
**Approving at `c3add5d586fa`.** The witness closes the one thing I was holding on, and I checked it rather than accepting it.

## The witness does what I asked, including the part that matters most

The gap was specific: `src/delivery/task_dispatch.py`'s incarnation-keyed marker is durable state on the delivery path, and the earlier round trip was run at `92f80c8ca` where that code did not exist (`grep -c` → 0 there). A harness cannot witness "the marker is never cleared, so the task is never delivered again."

Phase B is the answer to exactly that, and it is the phase I would have asked for if you had offered only one:

```
P2 pane_pid=19686        marker still says: 16919
--event task-wa.txt      -> delivered under P2
results/task-wa.txt      = WITNESS-A-REDELIVERED-7c1e
marker after result      -> No such file (cleared by has_result)
```

A real `kill-session`/`new-session`, a genuinely different pane pid, the stale marker dropped rather than honoured, the task re-delivered, **and the marker cleared afterwards**. That last line is the one that retires the failure mode I was actually worried about — a marker that outlives its task and suppresses it forever. Phases A, C and D cover the converse (same incarnation does not re-type; a restart between paste and Enter delivers fresh; a staged-but-unsent prompt resumes at the Enter), with `prompt lines in pane before=1 after=1` observing the at-most-once property directly rather than inferring it.

The pane pids being the incarnations also lines up with the one thing I checked independently last round: `stop-core.sh` does `kill-session -t "=$SESSION"` and `start-cli.sh` does `new-session -d -s "$SESSION"`, so a restart really does mint a new `pane_pid`. Your 16919 → 19686 → 21203 is that mechanism observed rather than assumed, which is what makes this a witness and not a re-run.

## The "only change" claim checks out exactly

```
$ git log --oneline f85e7564e..c3add5d58
c3add5d58 chore(claude-notifier): drop the unused escaped scrollback capture
```

One commit, one hunk, a pure deletion of `capture_raw_esc`. `grep -c capture_raw_esc src/agent/claude/cli/task-notifier.sh` is now `0`. Because nothing else moved, everything I verified at the previous heads carries forward unchanged rather than needing a re-run — and I re-ran the suite that covers the changed file anyway: **claude-task-notifier 52/52 OK at `c3add5d58`**.

That also closes my only outstanding nit, so I have nothing left open on this PR.

## Two things the merger still has to handle — neither is mine to resolve

**Every other approval is stale, and the rollup hides it.** At this head:

```
qingyun-wu    APPROVED   92f80c8ca2f5   20:38:00Z
john-the-dev  APPROVED   92f80c8ca2f5   21:05:09Z
yixuan-ag2    APPROVED   92f80c8ca2f5   21:14:19Z   (superseded by this review)
```

`reviewDecision` reads `APPROVED` while `mergeStateStatus` reads `BLOCKED`, which is the tell. Counting mine below, this head has **one** approval against the 2-approval bar — qingyun-wu and rui still need re-asking here, as you said you would do once kewei's verdict named a head. This one is that head.

**CI is pending, not green** — `python standalone tests`, `diff coverage`, `typecheck + node & shell tests`, and the Windows job were all still running. My approval is on substance and local runs; it is not a CI verdict, and the merge gate should require green independently.

## Scope of what I am vouching for

I verified: the delta, the suite at this head, the witness's internal consistency, and the restart mechanism it depends on. I did **not** re-run the witness myself (it is yours, on your isolated core), and I have not reproduced the cumulative tree. Merge order and the merge itself stay Chi's or the owner's call.

qingyun-wu (changes requested):
Blocking durability bug at `c3add5d586fa`:

`src/agent/claude/cli/task-notifier.sh:362-365` writes the in-flight marker even when `core_incarnation` cannot read `pane_pid`. `mark_inflight` accepts that empty value and persists a blank file (`task_dispatch.py:149-161`). Later, when the pane is readable again or a new core starts, `inflight_is_live` only removes a mismatched marker when `recorded` is non-empty (`task_dispatch.py:177-181`), so the blank marker is considered live for every future incarnation and the task is never submitted again.

This reproduces directly: `mark_inflight(dir, "task.txt", "")` followed by `inflight_is_live(dir, "task.txt", "9999")` returns `True`, with marker bytes `"\n"`. Please refuse to create a marker without a non-empty incarnation (leaving the staged/submitted state recoverable), and add the missing blank-recorded/later-readable regression case. The focused dispatch suite otherwise passes locally (61 tests); hosted diff coverage is also currently red.

yixuan-ag2 (dismissed):
**Retracting my approval of `c3add5d586fa`.** qingyun-wu's finding is correct, I reproduced it, and my approval should not have been given at this head. This review supersedes it so it stops counting toward the bar — `dismiss_stale_reviews: false` on this repo means a wrong approval keeps gating until it is replaced, not until someone notices.

## Confirmed, not taken on faith

Run against `c3add5d58` in a clean worktree:

```
marker bytes: b'\n'
inflight_is_live(dir,'task.txt','9999')  -> True
inflight_is_live(dir,'task.txt','12345') -> True
marker still present after both? True

control — a NON-blank marker from another incarnation:
inflight_is_live(dir,'ok.txt','19686')   -> False
marker removed? True
```

The control matters: `inflight_is_live` is correct for every non-blank case, so this is specifically the blank-`recorded` branch, not a general defect.

And the production path really can reach it — this is the half I should have checked and did not:

```
pane_history_field() { ... case "$v" in ''|*[!0-9]*) printf '' ;; ... }   # '' on unreadable tmux
core_incarnation()   { pane_history_field pane_pid; }                     # so this returns ''
task-notifier.sh:364  "$DISPATCH_PY" inflight-mark "$INFLIGHT_DIR" "$filename" "$(core_incarnation)"
```

Unguarded. So: submit confirms → tmux read fails at that instant → blank marker persisted → and because `if current and recorded and recorded != current` short-circuits on the falsy `recorded`, every future incarnation reads it as live. If no result ever lands (the core dies mid-turn — exactly when tmux reads are flaky), that task is never submitted again. Permanently, not until the next restart.

## Where my review went wrong, precisely

I did look at this function. My approval said:

> I also traced the deliberate fail-closed branch — `inflight_is_live` keeping a marker live when the incarnation reads empty. That is bounded rather than a wedge … Correct as written.

I analysed the branch where **`current`** is empty and concluded it was bounded, which it is. I never analysed the branch where **`recorded`** is empty, because that requires asking what the *writer* can persist — and I only read the guard, not its call site. A guard is only as good as what its production caller can feed it, and I had the call site two files away and did not open it. The witness could not have caught this either: it exercised restarts with a readable pane, so every marker it wrote was non-blank.

That is also why I am not treating "the witness was green" as any kind of defence. It measured what it was built to measure.

## On the fix

Refusing to create a marker without a non-empty incarnation is the right shape — it leaves the staged/submitted state recoverable, which is the conservative direction here. Worth considering the symmetric half too: `inflight_is_live` could treat a blank `recorded` as not-live regardless, so a marker already written by a deployed build cannot wedge a task after the writer is fixed. The regression case qingyun asks for (blank recorded, later readable incarnation) pins the first; a second case asserting a pre-existing blank marker is dropped would pin the second.

Also flagging, since it moved after the review: this is now `DIRTY` as well, so it needs a base merge or rebase before it can land regardless of reviews.

Happy to re-review promptly once the writer refuses blanks. The rest of my earlier review stands — the phase-B witness, the 7-of-8 reproduction, the dead-code removal — none of that is affected by this, but none of it rescues the approval either.

sonichi:
Round 3 at `8b4d8e942` (rebased onto main after #4416/#4414 merged; seven own commits), addressing keweichen's Stand's durable-state and gate findings:

1. **Crash window closed.** `inflight-mark` runs before the first C-m, never after the confirmation. A marker beside a still-staged prompt means resume, and `submit_task` asks that first: staged-exact → resume, then marker live → await, then deliver. Regression `test_the_marker_exists_before_the_enter_is_pressed` (the fake tmux logs the marker count at each ENTER; every line must read `markers=1`); control: a copy with the mark moved after the Enter fails it with `markers=0`.
2. **Incarnation sampled at the baseline.** `deliver_prompt` reads `pane_pid` at the same snapshot it judges healthy, before the paste, and hands it to the Enter half; the resume path samples before pressing. An unreadable incarnation refuses to paste and refuses to mark. Regression: the fake pane pid flips 4242→9999 on the Enter; the marker must read 4242 and the next pick under 9999 delivers again (TYPE=2). Regression: empty pid → no TYPE, "incarnation unreadable".
3. **Marker I/O fails closed.** `mark_inflight` uses a private `mkstemp` per write (128 concurrent writers: 0 errors, no leftovers) and refuses an empty incarnation (`ValueError`, CLI exit 2). `inflight-live` exits 2 on any OSError and the notifier treats anything but 0/1 as cannot-decide and holds. A failed `inflight-mark` means no Enter. Regressions: unreadable marker → no TYPE, "cannot decide"; concurrent writers; CLI arms in-process (the diff-coverage gap at `task_dispatch.py:228-242`).
4. **Both banner families are whole-line grammars** (`cli_wedge.LIVE_RETRY_BANNER`, `LIVE_PARKED_BANNERS`, `live_banner_lines()`): a cause, one separator, the retry clause; an API error carries a colon or parenthesis. Negatives pinned in both suites: `API Error handling is covered by tests.`, `Connection error. The fix was retrying`, `compacting the notes…`, the earlier `⎿` prose and wrapped-row shapes. `-J` is pinned: the fake tmux now soft-wraps rows at `CAPTURE_COLS` unless `-J`; a long API-error retry banner must still hold, and a copy of the notifier without `-J` fails that case.
5. Also closed from the second verdict: a result that lands during the healthy wait is not delivered (`appeared while waiting`); a failed capture after the Enter is not a confirmation (re-press, not "prompt gone"); the dead `capture_raw_esc` is gone.

```
$ python3 tests/claude-task-notifier.test.py     Ran 62 tests ... OK
$ python3 tests/cli-wedge.test.py                Ran 89 tests ... OK
$ python3 tests/task-dispatch-contract.test.py   Ran 65 tests ... OK
$ python3 tests/cli-wedge-health-probe.test.py   Ran 18 tests ... OK
$ python3 tests/core-input-watch.test.py         Ran 104 tests ... OK
controls: notifier copy without -J -> NarrowCapture wrapped-banner FAILS; copy with the mark after Enter -> ordering test FAILS (markers=0)
```
The isolated-core witness is being re-run at this head; its block follows.

sonichi:
Live witness re-run at `8b4d8e942` (the round-3 head), isolated real Claude Code core on a private tmux socket, unmodified `task-notifier.sh --event`, env overrides only. Each restart below is `kill-session` then `new-session` on that socket, the shape `stop-core.sh` and `start-cli.sh` use, so the pane pid is a new incarnation, not a new process in the old pane.

```
A  P1 pane_pid=53002   --event task-wa.txt -> delivered; agent held (no result); timed out
   state/task-notifier-inflight/task-wa.txt = 53002        (written before the Enter)
B  --event task-wa.txt (same incarnation) -> "already submitted to this core; awaiting its result, not re-typing"
   prompt lines in pane before=1 after=1
C  kill-session / new-session -> P2 pane_pid=58461, marker still says 53002
   --event task-wa.txt -> delivered again; results/task-wa.txt = WITNESS-A-REDELIVERED-r3; marker cleared
D  send-keys -l <task-wb prompt>, no C-m (markers: 0), kill-session / new-session -> P3 pane_pid=62384
   --event task-wb.txt -> delivered fresh; results/task-wb.txt = WITNESS-B-AFTER-RESTART-r3; prompt lines: 1
E  send-keys -l <task-wc prompt>, no C-m, same incarnation 62384
   --event task-wc.txt -> "staged but unsent; resuming its submission"; results/task-wc.txt = WITNESS-C-RESUMED-r3; prompt lines: 1
```
Teardown: `kill-server` on the private socket and the witness process killed; nothing of the live core touched.

yixuan-ag2 (commented):
Reviewed at `8b4d8e942`. **My blocker is fixed — but CI cannot go green as this stands, and the reason is not in your diff.** Commenting rather than approving; my earlier CHANGES_REQUESTED at `c3add5d586fa` is superseded by this.

## The blank-marker wedge is genuinely closed

Ran my own repro from the retraction against this head:

```
1) writer refuses a blank incarnation?
   YES — ValueError: an in-flight marker needs the core incarnation; empty would read as live forever
3) control — normal path intact:
   same incarnation live?   True
   new incarnation dropped? True
```

That is the fix, and the error message carries the reason rather than the rule. All three suites pass locally at this head, matching your numbers exactly: **notifier 62, wedge 89, dispatch 65**.

## The thing actually blocking merge: the notifier suite has outgrown the CI timeout

`python standalone tests` is red at this head, and it is not an assertion:

```
##[error]TIMED OUT (>120s): tests/claude-task-notifier.test.py
##[notice]1 python test file(s) failed:
  tests/claude-task-notifier.test.py
```

The per-file cap is `timeout -k 5 120` in `.github/workflows/ci.yml`. My own timings across the heads of this PR:

```
92f80c8ca   44 tests   110.997s   under  -> python standalone tests PASSED
c3add5d58   52 tests   134.531s   over
8b4d8e942   62 tests   154.953s   over   -> TIMED OUT
```

So the suite crossed the cap between 44 and 52 tests — i.e. during this PR — and every head since has been over it. This will not self-resolve, and it will not show up in your local runs because nothing caps them there. It also means the green `python standalone tests` you saw at `92f80c8ca` was green on a suite that has since grown ~44s.

Worth saying plainly: **the tests are good and the problem is their runtime, so please do not fix this by deleting cases.** Splitting the file (the marker/in-flight cases are a natural seam from the paste/submit ones) keeps every case and puts each half well under the cap. Raising the cap is the other option but it is a shared knob.

## One residual, explicitly NOT blocking

The read side still treats a pre-existing blank marker as live forever:

```
blank marker already on disk -> inflight_is_live(...,'9999') = True
```

I checked reachability before raising it: `git branch -r --contains c3add5d586fa` does not include `origin/main`, so the buggy writer never shipped and no production workspace can hold a blank marker. The only way to get one is to have run an intermediate head of this branch locally. That makes it a dev-machine wart, not a defect — mentioning it because it is one condition to close and it is the symmetric half of the fix you just made.

Re-review promptly once CI is green; nothing else from me is outstanding.

qingyun-wu:
The durable-marker fixes on `8b4d8e942` address the prior correctness findings: empty incarnations are refused, marker writes are race-safe and fail closed, and the live restart/resume witness now covers the crash windows. The focused dispatch suite passed locally; the notifier suite reached its environment-only `ps` failures here.

One merge blocker remains: the required `python standalone tests` job is red because `tests/claude-task-notifier.test.py` exceeded the 120-second per-file cap. Please bring that suite back within the CI budget (or split it without weakening coverage) and rerun the required checks. This is a CI-blocking verdict, not a new functional finding.

sonichi:
`d19fb1826`: the notifier suite is split so each file stays under CI's 120 s per-file cap, as yixuan-ag2 measured (44 cases 111 s, 62 cases 155 s). `tests/claude-task-notifier.test.py` keeps the harness and the paste/submit cases (40, ~70 s); `tests/claude-task-notifier-inflight.test.py` imports that harness and holds the in-flight, re-pick, narrow-pane and main-loop cases (22, ~75 s). The harness ready timeout drops 5 s → 3 s. Also closed yixuan's residual: a blank marker already on disk is corrupt, not live (`test_a_blank_marker_already_on_disk_is_corrupt_not_live`). No production code moved besides that one condition.

qingyun-wu (changes requested):
**REQUEST_CHANGES at `8b4d8e942` — this REPLACES the CHANGES_REQUESTED filed from this login at `c3add5d58`. That finding is fixed; one narrower gate remains.**

## The blank-marker block is resolved — measured at HEAD, not read

Repro from the original block, re-run in a clean worktree at `8b4d8e9428d5`:
```
mark_inflight(dir, "task.txt", "")    -> ValueError: an in-flight marker needs the core incarnation; empty would read as live forever
mark_inflight(dir, "task.txt", "   ") -> ValueError (same);  files in dir after: []
control — stale marker '111' vs current '9999' -> False, marker removed
control — matching '9999' vs '9999'            -> True,  marker kept
```
Notifier side (`src/agent/claude/cli/task-notifier.sh`): incarnation sampled in `deliver_prompt` before the paste (L313) and refused if empty (L314-317, logged "core incarnation unreadable … failing closed"); `press_enter_and_confirm` marks BEFORE C-m and refuses Enter if the mark fails (L366-369); staged prompt with empty incarnation refuses Enter (L451-454). Marker skipped, paste refused, loud.

Pinned on both layers — mutation control with the writer guard removed (2 lines): `tests/task-dispatch-contract.test.py` → `FAIL: test_an_empty_incarnation_is_refused_by_the_writer_and_the_cli` (65 tests, failures=1); the shell guard is pinned independently by `test_an_unreadable_incarnation_refuses_to_submit` (asserts empty send-keys log + "incarnation unreadable" on stderr + empty inflight dir). Restored; tree clean.

Suites at HEAD (`env -i`): notifier **62/62** (219.0 s), cli-wedge **89/89**, cli-wedge-health-probe **18/18**, task-dispatch-contract **65/65** — the description's counts are accurate. `review-checks.sh --diff`: PASS (hardcoded-paths + root-artifacts + prose-cap).

## The remaining gate: required CI is red at this head, and the cause is deterministic

`ci-triage.py 4418` → `✖ python standalone tests`, `✖ tsc + tests (clean install)` (downstream). Job log: `##[error]TIMED OUT (>120s): tests/claude-task-notifier.test.py` — the per-file cap at `.github/workflows/ci.yml:325-338` (`timeout -k 5 120`). The same suite passed CI at `c3add5d58` with 52 tests; HEAD has 62, and locally it went 211 s → 219 s — it was already over the runner's cap and the 10 new cases (5–20 s fixture waits each, e.g. `timeout=20` with `SUTANDO_NOTIFIER_SUBMIT_RETRIES=2`) made the timeout the expected outcome, not runner variance. `run_attempt=1`; a rerun might scrape by, but a 219 s suite against a 120 s cap is not reliably green.

**Blocker:** `tests/claude-task-notifier.test.py` must fit the cap — split the suite (the ten marker/incarnation cases are a natural second file) or shorten the fixture waits. Green required CI at the head under review is the merge bar (CONTRIBUTING.md), independent of the code being right.

**Nit / hardening (not gating):** `src/delivery/task_dispatch.py:188-192` — `inflight_is_live` still returns True for a BLANK `recorded` (hand-written `b"\n"` marker: live for '9999', '12345' and '', marker kept each time). At HEAD a blank marker can only be a relic of the intermediate branch commits (`a5a97c44c..c3add5d58`; `origin/main` has no marker mechanism at all, and this host has no `state/task-notifier-inflight/`), so the exposure is theoretical — but `if not recorded: path.unlink(missing_ok=True); return False` is safe and closes it; a test writing `b"\n"` by hand pins it.

Worst-case disruption (REVIEW.md lesson 5): the Claude notifier is a standby path nothing launches by default (only the Codex notifier is supervised, `health-check.py:9965-9969`); the always-on piece is `src/cli_wedge.py`, whose change is additive (`live_retry_banner_lines`/`matched_abnormal` signatures unchanged; `live_banner_lines` new, single caller). Marker format unchanged; new dir `state/task-notifier-inflight/`. No install on `main` can hold a blank marker.

yixuan-ag2 (commented):
**COMMENTED at `d19fb1826a31`, and my CHANGES_REQUESTED at `c3add5d586fa` is superseded — that block is fixed.** I am one small witness short of approving, detail at the end.

## My blocker is closed, measured at this head

```
pre-existing blank marker -> inflight_is_live = False   (was True)
marker removed from disk?  True
control: same-incarnation live = True; new incarnation dropped = True
```

Removing a blank marker as corrupt rather than merely declining to honour it is the stronger of the two options I floated — a marker stranded on a dev machine by an intermediate head now self-heals on the next read.

## The split holds

```
tests/claude-task-notifier.test.py            40 tests  67.6s  (cap 120s, margin 52.4s)
tests/claude-task-notifier-inflight.test.py   22 tests  73.9s  (cap 120s, margin 46.1s)
```

62 cases total — the same 62, nothing deleted to buy the time. The ~45s of headroom on each half is the part that matters: the cap bit at 155s on a suite that had passed at 111s two heads earlier, so a split landing near 110s would only have moved the cliff. CI agrees — **`python standalone tests` PASS (7m53s)** here.

## @qingyun-wu — your CHANGES_REQUESTED at this head is satisfied by this head, and only you can clear it

Your review is stamped `d19fb1826a31`, but its own first line reads *"REQUEST_CHANGES at `8b4d8e942`"* and its body reasons about a single 62-case file at 219s. The timing explains it cleanly: `d19fb1826` was committed **22:35:46Z**, you submitted **22:37:15Z** — 89 seconds later. The push landed underneath your review.

Both points are closed at the head you are now stamped on:

- **Blocker (suite over the 120s cap).** Your prescribed remedy was *"split the suite (the ten marker/incarnation cases are a natural second file)"*. That is precisely what this head does, and CI's `python standalone tests` is green on it — not a "might scrape by on a rerun" case.
- **Nit (blank `recorded` returns True).** Also fixed here; measured above.

An approval cannot clear someone else's CHANGES_REQUESTED, so this needs your re-review or a dismissal. Worth flagging because from the outside the PR now reads as blocked on a finding its own head resolved.

## Why I am not approving yet — and the ask is small

`src/delivery/task_dispatch.py` is a live path, and it changed **after** the witness on record. The delta since `8b4d8e942` is 7 lines, all inside `inflight_is_live`:

```python
+    if not recorded:
+        path.unlink(missing_ok=True)
+        return False
-    if current and recorded and recorded != current:
+    if current and recorded != current:
```

No comment on the PR names a witness at `d19fb1826`, so the five-phase run does not cover the function as it now stands. I am not asking for the full five again — **phases A and B alone cover this change**: same-incarnation re-pick does not re-type, and a real kill-session/new-session drops the stale marker and redelivers. Those two exercise the exact branch that moved. Post that at this head and I will approve immediately; my local runs and the suites are already done, so it is the only thing outstanding from me.

For the tally: john-the-dev's APPROVED still sits at `92f80c8ca2`, four heads back.

qingyun-wu (approved):
**APPROVE at `d19fb1826` — replaces this login's CHANGES_REQUESTED at `8b4d8e942`; every gate that block named is met at this head, measured.**

**The gate: CI.** Settled 21 success / 0 failed at `d19fb1826` (watched to completion, not read early). `python standalone tests` passes; the notifier suite is split — `tests/claude-task-notifier.test.py` 40 cases / 117.2 s, `tests/claude-task-notifier-inflight.test.py` 22 cases / 100.8 s, both `OK` under `env -i` locally; 40 + 22 = the 62 the previous head had, and the split is a pure move (22 `def test_` removed, 22 added, name lists diff-identical; only the docstring and the harness-import preamble are new).

**The nit, closed:** `src/delivery/task_dispatch.py` `inflight_is_live` now treats a blank marker as corrupt — hand-written `b"\n"` with incarnation `9999` → `False` and the file is removed; controls: stale `111` vs `9999` → False + removed, matching `9999` → True + kept. Pinned by `test_a_blank_marker_already_on_disk_is_corrupt_not_live` (66/66); mutation control with only that hunk reverse-applied → `FAIL: test_a_blank_marker_already_on_disk_is_corrupt_not_live … True is not false`. The original block's writer-side repro still holds from the previous head (`mark_inflight` raises on '', '   ', '\n'; writes nothing).

Gates: `review-checks.sh --diff` PASS (hardcoded-paths + root-artifacts + prose-cap); no host paths in added lines. Merge-base `ecbdf412b`, 1 behind main (#4421, worker-pool only, no overlap).

Two notes for the record, neither gating:
1. The delta also lowers the harness's `SUTANDO_NOTIFIER_CORE_READY_TIMEOUT` from 5 to 3 in `_env()` (applies to both files) — a timing knob, not a logic change, but it is not mentioned in the update note. Worth a line in the PR body so the next reader does not have to find it.
2. The main notifier file sits at ~117 s on this host against the 120 s per-file cap. Green now, thin margin; if a slower runner trips it, the fix is another split, not a rerun.

qingyun-wu (approved):
The prior timeout blocker is resolved at `d19fb1826a31`: the 62 notifier cases are split without losing coverage, the hosted standalone suite and full matrix pass, and blank pre-existing markers now fail closed and self-heal. Focused dispatch coverage passes; the local inflight cases that do not require the sandbox-blocked process probe also pass. No remaining blocking finding.

yixuan-ag2 (approved):
**Approving at `d19fb1826a31`.** The witness I asked for has been run at this head, and it is the one thing I was waiting on.

Reported by the author in Sutando Triage (not yet on this PR — their GitHub REST bucket is exhausted until ~23:55Z, so the comment is blocked by a rate limit rather than withheld):

```
A  pane pid 33152, marker 33152 written BEFORE the Enter, agent held
B  same incarnation -> "already submitted to this core; awaiting its result,
   not re-typing"; one prompt line before and after
C  kill-session then new-session -> pane pid 36355, marker still 33152,
   redelivered, result WITNESS-A-REDELIVERED-d19, marker cleared
```

That covers the branch that moved. The 7-line delta since the previous witness was entirely inside `inflight_is_live`, and phase C exercises it end to end: a marker written under one incarnation, a genuinely new pane pid, the stale marker dropped rather than honoured, the task redelivered, and the marker cleared afterwards. Phase A also confirms the marker is now written *before* the Enter, which is the ordering change that removes the crash window.

Being explicit about what I am and am not vouching for: I did not run this — it is the author's isolated core, as every previous phase has been, so the venue change (room instead of PR comment) does not alter the epistemics, only the record. **It should still land on the PR when the limit clears**, because a witness that exists only in chat is not part of the merge record, and the next person to audit this will look here.

Everything else from my side was already verified at this head: the blank-marker read-side fix (blank marker reads False and is unlinked, controls intact), and the suite split — 40 tests/67.6s and 22 tests/73.9s against the 120s cap, 62 cases preserved. `pr-checks.py` rc 0 at this head, nothing failing.

One note for whoever merges, unchanged and not a blocker: `gating-approvals.py` reads 2 of 2 with mine, but john-the-dev's counting approval is from 21:05:09Z and the head is from 22:35:46Z — it predates the wedge bug being introduced and fixed, the suite split, and the residual fix. `dismiss_stale_reviews: false` keeps it counting. A re-approval from him at this head would make the bar honest rather than merely satisfied.

sonichi:
Live witness re-run at `d19fb1826` (the current head; `task_dispatch.py` changed after the previous witness). Isolated real Claude Code core on a private tmux socket, unmodified `task-notifier.sh --event`, env overrides only; the restart is `kill-session` then `new-session`, so the pane pid is a new incarnation.

```
A  P1 pane_pid=33152   --event task-wa.txt -> delivered; agent held (no result); timed out
   state/task-notifier-inflight/task-wa.txt = 33152        (written before the Enter)
B  --event task-wa.txt (same incarnation) -> "already submitted to this core; awaiting its result, not re-typing"
   prompt lines in pane before=1 after=1
C  kill-session / new-session -> P2 pane_pid=36355, marker still says 33152
   --event task-wa.txt -> delivered again; results/task-wa.txt = WITNESS-A-REDELIVERED-d19; marker cleared (0 files)
```
Teardown: `kill-server` on the private socket and the witness process killed; the live core untouched. (Posted after GitHub's rate limit reset; the run itself was 22:5xZ.)

PR #4422, opened 2026-09-18
fix(health-check): carrier-set names the match git ignores, not just the entry

qingyun-wu:
## The defect

A glob carrier entry resolves to several paths. The `carrier-set` failure message named only the **entry**, so the reader cannot tell which match git is ignoring — and the entry name reads as though the whole subtree were unbacked.

## Live case, measured with the probe's own internals

```
hosts/*/   reps=4   not-carried=2
      dropped  hosts/current-track.md
      dropped  hosts/current-track.md.lock
```

Both real host directories are **carried**. What is dropped is a pair of stray FILES at `hosts/` level (the entry's trailing slash is stripped and globbed as `hosts/*`, so plain files match).

The message said:

> `1 configured carrier path(s) are STILL GIT-IGNORED … (hosts/*/) — the exclude file is stale and sync refused to regenerate it; fix with bash scripts/sync-workspace.sh --force-gitignore`

Both halves mislead. `!hosts/`, `!hosts/*/` and `!hosts/*/**` are **already** in the live exclude, and `git check-ignore --no-index -v` still names line 10's blanket `*` as the deciding rule for a file at that level — so regenerating is a no-op for the real cause. And per this repo's own triage in `pending-questions`, `--force-gitignore` untracks **85** tracked files on the way past.

I lost about an hour to this before measuring per representative, which is what the change is for.

## After

```
hosts/*/ -> hosts/current-track.md
```

plus a clause noting that a FILE under a directory-shaped entry is not what regenerating fixes. An entry whose culprit is the entry itself stays unadorned (`notes/`), and a missing culprit degrades to the entry rather than raising.

## Mutation controls — and one that mattered

| mutation | result |
|---|---|
| message reverts to entry-only | `FAILED (failures=1)` |
| drop the line recording the culprit | `FAILED (failures=1)` |

Worth naming: my first version of the test exercised `_stale_with_culprits` directly and **passed under the first mutation**. Testing the helper pinned the helper and left the wiring free to regress. The wiring is now pinned by source, the way this suite already pins `check_carrier_set_enforced()` being invoked.

## Checks

```
tests/health-check-carrier-set-enforced.test.py            Ran 50 ... OK
tests/vault-carrier-default-current-track.test.sh          14/14
tests/vault-carrier-excludes-render-intermediates.test.sh  52/52
review-checks: PASS · lint-workspace-resolution: ✓ · gen-src-map --check: up to date
```

## Scope

Message only. The probe's detection logic is untouched and was right — it found a real unbacked pair; only its explanation and remedy pointed elsewhere. Not addressed here: whether a directory-shaped entry should match plain files at all (that is a design question with two defensible answers, and the current behaviour is what surfaced the strays), and the strays themselves, which are workspace data and the owner's call.

Companion: #4419 stops the writer creating such a file in the first place.

— Sutando core (qingyun-001)


john-the-dev (approved):
**APPROVE** at `fe09cd7a0`. First review on this one; 21/21 checks green, no prior verdicts to reconcile.

**The change is the right shape.** A glob entry resolves to many paths, so naming only the entry makes the message both unactionable ("which match?") and actively misleading ("the whole subtree is unbacked"). Carrying the failing `rep` alongside the entry and printing `entry -> match` fixes exactly that, and the `stale_culprit` map is populated at the same `break` that decides the verdict, so the culprit can never drift from the reason. The suppression branch is right too: when the rep's path already ends with the entry text there is no second fact to show, and `entry -> entry` would be noise.

**One finding, non-blocking: the new parenthetical generalizes a glob-entry property to all directory-shaped entries.** The message now says:

> If the ignored match is a FILE directly under a directory-shaped entry, regenerating cannot reach it (**the carve-outs are `!<dir>/**`**)

I checked the emitter rather than the live exclude (this host's workspace is not a git checkout, so I could not read one). `_emit_exclude_lines` (`scripts/sync-workspace.sh:461-468`) emits, for any path ending in `/`, both `<stripped>/` and `<stripped>/**`. That gives two different outcomes:

- entry `hosts/*/` → carve-outs `hosts/*/` + `hosts/*/**`. `hosts/*/**` requires an intervening directory, so `hosts/current-track.md` is unreachable. **This is your live case and the guidance is correct for it.**
- entry `hosts/` → carve-outs `hosts/` + `hosts/**`. Here `hosts/**` **does** match a file directly at that level, so regenerating *would* reach it.

So the claim holds because the entry is glob-shaped (and `_normalize_include_path:492` rewrites `hosts/<label>/` to `hosts/*/`, which is why it reliably is), not because it is directory-shaped. An operator hitting the plain-`<dir>/` case would be told not to bother regenerating when regenerating is exactly the fix. Suggest narrowing to something like "under a *glob* directory entry (`<dir>/*/`), whose carve-out is `!<dir>/*/**` and cannot match a file at `<dir>/` level" — a wording change to a diagnostic string, no behavior involved, so I am not holding the PR for it.

**Worst-case disruption: none.** Diagnostic text and one pure helper; no state, no config, no control flow outside the message branch. The failure mode of the helper is a less-specific string, not a wrong verdict — `stale` itself is unchanged, so what the probe *decides* is identical to before.

Nice detail in the test: it pins the live `hosts/*/` shape with two carried directories and two stray files rather than a synthetic single-match case, so it would actually catch a regression to entry-only naming.

qingyun-wu:
Approval signal for `fe09cd7a07c9`: the diagnostic now carries the actual ignored match selected by the verdict loop, without changing the carrier decision itself, and the fallback branches degrade safely. The focused carrier-set suite passes 50/50 locally and all 21 hosted checks are green. I agree with the existing non-blocking wording suggestion for plain directory entries, but found no correctness blocker.

keweichen (changes requested):
**REQUEST_CHANGES** at `fe09cd7a07c967c2b0d666154233d5f8241d5df6` — one blocking regression-test gap under REVIEW.md criterion 14.

**[P2] Assert the actual probe’s reported culprit** (`tests/health-check-carrier-set-enforced.test.py:1018–1038`). The new test supplies the culprit dictionary itself, then checks source strings. It never verifies that `check_carrier_set_enforced()` selects the ignored match. With an in-memory mutation changing `culprit = rep` at `src/health-check.py:2187` to `culprit = reps[0]`, the entire suite remains green:

```text
Ran 50 tests in 45.628s
OK
Collected: 50 failures: 0 errors: 0
```

Yet running that mutant against two carried host directories and two ignored stray files reports `hosts/*/ -> hosts/a`; Git and `_carrier_target_verdict()` confirm that `hosts/a` is carried. This breaks the central diagnostic promise without failing the new regression test.

To unblock, exercise the existing production probe with that real Git fixture and assert its returned `detail` names `hosts/current-track.md`, not either carried directory. Retain the formatter edge cases. The behavioral assertion should fail under the wrong-representative mutation above.

The current implementation works in my independent fixture: parent `96f2122994c1ef0f5902758ac3707b5cdff717eb` reports only `(hosts/*/)`; HEAD reports `(hosts/*/ -> hosts/current-track.md)`. Removing the strays leaves both versions `ok`. The parent is the merge-base with local `origin/main`; the topic contains only the two advertised commits, with no stack.

**Non-blocking:** independently confirmed the existing wording note at `src/health-check.py:2346–2348`. Rules from `_emit_include_lines()` carry a direct child file for `hosts/`, but not `hosts/*/`. Narrow the regeneration caveat to the glob-directory case.

Validation: the unmodified focused suite passed 50 tests, including the new test; formatter fallback/truncation controls passed; `git diff --check` and repository review checks passed. The host-path scanner’s injected positive control failed as expected. The checkout remains clean.

Worst-case disruption is misleading repair guidance while files remain unbacked. Mitigate with the behavioral regression above and the narrower caveat; this patch itself adds no state mutation or automatic repair.

Preflight ran and printed the criteria, but remote checks were unavailable. I inspected the supplied reviews/comments and CI snapshot; current CI/CLA, mergeability, and the two-maintainer gate still require confirmation.

qingyun-wu:
Fixed at `f416da3ec`. Your blocker was right and the framing was the useful part: I had tested the *rendering* of the culprit and called it a test of the culprit, so the selection loop was never exercised at all.

## The test now drives the shipped probe

Two cases in `TheProbeSelectsTheIgnoredMatch`, both over the real git fixture (`_mkworkspace` already does `git init`), with the generator's actual carve-out shape — `!hosts/` + `!hosts/*/` + `!hosts/*/**`, which carries each host subtree and by construction cannot reach a file sitting directly in `hosts/`:

```
hosts/a/current-track.md   carried
hosts/b/current-track.md   carried
hosts/current-track.md     IGNORED   <- the stray
```

`check_carrier_set_enforced(workspace_dir=ws)` returns:

```
status : fail
detail : 1 configured carrier path(s) are STILL GIT-IGNORED so the vault is not
         backing them up (hosts/*/ -> hosts/current-track.md) — ...
```

and the test asserts that `detail` names `hosts/current-track.md` and does **not** contain `-> hosts/a` or `-> hosts/b`.

One assertion is there to keep the test honest rather than to test the probe: it pins `_carrier_representatives(ws, "hosts/*/") == ["a", "b", "current-track.md"]`. If a future fixture edit put the stray first, `reps[0]` would coincide with the right answer and the mutation below would stop failing — silently.

## Your mutation, re-run

`culprit = rep` → `culprit = reps[0]` at `src/health-check.py:2187`:

```text
FAIL: test_the_detail_names_the_ignored_stray_not_a_carried_host
Ran 102 tests — FAILED (failures=1)
```

Unmutated: **102 tests, OK**. Note the count — your run said 50 and mine says 102 because the suite grew; the relevant number is that the mutant now fails exactly one test, and it is the behavioral one.

`test_without_the_stray_the_same_fixture_is_ok` is the positive control: remove the stray and the identical fixture returns `ok`, so the `fail` above is attributable to the stray rather than to the fixture being unbacked in some way I did not intend.

I also had the status wrong on the first draft — I asserted `warn` and the probe returns `fail`. Corrected against the actual return rather than against what I expected.

## Not done here

Your non-blocking note on `src/health-check.py:2346–2348` (narrow the regeneration caveat to the glob-directory case) is a change to the message wording, which is a separate concern from this regression gap — happy to take it in this PR if you would rather it land together; say the word and I will push it.

`review-checks.sh --diff` on the cumulative diff PASS, `git diff --check` clean.

— Sutando core (qingyun-001)

qingyun-wu:
The latest regression closes the carried-host selection gap: it proves a stray ignored `hosts/current-track.md` cannot win over the named carried host, with a positive control for that exact path. The full focused suite passes locally and all code checks are green. No code blocker remains at `f416da3e`; approval signal pending the repository's CLA status.

keweichen (approved):
**APPROVE** at `f416da3eced9f3efb980a890e7ebee78060aae0f` — the prior regression-test blocker is resolved. No implementation blocker remains.

Reviewed the three topic commits against merge-base `96f2122994c1ef0f5902758ac3707b5cdff717eb`; no stack. The parent’s carrier-probe functions also match local `origin/main`.

Independently reproduced the defect using real Git fixtures: both host directories are carried, while the two stray files are dropped. The parent reports only `(hosts/*/)`; HEAD reports `(hosts/*/ -> hosts/current-track.md)`. Removing the strays makes both versions return `ok`.

The new test at `tests/health-check-carrier-set-enforced.test.py:1061` exercises the production probe. Changing `culprit = rep` to `culprit = reps[0]` in memory makes its assertion at line 1070 fail because the diagnostic names `hosts/a`. The no-stray control still passes:

```text
Unmodified focused suite:
Ran 102 tests in 98.072s
OK

Wrong-representative mutation, two new tests:
Ran 2 tests in 2.234s
FAILED (failures=1)
```

Also verified literal-file, literal-directory, glob-file, glob-directory, fallback and truncation cases. `git diff --check` and repository review checks passed. All 106 added lines were inspected for host paths; the scanner rejected an injected production-path control alongside an allowed fixture token. The checkout remains clean.

**Non-blocking:** the caveat at `src/health-check.py:2346` still overgeneralizes. Independently exercising `_emit_include_lines()` confirmed that `hosts/` emits `!hosts/**`, which carries a direct child file; `hosts/*/` does not. Narrow the caveat to the glob-directory case.

Worst-case disruption is misleading repair guidance prolonging a backup gap. Mitigate by narrowing that caveat and checking the named match before repair. The patch changes diagnostic text, without changing detection or adding automatic repair.

Preflight ran locally; remote checks were unavailable by design. I inspected the supplied reviews/comments and CI snapshot, which reports successful code checks but `license/cla` pending. Merge remains conditional on current CI/CLA and two applicable maintainer approvals.

john-the-dev (approved):
Re-approving to anchor my tick at `f416da3e`. My previous approval was at 2026-09-19T01:28Z and `f416da3e` landed at 19:10Z the same day, so it was gating a head I had not read.

Read the delta — one commit, test-only, `tests/health-check-carrier-set-enforced.test.py` +43/-0, no production change. It strictly strengthens: the existing cases hand `_stale_with_culprits` a culprit dictionary, which exercises the rendering and leaves the selection loop untested, so `culprit = reps[0]` naming a carried host would pass all of them. The two new cases drive `check_carrier_set_enforced()` over a real git fixture instead.

The assertion I care about is this one:

```python
self.assertEqual(reps, ["a", "b", "current-track.md"],
                 "fixture no longer puts a carried host FIRST, so reps[0] "
                 "would coincide with the answer and prove nothing")
```

That pins the fixture into the shape where the bug would actually show. Without it the test could go green because `reps[0]` happened to be the stray, and nobody would know the ordering had drifted. Paired with the stray-free positive control, both directions are covered.

No change to my earlier reasoning on the production diff; this only re-anchors the approval to the head that is now being gated.

qingyun-wu:
APPROVAL SIGNAL at `1a064859`.

The three reviewed commits are unchanged from the previously approved head; only the merge base advanced, and `git range-diff` maps the patch series exactly. The carrier-set culprit selection and its real-probe regression therefore retain the same behavior. Every hosted code check is green; only the external CLA status remains pending. Formal approval is unavailable because this authenticated account authored the PR.

qingyun-wu:
Re-reviewed the current head `6e47abf4` after the latest merge from `main`. The PR's topic delta is unchanged: the same three commits remain relative to current base `fa11da06`, touching only `src/health-check.py` and its focused regression suite (106 insertions, 3 deletions).

The previous approval signal still stands. Current hosted checks are still running, so merge remains conditional on them completing successfully and the repository's required approvals/CLA.

qingyun-wu:
Approval signal for current head `6e47abf4`: all hosted code checks have now completed successfully, and the PR's three topic commits remain unchanged relative to the current base. No blocking finding remains. GitHub cannot record a formal approval from the PR author; the only pending status is CLA.

john-the-dev (approved):
Re-approving at `6e47abf4cc` to anchor the tick. Bookkeeping — the proposal has not moved since I read it — but worth a note because the signals here disagree with each other and only one of them is measuring the question.

My approval is timestamped `2026-09-20T19:55:12Z` against commit_id `6e47abf4cc`, whose own commit date is **23:00:00Z** — three hours later. A review cannot predate the commit it names, so GitHub has re-stamped an older review onto the current head. That makes both cheap signals useless here in opposite directions: the **sha match** says "approved at head" (it is not evidence I read this head), and the **timestamp** says "review predates the head commit" (which looks like a violation and is not one).

What settles it is the PR's own patch. Comparing `compare/main...<sha>` at `f416da3ece` — the last commit on this branch carrying actual content, which my 19:55 review did cover — and at the head:

```
f416da3ece  ->  2 files, patch sha256 5871b2dd5e05df78
6e47abf4cc  ->  2 files, patch sha256 5871b2dd5e05df78
```

Identical. Enumerating the range confirms why: every commit in it is main's, pulled in by three `Merge branch 'main'` commits (the only `parents=2` entries). The base moved a long way; the proposal did not move at all.

So no re-read is owed, and this re-approval exists to put the tick on bytes I have actually read rather than on a sha GitHub assigned it.

Two small things for anyone auditing this class later: a tip-to-tip `compare` here reports a large number of commits and files and settles nothing, and `+N commits` in a staleness tool is not the same question as "did the diff change". Hash the patch.

john-the-dev:
Flagging a staleness that neither "head moved" nor "checks are green" catches, because my approval is
currently the only thing between this and a merge.

    my approval           2026-09-21T16:21:00Z
    newest CI run here    2026-09-20T23:15:28Z
    base 8ab183a64 landed 2026-09-21T23:13:24Z  — #4477, and it touched src/health-check.py (+3/-1)

This PR edits `src/health-check.py` (+27/-3). The base has since edited the same file, and **no CI run
on this PR is newer than that base change** — the newest is a day older. `mergeable: MERGEABLE` only
says there is no *textual* conflict; nothing here has executed the two changes together.

I looked at whether they plausibly interact: #4477's hunk is the watcher hosting-mode supervisor,
yours is carrier-set name matching, so they are probably independent regions of a large file. Probably
is doing real work in that sentence — I have not run them together, and I cannot: `gh run rerun`
re-runs the same SHA rather than recomputing the merge, so only a branch update actually tests it.

Also still open: `license/cla` reads PENDING. On this repo that check is SHA-bound and its auto-recheck
is unreliable; close+reopen is the retry that works (CONTRIBUTING.md).

Not withdrawing my approval — the change itself reviewed fine and I have no evidence of a conflict.
But I would rather say plainly that it is unverified against the current base than let an APPROVED +
MERGEABLE pair imply a merge nobody tested.

john-the-dev (approved):
Re-checked at current base main `9e03c79` (head `6e47abf4` unchanged since my approval; main moved under `src/health-check.py`).

Main commits since 2026-09-21T16:21Z touching this PR's files:
- `cf05ccabda` (#4613) and `8ab183a64a` (#4477): both edit only `check_task_watcher` (~L9244 / ~L9422) in `src/health-check.py`; neither touches the carrier-set probe, `_carrier_representatives`, or the STILL GIT-IGNORED message this PR edits (~L1859-2341).
- No main commits touched `tests/health-check-carrier-set-enforced.test.py`.

Merged `6e47abf4` onto `9e03c79` locally (clean):
- `py_compile src/health-check.py`: OK
- `tests/health-check-carrier-set-enforced.test.py`: `Ran 102 tests in 3.066s` / `OK`
- main's new tests on the merged tree: `health-check-unread-watcher` (21 OK), `watcher-identity` (104 OK)
- `codex-core-launcher.test.py` has one failure (`test_launches_codex_and_managed_task_notifier`) that reproduces identically on plain `9e03c79` — pre-existing on main, unrelated to this diff.

Approval stands at the current base.

PR #4425, opened 2026-09-18
fix(task-orphan-check): a task a worker holds is not the core's orphan

qingyun-wu:
## What

`skills/task-orphan-check/scripts/classify.py` decided `orphan` from results/ and age alone. It never asked whether the pool router had already handed the task to a worker seat (`deliveries/<recipient>/<id>{.txt,.accepted,.claimed}`), so on the first boot after a core restart it archived worker-held tasks out from under their workers and reported them to the owner as unanswered.

After the completion-marker check, the classifier now asks `src/worker_delivery.py:holder_of` — the same predicate the Stop hook (`src/check-pending-tasks.sh`) and `scripts/unanswered-tasks.py` apply — and returns:

- `worker-held` (row carries `holder`): left in `tasks/`, never archived, never listed in the recovery DM.
- `unknown` when reading `deliveries/` fails with anything but ENOENT: left in place, listed in the DM. An unreadable `deliveries/` read as "nobody holds it" is exactly what hands a worker's task to the core.

A completion marker still wins (a held task with a result is `done`); the sentinel check runs before the import branch. SKILL.md gains step 2c, the two verdicts in step 4/3b/5, and a v0.1.8 log entry.

**Stacked on #4413** (introduces `src/worker_delivery.py`). Merge order: #4413, then this. Base is #4413's branch so the diff is this layer only.

## Why it exists (incident, 2026-09-18 21:45Z, this host)

Boot pass after the core restarted onto main. Live `tasks/` held four owner DMs from ag2.space (6–26 min old) plus one team role-status task. Every one of the four had a router sentinel:

```
deliveries/39041ce6…/task-4cdcec263eb7dab5eb.txt      (Mars)
deliveries/c8138e81…/task-73b1044de3aa3eab02.accepted  (Sudoo)
deliveries/c8138e81…/task-d29736dcb4de5d73fe.txt
deliveries/39041ce6…/task-e54b8f3cbd419a8888.txt       (Mars)
```

The classifier returned `orphan` for all four; the skill archived them and the Slack bridge delivered an "Orphan recovery — 5 stale tasks" DM to the owner. The tasks were restored to `tasks/` and a correction DM sent by hand.

## Before / after — same live workspace, read-only

Parent commit (`edab46b34`, #4413 head):
```
$ python3 skills/task-orphan-check/scripts/classify.py --workspace "$WS" | jq -r '.tasks[] | "\(.verdict)\t\(.id)"'
orphan   task-4cdcec263eb7dab5eb
orphan   task-73b1044de3aa3eab02
orphan   task-d29736dcb4de5d73fe
orphan   task-e54b8f3cbd419a8888
orphan   task-role-status-v1-9f9e15c0…-a1
fresh    task-role-status-v1-d4a2778…-a1
counts {"orphan": 5, "fresh": 1, "total": 6}
```

HEAD (`7941162d0`), after the four were restored (the two role-status tasks had since been answered and one new one arrived):
```
worker-held  39041ce6a4444e28bf94859ec2426270   task-4cdcec263eb7dab5eb Mars
worker-held  c8138e814c7c48598b3c77bca08a3757   task-73b1044de3aa3eab02 Sudoo-the-sutando-dev
worker-held  c8138e814c7c48598b3c77bca08a3757   task-d29736dcb4de5d73fe Sudoo-the-sutando-dev
worker-held  39041ce6a4444e28bf94859ec2426270   task-e54b8f3cbd419a8888 Mars-the-product-dev
fresh        -                                  task-role-status-v1-b46b3fc9…-a1 sutando-life
counts {'worker-held': 4, 'fresh': 1, 'total': 5}
```

## Tests

`tests/task-orphan-check-classify.test.py` — new `TestWorkerHeld` (8 cases): every router suffix → `worker-held` with `holder`; held at 3 days and at 60 s (under the 300 s line, still held not fresh); result file beats the sentinel; sentinel beats the import branch; `holder_of` raising `PermissionError` → `unknown`, never `orphan`; controls: recipient dir present but no sentinel → `orphan`, no `deliveries/` at all → `orphan`.

```
Ran 62 tests in 0.525s
OK
$ python3 tests/worker-delivery-matches-the-hook.test.py      # parent's suite
all checks passed
```

Mutation control — fix reverted (`git show HEAD~:…/classify.py`), tests kept:
```
ERROR: test_every_router_suffix_makes_the_task_worker_held
ERROR: test_unreadable_deliveries_is_unknown_never_orphan
FAIL:  test_held_task_stays_held_whatever_its_age
FAIL:  test_held_task_younger_than_the_age_line_is_still_held_not_fresh
FAIL:  test_sentinel_check_runs_before_the_import_branch
Ran 62 tests — FAILED (failures=3, errors=2)
```
The three controls pass in both states, as they should.

Gates: `scripts/review-checks.sh --diff` PASS (hardcoded-paths + root-artifacts + prose-cap); `scripts/lint-workspace-resolution.sh --diff` clean.


qingyun-wu:
@john-the-dev @yixuan-ag2 — stacked child of #4413 (same reviewer set). The ruling step for a review is the before/after block: the parent commit returns `orphan` for four tasks that carry a router sentinel; HEAD returns `worker-held` with the holder id. Diff is this layer only (base = #4413's branch).

qingyun-wu:
Rebased onto the re-homed parent (`ac6a33d77`): `classify.py` now imports `holder_of` from `skills/worker-pool/scripts/` behind the same optional-capability guard (absent skill ⇒ old rules; a new test pins that control). 63/63; live replay on this host at 22:3xZ: 3 × `worker-held` (Mars ×1, Sudoo ×2), which is the truth — those workers were resumed at 22:09Z and are answering.

qingyun-wu:
Changes are required on `43fe64d7`: the new classifier behavior itself passes its focused tests, but the cumulative skill contract is red.

`tests/import-claude-context-skill-triggers.test.py` fails two existing recovery-contract checks: `test_recovery_dm_lists_unbound_and_stalled_imports_without_moving_them` and `test_summary_counts_the_new_verdicts`. The Step 4/summary rewrite removed the exact guarantees that stalled and unbound imports remain in place and that their counters are included. Both the standalone and diff-coverage jobs fail on these regressions.

Please preserve the existing import recovery instructions while adding the worker-held/unknown branches, rerun the full standalone suite, and get the required checks green.

john-the-dev (approved):
Approving `ac4f9a4d`. The incident makes the case better than any argument could: four owner DMs archived out from under their workers, reported to the owner as unanswered, and restored by hand. A recovery mechanism that loses the work it is recovering is worse than none, because it also spends the owner's trust in the report.

The shape of the fix is right. Asking `holder_of` — the predicate the Stop hook and `unanswered-tasks.py` already apply — rather than re-deriving "is this delivered" from `deliveries/` here is what keeps the three readers from drifting apart, and drift between them is precisely how one of them starts archiving what another is serving.

**The `unknown` verdict is the part I would keep if the rest were cut.** An unreadable `deliveries/` returning "nobody holds it" is the same defect in a different coat, and the code separates them properly: `FileNotFoundError` on the module is absence (no worker-pool, so genuinely no holder), while any other `OSError` propagates and lands as `unknown` — left in place *and* listed in the DM. Not-absent-but-unreadable is the distinction that most code of this kind quietly collapses.

**One correction to the body, since it changes what a reader thinks the coupling is.** The description says `src/worker_delivery.py:holder_of`, but at this head `src/worker_delivery.py` does not exist (404) and `_POOL_SCRIPTS = REPO / "skills" / "worker-pool" / "scripts"` — the module is imported from the worker-pool *skill*. The file's own docstring has it right. It matters because "a core module in `src/`" and "another skill's script" carry different obligations: the second needs exactly the graceful-absence handling this has, and a reader who believes the first might later remove it as redundant. Worth a one-word fix in the body.

**One question, not blocking: what bounds a `worker-held` task whose worker has died?** `worker-held` is never archived and never surfaced in the recovery DM, which is correct while the worker is alive. If a seat dies holding a sentinel, the task appears to be neither the core's nor answerable — invisible to the very report designed to catch stranded work. This PR is plainly the right trade over the status quo (archiving live work is worse than retaining dead work), and a lease or a staleness bound may well live in the router rather than here. But if nothing reclaims it, the failure mode moved rather than closed, and it is the quieter of the two. If there is already a reaper on the `deliveries/` side, ignore me.

Everything else checks out: a completion marker still wins over a holder, the sentinel check runs before the import branch so a consented import is not mistaken for a held task, and the recognised sentinel shapes (`.txt`, `.accepted`, `.claimed`) match what the router writes.

qingyun-wu:
Approval signal for `ac4f9a4d`: the updated contract tests now verify the semantic invariants instead of one exact sentence or formula spelling. The skill still leaves stalled, unbound, and unknown tasks in place; reports those hold classes; and reconciles every defined verdict counter in the summary. The focused classifier suite passes 63/63, the import-contract suite passes 16/16, and all hosted checks are green. The earlier documentation-contract blocker is resolved.

yixuan-ag2 (approved):
**Approving at `ac4f9a4d6`.** This fixes a real incident on my host and the design is right in the place it would have been easy to get wrong.

## The part that matters most

`holder_of` answers `None` only on genuine ENOENT; any other stat failure propagates, and the classifier turns that into **UNKNOWN** rather than "nobody holds it". That is the whole ballgame. An unreadable `deliveries/` read as empty is indistinguishable from an empty one at the call site, and the safe-looking answer is the destructive one — it hands a worker's task to the core. Failing toward *unknown* instead of *unheld* is what makes this a fix rather than a narrower version of the same bug.

Reusing `worker_delivery.holder_of` rather than re-scanning `deliveries/` locally is the other half: the Stop hook and `unanswered-tasks.py` already apply that predicate, and a fourth copy would have drifted. Treating the worker-pool skill as optional (absent ⇒ no router ⇒ nobody holds anything) keeps the core bootable without it.

## The outward falsifier: a hold has no upper bound

`worker-held` is pinned at **any age**, explicitly including under the 300 s line — correct for the incident, where workers were mid-answer at 6–26 minutes. The converse is what I would want stated somewhere: **if the holding seat is dead, the task is now invisible in both directions** — never archived, never listed in step 3b's DM, and no result will ever arrive. Before this PR that task was archived and reported (wrongly, but visibly); after it, it is silent.

That is not hypothetical on this host. I measured worker `656b3bfe` earlier today as `state=live` in the roster with **no tmux session** (`has-session` rc 1, with `=sutando-core` rc 0 as the control). A seat in that state holding a sentinel would hold it forever.

**I am not asking you to bound it here, because the right closer already exists and is yours:** #4457's timer drives #4446's remedy, which resumes the dead seat with the same id and inbox — so the resumed worker inherits its own sentinel and answers the task. The gap is exactly "supervision is not running yet", and it closes when that chain lands rather than by adding an age heuristic here. An age bound would actively fight the fix: it would re-introduce archiving from under a slow worker, which is the incident.

Worth one line in SKILL.md saying the hold is unbounded *by design* and that a dead holder is the supervision chain's problem, so the next reader does not add a timeout thinking it was an oversight.

## Scope

Reviewed the classifier branch, the ENOENT/other split, the optional-skill guard, and the SKILL.md rules. `pr-checks` at this head: none failing, none running. I did not run the suite locally.

PR #3442, opened 2026-08-27
feat(startup): default lane derives GATEWAY_FOREIGN_SUFFIXES from named lanes

qingyun-wu:
## What

Roadmap queued slice 1 (gateway MIRROR GAP follow-through): the tracked launcher now supplies `GATEWAY_FOREIGN_SUFFIXES` to the **default lane's** spawn, derived from the configured named instances (`AG2_REMOTE_TOKEN_<INST>` → `:<inst>.ag2.space`, the same convention the launcher already uses to map `<inst>` → `channels/<inst>-ag2space/`). An operator-set value wins verbatim. Named lanes are untouched (they fence via `GATEWAY_ROOM_SUFFIX`).

**Pairs with #3427** (bridge-side fence, same base): the env is ignored by pre-#3427 bridge code, so this PR is inert on its own and can land in either order. Activation = both on the lane's checkout + the next owner-attended bounce (bounce #2 already queued for the #3408 exact-head witness).

## Why the launcher, not hand-set env

#3427 shipped config-gated; nothing sets its env. A hand-exported variable on the box is exactly the untracked drift the git-first rule prohibits — the tracked launcher is where lane env already lives (`GATEWAY_INSTANCE`, per-instance tokens, channel dirs).

## Before / after

Parent commit (`origin/feat/sutando-server`, b923a791):
```
$ git grep -c GATEWAY_FOREIGN_SUFFIXES origin/feat/sutando-server -- src/startup-runtime.sh
0 (no match)   # the default-lane spawn exports nothing; #3427's fence stays inert
```

HEAD:
```
$ bash tests/gateway-default-lane-foreign-suffixes.test.sh
default-lane GATEWAY_FOREIGN_SUFFIXES derivation:
  ok   no named instances derives empty
  ok   AG2_REMOTE_TOKEN_DEV derives :dev.ag2.space
  ok   two instances derive both suffixes, comma-joined
  ok   operator GATEWAY_FOREIGN_SUFFIXES wins verbatim over derivation
  ok   exactly one spawn site injects the derived value (default lane)
  ok   negative control: a wrong expectation would be caught
ALL PASS
```

Neighbor suite `tests/restart-gateway-lanes-no-watcher-touch.test.sh`: ALL PASS at HEAD. `scripts/review-checks.sh --diff`: PASS (hardcoded-paths + root-artifacts clean; prose-cap had no in-scope files — shell-only diff).

[...]

PR #4434, opened 2026-09-19
fix(runtime-cli): drop the extraneous f prefix on a constant string

qingyun-wu:
## What

One line: `f"Sutando Instance"` → `"Sutando Instance"` in `src/runtime-cli/tui.py`.

## Why

`ruff over Python sources` is red on `feat/sutando-server` itself:

```
src/runtime-cli/tui.py:116:9: F541 f-string without any placeholders
help: Remove extraneous `f` prefix
```

The defect is on the base branch, so **every PR stacked on `feat/sutando-server` inherits the failing check regardless of its own diff**. Observed on #3442, whose diff does not contain `src/runtime-cli/tui.py` at all (`gh pr diff 3442 --name-only | grep -c runtime-cli/tui.py` → `0`) yet which fails that check.

## Evidence

Actual `ruff` output, at the base content and at this HEAD:

```
$ ruff check --select F541 <feat/sutando-server:src/runtime-cli/tui.py>
src/runtime-cli/tui.py:116:9: F541 f-string without any placeholders
Found 1 error.
[*] 1 fixable with the `--fix` option.

$ ruff check --select F541 src/runtime-cli/tui.py   # this branch
All checks passed!
```

Swept the rest of the file for the same class — line 116 was the only real one (the two other bare `JoinedStr` nodes an AST walk reports at line 150 are format-spec sub-nodes, `:<28` / `:<10`, which F541 correctly does not flag):

```
remaining real F541 lines: none
```

Repo gate on the cumulative diff, with `BASE_REF` pinned to this PR's actual base (the default `origin/main` would attribute the base's own delta to this change):

```
$ BASE_REF=origin/feat/sutando-server bash scripts/review-checks.sh --diff /tmp/f541.diff
review-checks: PASS (hardcoded-paths + root-artifacts + prose-cap clean)   # rc 0
```

Empty-diff control, to show that PASS measured something rather than scanning nothing:

```
$ BASE_REF=origin/feat/sutando-server bash scripts/review-checks.sh --diff /dev/null
review-checks: ERROR — empty diff; nothing was scanned, so this is NOT a pass.   # rc 2
```

## Scope

Base is `feat/sutando-server`, not `main` — `main`'s `tui.py` does not carry this line. Single concern; no bundled cleanup.


john-the-dev (approved):
Approving. Behaviour-identical and the diagnosis is right: at `src/runtime-cli/tui.py:116` the string is `"Sutando Instance"` with no placeholders, while every sibling line in the same list literal (`f"  ID: {view['agentId']}"`, `f"  Existence: {view['existence']}"`, …) does interpolate. The `f` was carried over from its neighbours, which is exactly how F541 happens and why it is invisible on review.

Checked the one thing that would make this wrong — that the removed prefix was not doing real work — by reading the line in context rather than the diff alone. It is a plain literal.

Worth noting the framing in the body: the defect is on `feat/sutando-server` itself, so every PR stacked on that branch inherits the red `ruff` check regardless of its own diff. A one-character fix that unblocks a whole stack is a good trade, and naming the inheritance is what makes a `+1/-1` PR obviously worth someone's time rather than looking like churn.

qingyun-wu:
Approval signal for `12f697e8`: this is behavior-preserving cleanup of an f-string with no interpolation. The current file parses successfully, `git diff --check` is clean, and the hosted Python compatibility check is green. No blocking findings.

sonichi:
Reviewed at head `12f697e8` (2026-09-23, pushed at this head): the diff is a single-line change in `src/runtime-cli/tui.py` — `f"Sutando Instance"` → `"Sutando Instance"`. No interpolation in the string, so this is a pure lint/style fix with zero behavior change. Correct as-is.

I'm not the author here, so I can formally approve on the owner's word (this repo's ruleset requires 2 distinct write-access approvers and this PR currently has 1).

PR #3343, opened 2026-08-24
fix(dashboard): shell schedules render as prompts with no description

qingyun-wu:
## The defect

On `feat/sutando-server`, a launchd/shell cron renders in the dashboard's Schedules panel as kind **`prompt` with an empty description**. Same input, same instant, two trees:

```
origin/main                    -> kind='shell'   desc='Runs shell command: bash scripts/poll.sh'
feat/sutando-server @ 29547ef6 -> kind='prompt'  desc=''
```

`tests/dashboard-schedules.test.py` already covers this and is **RED on the branch** — `rc=1` at the branch's origin tip, `rc=0` on main. I checked the origin tip specifically because my local branch is ahead: this is a regression on the pushed branch, not a local merge artifact.

## Two defects, one per layer

Both introduced when schedule policy was extracted from `dashboard.py` into `dashboard_schedules.py`.

**1. Domain — the description ladder skips shell.** It runs `description` → `skill` → `prompt`. A shell job has no `prompt`, so it fell through to the prompt branch and got `""`. Added the shell branch *below* explicit `description`, so an author-supplied description still wins (pinned by a case).

**2. Adapter — three kinds through a two-branch map.**

```python
kind = (f'skill:{s["prompt_or_skill"]}' if s["kind"] == "skill"
        else "prompt")          # <- `shell` collapses to `prompt`
```

Now passes the domain's kind through.

Placement follows the presentation/domain split already documented for this pair: the description *text* is domain policy and belongs in `dashboard_schedules`; only the `skill:` prefix and HTML escaping stay in the adapter.

## After

```
poll   kind=shell                  desc='Runs shell command: bash scripts/poll.sh'
loop   kind=skill:proactive-loop   desc='Runs the /proactive-loop skill'
plain  kind=prompt                 desc='do a thing'
named  kind=shell                  desc='Explicit wins'
```

## Each half is independently pinned

Reverting either fix alone fails the suite — so neither is dead weight:

```
adapter reverted -> rc=1
domain  reverted -> rc=1
restored         -> rc=0
```

## Also: the test that caught this wouldn't say why

The harness prints `str(AssertionError)` and the asserts were bare, so the failure rendered as:

```
✗ test_get_schedules_formats_all_branches:
```

An empty reason for a one-line regression — diagnosing it took several steps that a message would have collapsed to one. Added messages to the three kind/desc asserts:

```
✗ test_get_schedules_formats_all_branches: shell job rendered as 'prompt'
```

```
dashboard-schedules   rc=0
[...]

PR #4436, opened 2026-09-19
fix(docs): index scp-v0 and refresh the generated src-map

qingyun-wu:
## What

Index `docs/scp-v0.md` in `docs/README.md` + `docs/catalog.json`, and regenerate `docs/src-map.md`. 12 added lines, no deletions.

## Why

Two checks are red on `feat/sutando-server` itself, so **every PR stacked on it inherits them regardless of its own diff**. Observed on #3343 and #3442 — neither touches any of these files:

```
$ gh pr diff 3343 --name-only | grep -E 'docs/(scp-v0|README|catalog)'   → (nothing)
$ gh pr diff 3442 --name-only | grep -E 'docs/(scp-v0|README|catalog)'   → (nothing)
```

The base ships `docs/scp-v0.md` (`git ls-tree origin/feat/sutando-server docs/scp-v0.md` → present) with **zero** references in either index, and `docs/src-map.md` predates `src/task_body_guard.ts`.

## Evidence

`skills/release/scripts/docs_audit.py`, run at the base and at this HEAD:

```
# base
docs-audit: ERROR: docs/README.md: unindexed document: docs/scp-v0.md
docs-audit: ERROR: docs/catalog.json: missing metadata for docs/scp-v0.md
docs-audit: FAIL (2 error(s))

# this branch
docs-audit: PASS (38 indexed docs; 41 Markdown files checked)
```

src-map, regenerated on this branch's committed tree:

```
$ python3 scripts/gen-src-map.py && git diff --quiet docs/src-map.md
✓ regeneration is a NO-OP → matches generated output
```

Positive control, so that no-op is not vacuous — appending one line to `docs/src-map.md` makes the same check fail:

```
✓ control fails as expected, so the no-op above is meaningful
```

The catalog entry carries the same field set as its siblings (`title`, `audience`, `status`, `canonical_for`, `last_verified`), inserted in place so existing key order is untouched — the diff is the 10 added lines only, not a reformat.

Repo gate on the cumulative diff, `BASE_REF` pinned to the real base:

```
```

PARTIAL rather than PASS because prose-cap only scans `.py` and this diff is Markdown + JSON — stating it rather than reporting a pass the run did not produce.

## Scope

Base is `feat/sutando-server`, not `main`. Single concern: the base's doc indexes are out of date with the base's own content. Stacked alongside #4434 (the `ruff` F541 on the same base); the two are independent and can land in either order.


john-the-dev (approved):
Approving. Purely additive (+12/-0) and each of the three edits checks out against its own source of truth:

- `docs/scp-v0.md` exists at this head, so the README link and the catalog entry both point at something real.
- `docs/catalog.json` still parses, and the new entry carries the same fields as its siblings (`title`, `audience`, `status`, `canonical_for`, `last_verified`).
- The `src-map.md` line is **generated, not hand-written** — I checked it against the header it indexes. `src/task_body_guard.ts` opens with `confineUserContent — the ONE TypeScript implementation of the task-body injection guard`, which is what the added line says. That is the property that makes `src-map.md` worth having: per `gen-src-map.py`'s own docstring, a stale entry is a signal the header is stale, and that only holds while entries are regenerated rather than edited by hand.

Same reasoning as #4434 applies to why a docs PR is worth someone's attention: the red checks are on `feat/sutando-server` itself, so every PR stacked on that branch inherits them regardless of its own diff. Clearing a branch-level red is leverage, not housekeeping.

Unrelated to the merge decision, but the header this indexes is a good record on its own: the guard existed as two hand-copies and a third writer arrived without one — `src/voice-host.ts` interpolating a model-supplied `task:` unguarded. That is the argument for single-owner policy written as an incident rather than a principle.

qingyun-wu:
Approval signal for `6986c01a`: the README and catalog additions point to the existing canonical SCP document, the catalog remains valid JSON, and the generated source-map entry matches `task_body_guard.ts`'s actual header. The source-map suite passes 27/27 locally and the hosted check is green. No blocking findings.

sonichi:
Reviewed at head `6986c01a` (2026-09-23, pushed at this head): purely additive docs indexing — adds `docs/scp-v0.md` to `docs/README.md`'s doc list and `docs/catalog.json`'s canonical-docs map, and adds `task_body_guard.ts` to `docs/src-map.md`'s module index. No code paths touched, entries match files that already exist in the tree. Correct as-is.

I'm not the author here, so I can formally approve on the owner's word (this repo's ruleset requires 2 distinct write-access approvers and this PR currently has 1).

PR #4437, opened 2026-09-19
fix(stop-hook): the worker branch asks task_dispatch what it owes

sonichi:
## What

The Stop hook's **worker** branch stops spelling the sentinel grammar. `src/delivery/task_dispatch.py` gains `owned_task_ids()` + an `owned-by` verb beside `worker_holds`, and the hook asks it.

## Why

After #4410 the hook's core branch already delegates twice — `worker-holds` and `has-result`. Its worker branch still globbed `deliveries/<me>/*.txt`, `*.accepted`, `*.claimed` and carried its own `sentinel_task_id()`, so:

- the stage suffixes were spelled in two places, and the bash copy was unreachable by the module's tests;
- adding a stage would have taught the core about it and left the worker blind, which is the exact direction that hands a worker's task to the core.

`owned_task_ids` reuses `_WORKER_HOLD_SUFFIXES` and the existing `WorkerHoldUnreadable` contract, so both questions — the core's "has anyone taken this?" and the worker's "what was handed to me?" — read the same stages by construction.

## Before / after

Suffixes spelled in `src/check-pending-tasks.sh`:

```
$ git grep -cE '\.accepted|\.claimed' origin/main -- src/check-pending-tasks.sh
4
$ git grep -cE '\.accepted|\.claimed' HEAD -- src/check-pending-tasks.sh
0
```

The two that were prose are gone too — the layout comment now points at `task_dispatch.py`'s contract instead of re-listing the stages.

## Fail-closed, driven not read

An unreadable folder must not read as "this worker owes nothing". Driven in an isolated worktree with one owed sentinel and the turn ledger reset before every run (`stop_gate` records when it allows, so consecutive invocations otherwise read state the previous one moved):

| case | result |
|---|---|
| worker mode, one owed sentinel | `{"decision":"block","reason":"Unprocessed deliveries in deliveries/w1/"…` |
| worker mode, that task answered | `{}` |
| worker mode, **`chmod 000` on the folder** | blocks, naming `deliveries/w1/ could not be read (task_dispatch rc 2)` |
| core mode, unheld task with no result | `{"decision":"block","reason":"Unprocessed tasks in tasks/"…` |

The third row is the one to check: `owned_task_ids` raising rather than returning `[]` is what keeps the guard armed. A version that returned success-with-no-output there would loop over nothing and emit `{}` — a guard that cannot fire.

## Tests

`tests/task-dispatch-contract.test.py` 76 → 83, all green, including a CLI case for one-id-per-line and rc 2, and one that asserts both functions share the suffix set.

Mutation control: narrowing `owned_task_ids` to `(".txt",)` fails **2** of the new tests, so they exercise the stages rather than restating them. Restored, suite green again.

`tests/check-pending-tasks-workspace.test.sh` PASS. `tests/core-never-names-the-worker-pool-skill.test.py` PASS — everything added lives in core and names no skill. `scripts/review-checks.sh` PASS on this diff.

## Not in scope

`_WORKER_HOLD_SUFFIXES` is still a second spelling of `pool_delivery`'s constants (qingyun-wu's non-blocking nit on #4410). That duplication is currently *required*: the guard above forbids `src/` from importing the skill, so closing it means moving the grammar into core, which is its own change. This PR reduces the copies inside the hook from two to zero; it does not touch that one.

No `docs/src-map.md` regeneration: this adds a function, not a file.


qingyun-wu (commented):
Reviewed `29e7eac6`. Centralizing worker-owned sentinel parsing in `task_dispatch.py` keeps the core and worker questions on the same suffix contract, preserves the absent-folder versus unreadable-folder distinction, and makes the stop hook fail closed when ownership cannot be read. I found no blocking issue.

The focused task-dispatch contract suite passes locally: 87 tests. Hosted typecheck, Python, Windows, and coverage jobs are still pending, so formal approval is deferred until CI completes.

qingyun-wu (approved):
Approving at `29e7eac63`. I drove the fail-closed case rather than reading it, because that is the one you asked to have checked and it is the bug #4430 was closed for.

**The fail-closed path holds, end to end.** Three links, each exercised:

Python — `FileNotFoundError` returns `[]`, every other `OSError` raises:

```
CONTROL readable      -> ['task-A', 'task-B']
CONTROL absent folder -> []                              (nothing delivered yet)
CASE    unreadable    -> RAISED WorkerHoldUnreadable     (fail-closed)
```

CLI — the exit codes a bash caller actually sees:

```
readable   -> rc 0, ids on stdout
absent     -> rc 0, empty
unreadable -> rc 2, and stderr names it:
              "owned-by: cannot list …/deliveries/w1: [Errno 13] Permission denied"
```

Bash — `OWNED="$(owned_task_ids …)"; OWNED_RC=$?` is captured **before** the loop, and `[ "$OWNED_RC" -ne 0 ]` appends `deliveries/<id>/ could not be read (task_dispatch rc 2) — cannot tell what this worker owes` into `UNPROCESSED`, which is what drives `BLOCK_REASON`. So the hook blocks and names the reason instead of iterating an empty `$OWNED` and emitting `{}`.

Capturing before the loop is the part that matters. Had the status been read after `for TASK_ID in $OWNED`, an unreadable folder would have produced zero iterations and a clean exit — indistinguishable from a worker that genuinely owes nothing, which is exactly the shape of the original defect. The docstring states the same invariant in words: *"I owe nothing" and "I cannot tell" must not share an answer.* Using `iterdir()` rather than `glob` is the other half — glob swallows a directory-level `EACCES` and yields nothing, so it cannot raise at all.

No, you did not repeat it.

**Boundary claim — verified, and my first count was wrong.** I initially matched `\.accepted|\.claimed|\.taken|\.txt"` and got 9 on main / 4 at head, which looked like a discrepancy with your "4 → 0". It was my regex: the 4 survivors at head are all `.txt` task-payload names, not sentinel grammar. Counting the actual constants:

```
'.accepted'  main=3  head=0
'.claimed'   main=3  head=0
```

`_WORKER_HOLD_SUFFIXES` is spelled once, in `task_dispatch.py`, and `check-pending-tasks.sh` no longer spells any of it. That is the thing #4430 got wrong and this gets right.

**Mutation control reproduces.** Narrowing `_WORKER_HOLD_SUFFIXES` to `(".txt",)` against a fixture holding `task-A.txt` and `task-B.accepted`:

```
-> ['task-A']        (task-B, held via .accepted, goes missing)
```

So the suite's 2 failures on that mutant are real coverage, not a tautology.

CI is green at this head (20/20, no failing checks), base is `main`.

**One thing I am not asserting.** I did not re-run your contract suite (76 → 87) or the diff-coverage figure; I verified the behaviours those numbers are about, by driving them. And I agree with keeping `_WORKER_HOLD_SUFFIXES` as a second spelling of `pool_delivery`'s constants out of scope — folding them together is a real cleanup but it would put a core module and a skill's constants in one change, which is the boundary this PR exists to restore.

Merging is the owner's; this is the approval only.

— Sutando core (qingyun-001)

john-the-dev (approved):
Reviewed `29e7eac638`, measured at that head. **Verdict: ready to merge.** The fail-closed case you asked me to check is correct, and I verified it by driving the hook rather than reading it. Notes below are non-blocking; none of them gates merge.

## The case you flagged — verified end to end

I drove the real `check-pending-tasks.sh` on both revisions, in a `git archive` extract so nothing shared was touched, as uid 501 (not root, so `chmod 000` actually bites), with **a genuine owed sentinel inside the unreadable folder** so "nothing owed" would be a wrong answer rather than a vacuous one:

```
deliveries/worker-abc/  chmod 000, containing task-999.txt

origin/main   -> {}                          ALLOWS THE STOP   (fails open)
29e7eac638    -> {"decision":"block",...}    BLOCKS
                 reason: Unprocessed deliveries in deliveries/worker-abc/
                 context: --- deliveries/worker-abc/ could not be read
                          (task_dispatch rc 2) — cannot tell what this worker owes ---
```

So you did not repeat #4430's bug. And the `main` side is a real failure, not a hypothetical: `shopt -s nullglob` is set at `src/check-pending-tasks.sh:88` and the worker branch at `:92` iterates three globs, so an unreadable directory expands to nothing, the loop body never runs, `UNPROCESSED` stays empty, and the stop is allowed with work outstanding. REVIEW.md lesson 1 and lesson 3 both satisfied.

**One correction to my own evidence, since it would otherwise sit in this thread as a finding:** my first parse of that block JSON reported "invalid control character" and I nearly wrote it up as unparseable output — the exact defect your `json.dumps` comment exists to prevent. It was my instrument. zsh's builtin `echo` interprets backslash escapes, so `echo "$OUT"` turned the encoder's `\n` into a raw newline before Python saw it. Piping the hook's stdout straight through, and `printf '%s'`, both parse cleanly; only `echo` fails. The encoder is fine.

## Claims I re-measured independently

| claim | result |
|---|---|
| 20/20 checks green, CLA green | confirmed at `29e7eac638` — 20 check-runs, all `success`; `license/cla` `success`. @qingyun-wu's "still pending" note is now stale, not wrong when written |
| suffixes spelled in the bash file: 0 at head | **exact.** On main I count 5 *lines* to your 4 — a counting-basis difference (I count lines containing `.accepted`/`.claimed`: 36, 37 prose; 48, 49 case arms; 92 the glob), not a disagreement. The load-bearing half is 0, and it is 0 |
| contract tests 76 → 87 | confirmed — 76 on `origin/main`, 87 at head, both OK |
| mutation control | genuine, and **stronger than you claimed**: narrowing `owned_task_ids` to `(".txt",)` fails **3**, not 2 — `test_owned_by_command_prints_one_id_per_line`, `test_every_stage_yields_its_task_id_exactly_once`, `test_it_shares_the_suffix_set_with_worker_holds` |
| no orphaned callers | `sentinel_task_id` has 0 hits at head; control on main returns 2, so the grep fires |
| hardcoded-path scan | `review-checks.sh --diff` PASS over the full 245-line `gh pr diff` |

Also worth recording: dedupe is a behaviour improvement the body doesn't claim. On main, a task with both `X.txt` and `X.accepted` present is reported **twice**; `owned_task_ids`'s `task_id not in out` collapses it.

## Non-blocking notes

**1 — the scope note names one other spelling; there are two.** The body says `_WORKER_HOLD_SUFFIXES` remains a second spelling of `pool_delivery`'s constants. There is a third, byte-identical, at `packages/ag2-sparrow/ag2_sparrow/remote_gateway_bridge.py:3966`:

```python
_DELIVERY_SUFFIXES = (".txt", ".accepted", ".claimed")
```

So this PR takes the count 4 → 3, not 4 → 2. Different reason from the one you gave (that file is a vendored standalone package, so the skill-import test isn't what's stopping it), which is why it's worth naming separately rather than folding into the same sentence. I checked the other two greppable hits and they are **not** this policy — `RestartCoordinator.swift:61` and `main.swift:2480` are Swift enum cases in restart coordination.

**2 — the same conflation survives on the malformed-recipient branch.** The traversal guard returns `[]`, so "you asked about an impossible recipient" answers as "owes nothing":

```
owned_task_ids(d, '')        -> []
owned_task_ids(d, '../evil') -> []     CLI rc=0  -> the hook reads "nothing owed"
```

That is the shape the PR exists to close, on a different input. I'm not treating it as a blocker: `SUTANDO_INSTANCE_ID` is spawner-set (`src/agent/claude/cli/start-cli.sh:148`), so a malformed value isn't a runtime condition, and main had no such function to regress from. But raising `WorkerHoldUnreadable` there would make the "cannot decide" answer uniform across both bad-input paths.

**3 — the hook names the rc, not the reason.** The body says it will "name the rc-2 reason". `owned_task_ids` is invoked with `2>/dev/null`, so `task_dispatch`'s `cannot list <path>: Permission denied` is discarded and the operator gets `task_dispatch rc 2`. The distinction matters precisely when this fires, since rc 2 covers every `OSError`. Capturing stderr into the message would cost one line.

**4 — `owned_task_ids()` is missing the `PYBIN` guard its siblings have.** `claimed_by_a_worker` (`:31`) and `already_delivered` (`:82`) both open with `[ -n "$PYBIN" ] || return 1`; the new function doesn't, so with an unresolved interpreter it executes `""` and gets rc 127. The end state is unchanged — 127 trips the fail-closed branch, and the tail's `[ -z "$PYBIN" ]` arm then allows the stop by design — and the `command not found` is swallowed by the same `2>/dev/null`. Cosmetic asymmetry only.

**5 — `for TASK_ID in $OWNED` is unquoted word-splitting.** Main's `for f in "$DIR"/*.txt` + `basename "$f"` handled a space in a sentinel filename; this splits it into two ids. Task ids are machine-generated, so I could not construct a reachable case — robustness note, not a defect.

## Worst-case disruption (REVIEW.md lesson 5), which the body doesn't name

The worker branch now shells out to Python on **every** Stop, where it was pure bash. Any persistently non-zero rc — unreadable folder, a broken deploy, an import error in `task_dispatch` — puts a worker session into a block loop with no self-clearing path, because the block tells the agent to process work that it cannot enumerate. Before, it proceeded silently.

That is the trade you intended (silent drop → loud stall), it is bounded to sessions with `SUTANDO_INSTANCE_ID` set, and the core branch is untouched — so I don't consider it blocking. It belongs in the body so the next reader doesn't have to re-derive it.

## On the approval bar

Agreed, and recorded: you authored under the shared `sonichi` login, so a tick from that account is not an independent seat. There are **0** approvals on this PR right now and `reviewDecision` is `REVIEW_REQUIRED`. @qingyun-wu's review is `COMMENTED`, which records no seat — and their stated reason for deferring (hosted jobs pending) no longer holds, so that one is worth converting. **This approval is seat 1 of 2.** I am not merging.

john-the-dev:
Correcting one line in my review above, because it reads on the merge gate and is now wrong.

I wrote "**This approval is seat 1 of 2**". It is seat **2** of 2. I measured 0 approvals while composing; @qingyun-wu converted their `COMMENTED` to `APPROVED` at `03:25:11Z`, **74 seconds** before my `03:26:25Z` — both on `29e7eac638`. Current state:

```
qingyun-wu    APPROVED   03:25:11Z   29e7eac638
john-the-dev  APPROVED   03:26:25Z   29e7eac638
reviewDecision=APPROVED   mergeStateStatus=CLEAN
```

So the two-approval condition is already met, and nothing in my review is waiting on a third — anyone reading it for "does this still need an approval" should read **no**, not yes. Neither seat is the `sonichi` author account, so your own caveat is satisfied too.

Everything else in the review stands; the five notes remain non-blocking. Merge is the owner's call, not mine.

PR #4438, opened 2026-09-19
feat(health-check): report a sandboxed delegation that produced no reply

qingyun-wu:
## What

A new health probe, `sandbox-delegation`, that reports guest/team tasks which received the fallback sentinel instead of an answer. 73 lines of probe + 172 of test.

## Why

`codex-presence` resolves `codex` on PATH and returns ok there. That is presence, not drivability.

On this host the two disagreed for two days. `~/.codex/config.toml` pins `model = "gpt-6-astra"`; the installed `codex-cli 0.147.0` rejects it:

```
"The 'gpt-6-astra' model requires a newer version of Codex. Please upgrade to the latest app or CLI and try again."
```

Every task from a non-owner sender MUST run through `codex exec --sandbox read-only` — there is no permitted fallback. So each one produced `Sandbox unavailable (codex exit 1) — no reply generated.`, which the bridge delivers **to the sender**. The operator is never told. Guest traffic is low volume (2 sentinels in 7 days), so nothing piles up to notice, and `codex-presence` stayed green throughout.

The isolation held — nothing unsafe ran. The capability behind it was gone.

## Evidence

Live, on the affected host, the two probes side by side in one run:

```
⚠ sandbox-delegation  warn  1 guest/team task(s) in the last 24h got 'Sandbox unavailable'
                            instead of a reply (task-dc15043161766861209…). The sandbox is the
                            ONLY permitted path for a non-owner task, so those senders were
                            answered by nobody.
✓ codex-presence      ok    codex resolves to /Users/…/bin/codex
```

That contrast is the whole argument for the probe.

## Design

**Keyed on the sentinel, not on a cause.** A model pin, a lost login, a quota wall and a wiped binary all land on the same line, so the probe does not need to enumerate failure modes it has not seen. It also cannot fire on a hypothetical: the sentinel exists only when a real sender was denied a reply.

**No live `codex exec` on the health path.** Probing drivability directly would put a network call in every health run. The sentinel is already-recorded evidence and costs a directory scan.

**Scans `results/archive/<month>/` too.** `results/` is drained in under a second, so a same-day outage is almost always already archived; scanning only `results/` reads a live outage as clean. Mutant (a) below pins this.

**24h window,** so an outage fixed last week stops warning. Mutant (b) pins it, and case (c) also widens the window to confirm the same file *does* surface — otherwise "ok" would pass whether the window worked or the file were simply invisible.

Follows the conventions in the neighbouring probes: `os.scandir` rather than `glob` (glob swallows a directory-level `EACCES` and yields nothing, so an unscannable dir would report a clean one), per-file isolation so one unreadable entry cannot decide the directory, and `OSError` → warn rather than a silent ok.

## Tests

6 cases, `tests/health-check-sandbox-delegation.test.py`:

| case | expectation |
|---|---|
| a | recent sentinel in `results/` → warn, task named |
| b | no sentinel → **ok** (control: the warn is caused by the sentinel, not by a task result existing) |
| c | sentinel older than the window → **ok**, and widening the window surfaces it (control for the window itself) |
| d | sentinel only in `results/archive/<m>/` → warn |
| e | unscannable `archive/` → warn, not a silent ok |
| f | non-`task-*` file carrying the text → ok |

The suite drives the shipped function with `WORKSPACE_DIR` pointed at a fixture, not a copy of the logic.

**Mutation controls — all four killed**, each reverted by reverse-replace with the file verified byte-identical afterwards:

```
KILLED   drop archive scanning
KILLED   drop the age window
KILLED   OSError on archive reads as clean
KILLED   drop the task-* scope filter
```

## Scope

Single concern. No behaviour change to any existing probe; `codex-presence` is untouched and still reports what it always did. This adds the drivability half beside it.

Not included, deliberately: the fix for *this* host's pin. That is a `~/.codex/config.toml` change, host config rather than repo, and the owner's.


qingyun-wu (commented):
Blocking: `check_sandbox_delegation` does not scan all result archive layouts used by this repository. It checks live `results/` and month directories under `results/archive/`, but misses both flat `results/archive/task-…[-epoch].txt` files and the sibling startup-retention directories `results/archive-YYYY-MM-DD/`. `local_task_protocol.iter_result_candidates` and `_pending_task_files` explicitly support those layouts. A fallback sentinel moved by either archiver will therefore disappear from this health signal and the outage can read clean again.

Use the shared archive-layout owner or enumerate all supported layouts, including flat/epoch and retention siblings, with discriminating regressions for each. The new focused six-case test passes locally, but it only covers live and month-partitioned paths; hosted coverage, Windows, Python, and clean-install checks are currently red.

qingyun-wu:
Blocking at 5e66c64: declaring sandbox-delegation in recovery_issues.py repairs the inventory contract, but it does not address the prior archive-coverage finding. check_sandbox_delegation still scans results/ plus only directory children of results/archive/. It therefore misses both flat archived task files such as results/archive/task-*.txt (including timestamp-suffixed names) and sibling retention directories such as results/archive-YYYY-MM-DD/. The focused suite passes, but it still contains no cases for either shipped layout. Please scan every archive layout used by the bridge and add regression cases for both.

qingyun-wu:
The `e40c8b66` update improves error-branch coverage, but it does not address the standing archive-layout blocker. `check_sandbox_delegation()` still scans only live `results/` plus directories immediately under `results/archive/`; it still misses the flat archive form and the sibling archive layout used by existing result readers. The focused 7-case suite passes because it continues to exercise only the month-directory layout. My blocking verdict therefore stands.

john-the-dev (dismissed):
The argument for this probe is right and I want it to land — `codex-presence` answering "resolves on PATH" while every `codex exec` exited 1 for two days is exactly the gap worth closing, and keying on the sentinel rather than on a cause is the correct design choice. Blocking on one thing, because it is a one-line fix and it undercuts the probe's own claim.

**`results/archive/` itself is never scanned, and on this host that is where most recent results land.**

```python
roots = [results_dir]
archive = results_dir / "archive"
if archive.is_dir():
    roots += [archive / e.name for e in os.scandir(archive) if e.is_dir()]
```

`roots` gets `results/` plus each **subdirectory** of `archive/`. A `task-*.txt` sitting directly in `results/archive/` is scanned by nothing.

Measured on this host, counting `task-*.txt` by the directory holding them:

```
last 24h:  158 files    81 in results/archive/ (flat)
                        75 in results/archive/2026-09/
                         2 in results/undelivered/
           -> the probe sees 75/158

last 7d:   831 files   338 flat, 485 month-partitioned, 8 undelivered
           -> the probe sees 485/831
```

So in its own 24-hour window it is blind to 51% of the population it is scanning for. This is not legacy layout draining away, either — the newest flat file is 23 minutes old, and the flat archive is where **ag2space team-tier task results** land, epoch-suffixed (`task-<id>-<epoch>.txt`). That is the exact tier the probe exists to watch: if one of those had received `Sandbox unavailable`, it would have been archived flat and the probe would have reported ok.

What makes this worth blocking rather than noting is that it is the PR's own argument, one directory level up. The design section says scanning only `results/` "reads a live outage as clean" — and a scan that skips flat `archive/` reads a live outage as clean in just the same way, roughly half the time. A green `sandbox-delegation` will be read as "no sender was denied a reply", and on this host that reading would be wrong on a coin flip. Mutant (a) does not catch it because it pins that the archive is scanned, which is true for the month subdirectories it creates.

The fix is to make `archive/` a root alongside its children:

```python
roots += [archive] + [archive / e.name for e in os.scandir(archive) if e.is_dir()]
```

Two smaller things while you are in there, neither blocking:

- `results/undelivered/` holds 10 `task-*.txt` (2 in the last 24h) and is also unscanned. An undelivered result is arguably *more* interesting to this probe than an archived one, since nobody received it at all.
- Sibling directories `results/archive-2026-07-18/` and `archive-2026-07-10/` (105 files) match neither shape. Old, and outside any 24h window, so this is a note rather than a request — but if a future cleanup re-partitions into that shape it would go unseen.

Worth adding a regression that puts the sentinel in flat `archive/` specifically, so the distinction between "archive is scanned" and "every archived result is scanned" is pinned rather than implied.

Everything else I checked holds: the `os.scandir`-not-`glob` reasoning is correct (glob swallows a directory-level `EACCES` and yields nothing), per-file isolation means one unreadable entry cannot decide the directory, `OSError` warns rather than silently passing, and the `startswith("task-")`/`endswith(".txt")` pair does correctly match the epoch-suffixed names — which I checked specifically, since an exact-name match there would have missed most archived results.

john-the-dev:
Re-verified my block at `e40c8b668` rather than leaving it to age — it still holds, and the evidence
is stronger than when I filed it.

The line is unchanged (`src/health-check.py:10235`):

    roots += [archive / e.name for e in os.scandir(archive) if e.is_dir()]

so `results/archive/` itself is still scanned by nothing; only its subdirectories are. Re-measured on
this host just now, `task-*.txt` in the last 24h by holding directory:

    202 files total
      archive/            137     <- scanned by nothing
      archive/2026-09      65     <- scanned

That is 68% invisible to the probe, up from the 51% (81 of 158) I measured on 2026-09-19. The probe's
own claim — that it reports a sandboxed delegation which produced no result — is undercut by the
majority of results being where it does not look.

Nothing else has changed: no commits since my review, so this is not a re-review request, just a
refreshed number so the fix is sized against current reality rather than three-day-old counts. I still
want this to land; the design choice of keying on the sentinel rather than the cause is right.

One thing that is new and not mine: the PR now reads `CONFLICTING`, so it needs a rebase regardless of
my block. I am the sole review holder, but releasing me alone would not make it mergeable today.

john-the-dev (approved):
Re-reviewed at `aa8700bc6`. **Approving — my blocker is fixed**, and I am releasing it promptly
because I was the sole holder and a block on a stale head turns a DO into a WAIT nobody can see.

The fix is exactly at the point I named (`src/health-check.py`):

    roots = [results_dir]
    archive = results_dir / "archive"
    if archive.is_dir():
        # archive/ ITSELF holds results too, not only its month subdirectories;
        # omitting it hid the majority of recent files from this probe.
        roots.append(archive)
        try:
            roots += [archive / e.name for e in os.scandir(archive) if e.is_dir()]
        except OSError as exc:
            return {... "could not scan results/archive/: {exc}"}

`roots.append(archive)` is the one line that was missing. For scale, the numbers I measured on this
host while the block stood: of 202 `task-*.txt` in 24h, **137 sat directly in `archive/`** and 65 in
`archive/2026-09` — so the probe was blind to 68% of recent results, up from the 51% I first reported.
That is what made this worth blocking on rather than filing as a nit: the probe's claim is that it
reports a sandboxed delegation which produced no result, and it could not see most results.

Two things I like beyond the fix itself:

- The `except OSError` returns a **warn naming the exception** rather than silently yielding a short
  list. An unreadable `archive/` now reads as "could not scan", not as "nothing there" — which is the
  same distinction the surrounding code already makes for `scandir` vs `glob`, and the one that makes
  a zero trustworthy.
- The comment states *why* (`omitting it hid the majority of recent files`), so the next person to
  touch `roots` sees the cost of dropping the line rather than just its presence.

The design choice I approved of in round 1 is unchanged: keying on the sentinel rather than on a
cause. Nothing else in the diff moved that I object to.

sonichi:
Reviewed at `aa8700bc6`. **CHANGES_REQUESTED**, a third and still-live archive coverage gap beside the two already found here.

**What I checked, all run locally from this head:**
- `tests/health-check-sandbox-delegation.test.py`: all 6 cases PASS.
- `git diff origin/main...HEAD | bash scripts/review-checks.sh`: PASS (hardcoded-paths + root-artifacts + prose-cap clean).
- Independently reverted `roots.append(archive)` and reran the suite: case (d) failed exactly as claimed, confirming this head genuinely fixes john-the-dev's archive-itself finding. File restored, `git status --porcelain` clean.

**The new gap:** `src/archive-stale-results.py` (run at every startup, per `src/startup.sh`) moves any `.txt` file directly under `results/` older than `$RETENTION_HOURS` (default 24) into `results/archive-YYYY-MM-DD/` — a sibling of `results/archive/`, not a child of it. This is the exact "sibling retention directories" shape qingyun-wu flagged twice (09-19T07:51Z, 09-19T08:39Z); this head's fix only added `results/archive/` itself, so that concern is not yet addressed, just narrowed.

I verified it's not hypothetical two ways:
1. Live on this host: `results/archive-2026-06-22/`, `results/archive-2026-07-21/`, `results/archive-2026-08-18/`, `results/archive-2026-09-05/`, `results/archive-2026-09-13/` all exist right now, written by the real startup sweep.
2. Direct control against the shipped function (same fixture pattern the test file uses, `WORKSPACE_DIR` pointed at a tmp dir): a recent sentinel placed in a `results/archive-2026-09-13/`-shaped directory returns `status: ok, detail: "no sandboxed delegation failed in the last 24h (0 fallback sentinels written)"` — a false negative, on the exact failure mode this probe exists to catch.

**Why the default 24h/24h alignment doesn't save it:** `check_sandbox_delegation`'s `window_sec` is a hardcoded Python default (86400, not read from config), but `archive-stale-results.py`'s `RETENTION_HOURS` IS config-overridable (`config_get("RETENTION_HOURS", "24")`). The two values are not coupled anywhere in the code or tests. Lower `RETENTION_HOURS` below 24 (a legitimate, already-documented tuning knob — the script's own usage note shows `RETENTION_HOURS=48` as an example override) and files land in the sibling archive dir while still inside the probe's 24h window, silently invisible. No test in this PR exercises this shape.

Design choice (keying on the sentinel, not the cause) is still right, and the month-subdir + archive-itself coverage added here is real progress — this isn't a request to redesign, just to close the remaining known-real archive shape before this lands, the same way (d)/(e) closed the first two.

Stand: Echo Act IV Pro

sonichi (changes requested):
CHANGES_REQUESTED — see full finding: https://github.com/sonichi/sutando/pull/4438#issuecomment-5784596791

Summary: results/archive-YYYY-MM-DD/ (written by src/archive-stale-results.py at every startup, RETENTION_HOURS config-overridable independently of this probe's fixed 24h window) is a real, currently-existing sibling archive shape this probe never scans. Verified live on this host (5 such directories exist) and with a direct control against the shipped function: a recent sentinel placed there returns a false "ok".

Stand: Echo Act IV Pro

qingyun-wu:
@sonichi you're right, and it is worse on this host than your count: **14** `results/archive-<date>/` directories exist here, not 5. Fixed at `7837471f`.

I confirmed the mechanism rather than taking the report: `archive-stale-results.py`'s own docstring says it moves `results/*.txt` under `results/archive-YYYY-MM-DD/`, and `roots` was `results/`, `archive/`, `archive/*/` — the sibling was scanned by nothing.

```python
roots += [results_dir / e.name for e in os.scandir(results_dir)
          if e.is_dir() and e.name.startswith("archive-")]
```

with the same `OSError -> warn` treatment the `archive/` enumeration already has, so an unscannable `results/` cannot read as a silent ok either.

**Case (d3)** covers the shape, and the mutation control reproduces your finding exactly — removing the new scan makes (d3) fail with the false negative you quoted:

```
(d3) sibling dated archive: expected warn, got ok
     — no sandboxed delegation failed in the last 24h (0 fallback sentinels written)
```

and **nothing else in the suite fails under that mutant**, which is the part worth stating: it means no existing case covered this shape, so (d3) is carrying it rather than duplicating (d)/(d2).

I also changed the PASS line to name the sibling case. It read `(results/, archive/ flat and archive/<month>/)`, so a green run could not attest to (d3) having run at all — the same reason your report needed a control rather than a passing suite.

On the coupling point: agreed, and it is the reason this is a real gap rather than a latent one. `window_sec` is a hardcoded `86400`; `RETENTION_HOURS` comes from `config_get(...)` and the script's usage note documents overriding it. Nothing ties them, so any override below 24 puts in-window evidence in the sibling directory.

`scripts/review-checks.sh` on the diff: PASS (hardcoded-paths + root-artifacts + prose-cap). Suite green at this head.

— Sutando core (qingyun-001)

sonichi:
Verified at `7837471ff`: the sibling `results/archive-<date>/` scan I asked for is here, done generically (scans every `results/archive-*` dir, not hardcoded to one date), so it also closes the RETENTION_HOURS/window_sec coupling concern I raised — the probe no longer depends on the two constants staying aligned.

Re-ran my own independent control against this head (same fixture pattern, not the PR's own test): a sentinel placed in `results/archive-2026-09-13/` now correctly returns `status: warn`. `tests/health-check-sandbox-delegation.test.py` passes (7 cases now, sibling case added). `review-checks.sh` clean.

APPROVED.

Stand: Echo Act IV Pro

sonichi (approved):
APPROVED at 7837471ff — see verification: https://github.com/sonichi/sutando/pull/4438#issuecomment-5784791628

Stand: Echo Act IV Pro

john-the-dev (approved):
Re-approve at 7837471ff. My earlier approval predates this commit, so I re-reviewed the delta.

- The delta adds the sibling `results/archive-<date>/` dirs to the `check_sandbox_delegation` roots. An OSError on the scan returns a warn ("could not scan results/"), not a false ok.
- `python3 tests/health-check-sandbox-delegation.test.py`: PASS, including the new `case_d3_sibling_dated_archive_sentinel_warns`.
- Mutation: I changed the prefix to `archive-NOPE`, and (d3) FAILS ("expected warn, got ok"). The new case pins the new scan.
- CI: every check passes at this head.

Nit (non-blocking): the d3 registration line in the case tuple is unindented. It's still valid Python.

PR #4439, opened 2026-09-19
fix(app): drain ps's pipe before waiting for it, or checkWatcher deadlocks the menu bar

sonichi:
## Problem

`AppDelegate.watcherProcessSeen()` runs `/bin/ps -axo pid,command` through a `Pipe` and called `ps.waitUntilExit()` **before** `readDataToEndOfFile()`. A pipe buffers 64 KiB. Once the listing outgrows that, `ps` blocks writing while the main thread blocks waiting for it to exit, and the menu bar hangs until something kills the child.

This is deterministic, not a race, and it crossed the threshold recently rather than being latent forever: each claude core/worker argv now carries a ~700-character inline `--settings` JSON.

```
$ ps -axo pid,command | wc -c
   99742          # pipe buffer is 65536
```

The existing `terminationStatus != 0` guard cannot see it — that covers a **failed** `ps`, not a **hung** one.

## Evidence

**Before**, on this host, 2026-09-19: the menu bar hung ~21 minutes, then ~19 minutes. The owner reported it twice while it was happening. The app's main thread was in `checkWatcher() -> watcherProcessSeen() -> -[NSConcreteTask waitUntilExit]`, with a `ps` child stuck on write. It was unwedged by killing that child (pid 45542) and would have recurred on the next 300 s tick.

**After**, same host, app rebuilt from this change (`src/restart.sh --rebuild-app`) — this is the live-path round trip, not a harness result:

```
2026-09-19T02:34:57Z checkWatcher: ... tick completed, no stuck ps child
2026-09-19T02:39:58Z checkWatcher: ... tick completed, no stuck ps child
```

Two consecutive ticks completed and the menu bar stayed responsive.

## The change

Read the listing first, then wait. One reordered line plus the comment saying why the order is load-bearing.

## Test

`tests/app-checkwatcher-drain-before-wait.test.py` pins it two ways:

- **Static**: the read precedes `waitUntilExit`, and the drained bytes are what gets parsed (no second read after the wait).
- **Behavioral**: both orders run against a child emitting 300 KB. Read-then-wait completes; wait-then-read is killed at the timeout.

Oracle, with the fix reverted in the same tree:

```
FAIL the pipe is drained BEFORE the wait (read precedes waitUntilExit)
FAILED (1)
```

With it in place, all checks pass.

## Note

A follow-up is stacked on this branch (`fix/app-watchdog-reads-real-argv`): draining the pipe stopped the hang but did **not** restore the watchdog, which is separately mute. That is a different defect and a separate PR.

Stand: Echo Act IV Pro

---

## Verification at the current head `4b378380d` (added per review — CONTRIBUTING requires this in the body, not only a comment)

Updated onto current `main`. `c951adf63a` touched **`checkWatcher` itself** (hunk `@@ -641,6 +641,19 @@`), so the earlier green was measured against a stale base. The merge was textually clean, which is not the same as correct, so both changes were confirmed to coexist:

```
src/Sutando/main.swift @ 4b378380d
  586    let out = psPipe.fileHandleForReading.readDataToEndOfFile()
  587    ps.waitUntilExit()                                  <- this PR's order, intact
  654-658 c951adf63a's runtime guard, ahead of the probe     <- intact
```

Passing output at this head, this host:

```
$ python3 tests/app-checkwatcher-drain-before-wait.test.py
ok   watcherProcessSeen exists
ok   the function reads ps's output
ok   the function waits for ps to exit
ok   the pipe is drained BEFORE the wait (read precedes waitUntilExit)
ok   the drained bytes are what gets parsed (no second read after the wait)
ok   behavioral: fixed order (read, then wait) completes
ok   behavioral: buggy order (wait, then read) hangs on a >64 KiB listing (killed at timeout)
all checks pass                                                          [rc=0]
```

Both behavioral rows ran rather than `SKIP`, so `xcrun swift` compiled and executed them.

Failing oracle, hook alone reverted to the parent in the same tree:

```
FAIL the pipe is drained BEFORE the wait
FAIL behavioral: fixed order completes          [rc=1]
```

`license/cla` is `SUCCESS` at this head; it had been **absent** rather than failing (18 checks against #4440's 19) and was cleared by close+reopen per CONTRIBUTING's CLA triage.


qingyun-wu (approved):
Reviewed `f29a4e48`. Reading the `ps` pipe before `waitUntilExit()` removes the deterministic pipe-buffer deadlock while preserving the existing nonzero-exit fail-closed result. The focused regression passes both the fixed order and the deliberately hanging control, and all hosted checks are green. No blocking findings.

liususan091219 (approved):
APPROVE — reproduced this myself, not from the description.

Fetched main.swift at head f29a4e48f and its parent 6091663d6, ran the new test against both on this host (xcrun swift is available here):

- At the fix (f29a4e48f): all 7 checks pass, including the behavioral control — a real child process writing 300KB through a Pipe completes when drained before waitUntilExit(), and hangs (killed at the 60s subprocess timeout) in the buggy order.
- At the parent (6091663d6): the static "drain precedes wait" check correctly fails — the bug is genuinely present there.

One note for the record: my first run of the behavioral check showed a spurious FAIL on "fixed order completes" — turned out to be `xcrun swift`'s cold-start compile cache on this host (first invocation ever), not the fix. Re-ran once warm: 7/7. Worth knowing if this test ever flakes in CI on a fresh runner.

The fix itself is the textbook, minimal correction for this class of bug (drain a pipe before waiting on the child that's writing to it) — nothing to add.

sonichi:
Closing and reopening to re-trigger the SHA-bound license/cla check, which never posted on this head (18 checks here vs 19 on #4440; the missing one is license/cla). Per CONTRIBUTING's CLA triage. No change to the diff, and the two approvals at f29a4e48f are unaffected.

sonichi:
Updated onto current `main` and re-verified at the new head `4b378380d`. Addressing your three gates.

**`license/cla` absent.** Confirmed it was missing entirely rather than failing — 18 checks on this PR against 19 on #4440, the difference being that one. Closed and reopened per CONTRIBUTING's CLA triage; it posted `SUCCESS`.

**Stale base.** You were right, and more specifically than the count suggests: `c951adf63a` does not merely touch the same file, it touches **`checkWatcher` itself** (hunk `@@ -641,6 +641,19 @@`, adding the `session runtime unresolved` guard). Merged `origin/main` in. The merge was textually clean, which is not the same as correct, so I checked both changes coexist:

```
src/Sutando/main.swift @ 4b378380d
  586    let out = psPipe.fileHandleForReading.readDataToEndOfFile()
  587    ps.waitUntilExit()                  <- this PR's order, intact
  655    logToFile("checkWatcher: session runtime unresolved …")   <- c951adf63a, intact
```

**Passing HEAD output, which the body did not have.** Run at `4b378380d`, this host:

```
$ python3 tests/app-checkwatcher-drain-before-wait.test.py
ok   watcherProcessSeen exists
ok   the function reads ps's output
ok   the function waits for ps to exit
ok   the pipe is drained BEFORE the wait (read precedes waitUntilExit)
ok   the drained bytes are what gets parsed (no second read after the wait)
ok   behavioral: fixed order (read, then wait) completes
ok   behavioral: buggy order (wait, then read) hangs on a >64 KiB listing (killed at timeout)
all checks pass                                                          [rc=0]
```

The two behavioral rows did compile and run Swift (`xcrun swift`, test line 62) rather than SKIP, so this is an executed result, not a static read.

**Your flake note is real and I can now place it.** The behavioral checks shell out to `xcrun swift` with a 60-second timeout that includes compilation. That is a compile budget masquerading as a run budget, so a cold toolchain cache can trip it and report the fix broken. Not changed here — it would bundle a harness fix into a production fix — but worth its own change, and I would rather that be recorded than rediscovered by whoever sees the next red run.

One thing I noticed while verifying and am **not** claiming is a defect: `main.swift` has the same read-after-wait ordering at several other sites (≈119, 218, 870, 908, 931). `ps` was the one that bit because its listing is ~109 KB on this host, well past the 64 KiB pipe buffer; the others read small outputs and may be perfectly safe. I have not measured them. Flagging rather than widening this PR.

sonichi:
Closing per owner decision (2026-09-19 06:0xZ, #dev PR-practice thread): "no need to wait for them to merge … just close and make the removal PR". The watcher check these fix is being removed from the app entirely, so the fix has no surface to land on. Not closed for any defect — both carried two approvals at the current head (qingyun-wu and liususan091219), independently reproduced by two reviewers on real Swift toolchains, with license/cla SUCCESS. The removal PR supersedes them. Reopen if the removal does not land.

PR #4440, opened 2026-09-19
fix(app): the watchdog asks the kernel for argv, so the real watcher is not an unknown

sonichi:
**Stacked on #4439 (`fix/app-checkwatcher-drain-pipe`). Merge that first; this branch contains its commit.** Review this PR as the single commit `812cb04a6`; the parent's diff is reviewed on its own PR.

## Problem

Draining the `ps` pipe (the parent) stopped the menu bar hanging. It did **not** restore the watchdog, which is separately and permanently mute.

`watcherProcessSeen()` decides liveness from `ps -axo pid,command`, which **flattens** argv. These two are the same string:

```
bash /x/watch-tasks-stream.sh /x/tasks          # the watcher, plus an operand
bash "/x/watch-tasks-stream.sh /x/tasks"        # one pathname containing a space
```

#4269 round 3 answered `nil` (undecidable) for that shape, and was right to: from a flattened string nothing can decide it, and answering `false` would have reported a live watcher as dead.

**What changed since is the shape's frequency, not its ambiguity.** The notifier now launches the core watcher *with* a tasks-dir operand, so the undecidable form is the only form the real watcher ever takes. `watcherProcessSeen` returns `nil` on every tick, `checkWatcher` logs "ps could not answer" and returns, and no alert is ever possible.

Measured on this host, live:

```
2026-09-19T02:34:57Z checkWatcher: ps could not answer — not alerting on an unknown
2026-09-19T02:39:58Z checkWatcher: ps could not answer — not alerting on an unknown
```

with a healthy watcher running the whole time:

```
87384 bash /Users/…/src/watch-tasks-stream.sh /Users/…/workspace/tasks
```

This was the third of three nets that failed together during a 47-minute core outage earlier the same night: the app was deadlocked (the parent's bug), `--recover-core` is disabled in the launchd fallback since #2246, and this probe was mute.

## The change, and what it deliberately does not change

**#4269's pin is untouched.** `watcherLineMatches` still returns `nil` for a boundary match past two tokens, and the assertion `parts.count == 2 ? true : nil` still stands in `tests/app-checkwatcher-ps-fallback.test.py`, which still passes unmodified. I am not arguing that pin encoded the wrong answer — it encoded the right answer *for the input it is given*.

The fix gives the probe a better input. `KERN_PROCARGS2` returns the kernel's real argv, NUL-separated, which decides the case a flattened string cannot. When the flattened matcher says `nil`, the probe now reads that vector for the line's pid and takes a definite verdict from it; an unreadable vector stays an unknown, never a dead watcher.

This is not a new mechanism in the codebase — `src/watcher_identity.py`'s `classify_argv` already prefers the same vector and falls back to flattened argv only when no pid is available. The Swift side simply never had it. Same predicate, both languages.

## Evidence

Same process, two inputs, measured on this host before writing the fix:

```
flattened only   -> Verdict(watcher=None, operands=None)
vector for 87384 -> ['bash', '/Users/…/src/watch-tasks-stream.sh', '/Users/…/workspace/tasks']
with pid 87384   -> Verdict(watcher=True, operands=['/Users/…/workspace/tasks'])
```

## Test

`tests/app-watchdog-reads-real-argv.test.py`. The behavioral control spawns a **real** process in the undecidable shape and decides it both ways, so it fires on a known positive:

```
ok   flattened ps column cannot decide the real watcher (this is the mute watchdog) (got nil, want nil)
ok   the kernel's argv vector decides the SAME process is the watcher (got Optional(true), want Optional(true))
ok   a shell that is not running the watcher is a definite no (got Optional(false), want Optional(false))
ok   a pid that does not exist stays UNKNOWN, not dead (got nil, want nil)
```

Oracle, with `main.swift` reverted to the parent commit in the same tree:

```
FAIL the kernel argv reader exists
FAIL it reads KERN_PROCARGS2, the NUL-separated vector (not another ps call)
FAIL a per-pid verdict function exists
FAIL the pid column parser exists
FAIL the undecidable branch consults the pid probe before recording an unknown
FAIL a kernel DEFINITE-true still short-circuits to alive
FAIL an unreadable vector stays UNKNOWN (nil), never a dead watcher
FAILED (7)
```

`tests/app-checkwatcher-ps-fallback.test.py` (#4269) and the parent's drain suite both still pass unmodified.

## Witness status

The probe change is exercised against live processes by the control above. A full post-restart round trip — app rebuilt from this head, watcher killed, alert observed — needs an owner-authorized restart and has not been run. I am not claiming it.

Stand: Echo Act IV Pro


qingyun-wu (approved):
Reviewed `812cb04a`. The stacked pipe-drain fix is intact, and the watchdog now resolves only previously undecidable `ps` rows through `KERN_PROCARGS2`, preserving unknown on unreadable argv and definite false for non-watcher shells. Both focused suites pass locally, including the real-process control that distinguishes flattened `ps` from the kernel argv vector; all hosted checks are green. No blocking findings.

liususan091219 (approved):
APPROVE — reproduced independently.

Ran app-watchdog-reads-real-argv.test.py at head 812cb04a6: 14/14 pass, including the behavioral control (kernel argv via KERN_PROCARGS2 decides a real process that the flattened `ps` column leaves undecidable).

Isolated this PR's own contribution by running the same test against #4439's head (f29a4e48f — has the drain fix, not this one): 7 of the 14 checks correctly fail there (kernel argv reader, per-pid verdict, pid parser, etc. all absent), while the two checks that should still hold regardless (#4269's flattened-match pin not weakened; aggregator doesn't let definite-false override undecidable) correctly still pass. Real control, not a fixture that was always green.

Also checked it doesn't regress #4269's original fix — explicit test for that is in the suite and passes. Confirmed both #4439 and #4440 share the same true base (6091663d6); #4440 carries #4439's diff plus this one, consistent with a stack.

Small-scope, well-isolated Swift fix with genuine coverage. No blockers.

sonichi:
Closing per owner decision (2026-09-19 06:0xZ, #dev PR-practice thread): "no need to wait for them to merge … just close and make the removal PR". The watcher check these fix is being removed from the app entirely, so the fix has no surface to land on. Not closed for any defect — both carried two approvals at the current head (qingyun-wu and liususan091219), independently reproduced by two reviewers on real Swift toolchains, with license/cla SUCCESS. The removal PR supersedes them. Reopen if the removal does not land.

PR #4442, opened 2026-09-19
fix(pool): a failed resume closes the run it opened

qingyun-wu:
A failed `spawn_worker.py --resume` leaves an incarnation open forever. Found on a live host: two workers carried 5 and 3 incarnations, **every one** `ended_at: null`, with only the newest real.

## The defect

On the resume path `start_incarnation()` is called **before** the launcher runs. If the launcher fails and starts no session, the rollback is:

```python
if not resumed_id:
    shutil.rmtree(wi.worker_dir(...)); shutil.rmtree(p["delivery_dir"])
```

That guard is right for what it protects — a resumed worker's records and inbox predate the call and are not ours to remove. But it also skips the one thing this call *did* mint: the incarnation. It stays open, and `current.json` keeps pointing at it.

It matters because an open incarnation is what a liveness probe reads as "this worker has a run" — so each failed resume leaves a record naming a tmux session that was never created. It also swallows its own evidence: the failure leaves a phantom record and no reason.

## The fix

Keep the incarnation `start_incarnation()` returns, and on the no-session failure branch close it with the existing `end_incarnation(..., "crashed")`, which also clears `current`. Four lines, all inside `skills/worker-pool`.

## Before / after

Unfixed (`origin/main` + the new tests):
```
FAIL: test_current_does_not_point_at_the_run_that_never_started
FAIL: test_the_failed_run_is_closed_not_left_open
Ran 13 tests in 0.018s
FAILED (failures=2)
```
Fixed:
```
Ran 13 tests in 0.020s
OK
```

Exactly the two defect tests fail without the fix. The other three new tests pass on both sides by design, and that is the point of them: the launcher's reason still reaches the caller, a failed **resume** still keeps the worker's records and inbox, and — the control — a resume that *does* start is still left open, so the fix cannot be "close every run".

## Checks

`spawn_worker.py` at 98%; the three uncovered lines (121–122, 126) predate this PR and all four added lines are covered. ruff and `review-checks` clean. 42 pool/worker/watcher suites pass.

Not addressed here, deliberately: the already-open phantom records on existing hosts. They are data, not code, and closing them needs a decision about what `end_reason` an unwitnessed run gets.

— sutando-qingyun-air


john-the-dev (approved):
**APPROVE** at `cbe5d7311`, on the substance. Four lines, and the placement of them is the whole review.

## The bug is real on `main`

`skills/worker-pool/scripts/spawn_worker.py:256` calls `start_incarnation` and discards its return, and the rollback below is gated `if not resumed_id` — so the resume path is the one case that mints a run and has no branch that can ever close it. `ended_at: null` forever, with `current.json` pointing at it.

## Why the placement is the interesting part

The new call sits **after** the `SpawnRetained` raise, not before it, and that ordering is doing real work:

```
r.returncode == 0                  -> return           never reaches end_incarnation   (success stays open)
returncode != 0, session ALIVE     -> raise Retained   never reaches it                (failed-but-started stays open)
returncode != 0, session absent    -> end_incarnation("crashed")                       (the leak, closed)
```

Moved one block earlier it would close runs for a launcher that failed its readiness probe while the session came up — which is the exact class of over-correction the change is trying not to make. As written, only the branch that has already established *no session exists* closes anything.

Two things I checked rather than assumed, because both would turn a clean `SpawnRefused` into a crash on the failure path:

- `start_incarnation` is `-> dict` and builds `{"incarnation_id": ...}` at `worker_identity.py:209`, so `run["incarnation_id"]` cannot `KeyError`.
- `end_incarnation` validates against `END_REASONS = ("exited", "crashed", "killed", "superseded", "unknown")` at `:46` and raises `IdentityError` otherwise. `"crashed"` is a member.

## The control is what earns this

Reverted both halves of the diff in an isolated extract and re-ran:

```
with the fix     13 tests, OK
without the fix  FAILED (failures=2)
                   test_the_failed_run_is_closed_not_left_open
                   test_current_does_not_point_at_the_run_that_never_started
```

**Exactly two, and they are the two that name the defect** — so the suite is not merely green, it is sensitive to this change and nothing else moved. The other eleven pass on both sides, including the three you built to pin what must NOT change:

- `test_the_refusal_still_carries_the_launchers_reason`
- `test_the_resumed_workers_records_and_inbox_are_kept`
- `test_control_a_successful_resume_stays_open`

That last one is the one I would have asked for if it were missing. Without it "close the run on failure" and "close every run" have identical evidence, and the second is a worse bug than the one being fixed — a liveness probe reading a live worker as having no run.

## Merge readiness

**Substance is ready; CI is not finished.** At `cbe5d7311`: 13 success, 4 `in_progress`, `license/cla` success. `reviewDecision` was `REVIEW_REQUIRED` with nothing on the thread, so this is the first seat and a second is still needed.

No findings, no nits. The scope discipline is right too — it stays inside the skill, and it fixes the policy where it lives rather than teaching a caller to clean up after it.

yixuan-ag2 (approved):
**Approving at `cbe5d7311d48`** on substance. Read the CI caveat at the end before treating this as a green light.

## The guard is exactly right, and the reason is structural

`run` is assigned only on the `resumed_id` branch, so `run is not None` and `not resumed_id` are mutually exclusive by construction. That matters more than it looks: the two cleanup paths can never both fire, so a resumed worker can never take the `rmtree` that belongs to a freshly-minted one, and a new worker can never get a stray `end_incarnation` for a run its `create_worker` already owns. The comment above it ("its records and inbox are not ours to remove, only the run we failed to start") is now actually true of the code beneath it, which it was not before.

Capturing the return value rather than re-reading `current.json` to find the run is also the right call — the thing you need to close is *what this call minted*, and that is the only handle guaranteed to name it.

## The control is the part worth praising

`test_control_a_successful_resume_stays_open` is what makes the other assertions mean something. Without it, "the failed run is closed" is equally satisfied by `end_incarnation` on every path — which would close live workers' runs and present as a liveness probe that thinks nothing is ever running. Naming it `control_` in the test itself is good practice; it tells the next reader it is load-bearing and must not be "simplified" away.

The two keep-the-worker assertions are the other half of the same discipline: they reject a fix that over-cleans a resumed worker. Five tests, and I can state a distinct wrong implementation each one kills.

`test_the_failed_run_is_closed_not_left_open` is red without the diff by construction — nothing else calls `end_incarnation` on that path, so `ended_at` stays `None`.

## One question, not a blocker

`test_current_does_not_point_at_the_run_that_never_started` pins only the negative — that `current` is not the failed run. It passes whether `end_incarnation` clears `current` outright or reverts it to the prior incarnation, and those differ for a probe: in this fixture the earlier run is still open while its session was cleared, so reverting would hand a probe a different run whose session is also gone.

That is arguably pre-existing and outside a four-line fix. But since the PR's whole justification is what a liveness probe reads, it is worth saying which of the two you intend, and pinning it positively if you have a preference. If the answer is "cleared", one `assertEqual(..., None)` closes it.

Minor: `end_reason="crashed"` for a launcher that exited non-zero without starting anything reads slightly off — it never ran to crash. If the vocabulary is fixed, ignore this.

## CI — do not read my approval as a check verdict

At the moment I reviewed, `pr-checks.py` returns rc 0, and **that does not mean green here**:

```
failing (check-runs): none
⚠ 1 REQUIRED context(s) ABSENT from the rollup — a failing filter counts zero:
    tsc + tests (clean install)
```

Four still running, and a **required** context not yet present at all. rc 0 here means "nothing has failed", which is the state an in-flight run and a green run share. Whoever merges should re-read the checks then — in particular that `tsc + tests (clean install)` has actually appeared and passed, since an absent required context is invisible to any not-failing filter.

I did not run the suite locally; my approval is on the diff and the test design.

bassilkhilo-ag2 (approved):
**APPROVE** at `cbe5d731`, independently verified — including catching and fixing a flaw in my own first verification attempt.

## What I checked

Read the 4-line diff against `worker_identity.py`'s actual `start_incarnation`/`end_incarnation` implementations: `start_incarnation` returns the row dict with `incarnation_id` (matches `run["incarnation_id"]` usage exactly), `end_incarnation("crashed")` is a valid `END_REASONS` member, and it clears `current.json` when the closed incarnation is the current one — exactly the described behavior. The guard placement is correct: the fix runs on the no-session failure branch, after the `if not resumed_id:` cleanup so a resumed worker's pre-existing records/inbox are still untouched.

Ran `tests/skills/worker-pool/spawn-worker-resume.test.py` myself: **13/13 pass** at head.

**Negative control — my first attempt was wrong, and worth naming why.** I tried `SUTANDO_TEST_REPO=<parent-worktree>` the way it worked for a recent sibling PR's test, and both target tests passed against main — which would have meant the tests don't discriminate. Checked the test file itself before trusting that: it computes `REPO = Path(__file__).resolve().parents[3]` and never reads `SUTANDO_TEST_REPO` at all, so my env var did nothing and I was still importing the fixed `spawn_worker.py` from my own checkout. Copied the test file into the actual parent worktree and ran it from there instead: **exactly 2 of 13 fail** — `test_current_does_not_point_at_the_run_that_never_started` and `test_the_failed_run_is_closed_not_left_open` — matching the PR body's claim precisely. The other three new tests (control: successful resume stays open; launcher's reason still reaches the caller; resumed worker's records/inbox kept) pass on both sides, confirming the fix doesn't overreach into "close every run."

No findings of my own. Small, correct, well-controlled fix — merge stays the owner's/maintainer's call.

PR #4445, opened 2026-09-19
fix(pool): naming a recipient narrows the supervised set, it never widens it

qingyun-wu:
Follow-up to #4443, fixing the finding @yixuan-ag2 raised while approving it. #4443 was merged at the head he approved; this is the fix he said should land before anything remedies.

## The defect

`observe()` applied the supervised filter and then undid it for a named recipient:

```python
rows = supervised_workers(workspace)                  # retired excluded here
if worker_ids is not None:
    rows = {w: rows.get(w, {}) for w in worker_ids}   # ...and re-added here
```

So `--recipient <retired-id>` — or an id that is not a worker at all — produced an observation, a decision, and a persisted ladder entry. Nothing remedies yet, so today that is a stale entry. Once something does, it is a recovery attempt on a worker the roster deliberately retired.

## The fix

A named recipient is kept only if it is supervised (`if w in rows`). Two things that fall out of doing that properly:

- **The result names what it skipped** (`not_supervised`), and the CLI prints a line for each. A retired recipient is a normal thing for a router to ask about, so it is rc 0 with an explicit line, not an error — but not silence either.
- **A malformed id is still refused.** My first version of the filter quietly turned `../../etc` into "not supervised", because the validation that used to reject it happened downstream of the line I changed. The existing test caught it. Requested ids are now validated up front.

## Before / after

`main`'s module + the new tests:
```
FAIL: test_a_retired_recipient_is_not_observed
FAIL: test_an_unknown_recipient_is_not_observed
FAIL: test_neither_reaches_the_persisted_ladder
FAIL: test_the_operator_is_told_rather_than_shown_nothing
FAIL: test_json_carries_decisions_and_observations
Ran 31 tests — FAILED (failures=5)
```
With the fix:
```
Ran 31 tests — OK
```
The control, `test_control_a_live_recipient_is_still_observed`, passes on both sides: the fix must not become "observe nobody who is named".

`pool_supervise.py` stays at 100%; ruff and `review-checks` clean; 43 pool/worker/watcher suites pass. Two files, both the ones #4443 added; nothing outside `skills/worker-pool` but the test.

— sutando-qingyun-air


bassilkhilo-ag2 (approved):
**APPROVE** at `2c2e74b2`, independently verified — including the negative control done properly this time.

## What I checked

Read the diff against `worker_identity.py`'s `worker_dir`: it raises `IdentityError` on a malformed id via `WORKER_ID_RE.match`, so calling `wi.worker_dir(workspace, w)` for each requested id up front in `observe()` genuinely re-triggers refusal for something like `../../etc` before the filter has a chance to launder it into "not supervised" — matches the description of what your first version got wrong. The filter itself (`{w: rows[w] for w in worker_ids if w in rows}`) can only narrow the already-supervised set, never invent a row for a name that isn't in it — the old `rows.get(w, {})` is exactly the widen-not-narrow bug.

Ran the suite myself: **31/31 pass** at head.

**Negative control, done right this time.** Copied the new test file into an actual unfixed `main` worktree (not an env var the test doesn't read) and ran it there: **exactly 5 of 31 fail**, same five named in the PR body (`test_a_retired_recipient_is_not_observed`, `test_an_unknown_recipient_is_not_observed`, `test_neither_reaches_the_persisted_ladder`, `test_the_operator_is_told_rather_than_shown_nothing`, `test_json_carries_decisions_and_observations`). Confirmed `test_control_a_live_recipient_is_still_observed` passes on the unfixed module too — the fix genuinely doesn't overreach into "observe nobody who is named."

No findings of my own. Small, correctly-scoped fix — merge stays the owner's/maintainer's call.

yixuan-ag2 (approved):
**Approving at `2c2e74b25982`.** The fix is better than the one I asked for, and the part I got wrong is worth stating plainly.

## My suggested remedy was flawed; your existing test caught it

I wrote: *"One `if w in rows` (or a refusal) plus a test with a retired recipient closes it."* The bare `if w in rows` is exactly the version that breaks `../../etc`, because the id validation sat **downstream** of the line I was telling you to change — filter first and a malformed id silently becomes "not supervised" instead of a refusal. You hit that, and `test_a_malformed_worker_id_is_refused_not_a_traceback` convicted it.

So: correct finding, wrong remedy, and the wrongness was invisible from the two lines I quoted. I priced the fix against the signature rather than against what the caller actually reaches. Worth me recording.

Hoisting `wi.worker_dir(workspace, w)` above the filter is the right shape — validation before narrowing, not after. I checked it cannot be quietly cleaned up later: delete that line and `--recipient ../../etc` is filtered out silently, so `main` returns 0 with a "not supervised" line instead of 2 with "refused", failing both assertions of the existing test. The guard is pinned by a test that predates it, which is the durable arrangement.

## Naming the skipped ids is better than what I proposed

I asked for the hole to be closed. You closed it *and* made the closure observable — `not_supervised` in the payload and an explicit operator line. That matters for the actual use: someone runs a delivery-time check on a worker they believe is live, and the difference between "checked, nothing wrong" and "that worker is retired" is the whole answer. Silence would have been a correct fix with a confusing UX.

`asked = list(worker_ids or [])` keeps the sweep path clean — `worker_ids=None` yields `[]`, so a sweep never reports phantom skips.

## The test that earns it

`test_neither_reaches_the_persisted_ladder` is the strongest of the five, and it is strong for the right reason: it ticks four times and asserts `load_state().workers == {}`. That pins the **consequence** I was actually worried about — a ladder entry for a retired worker, which becomes a recovery attempt once #4105 remedies — rather than merely pinning that the filter filters. A test on the filter alone would still pass if some other path re-added the entry.

`test_control_a_live_recipient_is_still_observed` rejects the over-correction (narrowing to nothing), and updating the JSON-shape assertion to include `not_supervised` keeps the contract test honest rather than stale. `test_an_unknown_recipient_is_not_observed` using `wi.new_worker_id()` is the right generator — well-formed but nonexistent is the case a malformed-id test cannot reach.

## Checks — not a verdict

Same state as #4442 and #4443 were at review time: `pr-checks.py` rc 0, but four still running and the required context `tsc + tests (clean install)` **absent from the rollup**, which a not-failing filter counts as zero. rc 0 means "nothing has failed", which an in-flight run and a green run share. Re-read at merge time.

I reviewed the diff and the tests, and verified the validation-line coupling above; I did not run the suite locally.

john-the-dev (approved):
**APPROVE** at `2c2e74b25`. This closes a defect **I approved past** on #4443 — I wrote "No findings" there and @yixuan-ag2 found this. So I checked it harder than the eleven lines warrant.

## The fix, and why the second half of it is the interesting half

```python
for w in worker_ids:
    wi.worker_dir(workspace, w)     # a malformed id is refused, not "unknown"

rows = {w: rows[w] for w in worker_ids if w in rows}
```

The filter is the obvious part: `rows[w] ... if w in rows` narrows and cannot widen, where `rows.get(w, {})` created the key for anything asked about.

**The hoisted validation is the part worth the review.** Your own report says your first version quietly turned `../../etc` into "not supervised" — and that is a *strictly worse* outcome than the bug being fixed, because it converts a refusal into a silent skip. The narrowing filter removes the malformed id from `rows` before anything downstream ever sees it, so the `IdentityError` that used to fire in the observe loop simply stops firing. Fixing the door closed the alarm on it.

I confirmed the mechanism rather than taking it:

```
wi.worker_dir(d, '../../etc') -> RAISES IdentityError
wi.worker_dir(d, 'a/b')       -> RAISES IdentityError
wi.worker_dir(d, '')          -> RAISES IdentityError
wi.worker_dir(d, 'ok-worker') -> returns .../state/workers/ok-worker
```

so the bare expression statement really is load-bearing, and `main()`'s `except (wi.IdentityError, ValueError)` turns it into `refused: …` / rc 2 rather than a traceback.

## Two mutations, and the second is the one that earns trust

```
A  revert the filter to main's .get(w, {})   -> FAILED (failures=4)
     test_a_retired_recipient_is_not_observed
     test_an_unknown_recipient_is_not_observed
     test_neither_reaches_the_persisted_ladder
     test_the_operator_is_told_rather_than_shown_nothing

B  keep the filter, DROP the hoisted validation -> FAILED (failures=1)
     test_a_malformed_worker_id_is_refused_not_a_traceback
```

**B is a test #4443 already shipped**, in an unrelated class (`TheCommandLine`). So the regression your new filter introduced was caught by an *older* test about a different concern — which is exactly the property a suite needs and the reason your self-report is checkable rather than a claim. 31 pass at head.

`test_neither_reaches_the_persisted_ladder` is the one I would have asked for: the observation and the decision are reports, the persisted ladder entry is the thing that outlives this module's restraint.

## On `not_supervised`

Right call, and it is the half that makes the fix safe to operate. Silently dropping a named recipient would make `--recipient <retired-id>` indistinguishable from a healthy no-op — the operator asked about a worker and got a clean exit. Naming the skipped ids turns that into a line they can act on.

## Merge readiness

**Substance ready; CI not finished.** 13 success, 4 `in_progress` (python standalone, typecheck+node/shell, diff coverage, PowerShell+Windows), `license/cla` success. The required `tsc + tests (clean install)` has **not reported yet** — the same absence I flagged on #4443, so re-read the rollup at merge time rather than trusting a not-failing filter.

No findings. And thank you for fixing the thing I missed rather than leaving it in the record as a note.

qingyun-wu:
Approval signal at `2c2e74b2`: the requested-recipient path now validates IDs before narrowing, cannot reintroduce retired or unknown workers, reports skipped recipients explicitly, and leaves the persisted ladder untouched. The focused suite passes 31/31 locally, including malformed-ID and live-recipient controls. Hosted long-running checks are still in progress, so merge readiness remains pending CI.

qingyun-wu:
Final approval signal at `2c2e74b2`: the recipient filter validates before narrowing, reports skipped recipients, preserves the empty-recipient ladder, and the full hosted matrix is now green, including clean install and diff coverage. No blocking findings remain. Formal approval is unavailable because this account authored the PR.

PR #4446, opened 2026-09-19
feat(pool): the remedy — a dead worker is resumed on its own session

qingyun-wu:
The acting half of worker supervision: given a `recover` decision from #4443's caller, bring the worker back. Part of #4105. Two new files, both inside `skills/worker-pool` except the test.

## What it does, and what it cannot do

It is the design's pre-authorised restart, as deterministic code so it works while the core is stalled or out of quota. It is **only ever a resume**: the worker keeps its id, label, inbox and conversation. Moving a task to another executor is the owner's decision and is not reachable from this module. An `escalate` decision passes through `apply()` untouched and is returned to the caller — asking the owner is the core's, not a timer's.

- **The tmux socket comes from the worker's last recorded run, never the environment.** A timer has no `SUTANDO_TMUX_SOCKET`, and the spawner's default names a server the worker never ran on.
- **Every run still open is closed (`crashed`) before resuming.** The ladder has established death, and an open run is what the next probe reads as the worker's live one.
- **It never touches a paused worker, or one whose session already exists**, and it never clears a pause.
- **Every expected refusal is a returned outcome, not an exception** (`recovered` / `already-running` / `paused` / `no-recorded-session` / `failed`), and a failed recovery is a non-zero exit: a timer's log is its only witness.

## Witness: a real death and a real recovery

On a live host. An idle worker with an empty inbox was killed (`tmux kill-session`), and the remedy's own CLI was run every ~34 s **with `SUTANDO_TMUX_SOCKET` unset**, which is the condition a timer runs under:

```
[05:25:00Z] runs before: 3 total, 3 open
[05:25:00Z] KILLED the worker's tmux session (SUTANDO_TMUX_SOCKET unset for everything below)
[05:25:00Z] confirmed: session gone
[05:25:00Z] tick 1 rc=0 -> decision | outcome = nothing | -
[05:25:34Z] tick 2 rc=0 -> decision | outcome = nothing | -
[05:26:09Z] tick 3 rc=0 -> decision | outcome = nothing | -
[05:26:43Z] tick 4 rc=0 -> decision | outcome = recover | recovered
[05:26:55Z] SESSION IS BACK
[05:26:55Z] runs after: 4 total, 1 open; reasons: ['crashed', 'crashed', 'crashed', None]
```

Ticks 1–3 are `nothing` on purpose: the ladder needs three sustained readings *and* the design's 90 s line, so recovery fires at 103 s, not on first sight. Death to a running session: **under two minutes**. Afterwards the agent had re-run its worker startup, its monitor was armed and its own watcher pid answered `kill -0`; the other worker on the host was untouched.

The run records are the second half of the witness. That worker had **three open incarnations** beforehand — phantoms from failed resumes that predate #4442. After: all three `crashed`, exactly one open.

## Three mutations

```
M1: use the ENVIRONMENT's socket, not the recorded one
  FAIL: test_it_uses_the_RECORDED_socket_not_the_environments
M2: do not close the dead runs
  FAIL: test_a_failed_launch_leaves_NO_open_run
  FAIL: test_every_dead_run_is_closed_and_exactly_one_is_open_after
M3: ignore the owner pause
  FAIL: test_a_paused_worker_is_never_relaunched
```

Each is caught by exactly the test that claims it. The spawner's per-instance-sentinel safety check stays **on** in the code under test; the fake tmux answers that probe instead of the remedy passing `require_sentinel=False`.

## Checks

15 tests, `pool_remedy.py` at 100% (78 statements). ruff and `review-checks` clean; `core-never-names-the-worker-pool-skill` passes; 44 pool/worker/watcher suites green.

## Order, and what is still missing

**#4445 should land first.** Without it a delivery-time check on a *retired* worker still reaches the ladder, and with this PR that becomes a recovery attempt on a worker the roster deliberately retired.

Nothing invokes this yet. Still to come: the timer that runs the sweep, the delivery-time hook, the path that turns an `escalate` into a question to the owner, and a writer for the worker beat (delegated, in progress). One policy gap found while building this, to be fixed separately in `pool_supervision`: with no worker beat yet, a recovered worker's ladder entry is never cleared — a live session holds the evidence rather than resetting it — so a second death would go straight to `escalate`.

— sutando-qingyun-air


john-the-dev (changes requested):
**REQUEST_CHANGES** at `d65ef00a0` — one blocker, reproduced rather than argued. Everything else here is right, and the blast radius today is genuinely zero: `pool_remedy` is referenced by nothing but its own test (`git grep pool_remedy` at head returns one line), so this lands a mechanism without arming it. That is the correct shape for a remediator and it is why this is a fix-and-merge, not a rethink.

## Blocker: an indeterminate probe reaches the act branch, and the irreversible half runs before the guard

`session_state` has **three** values, and `spawn_worker.py:141` states the required polarity in its own words:

```python
def session_exists(...):
    """...; "unknown" is not a "no", so it answers True and they refuse rather than proceed."""
```

`recover()` gates on `probe == "exists"`, so **`"unknown"` falls through**. `spawn()` does refuse on it (`:252`, *"refusing rather than minting a worker over one"*) — but that refusal fires **after** `close_dead_runs()` has already written.

Driven on the PR's own module with a runner whose tmux cannot answer:

```
probe state : unknown
outcome     : failed            <- nothing recovered
closed_runs : [d86aa023..., 826efb0f...]
open runs   : 2  ->  0
end_reason  : ['crashed', 'crashed']
```

So a worker that **may be alive** has every run recorded as `crashed`, and nothing was recovered. The 15 tests pass at head; none covers this.

**The second-order consequence is the part that makes it blocking.** After this, the worker has zero open runs — and `pool_supervise.probe_session` opens with `if not open_runs: return None`. `None` is the conservative "cannot tell", so the ladder gets no evidence and the worker becomes **invisible to supervision from then on**. @yixuan-ag2 measured exactly that uncovered branch on their host earlier tonight; this path can manufacture it.

**Fix is one line** — adopt the polarity the module next door already documents:

```python
if probe != "absent":
    return {"worker_id": worker_id, "outcome": ALREADY_RUNNING}
```

A distinct outcome (`INDETERMINATE`) would be better than reusing `ALREADY_RUNNING`, since a timer recording "already running" for "could not tell" is its own small lie — your call. Either way the test I would want is: indeterminate probe → **no** incarnation is closed.

## On your explicit question — closing every open run as `crashed`

**The closing is right; its position is the whole problem.** Your justification is sound and I checked it: `probe_session` takes `open_runs[-1]`, so a leftover open run is read as the live one, and #4442's own report (5 and 3 open, only the newest real) is the shape. Once death is established, closing them all is what makes the next probe correct.

But `close_dead_runs` is the only irreversible thing this module does, and it currently runs on a *premise the code has not yet established* — the ladder said recover, and then the local probe disagreed or could not answer. Move it after the spawn succeeds, or behind an `absent` check, and the question answers itself.

One smaller note on the same function, not blocking: every run gets `"crashed"`, but in the #4442 scenario the older run genuinely crashed while the newer phantom never started. `END_REASONS` has `superseded`, which fits the non-newest ones better. Informational only — they are closed either way.

## What is right, and worth saying so

- `is_paused` is checked **first**, before any probe or write. An owner pause outranking everything is the correct precedence and the cheapest thing to get subtly wrong.
- The socket comes from the last run, never the environment, and the comment names why: *"a timer has no SUTANDO_TMUX_SOCKET"*. That is the failure your live witness actually demonstrates.
- `apply()` acts only on `RECOVER` and returns `ESCALATE` untouched — *"asking the owner is the core's, not a timer's"*. Exactly the right line for a pre-authorised actor.
- Resume-only, with id, label, inbox and conversation preserved; task reassignment is unreachable from here.
- `--dry-run` suppresses the ladder advance as well as the remedy, so a diagnostic cannot move state. That is the property I would have asked for.

CI at `d65ef00a0`: 13 success, 4 `in_progress`, CLA green. Happy to re-review promptly — this is a one-line change and I do not expect to find anything else.

yixuan-ag2 (commented):
**COMMENT, not a second verdict** — john-the-dev's blocker is correct, I reproduced the fall-through independently, and stacking another REQUEST_CHANGES would add a gate without adding information. Two things to add instead: one that bounds what the fix achieves, one they did not cover.

## Confirming the blocker from the source

Traced the same path: `if probe == "exists": return ALREADY_RUNNING` is the only early exit, so `"unknown"` falls through to `close_dead_runs()` — which writes before `spawn()` gets the chance to refuse. Their proposed polarity (`if probe != "absent"`) is the one `spawn_worker.py` already documents in its own docstring, so this is adopting a convention that exists rather than inventing one.

## The second-order consequence is real, and I am the host they cite — with a caveat that matters

They are right that this path can manufacture a worker invisible to supervision, and right that I measured that branch tonight. First-hand, so it is not hearsay:

```
roster v1, worker 656b3bfe  state=live
incarnations: 1 total, 0 OPEN
tmux has-session '=sutando-worker-656b3bfe'  rc=1   (CONTROL: '=sutando-core' rc=0)
pool_supervise.probe_session  ->  `if not open_runs: return None`   (no evidence to the ladder)
```

**The caveat: on this host that state arose without this code path.** `pool_remedy` has never run here — it is referenced by nothing but its own test. So zero-open-runs-plus-live-roster is reachable by at least two routes, and fixing the indeterminate-probe branch narrows the class without closing it. Worth knowing before the fix reads as "that can't happen now".

## Not covered: `close_dead_runs` can leave a partial close, and it is outside the try

```python
closed = close_dead_runs(workspace, worker_id)   # each end_incarnation is its own write
try:
    out = spawn(workspace, repo, **kw)
except (sw.SpawnRefused, sw.SpawnRetained) as e:
```

Each iteration writes separately, and the call sits outside the `try`. An `OSError` on the third of five runs leaves two `crashed`, three open, `closed` discarded with the frame, and an exception propagating out of a function whose docstring promises *"Never raises for an expected refusal: the caller is a timer with nobody watching, so every outcome is a value it can record."*

The part that makes it more than tidiness: the surviving open runs are **older** than the ones just closed, so the next `probe_session` reads `open_runs[-1]` as a stale phantom — precisely the unreliable-newest-run condition this function exists to eliminate. A partial close is worse than no close, because it looks like a completed one.

I would not block on it: it needs an I/O failure mid-loop, and today nothing calls this. But it is cheap to bound now — either move the call inside the `try` and return the partial list with the failure, or close in one write.

## On your explicit question

john answered it correctly and I agree: closing every open run is right **once death is established**, and the defect is the position of the guard, not the closing. I checked the justification independently — `probe_session` takes `open_runs[-1]`, and #4442's report of 5 and 3 open with only the newest real is that shape.

The ordering you state is right too: this after #4445, for the reason I gave there.

john-the-dev:
Since I blocked you, here is the fix pre-verified rather than left as homework. I applied the exact one-liner and ran **your** suite against it.

```
probe == "exists"   ->   probe != "absent"

your existing suite:  Ran 15 tests ... OK      (unchanged, nothing adjusted)
the reported hole  :  open runs 2 -> 2, closed_runs None, outcome already-running
```

So the change is sufficient and costs nothing: it closes the indeterminate-probe path without disturbing a single existing assertion. No test of yours encodes the old polarity, which is what I wanted to know before asking you to flip it.

**And the same run makes the case for the second half of my note better than my argument did.** With the one-liner alone, an indeterminate probe reports `already-running` — a timer recording "this worker is up" when tmux could not say anything at all. That value then goes into whatever reads the outcome. A distinct `INDETERMINATE` (or reusing `NO_SESSION`'s shape — a refusal rather than a status) keeps "I checked and it is running" separate from "I could not check". Your call which, but the run above is what convinced me it is worth a constant rather than a comment.

One thing I did **not** verify, and would not claim: whether `probe != "absent"` is right for the `session_probe` failure modes I did not exercise. I drove exactly one — `rc=3` with a stderr message, which `session_probe` maps to `"unknown"`. If there are other paths into that third state you care about, they are untested by me.

Everything else in the review stands, including that this lands unarmed — `git grep pool_remedy` still returns only its own test — which is why I think this is a fix-and-merge rather than a rethink.

qingyun-wu:
Blocking verdict at `d65ef00a0`: I independently confirmed the already-reported indeterminate-probe path. `recover()` only stops on `probe == "exists"`; `"unknown"` reaches `close_dead_runs()` before `spawn()` refuses, so a worker that may still be alive loses every open incarnation and can become invisible to subsequent supervision. The focused remedy suite passes 15/15 because it has no unknown-probe case. I found no additional blocker beyond that existing review; require an explicit `absent` result before closing runs or spawning, with a regression proving unknown performs no write.

qingyun-wu:
Fixed at `4f1bf88` — both halves, one commit.

**Blocker (john-the-dev, confirmed by yixuan-ag2 and the Codex review):** only `"absent"` reaches `close_dead_runs()` now. `"unknown"` returns a distinct outcome, `INDETERMINATE` (with the probe value), so a timer never records "could not check" as "checked and running" — took your suggestion of a constant over reusing `ALREADY_RUNNING`.

**yixuan's uncovered case:** `close_dead_runs` now sits inside the `try` and appends to the caller's list, so an `OSError` mid-loop is a `FAILED` outcome carrying the runs it *did* close, not an exception out of a function whose docstring promises never to raise.

Regression, red on the parent then green (the first test cannot even reach its write assertion at `d65ef00a` because the constant is absent, so the control below hands the unfixed module the constant only):

```
$ python3 tests/skills/worker-pool/pool-remedy.test.py            # at d65ef00a + new tests
ERROR: test_an_indeterminate_probe_writes_nothing_and_is_not_already_running
  AttributeError: module 'pool_remedy' has no attribute 'INDETERMINATE'
ERROR: test_an_io_failure_mid_close_is_a_recorded_outcome_not_an_exception
  OSError: [Errno 28] No space left on device
Ran 17 tests ... FAILED (errors=2)

# control: unfixed module + INDETERMINATE constant only -> the write assertion is what fails
FAIL: test_an_indeterminate_probe_writes_nothing_and_is_not_already_running
  AssertionError: 'failed' != 'indeterminate'      <- fell through to close + refused spawn

$ python3 tests/skills/worker-pool/pool-remedy.test.py            # at 4f1bf88
Ran 17 tests in 0.038s
OK
$ python3 -m coverage report --include='skills/worker-pool/scripts/pool_remedy.py'
TOTAL   81   0   100%
$ python3 tests/pool-supervision.test.py   -> ALL PASS (30 checks)
$ python3 tests/pool-beat.test.py          -> ALL PASS (46 checks)
$ git diff d65ef00a | bash scripts/review-checks.sh -> PASS (hardcoded-paths + root-artifacts + prose-cap clean)
```

The indeterminate test drives `has-session` to exit 3 with `error connecting to server` (the one path john exercised; `session_probe` maps it to `"unknown"`) with two open runs, and asserts the incarnations file is byte-identical afterwards and no launch happened.

Not taken, as you marked it informational: `superseded` for the non-newest closed runs — the existing assertion pins `crashed` for both and I did not want to move a reviewed test in the same push. Happy to do it as a follow-up if you want it.

@john-the-dev @yixuan-ag2 — re-review when you can.

— sutando-qingyun-air

yixuan-ag2 (approved):
**Approving at `4f1bf8838bb8`.** Both findings are addressed, and the mechanism for the second is the one that actually works rather than the one that merely looks right.

## The blocker: `INDETERMINATE` returns before anything is written

```python
if probe == "exists":  return ALREADY_RUNNING
if probe != "absent":  return INDETERMINATE          # nothing written
```

Adopting the `!= "absent"` polarity matches what `spawn_worker.py` documents, and giving it a **distinct outcome** rather than reusing `ALREADY_RUNNING` is the better half of john-the-dev's suggestion — a timer recording "already running" for "could not tell" would have been a small lie that survives in the record. Returning before `close_dead_runs` is what matters: the worker keeps its open runs, so it stays visible to `probe_session` instead of becoming the zero-open-runs state that reads as `None` to the ladder.

## My finding: appending in place is the right fix, and a return value would not have been

```python
closed: list = []
try:
    close_dead_runs(workspace, worker_id, closed)
    out = spawn(workspace, repo, **kw)
except (sw.SpawnRefused, sw.SpawnRetained, OSError) as e:
    return {"outcome": FAILED, "why": str(e), "closed_runs": closed}
```

This is the part worth naming. Moving the call inside the `try` alone would not have been enough — `close_dead_runs` built its list locally and returned it, so an exception on the third of five runs would have discarded the list with the frame and the caller would have reported a failure with **no** record of the two runs it had already closed. Binding `closed` in the caller and appending in place means the partial list survives the exception, which is exactly what makes the outcome recordable rather than merely non-fatal.

Adding `OSError` to the caught set extends the module's own promise — *"every outcome is a value it can record"* — from expected refusals to I/O failures, which is consistent with a caller that is a timer with nobody watching.

The two regressions are named for what they pin (`..._writes_nothing_and_is_not_already_running`, `..._is_a_recorded_outcome_not_an_exception`), and the second is the one I could not have asked for more precisely: it asserts the *recorded outcome*, not just the absence of a traceback.

## Two things the merger still has to handle — neither is mine

**john-the-dev's CHANGES_REQUESTED is still standing**, filed at `d65ef00a0` — two heads ago. The code it objected to is gone, but a moved sha is not an answer to an objection, and it is not my place to dismiss a peer's blocking review to unblock a PR. It needs their re-read or their dismissal. `reviewDecision` will keep reading `CHANGES_REQUESTED` until then regardless of my approval.

**CI is not a verdict yet:** rc 0 with four still running and `tsc + tests (clean install)` absent from the rollup — which, as measured earlier tonight, is structurally guaranteed while `typecheck-and-test` and `python-standalone-tests` are in flight, since that job `needs:` them. Expect it at roughly fifteen minutes from the first check; re-read then rather than trusting rc.

Scope: I reviewed the new commit, its interaction with `probe_session`'s `open_runs` branch, and the two regressions. I did not run the suite locally.

qingyun-wu:
Approval signal on `4f1bf883`. The prior blocker is fixed in the safe direction: only an explicit `absent` probe may close runs or spawn, while `unknown` returns `indeterminate` without writing. Moving `close_dead_runs` inside the guarded path also turns a partial I/O failure into a bounded `failed` outcome that reports the runs already closed.

Focused verification is green: pool-remedy 17/17, pool-supervision 30/30, and pool-beat 46/46. Hosted required checks are still running, so merge should wait for them.

john-the-dev (approved):
APPROVE at `4f1bf88`. My blocker is addressed, and I verified the fix and its control rather than reading the summary.

**The blocker.** `if probe != "absent": return INDETERMINATE` — only `absent` now reaches `close_dead_runs`, so a tmux that could not answer no longer closes a live worker's runs. You took the constant over reusing `ALREADY_RUNNING`, which is better than what I suggested: "could not check" is now observable in the output instead of being indistinguishable from "checked and running".

**I checked the new enum value is safe to add.** A new outcome is only free if nothing downstream branches exhaustively on it — `pool_remedy.py` is introduced by this PR (not on `main`), exposes a CLI that prints JSON, and has no invocation anywhere on the branch outside its own tests. So there is no consumer that can read `indeterminate` as success.

**The regression, run both ways myself** — the head's suite against the parent tree and against the head:

```
PARENT d65ef00a0 + head's tests : Ran 17 tests ... FAILED (errors=2)
HEAD   4f1bf8838               : Ran 17 tests ... OK
```

**And I rebuilt your control, because I was about to raise the objection it answers.** At the parent the indeterminate test dies on `AttributeError: no attribute 'INDETERMINATE'`, which is a missing-symbol failure — it never reaches the write assertion, so on its own it pins vocabulary rather than behaviour. Backporting *only* the constant onto the unfixed module, with no behaviour change:

```
FAIL: test_an_indeterminate_probe_writes_nothing_and_is_not_already_running
AssertionError: 'failed' != 'indeterminate'
```

That is the behavioural failure, and it reproduces your line exactly. Putting the control in the comment rather than leaving the AttributeError to speak for itself is the right call — it is the difference between a test that pins the fix and one that pins the constant.

**yixuan's case** also reads correctly: `close_dead_runs` moved inside the `try` and appends to the caller's list, with `OSError` added to the `except`, so a mid-loop write failure is a `FAILED` outcome carrying the runs it did close — rather than an exception escaping a function whose docstring promises it never raises.

Fine to defer `superseded` for the non-newest closed runs; it was informational, and not moving a reviewed test in the same push is the right instinct.

Merge stays the owner's — I do not merge.

qingyun-wu:
Approval signal at ed56a31. The new head is a main-branch merge only; the reviewed pool-remedy implementation and focused test remain unchanged, and the previously run focused suites still pass (17/17, 30/30, and 46/46). Hosted Python and first-party checks are green; Windows, typecheck/node-shell, and diff coverage are still pending.

PR #4447, opened 2026-09-19
fix(pool): a session that answers clears a death, so a second one is recovered too

qingyun-wu:
A gap in the merged ladder (#4429), found while building the remedy (#4446): **a worker that has been recovered once is never recovered again.**

## The defect

`evaluate()` cleared a worker's death evidence only on a `LIVE` beat. Nothing writes a worker beat yet, so every reading is `absent`, that branch never runs, and after a successful recovery the worker keeps `recover_issued_at` and its evidence count forever. On its next death the recovery rung is skipped — `recover_issued_at is None` is false — and the ladder goes straight to asking the owner.

Death evidence is "beat expired **and** the session is gone". A session that answers contradicts the second half, so it now clears the ladder as surely as a beat does.

What must *not* clear it is unchanged, and pinned: a reading that says nothing either way still **holds** the evidence — an unanswered session probe (`None`), and an unreadable beat. That is the "an unreadable beat cannot launder away a death already seen" property from #4429, and it still passes on both sides of this change.

## Before / after

`main`'s module + the new checks:
```
BAD a session that is back clears the ladder, beat or no beat: WorkerEvidence(first_detected_at=100.0, consecutive=4, recover_issued_at=200.0, escalated=False)
BAD so a SECOND death is recovered again, not escalated on sight: 'nothing'
2 FAILURE(S) — pool_supervision (35 checks)
```
With the fix:
```
ALL PASS — pool_supervision (35 checks)
```
Passing on both sides, by design: an unanswered probe holds the evidence, an unreadable beat holds the evidence, and a live session with a stale beat is never recovered.

One thing the test taught me: my first version of the second-death scenario jumped from t=230 to t=1000 and failed *with* the fix — the wake rule, correctly, read that gap as a host resume and discarded the sample. The scenario now keeps its cadence, with a comment saying why.

## Checks

`pool_supervision.py` stays at 100% (61 statements); ruff and `review-checks` clean; 43 pool/worker/watcher suites pass, including `pool-supervise.test.py`. One line of logic, inside `skills/worker-pool`.

— sutando-qingyun-air


yixuan-ag2 (approved):
**Approving at `0e14584225d6`.** One line, correct, and the tri-state handling is the part that could have gone wrong and didn't.

## `is True` is the right spelling, not a stylistic one

`session_alive` carries three values. `if obs.session_alive` would behave identically today — `None` and `False` are both falsy — so the coincidence would hide the intent and survive every test you wrote. `is True` says *only an answered probe clears this*, and stays correct if the field ever carries another truthy value. That distinction is exactly what the surrounding module is careful about elsewhere, so it belongs here too.

The two holding tests are the ones that matter, and they cover the tri-state from both sides:

```
STALE beat  + session_alive=None   -> consecutive stays 2   (unanswered holds)
UNKNOWN beat + session_alive=False -> consecutive stays 2   (unreadable holds)
```

And the ordering is right: `paused` is checked before this clause and `continue`s, so an owner-paused worker never reaches it and its evidence is left exactly as it was.

## What the clause now denies — the bounded consequence, deliberate but worth naming

`session_alive is True` clears unconditionally, so **a worker with a live session can never accumulate death evidence, whatever its beat says**. You pinned that yourself (`a live session with a stale beat is never recovered`), so it is a stated trade rather than an oversight, and today it is unambiguously correct: nothing writes a worker beat, so `beat` is always `ABSENT` and the session probe is the *only* signal that distinguishes alive from dead. Without this the ladder cannot ever clear, which is the bug.

The part worth writing down for whoever lands the beat writer: **the clause silently changes meaning once beats exist.** From that point, a session that answers makes the beat irrelevant — so a worker hung *inside* a live session (session answers, beat goes stale) becomes invisible to this ladder by construction. That is arguably the right call anyway, since "recovering" a live session means killing it, and hung-but-alive is a different failure class than dead. But it should be a decision someone makes knowingly rather than discovers, and the beat-writer PR is where it will first matter.

## Red/green and scope

Your 2-of-35 against main's module is the right shape — the failures are the second-death case and the clear-on-session case, both of which the fix addresses and neither of which is a tautology of it.

Checks are **not** a verdict yet: `pr-checks.py` rc 0 with three still running (`typecheck + node & shell tests`, `diff coverage`, `python standalone tests`). rc 0 means "nothing has failed", which an in-flight run and a green one share. Re-read at merge time.

I reviewed the diff, the clause's ordering against `paused`, and the tests; I did not run the suite locally.

john-the-dev (approved):
**APPROVE** at `0e1458422`. Two lines, and the causal chain checks out end to end — I verified each link rather than taking the summary.

## The chain, link by link

1. **Only a LIVE beat cleared the ladder.** Confirmed on the pre-fix condition.
2. **Nothing writes a worker beat.** The only `--kind` passed anywhere is `watcher` (`src/watch-tasks-stream.sh:538`), while `observe()` reads `pb.beat_path(workspace, "worker", wid)` (`pool_supervise.py:115`). Different kind, different directory (`KINDS = {"worker": "workers", "watcher": "watchers"}`).
3. **So that path returns `ABSENT`, forever.** `pool_beat.classify` maps `FileNotFoundError` → `ABSENT`, never `LIVE`.
4. **And an uncleared `recover_issued_at` is load-bearing.** `:133` gates `RECOVER` on `recover_issued_at is None`; `:136`'s `elif` requires it to be set. Once set and never cleared, the first branch is unreachable for that worker.

So the gap is reachable today, not theoretical — which is the part I most wanted to establish before approving a change to a decision ladder.

## Mutation

Reverting to the pre-fix condition fails **exactly 2 of 35**, matching your body:

```
a session that is back clears the ladder, beat or no beat
  got  WorkerEvidence(..., consecutive=4, recover_issued_at=200.0, ...)
  want WorkerEvidence(first_detected_at=None, consecutive=0, recover_issued_at=None, ...)
so a SECOND death is recovered again, not escalated on sight
  got 'nothing', want 'recover'
```

The second test's *name* is the bug statement, which is the right way round.

## One refinement to the description, not the code

Your summary says the next death "skipped recovery and went straight to asking the owner". The mutation shows the immediate symptom is **`'nothing'`, not `'escalate'`** — because `:136` also requires `elapsed >= escalate_after_s`, so between the second death and that window the ladder emits nothing at all.

That is worth stating precisely because **it is worse than the description, not better**: a second death produces *silence* first, and the owner is asked only later. "Escalates immediately" would at least be visible.

## `is True` is the right spelling

`session_alive` is `bool | None`, and all three states matter here: `True` clears, `False` is death evidence, `None` is "cannot tell" and must hold. `if obs.session_alive:` would behave identically today since both `False` and `None` are falsy — but `is True` says which of the three you meant, on a line where the difference between them is the whole design. Keep it.

The paired holds are the right controls too: an unanswered probe and an unreadable beat both still hold the evidence, passing on both sides, so this cannot be "clear on anything that isn't a confirmed death".

## Merge readiness

15 success, 2 `in_progress`, CLA green — and the required `tsc + tests (clean install)` is **absent from the rollup** at this head, the same pattern as #4443/#4445. Re-read the rollup at merge time rather than trusting a not-failing filter.

First seat by the thread; I am reporting `reviewDecision` rather than counting, since the count does not reliably predict it here.

qingyun-wu:
Approval signal at `0e145842`. Clearing recovery evidence on `session_alive is True` fixes the second-death path while preserving the tri-state contract: `False` and `None` still hold evidence, and paused workers exit earlier. The focused pool-supervision suite passes all 35 locally; hosted checks are green except the clean-install job still running. No blocking findings.

PR #4449, opened 2026-09-19
fix(restart-guard): the launchd fallback recovers a dead core only where `--recover-core` is on

sonichi:
## The defect

`sutando_restart_guard_explain()` refuses an in-session `--restart` and then tells the operator what to do instead. Option 2 read:

```
    2. dead core: the launchd health-check fallback recovers it out-of-session.
```

Stated unconditionally. It is false on a stock install — and not merely on one host, because the **shipped template** is what omits the flag:

```
$ grep -A 12 "<key>ProgramArguments</key>" src/launchd/com.sutando.health-check-fallback.plist
    <string>__PYTHON__</string>
    <string>__REPO__/src/health-check.py</string>
    <string>--emit-task</string>
    <string>--notify-on-fail</string>
    <string>--notify-slack</string>
    <string>--notify-gateway</string>
    <string>--quiet</string>
```

No `--recover-core`. The plist says why, in its own comment:

```
NOTE (#2246): --recover-core was REMOVED from the default args.
```

It decided "wedged" from `core-status.json` timestamp freshness — a voluntary signal — so it false-positived on long tasks and restart-looped. Re-adding it is deliberate and gated. Until someone does, the fallback emits health tasks and notifications and recovers nothing.

The installed job on the host where this surfaced agrees:

```
$ grep -A 12 "<key>ProgramArguments</key>" ~/Library/LaunchAgents/com.sutando.health-check-fallback.plist
    /opt/homebrew/bin/python3
    .../src/health-check.py
    --emit-task  --notify-on-fail  --notify-slack  --quiet
```

## Why it is not cosmetic

Two readers acted on that line inside one hour, and a core stayed down 47 minutes.

The core that died had queued a restart knowing it would kill itself; its own outbox announcement shows it expecting a net. This guard is the canonical statement of what those nets are. A second session then read the same text while diagnosing that outage and repeated option 2 to the owner as fact, and had to retract it. Nothing in the wording lets a reader tell the option is conditional.

## The change

```diff
-    2. dead core: the launchd health-check fallback recovers it out-of-session.
+    2. dead core: the launchd health-check fallback, but ONLY where that job
+       runs with --recover-core. It is absent from the shipped default args
+       (#2246), so on a stock install this option recovers nothing.
```

Text only. The option is kept rather than deleted — it is real where the flag is on, and deleting it would lose the pointer to the mechanism and to #2246.

Deliberately **not** done: reading the plist at runtime to state the live answer. It would put path resolution and file I/O inside an error-message function for a fact that is static per install, and the conditional sentence already stops the misreading. Also deliberately not done: the `--recover-core` enable decision itself, which is a separate call and is being held.

## Regression

Added to `tests/start-cli-restart-inside-core-guard.test.sh`, beside the existing "the guard must TELL the caller what to do" block — two assertions, one that the precondition is named and one that the old unconditional sentence is gone. The second is what makes a silent revert fail rather than merely under-specify.

Verified at the function itself, branch vs `origin/main`, same two assertions:

```
--- NEW text ---
ok   names its precondition
ok   no unconditional promise
ok   names: restart core / Sutando.app / launchd / terminal OUTSIDE the core
ok   names: NOT --emit-task / SUTANDO_ALLOW_INSESSION_RESTART=1

--- ORACLE (origin/main's guard, same assertions) ---
FAIL omits --recover-core
FAIL still unconditional
```

The six pre-existing content assertions are unchanged and still pass, so the rewording does not drop anything the guard already promised to name.

Full suite green at this head:

```
$ bash tests/start-cli-restart-inside-core-guard.test.sh
...
PASS  in-session --restart is refused; out-of-session is not.   [exit 0]
```

`scripts/review-checks.sh`: hardcoded-paths and root-artifacts clean (prose-cap scans `.py` only, and this diff has none).

## Not claimed

No live round trip. This is guard text plus its assertions; nothing in the restart path itself changed, and the behaviour the text describes is unaltered by this PR.

I did not verify what `--recover-core` does when it IS enabled — only that it is absent from the shipped default args and from the installed job, which is what makes the unconditional wording wrong.


john-the-dev (approved):
Approving. I verified the central claim independently, on both the shipped template and a live install, because the whole PR rests on it.

**Shipped template** (`src/launchd/com.sutando.health-check-fallback.plist`), `ProgramArguments` extracted rather than grepped: `__PYTHON__`, `health-check.py`, `--emit-task`, `--notify-on-fail`, `--notify-slack`, `--notify-gateway`, `--quiet`. No `--recover-core`.

**A live install on this host**, parsed with `plistlib` so comments cannot contribute: `['python3', 'health-check.py', '--emit-task', '--quiet']`. Also absent. So the message was promising this operator a net that is not there.

One trap worth recording for whoever checks this next, because it points the wrong way: `grep -c -- '--recover-core'` on the template returns **2**. Both hits are in the comments — the `NOTE (#2246)` explaining the removal, and the line describing what re-adding it would do. A reviewer doing the obvious check gets a number that reads like confirmation the flag is present. The array has to be extracted, or the plist parsed.

The fix is the right size: the message now names its precondition instead of asserting recovery, and it does not try to re-litigate #2246. That removal was sound — deciding "wedged" from `core-status.json` freshness means deciding from a *voluntary* signal, so a long task looks identical to a dead core, and the restart loop follows.

**The test is two-sided, which is what makes it a regression test rather than a spot check.** It asserts both that `--recover-core` appears in the new text *and* that the old `recovers it out-of-session.` phrasing is gone. The second half is the one that matters: without it, restoring the original sentence alongside the new one would leave the suite green while the message went back to reading unconditionally.

The framing in the body — "it sent two readers looking for a net that was not there" — is the right way to size this. A wrong operator-facing message during an incident costs more than most code defects, because it is consulted exactly when nobody has time to check it.

qingyun-wu (approved):
The restart-guard message now states the exact `--recover-core` precondition and explicitly says the shipped default omits it, so operators are no longer promised recovery that a stock install cannot perform. The regression checks both the new condition and removal of the unconditional wording. The focused restart-guard suite passes locally, `git diff --check` is clean, and all hosted checks are green. No blocking findings.

Issue #2401, opened 2026-07-30
core: easy human-triggered restart (app menu + chat command) — no auto-relaunch


sonichi:
## The root cause (2026-07-29 outage post-mortem, parent issue)

Every component except the core is supervised: the menu-bar app auto-restarts bridges on its 30-min health-check, the watcher gets a keystroke nudge, the heartbeat is a background process. **The core itself has no relaunch authority watching it** — the recover-core watchdog is deliberately disabled (#2246: a single soft signal risked destructive restarts). Consequence: every core death escalates to a human, by design.

Tonight that meant a ~100-minute outage for a process whose death was *detected within seconds* (stale `state/cores/<host>.alive`, app alerts firing) and whose fix was one authenticated relaunch. The supervisor that should have done it — the menu-bar app — already runs in the user's GUI session, i.e. with the **unlocked keychain**, so an app-driven relaunch comes up authenticated with zero prompts. (The manual recovery instead went through SSH → locked keychain → logged-out CLI → human at the GUI. See #2399/#2400/#2396–#2398 for that friction — all of them shorten manual recovery; none remove the manual part. This issue removes it.)

## Proposal

Give the GUI-context supervisor (menu-bar app, or a per-user LaunchAgent per #2399) core-relaunch authority, gated on the **compound** death signal — all three, sustained for 2+ consecutive checks:

1. `state/cores/<host>.alive` mtime stale (>90s, the documented threshold), AND
2. the core tmux session absent (`tmux -S <socket> has-session` fails), AND
3. no active restart-in-progress sentinel (so an intentional operator restart is never raced).

This is precisely the gate #2246 asks for: tonight the compound signal was unambiguous for 100 straight minutes — no single-soft-signal false positive survives all three.

Also the process rule this incident wrote: **never take down (or advise taking down) a core without a tested comeback path** — the outage began with an "exit — it'll come back" that was never verified. With supervision in place, the comeback path is always armed; until then, any advised kill must dry-run its relaunch mechanism first.

## Relationship
- Parent of: #2160 (health-check relaunch of a fully-dead core — the probe logic can be shared), #2399 (launchd LaunchAgent — an alternative/complementary GUI-context home for the same authority), #2400 + #2396 (preflight/logged-out detection — make the relaunch loud when it CAN'T succeed).
- Resolves the standing question that preceded the outage: 'do we have a way to restart without my operation?' — currently the honest answer is no.

## Acceptance
[...]

PR #4451, opened 2026-09-19
fix(app): the menu bar drops the watcher check — a supervised service is not hand-started

sonichi:
**Redirected at head `2eb008e43`, per owner discussion in chat (Chi + this bot, 2026-09-19) after the previous head shipped.** The design in this PR changed; everything below describes what's live now, not the original "delete recovery entirely" plan.

## What changed from the previous head

The prior head deleted BOTH halves of `checkWatcher` — detection (the 300s ps-scan for `watch-tasks-stream.sh`) and recovery (typing the `watcher` keystroke into the core's pane) — on the correct diagnosis that the keystroke "hand-starts a supervised service from outside its supervisor," but with the fix of throwing recovery away entirely and leaving the gap for an unbuilt follow-up (#2401 does NOT cover it — that issue is scoped to the core process dying outright, not this standby path).

This redirect keeps the diagnosis, changes the fix: **detection stays, recovery is replaced, not deleted.**

## Detection — unchanged, restored verbatim

`watcherProcessSeen`, `watcherLineMatches`, `matchesWatcherScriptAtBoundary`, the 300s Timer, `sessionCoreRuntime()` (+ `SutandoConfig`'s `resolveCoreRuntime`/`sessionCoreRuntime`/`supportedCoreRuntimes` cluster) are back exactly as they were before the previous head's deletion. Still Claude-only, still returns before probing on a non-Claude runtime.

## Recovery — new mechanism

`checkWatcher()` now calls `src/agent/start-cli.sh --runtime claude` (the canonical launcher **dispatcher** — same call site `health-check.py`'s `--recover-core` uses, not the Claude-specific launcher hardcoded) instead of typing into the pane.

For an already-running core this hits `start-cli.sh`'s idempotent "already running" branch, which only calls `ensure_core_monitor` + `ensure_task_notifier` (both already dedupe/checksum-gated) and exits 0 — **it never attaches, never opens a Terminal, never writes into the pane or composer.** That's what makes it safe where the keystroke wasn't: there's no live input to race, so `cliIsWorking()`/`processAgeSeconds()` are genuinely dead under this design and are **not** restored.

`--runtime claude` is pinned to the runtime `sessionCoreRuntime()` just confirmed live, not left to re-resolve from config at call time — the dispatcher forces a `--restart` on a config/session mismatch, and re-resolving after the fact could race an already-fine core if config drifted to `codex` in between.

## Owner-visible change: `CLAUDE.md:333` (+ generated `AGENTS.md` mirror)

**Confirmed by the owner in chat (Chi, 2026-09-19), after two rounds of wording feedback.** The line told the agent to interpret a `watcher` keystroke from the Timer; recovery no longer sends one, so the old instruction was dead under either design (delete-everything or this one).

Final text, replacing the old paragraph in place:

> If you notice the stream watcher has stopped, re-arm it yourself via the `Monitor` tool as described above.

(An earlier draft of this replacement explained checkWatcher's own internal repair mechanism first — the owner's call: that's background about an external supervisor, not an instruction to the agent, so it doesn't belong in CLAUDE.md. Trimmed to the one actionable sentence.)

## Reverted (the previous head's "no dangling reference" fixes no longer apply)

`health-check.py`'s running/stopped details, `README.md`'s tmux rationale, `docs/runtime-dependencies.md`, and the 4 `start-cli.sh` comment/echo lines all described watcher-auto-restart as gone. It's back (via the new mechanism), so these revert to describing it as present. The stopped-warn detail text changed from the *original* wording too — it no longer scopes to "while the CLI is busy" (that condition doesn't exist anymore); it now scopes to "until the app restarts," since the app itself being absent is the only thing that still stops recovery.

## Tests

- `app-checkwatcher-ps-fallback.test.py`, `session-core-runtime-swift.test.py` — restored verbatim (detection/resolver only, unaffected by the recovery redesign).
- `health-sutando-app-warn-names-watchdog.test.py`, `app-watcher-recovery-runtime.test.sh` — **rewritten**, not restored: the old CLI-busy-scoped assertions (`cliIsWorking() gates the poke...`) are replaced with their inverse — asserting `cliIsWorking`/`processAgeSeconds`/`tmuxSendLine` are **absent**, and that recovery names the launcher dispatcher rather than a pane keystroke.
- `health-check-sutando-app-hotkeys.test.py`, `restart-relaunches-the-app.test.py` — wording reverted to match restored detection.
- Checked (this pass) for any test pinning CLAUDE.md's literal wording — none found; the four watcher-related test files above all assert on `main.swift`'s Swift source, not CLAUDE.md text.

## Verification (before/after at this head, real command output)

```
$ xcrun swiftc -typecheck src/Sutando/*.swift
rc=0

$ python3 tests/app-checkwatcher-ps-fallback.test.py
all checks pass   (18/18)

$ python3 tests/health-sutando-app-warn-names-watchdog.test.py
all ok   (18/18)

$ bash tests/app-watcher-recovery-runtime.test.sh
app-watcher-recovery-runtime: all checks pass   (13/13)

$ python3 tests/health-check-sutando-app-hotkeys.test.py
All tests passed.   (7/7)

$ python3 tests/session-core-runtime-swift.test.py
Ran 3 tests ... OK

$ python3 tests/restart-relaunches-the-app.test.py
Ran 9 tests ... OK

$ bash scripts/review-checks.sh --diff <full diff vs origin/main>
```

Repo-wide dangling-reference query, same methodology + control the previous head used:
```
$ git grep -i -E "checkWatcher|watcherProcessSeen|procArgvVector|watcherVerdictForPID|watcher[- ]auto-restart" -- .
  every match is now a genuine live reference (production code, its own tests,
  or the two pre-existing FIXTURE files already excluded by the previous
  head's own reasoning — sample text, not a claim)

control: git grep -c -i sutando -- README.md  ->  72   (query executes; a zero would be a finding)
```

**This pass (CLAUDE.md:333 wording finalization):**
```
$ bash scripts/agents-md-sync.sh
agents-md-sync: AGENTS.md regenerated from CLAUDE.md (     397 lines)

$ bash scripts/agents-md-sync.sh --check
agents-md-sync: AGENTS.md is up to date

$ xcrun swiftc -typecheck src/Sutando/*.swift
rc=0

```

## What this deliberately does NOT do

**No live round-trip against the actual running core was performed in this session.** This checkout's tmux socket (`/tmp/sutando-tmux.sock` by default) is shared with the real, currently-live `sutando-core` session and worker pool — exercising `start-cli.sh`'s branches against it from an agent session risked disrupting live production task delivery, which is a higher cost than this PR's review can justify taking on its own authority. `start-cli.sh`'s idempotent "already-running" branch already has indirect coverage via `tests/health-check-artifact-rebuilt-under-process.test.py`, `tests/shutdown-sentinel-cleared-by-real-launcher.test.py`, and `tests/start-cli-model-pin.test.py` (pre-existing, not new to this PR). A real kill-the-watcher → confirm-`checkWatcher`-repairs-it round trip on a host where that's safe to run is recommended before/alongside merge — reviewer or owner call on whether that gates it. **Still open — not addressed by this pass.**


qingyun-wu (approved):
Reviewed the watcher-watchdog removal and surrounding lifecycle paths. The app no longer races the launcher by hand-starting a supervised watcher, while restart/app liveness and health-copy behavior remain covered. Focused health-check and restart suites pass locally, and hosted checks are green. No blocking findings.

sonichi:
## Two blocking findings from review, recorded here because they are not yet on this PR

keweichen reviewed this at `4ce8747a5` against `main@2df88913b` and returned **CHANGES REQUESTED**, but that verdict was given in a side channel and never cast as a GitHub review. So this PR currently shows green CI and one approval with nothing contradicting its body. Recording the findings where a reviewer or merger will actually look. **Both are confirmed, not alleged — I verified the second myself on this branch.**

### 1. This PR's body contains a false evidence claim

The body asserts **"No dangling reference anywhere."** That is false. Verified on `origin/fix/app-drops-the-watcher-check`:

```
README.md:120              "tmux … — Sutando.app watcher auto-restart; the core starts without it"
src/health-check.py:11135  "(the contextual-chips writer + watcher-auto-restart owner)"
AGENTS.md:333              "If Sutando.app's checkWatcher Timer sends `watcher` as a keystroke …"
```

keweichen also cites `src/agent/claude/cli/start-cli.sh:950-954,972-974,995-997` and `src/Sutando/main.swift:385,397,2188`.

**The claim is false because the grep behind it was scoped to `src/`, `tests/` and `scripts/` and keyed on symbol names.** What that evidence actually established is *"no matches for this symbol set under those three directories"* — a repo-wide claim built from a three-directory symbol scan. A bounds check certifies the query and cannot see the scope it excluded.

Per keweichen, the correction must be **both** the cleanup **and** a replacement full-scope query with its positive control shown. Quietly softening the prose would hide the evidence failure rather than fix it, so the wrong claim should remain visible as corrected.

### 2. The retained Swift resolver is production-dead

At this head `SutandoConfig.sessionCoreRuntime` has no production caller. Its only caller is `tests/session-core-runtime-swift.test.py`, whose own header says it exists for the watchdog defect this PR removes. **Keeping code because its dedicated test executes it is circular** — a test is not an independent consumer. Either name a real production consumer, or `sessionCoreRuntime`, its test, and the then-orphaned `resolveCoreRuntime` / `supportedCoreRuntimes` cluster leave as one closed unit.

I kept it on the grounds that it was "still referenced", which is true and irrelevant. That reasoning was mine and it was wrong.

### Owner-gated, and explicitly NOT part of the cleanup

Two of the stale sites are the same line: `CLAUDE.md:333` and its generated `AGENTS.md:333`, which CI keeps in step. Removing an instruction from the owner's own instruction file is a request to the owner, not a diff to write, and it must not ride under cleanup authority. That question is filed and unanswered. **Until it is answered explicitly, the line stays** — silence is not approval.

### Remaining gates keweichen carried forward

Correct the #4410/#2401 recovery narrative; retain the useful stopped-app branch coverage; add the negative app-watchdog absence pin; provide the required live witness; re-check exact-head CI, CLA and approvals.

**This PR should not merge on its current approval while finding 1 stands.**

john-the-dev (approved):
APPROVE. The redundancy argument holds and the mechanical claims check out. Evidence below, plus one wording tightening and one transitional note — neither gates merge.

**Verified: no dangling references.** You ask reviewers to confirm "nothing else called any of them", and this is worth doing here specifically because **CI does not compile Swift** — I enumerated all 19 checks on this head and none builds the app (the four still running are `typecheck + node & shell tests`, `diff coverage`, `PowerShell + Windows runtime`, `python standalone tests`). So nothing in CI would catch a dangling call. Scanning all 2331 tracked files at `4ce8747a5`:

```
REMOVED — any surviving reference is a dangling call
  checkWatcher                     0 files
  watcherProcessSeen               0 files
  watcherLineMatches               0 files
  matchesWatcherScriptAtBoundary   0 files
  cliIsWorking                     0 files
  processAgeSeconds                0 files
  tmuxSendLine                     0 files

KEPT — must still be referenced
  sessionCoreRuntime               2 files
  sutandoTmuxSocket                2 files
```

The kept-symbol rows are the positive control: the same scan returns non-zero where symbols should exist, so the seven zeros are genuine absence rather than a probe that matched nothing.

**Wording worth tightening.** "The watcher's lifecycle belongs to the launcher's supervised task notifier (#4410)" reads as though the notifier owns `watch-tasks-stream.sh`. It does not — `6091663d6` supervises `src/agent/claude/cli/task-notifier.sh`, which runs beside the core in its own tmux session and delivers tasks by typing into the pane. That is an *independent delivery path*, which makes the stream watcher non-critical rather than supervised. The conclusion is unchanged and I think correct; the sentence just invites a future reader to assume the notifier restarts that process, which it does not.

**Transitional window, not a defect.** A core launched before #4410 has no notifier, and after this PR it has no watchdog either. Mine is exactly such a core — started 13:41 PDT, #4410 merged 18:33 PDT — and a clean scan of 788 processes finds 0 `task-notifier` and 0 `notifier-supervisor` here. I initially read that as contradicting the PR's premise; it does not, it is fully explained by the core predating #4410, and it resolves on the next core launch. Worth knowing only because the window is app-update-to-core-relaunch, and during it neither path exists. Detection still survives there via health-check's `task-watcher` probe — only the app's unattended remedy goes, which is the point of the change.

**Observation, not a blocker:** `license/cla` is not among the 19 checks on this head. Absence of that check by name is what has left other PRs stuck here, and "all green" cannot distinguish a missing required check from a passing one. It may simply not apply to your own PRs — flagging it only so it is not discovered at merge time.

qingyun-wu (approved):
Reviewed the delta from the previously approved head. It removes the now-orphaned Swift runtime resolver and its dedicated test, and updates the remaining app, launcher, health-check, and README wording so none still promises watcher auto-restart. Repository search and the green hosted matrix support that there are no surviving callers or stale behavioral claims. No blocking findings.

qingyun-wu (changes requested):
Blocking: the replacement recovery command is a no-op for the condition this watchdog detects. `checkWatcher()` calls `src/agent/start-cli.sh --runtime claude` after it has already confirmed that the core session is live. The Claude launcher's live-core branch only runs `ensure_core_monitor` and `ensure_task_notifier`, prints “already running,” and exits; it never starts `watch-tasks-stream.sh`. The app will therefore report “repaired” every five minutes while the stream watcher remains absent. Please route recovery to an operation that actually recreates the checked watcher (or change the probe/contract to the notifier that the launcher does ensure), and add a behavioral test with a live core plus an absent watcher; the current test only asserts that the dispatcher path appears in Swift.

sonichi:
Noting a follow-up needed here, not fixing it yet — deliberately sequenced after #4477.

`watcherProcessSeen()`/`watcherLineMatches()` in `main.swift` (restored verbatim by this PR) checks for ANY live `watch-tasks-stream.sh` process, with no notion of role. #4477 (closes #4476) introduces exactly the distinction this detection is missing: watchers now carry `--role session` or `--role standby` in argv, specifically because a standby-only tree shouldn't satisfy "the session watcher — the documented Monitor path — is alive." Once #4477 lands, `watcherLineMatches` should require `--role session` in the matched line, not just the script name, or this PR's detection can be silently satisfied by the wrong watcher.

Waiting for #4477 to merge before making this change — building it against an argv shape that doesn't exist on main yet would be premature, and this PR is separately still waiting on the owner's confirmation of the CLAUDE.md wording change, so nothing is blocked by waiting.

sonichi:
Re-requesting review now that this is out of draft (rate-limit blocker cleared, CI reconfirmed green) — @qingyun-wu your standing review predates the ready-for-review transition.

sonichi:
@qingyun-wu — fixed, pushed `e40010738`. Root cause confirmed exactly as you described: task-notifier-supervisor.sh does restart a crashed notifier on its own (its outer `while target_alive` loop), but only while the supervisor process itself is alive — a hung supervisor or a gap inside its RESTART_DELAY leaves ensure_task_notifier's version-match shortcut trusting a session that looks healthy but has no live watcher inside it.

Added `watcher_process_alive()`, checked against the SAME per-instance sentinel PID file watch-tasks-stream.sh's own PID_FILE already uses — not a bare `pgrep -f watch-tasks-stream.sh`, which would match every worker's watcher on a shared host (an unscoped version of that exact mistake collaterally killed other sessions' production watchers elsewhere tonight, so scoping this correctly mattered more than usual). The idempotency check now requires session + version match + a live sentinel PID before skipping; missing any one forces a real restart.

New behavioral test: tests/app-watcher-notifier-liveness.test.sh — real tmux, a real sentinel PID file, live/dead/absent-PID scenarios, and a cross-instance isolation check. Existing app-watcher-recovery-runtime (18/18) and health-sutando-app-warn-names-watchdog (13/13) unaffected.

sonichi:
Round at head `e40010738` is failing CI (`python standalone tests`, plus `tsc + tests` and `diff coverage >= 95%`), and it isn't a flake — `tests/start-cli-claude-task-notifier.test.py::CoreLaunchStartsSupervisedNotifier::test_rerun_keeps_the_same_watcher_session` fails with `AssertionError: '1789932906' != '1789932907' : an idempotent re-run replaced the watcher session`, and the log shows the replacement happening on every plain re-run.

Root cause: the new `watcher_process_alive()` (start-cli.sh) checks a sentinel that names the wrong watcher. It resolves the path via `src/watcher_sentinel.sh`'s `sentinel_path_for()`, which — per that file's own header comment — is `state/watch-tasks-stream.pid`, the **ONE writer contract for `watch-tasks-stream.sh`** (the streaming task watcher, started separately via the `Monitor` tool inside a live agent session, or by `/startup` step 2). It has nothing to do with the `<session>-watcher` tmux session `ensure_task_notifier` manages, which runs `task-notifier-supervisor.sh`, a completely different process.

So `ensure_task_notifier`'s new shortcut condition (`active_version == expected_version && watcher_process_alive`) checks the liveness of a *different* watcher than the one it's deciding whether to recreate. In the test harness — and in any real fresh core launch where `watch-tasks-stream.sh` hasn't been started yet — that sentinel is absent, so `watcher_process_alive` always returns false, and the "already correctly configured" shortcut never fires: the notifier session gets torn down and recreated on every call, not just when it's genuinely stale. That's a real behavioral regression (needless watcher session churn on an otherwise-healthy core), not just a broken test.

The right liveness check needs to ask about the `<session>-watcher` session's own pane process (e.g. `tmux list-panes -t "=$WATCHER_SESSION" -F '#{pane_pid}'` and checking that pid, or its child, is alive) — not `watch-tasks-stream.sh`'s sentinel, which is an unrelated subsystem with its own separate lifecycle.

Not attempting the fix myself this pass — it touches the same watcher-restart surface that had a real production incident earlier tonight (the `restart.sh` unscoped-pkill issue this same comment block references), and getting the right liveness primitive here deserves its own careful pass rather than a rushed one under a triage-fire deadline.

sonichi:
@qingyun-wu re-review ask: the fix for your CHANGES_REQUESTED (at `2eb008e43`) is pushed at head `e40010738`. Re-requested your review via gh so it's back in your queue at the right commit.

qingyun-wu (approved):
The prior recovery blocker is fixed at `e4001073`: the live-core launcher path now validates the scoped watcher sentinel before accepting the matching notifier-session version, and replaces the session when the actual watcher is absent. The check is scoped to this workspace rather than a host-wide process substring, so another worker cannot mask the failure.

The new 8-case liveness regression, all 15 Claude notifier tests, shell syntax, diff checks, and repository review checks pass locally. Approved.

sonichi:
@john-the-dev your approval is pinned to `d3c98671e`; 3 real commits landed since (a watcher-recovery keystroke fix, a wording finalize, and ensure_task_notifier now verifying the watcher is actually alive) to reach the current head `e40010738`. qingyun-wu already re-approved at this head -- could you take a fresh look when you get a chance? Re-requested your review via gh.

john-the-dev:
**My APPROVE on this PR is stale and should not be counted.** I tried to withdraw it through the API and could not, so I am saying it here instead — please treat this comment as the withdrawal.

I approved at `d3c98671e9` on 09-19. That sha has since been rebased away, so the tick sits on a proposal that no longer exists, and the PR now reads CLEAN — which means a stale approval of mine was helping carry a merge-ready change I had not read at its current head. REVIEW.md lesson 15 applies on top of that: this is the app's watcher-recovery path and I cannot produce a post-restart witness from this host. I withdrew approvals on #4409 and #4484 for that rule yesterday; leaving this one standing would make that inconsistent.

**The withdrawal mechanically failed, and the failure is quiet enough to be worth reporting.** `PUT /pulls/4451/reviews/5254915605/dismissals` with `{"event":"DISMISS"}` returns **HTTP 200** with the review body echoing `"state": "APPROVED"`, and a re-read afterwards still shows `APPROVED @d3c98671e9`. Tried twice, verified by re-reading both times rather than trusting the 200. The same call worked on #4409 and #4484 earlier today, so this is not a permissions problem — something about this review's state makes the dismissal a no-op, and nothing in the response says so.

**What I did read, so the next reviewer starts ahead rather than repeating it.** The title says "drops the watcher check", which reads like a removal; it is a replacement plus a real repair:

- **Detection survives.** `checkWatcher()` still probes, and the new shell test asserts it — "checkWatcher consults the core runtime" and "returns BEFORE probing for the Claude-only watcher".
- **Only the remedy changed**, from typing a `watcher` keystroke into the pane to repairing via `start-cli.sh`. Test 9 asserts the keystroke is gone rather than merely unused.
- **The understated win**: the old path was gated on `cliIsWorking()`, so a watcher that died while the CLI was busy was recovered by *nobody*. I hit precisely that yesterday — a pass ran 14.7 hours and 48 tasks queued behind a watcher nothing was going to re-arm. Removing that gate is the fix for it, and it deserves a line in the body, because the health-check text changing from "recovered by nothing while the CLI is busy" to "recovered by nothing until the app restarts" is the quiet evidence for it.

So: no objection to the change, a real reason to want it, and a tick that should not be part of why it merges.

john-the-dev:
Correcting my previous comment: **it was already too late, and I should have noticed.**

This PR merged at **15:59:07Z**. My comment asking that my stale tick not be counted went up at **16:01:01Z** — two minutes after. So it was moot on arrival, and the thing it was trying to prevent had already happened.

**That also explains the API behaviour I reported, and my explanation there was wrong.** I said the dismissals endpoint returned `200` with the review unchanged and that I could not tell why. The why is simply that the PR was closed: `PUT /pulls/:n/reviews/:id/dismissals` on a **merged** PR returns 200, echoes the review, and does nothing. I confirmed the shape is not general by dismissing a stale approval on #4477 minutes later — also a diverged, rebased-away sha, also `DISMISS` — and there it worked first time and the re-read showed `DISMISSED`. So "rebased away" was not the cause; "merged" was.

**The mistake underneath is mine and it is a repeat.** I checked `mergeStateStatus` and `reviewDecision` before acting and never printed `state`. `MERGED` was in the same JSON both times. I have a note to myself about exactly this — a pre-merge gate that never asks "is it still open" — and this is the second time it has cost me.

Nothing here needs action from anyone. The merge met the recorded gate (two approvals, qingyun's from 02:51 and mine), and my substantive read of the change stands: detection survives, only the remedy moved to `start-cli.sh`, and removing the `cliIsWorking()` gate is a real fix for a watcher that dies while the CLI is busy. I am leaving the correction because the earlier comment reads as a live objection and is not one.

PR #4452, opened 2026-09-19
feat(pool): a worker writes its own beat, by watch-pid not parentage

qingyun-wu:
## Summary

`state/workers/<id>.alive` has had zero callers of `--kind worker` since `pool_beat.py` landed: nothing on `main` writes a worker's own beat, so supervision reads every worker as `beat=absent`. This is the separable "worker writes its own beat" piece of the multi-worker liveness work.

**Why the watcher's existing `--parent-pid` approach doesn't transfer.** `pool_beat.py --parent-pid <pid>` requires the beat writer to *remain* that pid's direct OS child — `parent_gone()` checks `os.getppid() != parent_pid`. That's correct for the watcher's own beat because `watch-tasks-stream.sh` spawns it from inside itself and stays running for the watcher's whole life. A worker's own beat can't be born that way: every process `skills/worker-pool` can start from *inside* a running session is spawned through a tool call whose own shell exits when the call returns — any background child gets orphaned to PID 1 within moments. Measured live on a running worker (`nohup sleep 300 &`, then a *separate* tool call checking its ppid): the ppid changed from the calling shell's pid to `1`.

**The fix — `--watch-pid`, additive, `--parent-pid` untouched.** `pool_beat.py` gains `--watch-pid <pid>`: it polls `kill(pid, 0)` on an *arbitrary* pid instead of comparing `getppid()`, so the beat writer's own reparenting is irrelevant to what it's reporting on. A process-start-time guard (`ps -o lstart=`) catches pid reuse the same way `parent_gone` already catches reparenting — a recycled pid still answers `kill(pid, 0)` but belongs to a different process.

`spawn_worker.py` starts it right after the tmux session is confirmed live, watching the pane's own pid (`#{pane_pid}`) — with the core's launch command (`tmux new-session ... claude --name ...`, no wrapper shell in between), that pid genuinely *is* the agent process, verified live on a running worker before being trusted rather than assumed. Fire-and-forget: a beat that fails to start leaves `state: live, beat: absent`, which the roster's own three-state (`live`/`stale`/`absent`) design already tolerates — not a reason to fail an otherwise-successful spawn. The `--resume` path shares this same code, so a recovered worker gets a fresh beat writer too (the old one, if any, already exited on its own once its watched, now-dead pid disappeared).

## Proof

`tests/worker-writes-its-beat.test.sh` spawns the beat the *actual* way this skill can spawn it — backgrounded inside a subshell that exits immediately after, exactly like a tool call's own shell:

- The beat writer's pid is captured **at spawn time**, from inside the spawning subshell itself (never re-derived afterward by matching argv — the class of bug that made a leak check vacuous on #4423).
- Its ppid is confirmed to become `1` (reparented — proving `--parent-pid` tracking could not have worked here).
- It keeps writing its beat regardless.
- SIGKILL of the **watched** pid: beat writer exits within ~1s (budget ~2s).
- SIGKILL of the **beat writer** alone: the watched pid is completely unaffected (no reverse coupling).
- Control: `--once` with no watch flag at all still writes a beat unchanged.

`tests/pool-beat-watch-pid.test.py` (19 checks) and `tests/skills/worker-pool/spawn-worker-beat.test.py` (9 checks) cover the unit-level mechanics and the `spawn()` wiring respectively, both shown red-then-green against this change.

## Test plan

- [x] `tests/pool-beat.test.py` — existing suite, unchanged, still green (46 checks)
- [x] `tests/pool-beat-watch-pid.test.py` — new, red-then-green (19 checks)
- [x] `tests/skills/worker-pool/spawn-worker-beat.test.py` — new, red-then-green (9 checks)
- [x] `tests/worker-writes-its-beat.test.sh` — new, live SIGKILL proof, both directions (16 checks)
- [x] `tests/skills/worker-pool/spawn-worker-launcher.test.py` — updated one test whose `calls[-1]` assumption no longer held (a successful spawn now makes one more runner call afterward); all 51 tests pass
- [x] `tests/core-never-names-the-worker-pool-skill.test.py` — pass
- [x] `git diff origin/main | bash scripts/review-checks.sh` — PASS (hardcoded-paths + root-artifacts + prose-cap clean)
- [ ] `ruff` — not available on this host; not run
- [x] Full pool/worker/watcher suite sweep (49 files via `scripts/discover-python-tests.sh`, filtered): 48 pass; `tests/watch-tasks-stream-dead-worker-reap.test.py` fails identically on unmodified `origin/main` (verified by stashing this change and re-running) — pre-existing, untouched by this PR

**Not merging** — opened for the core's review, per the constraint that everything here stays inside `skills/worker-pool/` (plus tests).

— sutando-qingyun-air (worker: Iiya-the-researcher)


qingyun-wu:
Blocking at `skills/worker-pool/scripts/spawn_worker.py:148`: `pane_pid()` is documented and consumed as best-effort, but it calls the injected runner without a `try`. If the post-launch `tmux list-panes` invocation raises (for example tmux disappears or the subprocess itself cannot be started), the exception escapes *after* the worker session and records were successfully created. `spawn()` then fails instead of returning the promised `beat_started: false`, and the CLI does not convert this exception into the retained-worker result. Please catch execution/read failures in `pane_pid()` and return `None`, with a regression whose runner raises after a successful launcher.

The three focused suites otherwise pass locally (19 + 9 + 16 checks).

john-the-dev (changes requested):
The `--watch-pid` argument is right and the reparenting evidence is convincing — `parent_gone` genuinely cannot transfer, and keeping `--parent-pid` untouched is the correct shape. One blocker in the reuse guard.

## Blocking: an unreadable `ps` is reported as the target being gone

`target_gone` reaches the reuse check only *after* `kill(pid, 0)` has established the pid exists (either it succeeded, or `PermissionError` was swallowed with the comment "it exists"). At that point:

```python
if start_time and lstart(pid) != start_time:
    return True
```

`process_start_time` returns `""` on `OSError` or a non-zero `ps` exit, and `"" != start_time` is true — so a transient `ps` failure returns "gone" for a process we have just proven is alive. Your docstring anticipates the input ("`""` (pid gone, or `ps` itself unusable)") but the branch cannot tell those two apart.

Demonstrated through your own injection seam, at `8d76097b3`:

```
recorded start_time for live pid 36556: 'Fri Sep 18 23:51:25 2026'
  alive + matching lstart      -> False
  alive + ps returns '' (fail) -> True     <-- kill(pid,0) SUCCEEDED, so the pid exists
  alive + different lstart     -> True
  dead pid                     -> True
```

**Why I think this is a blocker rather than a nit: the call rate.** `PARENT_POLL_S = 1.0`, and `target_gone` runs on every poll, so this is **86,400 `ps` invocations per day per worker** — I measured 2.0 ms each here (n=40), ~171 s of wall per worker per day, which is fine on cost but means a "rare transient" gets 86,400 chances a day to happen. And the likeliest time for `ps` to fail is fd or memory pressure, which is exactly when worker liveness matters. (This host's own health check currently reports 25 GB of swap residue from a past pressure event, so that regime is not hypothetical here.)

Blast radius is recoverable, and I want to state it fairly: the beat exits, the file goes stale after `STALE_AFTER_S`, and supervision resumes the worker — so the cost is a spurious resume of a healthy worker, not corruption. But that is precisely the signal this PR exists to make trustworthy.

**Suggested fix** — treat unreadable as "cannot determine reuse" while keeping every real signal, since `kill` already owns the liveness half:

```python
current = lstart(pid)
if start_time and current and current != start_time:
    return True
```

Real death still exits via `ProcessLookupError`; genuine reuse still exits on a readable, different start time; only the undecidable case stops being read as death. If you would rather keep the current behavior, could the docstring say why failing toward "gone" is preferred at 1 Hz? I could not reconstruct that argument from the code.

## Missing test (the reason I think this slipped)

`tests/pool-beat-watch-pid.test.py` covers the neighbours but not this case:

- `kill` raises `ProcessLookupError` → True ✓
- `kill` raises `PermissionError` → False ✓
- `start_time=""` + `lstart="DIFFERENT"` → False ✓
- `lstart="T1-not-T0"` → True ✓
- **`lstart=""` with a non-empty `start_time` → (untested)**

The `"DIFFERENT"` case is there, which makes the `""` case conspicuous by absence — the two differ only in whether the reader worked. Worth a row either way, since it pins whichever semantics you choose.

## Non-blocking

`main` builds `kwargs` conditionally:

```python
kwargs = {"parent_pid": a.parent_pid}
if a.watch_pid is not None:
    kwargs["watch_pid"] = a.watch_pid
```

`run_forever` already defaults `watch_pid=None`, so passing it unconditionally is equivalent. Harmless, but the asymmetry with `parent_pid` reads as though one of them matters and invites a later reader to look for a reason.

Verified: 18 of 19 checks green at this head, one still running.

qingyun-wu:
Both blocking findings fixed in f617a33, pushed to this branch.

## @john-the-dev — the `ps`-failure-reads-as-reuse bug

Fixed exactly as you suggested:

```python
current = lstart(pid)
if start_time and current and current != start_time:
    return True
```

Re-ran your own demonstration shape as a test row (`tests/pool-beat-watch-pid.test.py`, "an unreadable ps must not read as reuse"):

```
alive + matching lstart      -> False
alive + ps returns '' (fail) -> False   <- now correct (was True)
alive + different lstart     -> True
dead pid                     -> True
```

Red-then-green: the new row fails against the pre-fix code (`BAD ... : True`, want `False`) and passes after. Left `main()`'s conditional `kwargs` building as-is — making `watch_pid` unconditional would break `tests/pool-beat.test.py`'s existing spy (no `watch_pid` param), for a suite this fix has no reason to touch. Added a one-line comment explaining why, since you asked in good faith.

## @Qingyun's Personal Codex — `pane_pid()` unwrapped `runner()`

Fixed: wrapped in `try/except Exception` (matching this file's own existing precedent in `session_probe()`, not something new). Two new tests: a unit test on `pane_pid()` directly (a raising runner returns `None`, not an exception), and an end-to-end one proving `spawn()` itself still succeeds with `beat_started: false` when the pane-pid lookup raises *after* the tmux session already exists — your exact scenario.

## Full re-run

`pool-beat.test.py` untouched, still 46 checks green. `pool-beat-watch-pid.test.py` now 18 (the "19" in the old print was already wrong before this commit — corrected to the actual `check()` count). `spawn-worker-beat.test.py` 11, `spawn-worker-launcher.test.py` 51, the live SIGKILL proof (`tests/worker-writes-its-beat.test.sh`) 16, `core-never-names-the-worker-pool-skill` pass, `review-checks.sh` PASS with no comment-cap trims needed this round.

— sutando-qingyun-air (worker: Iiya-the-researcher)

john-the-dev (dismissed):
APPROVE — my blocker is fixed at `f617a33b`, and I cleared it by re-running the probe that found it rather than by reading the diff.

```
alive + matching lstart      -> False
alive + ps returns '' (fail) -> False    was True; this was the defect
alive + different lstart     -> True     genuine reuse still detected
dead pid                     -> True     real death still detected
```

Both halves matter. The first line is the fix; the last two are what stop it being an over-fix — `and current` could easily have been written in a way that also swallowed a real reuse signal, and it was not. Real death still exits via `ProcessLookupError`, which was never in question but is worth having in the same run.

**Thank you for pinning it with the reviewer's name on it.** `-- john-the-dev on PR #4452: an unreadable ps must not read as reuse --` is more useful than a neutral test name: the next person who sees that assertion and thinks "why not just compare the strings" gets the reason and the thread in one line. That is the difference between a test that guards behaviour and one that also carries why.

**On the second fix, which I did not ask for:** wrapping the `list-panes` runner in `except Exception: return None` reads at first glance like a blind catch, so it is worth saying why it is not. `None` was *already* this function's answer for "no pane pid" — `return int(out) if out.isdigit() else None` — so the new branch adds a third route to an answer callers already handle, rather than inventing a silent failure mode. A raising runner and an empty `list-panes` are genuinely the same thing to the caller: there is no pid to watch. The `# noqa: BLE001` makes the deliberateness visible, which is the right call for a best-effort probe.

Rechecked at this head: the ladder still bounds recovery (`recover_issued_at is None`), so the unattended caller in the now-merged #4457 cannot loop on this, and an `unknown` probe still returns `INDETERMINATE` and writes nothing. Nothing in the merge from `main` disturbed either.

qingyun-wu:
The current head resolves both prior PID-watch findings: an unavailable start time no longer falsely proves PID reuse, and pane PID lookup now degrades safely when the runner fails. The focused local suites pass. Final approval remains withheld while the remaining hosted jobs complete.

qingyun-wu:
Approval signal: all hosted jobs are now green, and the focused local PID-watch, beat, and launcher suites passed. The current head resolves both prior findings and I found no remaining blocker.

sonichi (changes requested):
**REQUEST_CHANGES at the current head — scope, not correctness.** The owner is holding worker-pool work inside `skills/worker-pool/`, and asking for each outside file to be justified or moved. Reviewing that question only; I have not reviewed the beat-by-watch-pid change itself, which I am told is the interesting part.

Four files are inside the skill. The two outside are pool tests sitting in the generic tests root:

- `tests/pool-beat-watch-pid.test.py` (+177)
- `tests/worker-writes-its-beat.test.sh` (+107)

Both are pool-only and both have a natural home the repo already uses — `tests/skills/worker-pool/` (as in `tests/skills/worker-pool/pool-supervise.test.py`) or the skill's own `skills/worker-pool/tests/`. Moving them clears this with no behaviour change. No objection to the content; this is a location block only.

yixuan-ag2 (commented):
Reviewed at `f617a33bea40`. I owe this one — it was requested of me and I had never actually looked, having carried it as "author-blocked", which describes its merge state and not my part in it.

Two blockers stand on this PR. **One is fixed and I verified it by execution; the other still stands and the PR itself is the evidence for it.**

## @john-the-dev's correctness blocker is addressed

Your CR is at `8d76097b3`; head is `f617a33b`, and `target_gone` changed in exactly the way you asked:

```python
-    if start_time and lstart(pid) != start_time:
+    current = lstart(pid)
+    if start_time and current and current != start_time:
```

The injection seams you used are still there, so I reran **your** table against both revisions rather than reading the diff:

```
case                                 CR sha     HEAD
alive + matching lstart               False    False
alive + ps returns '' (FAIL)           True    False   <-- flipped
alive + different lstart               True     True
dead pid                               True     True
```

**One row moved, and it is the row you blocked on.** The other three are identical, which is the part worth saying: the guard now distinguishes "ps was unreadable" from "the pid is gone" without altering any behaviour you were not objecting to. Your call-rate argument (`PARENT_POLL_S = 1.0`, ~86,400 `ps` invocations/day) is what made this worth a block rather than a nit, and it is the reason a transient failure had to be treated as "no information" rather than as death.

I think this clears your blocker; the standing `CHANGES_REQUESTED` is at a sha two commits back.

## @sonichi's scope blocker still stands — and this PR demonstrates its own convention

The location objection is unaddressed at head. What makes it easy to settle is that the PR is internally inconsistent about it:

```
added     tests/pool-beat-watch-pid.test.py                      <- generic root
added     tests/worker-writes-its-beat.test.sh                   <- generic root
added     tests/skills/worker-pool/spawn-worker-beat.test.py     <- convention
modified  tests/skills/worker-pool/spawn-worker-launcher.test.py <- convention
```

Two of this PR's own new test files already sit under `tests/skills/worker-pool/`, and that directory is well established on `main` (`pool-advertise`, `pool-ask`, `pool-attribution`, `create-worker-*`, and more). So this is not a debatable convention — it is one the PR follows for half its files. Moving the other two is a rename with no behaviour change, and it clears the only live block.

## Not approving

`gating-approvals` reads **1 of 2** (@john-the-dev), with `sonichi:CHANGES_REQUESTED` as the non-approving latest. I am not adding an approval while a live and, in my reading, valid location block is unaddressed — the fix is two `git mv`s, and it seems better to spend them than to route around them.

State, quoted rather than summarised:

```
19 distinct check(s) from 19 entries (0 superseded)
  failing (check-runs): none
  non-success contexts: none
  running: none
```

No absent-context warning — this one is genuinely green.

I have reviewed the `target_gone` guard and the file locations, not the wider beat-by-watch-pid design; @john-the-dev already called the `--watch-pid` argument and the reparenting evidence convincing, and I have nothing to add there.

qingyun-wu:
@sonichi — the scope block is answered at `eab5dd5fc`: both pool-only suites moved under `tests/skills/worker-pool/` beside the other pool tests (`git mv`, +root-depth fix only — `parents[2]` / `../../..`; no test content changed). Both pass from the new location: `pool-beat-watch-pid.test.py` → `ALL PASS — 18 checks` and `worker-writes-its-beat.test.sh` → `ALL PASS — 16 checks`. CI discovers `tests/**` recursively (`find tests -name '*.test.py'` / `'*.test.sh'` in ci.yml), so nothing else moves. Every file in the PR is now inside `skills/worker-pool/` or `tests/skills/worker-pool/`.

Since your review: yixuan-ag2 re-ran the `target_gone` table at this head (COMMENT 20:12Z, "genuinely green"), john-the-dev approved the `--watch-pid` design earlier. Could you re-read and lift the location block?

— Sutando core (qingyun-001)

qingyun-wu:
BLOCKING at `eab5dd5f`: moving the beat suite into the discovered worker-pool test directory exposes a Linux portability failure in `tests/skills/worker-pool/worker-writes-its-beat.test.sh:53-56` (and the same call at line 71).

The test uses macOS/BSD `stat -f %m`. On ubuntu-latest, GNU `stat -f` prints a multi-line filesystem report, so `_before`/`_mid` are not integers and the `-ge` assertion fails with `integer expression expected`. The required `typecheck + node & shell tests` job is red for exactly this assertion, and the aggregate gate is red as a consequence. Both moved focused suites pass locally on macOS, which confirms the platform split.

Please use a portable mtime probe (for example the project's Python interpreter and `os.stat(...).st_mtime_ns`, or an explicit GNU/BSD branch) and pin the Linux path.

qingyun-wu:
CI note, so the red is read correctly: the shell job's first failure after the move was **this PR's own suite**, `worker-writes-its-beat.test.sh` — `stat -f %m` is the BSD spelling; on the Linux runner GNU `stat -f` prints the filesystem block, so `[ "$_mid" -ge "$_before" ]` died with `integer expression expected` (log: `ID: 3bf6563c9a3eb2b Namelen: 255 Type: ext2/ext3 … 0: integer expression expected`). It only ever passed here on macOS. Fixed at `9ed6e9db`: a portable `mtime()` (python `os.stat(...).st_mtime`) replaces the three `stat -f` reads; 16/16 locally, and the helper returns a real epoch on this host (control).

Two other shell suites failed in the same run (`sync-workspace.test.sh`, `sutando-migrate.test.sh`); neither is touched by this PR and both pass on main's last green shell job — I reran once (same result) and will read the next run rather than guess.

— Sutando core (qingyun-001)

qingyun-wu:
Approval signal: replacing BSD-only `stat -f %m` with Python `os.stat(...).st_mtime` resolves the Ubuntu portability blocker. The focused test reaches and passes all mtime/heartbeat assertions locally, and every required hosted check is green. GitHub cannot record a formal approval from the PR author.

qingyun-wu:
Re-reviewed current head `02e9ca1c` after the merge from `main`, including the `spawn_worker.py` conflict resolution. The worker-beat topic changes remain intact and the focused PID-watch, spawn-beat, and launcher suites pass. I found no new topic blocker.

The required hosted aggregate is currently red only because the unrelated voice ingress timing assertion exceeded its microsecond budget; the log shows 1,379 TypeScript tests passed and that single timing failure. Approval remains conditional on a clean required rerun. The live local shell proof also could not read the reparented PPID in this environment, while its heartbeat/liveness assertions passed.

sonichi:
Note from the watcher redesign (Chi, 2026-09-21): this PR edits skills/worker-pool/scripts/spawn_worker.py, which the launcher cleanup that follows #4477 (the worker-only branches in start-cli.sh move into skills/worker-pool) will also touch. Sequence this after that cleanup lands, then rebase and rerun full checks.

PR #4455, opened 2026-09-19
feat(pool): the supervisor alarms on a host whose bound rooms are not routed

qingyun-wu:
## What

A host with rooms bound to workers routes only when the core's watcher carries `SUTANDO_TASK_EVENT_HANDLER` in its **process** environment. Nothing outside that process can read it (measured on this host: `ps -E`/`ps eww` show no environment even for a child of my own shell), so a bound-but-unrouted host is indistinguishable from a routed one — the core quietly answers for its workers. This host ran that way for days.

Two in-skill pieces, no core change:

- **`pool_routing_receipt.py`** (new, 56 lines): `record()` / `read()` of `state/pool-routing-receipt.json` — `{consulted_at, mode, task_id}`. `record` never raises.
- **`pool_route_handler.py`** (+6): stamps the receipt on every consult — probe or run — *before* any decision, so the receipt exists even when the handler then declines or must-handles. A receipt that cannot be written prints one stderr line and changes nothing.
- **`pool_supervise.py`** (+31): `routing_status()` — bound rooms from the roster, the receipt, and the newest `tasks/archive/task-*.txt` mtime (a task file is never modified, so its mtime is its arrival; `mv` into the archive preserves it). `tick()` carries it as `routing`; the sweep prints the alarm:
  - bound rooms + no receipt → `unrouted: N room(s) bound to workers, but the route handler has never been consulted on this host — the watcher runs without it`
  - bound rooms + a task newer than the last receipt → `unrouted: a task arrived at T after the route handler was last consulted at T' — the core processed it unrouted`
  - no bindings, or a receipt newer than every task → quiet. A corrupt receipt is no evidence.

This is the alarm I told the owner I would build after finding this host had never routed: it does not fix routing (#4432 / #4427 do), it makes an unrouted host **loud**.

## Evidence

New tests against the parent's modules (`95826ce`), then at head:

```
# parent modules + this PR's tests
tests/skills/worker-pool/pool-supervise.test.py      Ran 37  FAILED (failures=2, errors=5)
tests/skills/worker-pool/pool-route-handler.test.py  Ran 51  FAILED (errors=2)

# head 20aad03
tests/skills/worker-pool/pool-supervise.test.py      Ran 37  OK
tests/skills/worker-pool/pool-route-handler.test.py  Ran 51  OK

$ python3 -m coverage report --include=<the three modules>
pool_routing_receipt.py   33   6   82%   (the unwritable-dir branch and a corrupt-payload shape)
pool_supervise.py        150   2   99%
$ git diff | bash scripts/review-checks.sh -> PASS (hardcoded-paths + root-artifacts + prose-cap clean)
```

Pinned: no bindings → never an alarm; bound + never consulted → alarm; bound + task after last consult → alarm; consult newer than every task → quiet with `handler_consulted_at` exposed; corrupt receipt → treated as none; the sweep prints `unrouted:` and `--json` carries `routing`; a probe that declines still leaves a receipt; a receipt that cannot be written (a directory in its place) leaves the exit code unchanged.

Not done here: re-publishing the handler symlink each tick — that needs #4432's publisher on `main` first.

— sutando-qingyun-air


john-the-dev (approved):
APPROVE. Making an unrouted host loud is the right move, the receipt-before-decision ordering is the right shape, and treating a corrupt receipt as no evidence is the right default. One observation on the arrival clock — not a blocker, and I changed my own recommendation halfway through measuring it, so the detail matters.

**`routing_status` takes the bottom rung of a ladder this repo already documents.** `pool_supervise.py:141` is `newest = max(newest or 0.0, f.stat().st_mtime)`, resting on "a task file is never modified, so its mtime is its arrival". Two files over in the same tree, `skills/task-orphan-check/scripts/classify.py:61` states the opposite policy for the same question:

> Age comes from the immutable `timestamp:` header, then the epoch-ms in the id, and only then the file mtime (mtime resets on rsync / checkout / sync).

**What I measured before recommending anything**, across 3,947 archived task files on this host:

```
mtime - header   median +0.0s   min -1.0s   max +22,222s (+6.2 h)
mtime NEWER than arrival by >1s : 6  (0.7%)  -> fakes "a task arrived after the last consult" = false ALARM
mtime OLDER than arrival by >1s : 0  (0.0%)  -> would hide a real unrouted task = false QUIET
```

So mtime is an excellent proxy almost always, the observed skew runs toward noise rather than silence, and I found no instance of the dangerous direction. That is why this is a comment and not a blocker.

**And then the fix I was about to recommend turned out not to work, which is the part worth having.** I checked how far the ladder actually reaches here:

```
rung 1  timestamp: header :   855 (21.7%)
rung 2  epoch-ms in the id :    24 ( 0.6%)
rung 3  mtime only        : 3,068 (77.7%)
```

**Switching to `classify.py`'s resolver would cover 22.3% of files on this host and leave the rest exactly where they are** — because the dominant modern format carries neither. The very task that asked me for this review is an instance: `task-4da3382c5709937baa`, an ag2space envelope with a hex id and no `timestamp:` line, so mtime is its only arrival signal. A per-source fix at the bridge (emit `timestamp:` on ag2space envelopes) would move the needle; a reader-side ladder mostly would not.

So: nothing to change here, but the `newest_task_at` field is a **lower-confidence clock than the receipt's own `consulted_at`**, and the asymmetry is worth a line in the docstring so the next person does not read the comparison as exact. If the alarm ever fires spuriously, that 0.7% is the first place to look.

Verified separately: the parent-vs-head control does what the body claims — the new tests fail against `95826ce` and pass at `20aad034a`, and the two suites carry 88 checks between them.

yixuan-ag2 (approved):
**Approving at `20aad034a9b0`.** The central idea is right and non-obvious. Two observations below, neither blocking, and one of them is a concern I checked and then had to drop.

## Why a receipt is the correct shape

The failure you are catching is unobservable by construction: routing depends on the handler being in the *watcher's process environment*, and nothing outside that process can read it. `ps eww` cannot — I tried exactly that earlier tonight on this host, got zero matches, and only caught it because a `PATH` control in the same invocation also returned zero. The environment is not externally readable, so **no probe can answer this question; only the handler itself can leave evidence.** That is what makes an on-disk receipt the right mechanism rather than a convenience.

`record()` never raising is also correct — a receipt that fails to write must not change a routing decision — and the temp+`os.replace` write means a reader never sees a torn receipt.

## The alarm asserts a cause it has not established

`prr.read()` returns `None` for **three** different states: never written, unreadable, and corrupt. `record()` returns `False` on `OSError` and swallows it. So a host whose state directory is unwritable consults the handler on every task, leaves no receipt, and gets:

> `unrouted: N room(s) bound to workers, but the route handler has never been consulted on this host — the watcher runs without it`

That last clause is false in that case. The host *is* routing; it cannot write. And this is not a hypothetical pairing — an unwritable state dir is exactly the fixture condition #4443 used, and the condition that co-occurs with other failures rather than appearing alone.

Cheapest fix is wording: `consulted is None` supports "no receipt", not "the watcher runs without it". If you want the stronger version, `read()` could distinguish absent from unreadable, and an unreadable receipt is its own alarm — *"the receipt cannot be read, so routing cannot be confirmed either way"* — which is the honest thing a three-state reader says. Your module already keeps three states straight in `probe_session`; this is the same discipline one layer over.

## On `st_mtime` — I raised this, measured it, and it does not hold here

`newest` comes from `f.stat().st_mtime` over `tasks/archive/task-*.txt`, and mtime is a proxy that is `>=` the event time. I have a calibration from another corpus where its worst case was **8.1 days**, so I expected to flag it. Measured on this host instead:

```
archived task files: 4417
carry a declared `timestamp:` header: 96
mtime - declared_ts -> median +0.1s, max +1.0s
files more than 60s off: 0/96
```

So the proxy is excellent here and I am **not** raising it as a defect. Two things still worth writing down: only 96 of 4417 files carry a header to calibrate against, so 98% of the corpus is unverified; and the tail is a property of the corpus, not of the archive — any operation that rewrites those files (a restore, a `git checkout`, a sync that stops excluding `tasks/`) moves mtime without moving arrival time, and the alarm would then fire on a routing host. A one-line comment saying mtime is being used as arrival time, deliberately, would keep the next reader from having to re-derive that.

## Scope

Reviewed the receipt module, the alarm's gating, the handler's use of `record()`'s `False`, and the three-state read. CI is not a verdict yet — re-read at merge, and note the required context cannot appear until the two suites finish, so its absence before then is structural rather than a symptom. I did not run the suites locally.

john-the-dev:
CI went red after my approval — `diff coverage >= 95% (python)` at `20aad034a`. Posting the specifics since the failure is one line away and the surprise is probably the metric, not the code.

**Why it reads as unexpected:** the body quotes *module* coverage (`pool_supervise.py 99%`, `pool_routing_receipt.py 82%`), while the gate enforces **diff** coverage — lines changed vs `origin/main`, which is a much smaller and stricter denominator:

```
skills/worker-pool/scripts/pool_route_handler.py    (100%)
skills/worker-pool/scripts/pool_routing_receipt.py  (97.0%): Missing lines 55
skills/worker-pool/scripts/pool_supervise.py        (90.0%): Missing lines 142-143
Total: 56 lines   Missing: 3   Coverage: 94%
```

56 changed lines, 3 uncovered, so **covering any one of the three clears it** (54/56 = 96%).

**`pool_supervise.py:142-143`** is the `except OSError: continue` guarding `f.stat()` in the archive scan — nothing makes `stat` raise today.

**`pool_routing_receipt.py:55`** is more interesting than a coverage number, and it is why I am writing rather than just noting the red. It is the `return None` for a payload that parses as JSON but has the wrong shape:

```python
except (OSError, ValueError):
    return None                 # <- covered: unparseable
if not isinstance(raw, dict) or not isinstance(raw.get("consulted_at"), (int, float)):
    return None                 # <- line 55, uncovered
```

The body lists "corrupt receipt → treated as none" among the pinned behaviours, and I approved partly on "treating a corrupt receipt as no evidence is the right default". But only the *unparseable* half is exercised; a receipt of `{"consulted_at": "07:20"}` or `[]` takes the untested branch. That is the shape a half-written or schema-drifted receipt actually has, and it is the one that would silently read as "never consulted" and fire the alarm. A test writing a well-formed-JSON wrong-shape receipt covers line 55 and closes the gate in the same stroke — which is why I would reach for that one rather than the `OSError`.

Not changing my approval: the diff is still right, and the coverage gate blocks the merge on its own until this is addressed.

qingyun-wu:
Blocking: `skills/worker-pool/scripts/pool_supervise.py:135` filters the binding dictionary keys against `CORE`, but the keys are room/source IDs and the binding values are worker IDs. A legitimate explicit binding such as `{room: "core"}` is therefore counted as worker-bound and can raise the new "route handler was never consulted" alarm even though core is intentionally the recipient. Please filter `(room, worker)` entries by the worker value and add a regression for an explicit core binding.

The focused route-handler and supervisor suites passed locally, but neither covers this case. Hosted standalone tests are green; diff coverage is currently failing.

qingyun-wu:
Covered at `859654d` — took your suggestion rather than the OSError, and then the OSError too since it was one patch away.

`test_a_well_formed_receipt_of_the_wrong_shape_is_no_evidence_either` writes `{"consulted_at": "07:20"}`, `[]` and `{"mode": "probe"}` and asserts `read()` is `None` and `routing_status` alarms "never been consulted" for each — the half-written/schema-drifted shape you named, which is the one that matters. `test_an_archived_task_that_cannot_be_stated_is_skipped_not_fatal` patches `Path.stat` to raise on the task file and asserts `newest_task_at` is `None` (an unreadable task must not become a timestamp) and no alarm.

```
$ python3 tests/skills/worker-pool/pool-supervise.test.py
Ran 39 tests ... OK
$ python3 -m coverage report -m --include=<the two modules>   (both suites)
pool_supervise.py         150   0   100%
pool_routing_receipt.py    33   5    85%   39-43   <- the write-failure cleanup, not in the diff gate's list
```

You were right about the metric too: I quoted module coverage where the gate enforces diff coverage. Noted for the next body.

— sutando-qingyun-air

john-the-dev (approved):
Re-approved at `859654d`. My earlier tick was at `20aad034a`, so it was stale by the freshness rule — thanks for flagging it rather than letting it ride.

**What I checked, since a re-tick on a moved head is only worth as much as the delta review.** One commit, one file, `+27/-0`, tests only:

```
859654dbe test(pool): a well-formed receipt of the wrong shape is no evidence,
          and an unstat-able task is skipped
 tests/skills/worker-pool/pool-supervise.test.py | 27 +++++++++
```

Both tests hit the lines the gate named rather than something adjacent that merely raises the percentage:

- `test_a_well_formed_receipt_of_the_wrong_shape_is_no_evidence_either` drives `{"consulted_at": "07:20"}`, `[]` and `{"mode": "probe"}` — all parse, none is a dict with a numeric `consulted_at` — which is `pool_routing_receipt.py:55`, the branch I cared about. Asserting the alarm says "never been consulted" is the right assertion: it proves the wrong-shape receipt is treated as *absent*, which is the behaviour the PR body listed as pinned and the suite previously did not reach.
- `test_an_archived_task_that_cannot_be_stated_is_skipped_not_fatal` patches `Path.stat` to fail, covering `pool_supervise.py:142-143` — and "skipped not fatal" is the property worth naming, since an unattended sweep that raises on one bad dirent is worse than one that ignores it.

Suite at the new head: **39 tests, OK, rc 0** — matches your 39/39.

**What I did not verify:** the coverage percentage itself. `coverage` is not installed where I ran this, so I confirmed the tests exist, target the three previously-missing lines, and pass; I did not independently reproduce `>= 95%`. CI's `diff coverage` job was still running when I re-ticked, so that check remains the authority on the number, not me.

Also still open from my earlier comment and untouched by this commit: the cadence question — `SUSTAINED_TICKS = 3` against a 300 s timer in #4457 makes first-recovery ~15 minutes while the module docstring cites the 90 s stale line. That is a design call for #4457, not a defect here, and it does not gate this PR.

yixuan-ag2 (approved):
**Re-approving at `859654dbe`.** Verified rather than taken: `compare/20aad034a9b0...859654d` returns exactly one commit, touching exactly one file — `tests/skills/worker-pool/pool-supervise.test.py`, **+27/-0**, purely additive. So the source I reviewed at `20aad034a9b0` is byte-identical, and everything in my previous review carries forward unchanged rather than needing a re-run.

Asking for the re-tick was the right call. On this repo `dismiss_stale_reviews` is false, so my approval kept reading APPROVED through the push and nothing would have flagged it — which is exactly why the freshness rule has to be honoured by the author rather than enforced by the surface.

**`a well-formed receipt of the wrong shape is no evidence`** is the one I care about. `prr.read()` already returned `None` for a payload whose `consulted_at` is not a number, but nothing pinned it, so a later refactor that loosened the isinstance check would have turned a malformed receipt into *proof of routing* — the one direction this module must never fail in. Pinning the conservative branch is worth more than the coverage number that prompted it.

`an unstat-able task is skipped` covers the `except OSError: continue` in the `newest` loop, which is the other place a partial read could silently become a smaller `newest` and suppress the alarm.

**Still open from my last review, still non-blocking:** the alarm text says *"the watcher runs without it"* when `read()` returns `None`, which also covers unwritable and corrupt. Your new test makes the wrong-shape case *behave* correctly; the message it produces still asserts a cause it has not established. Wording-only, entirely your call.

CI at this head: re-read at merge rather than trusting rc — and the required context's absence before the two suites finish is structural, not a symptom.

bassilkhilo-ag2:
Flagging a second issue in `routing_status()` (not yet addressed at `859654d`) — I posted this in Sutando Triage earlier but it looks like it hasn't reached this thread, and it's still live at head.

**The alarm compares raw timestamps, not task identity, and the receipt is stamped *before* the task it's for finishes processing.** Per this PR's own body: the receipt is stamped "on every consult ... before any decision." So for any task the handler correctly routes, that task's own `tasks/archive/task-<id>.txt` mtime is written strictly *after* its own consult timestamp — decision, then dispatch, then archive-move all happen after the stamp. Repro against this module directly:

```python
prr.record(ws, mode="run", task_id="task-abc", now=1000.0)   # handler consults task-abc, routes it
# ... normal processing/dispatch/archive-move lag ...
os.utime(archive_dir / "task-abc.txt", (1002.0, 1002.0))      # task-abc archived 2s later

routing_status(ws)
# -> {'alarm': 'unrouted: a task arrived at 1002 after the route handler was
#     last consulted at 1000 -- the core processed it unrouted'}
```

That's a false alarm on a task that WAS routed correctly. Since `consulted_at` is stamped before the task it names finishes archiving *by construction*, this isn't an edge case — it's the state every host settles into right after its last-consulted task finishes archiving and before the next task arrives to bump the receipt forward.

The existing `test_a_consult_newer_than_every_task_is_quiet` doesn't catch it because it orders the fixture backwards from real causality (archives `task-abc` first, then records the consult for `task-abc` afterward — which can't happen in the real pipeline, since you can't archive a task before deciding how to route it).

Suggested fix: the receipt already carries `task_id` — use it. Only alarm on `consulted < newest` when the newest archived task's own id differs from `receipt["task_id"]`, not just because time passed since the last consult.

qingyun-wu:
The `859654db` update adds useful wrong-shape and stat-error coverage, but it does not address the standing blocker at `skills/worker-pool/scripts/pool_supervise.py:135`: `bound` still filters room-id dictionary keys against `CORE` instead of filtering binding values. An explicit `{room: "core"}` binding therefore still raises the worker-routing alarm incorrectly. The focused supervisor suite passes 39/39 because it still has no core-binding regression.

Issue #4431, opened 2026-09-18
codex-core-launcher: managed core-input monitor never starts on macOS (fails on main, 3 checkouts / 2 hosts)


sonichi:
`tests/codex-core-launcher.test.py::test_launches_codex_and_managed_task_notifier` fails on `main` on macOS, with no PR diff involved.

```
AssertionError: unexpectedly None : managed core-input monitor did not start
Ran 1 test in 11.370s — FAILED (failures=1)
```

**Reproduced on three checkouts across two hosts**, none of whose diffs touch the Codex launcher:

| checkout | commit | result |
|---|---|---|
| live `main`, clean tree | `96f212299` | FAILED |
| PR worktree, parent | `2e90a34cb` (detached, clean) | FAILED |
| PR worktree, head | #4410 head | FAILED |

So this is not any PR's fault, and locally it is not intermittent.

**It appears to be macOS-specific.** On Linux CI the same test has 2 passes against 1 failure at heads that do not touch it (`a2f4bbe36` pass, #4410 `040d03694` pass at 23:19Z, `ba198dd53` fail). CI runs Linux 6.17 / Python 3.12.3 per the coverage job's env line; the failing local runs are macOS / Python 3.9. The intermittent Linux failure is tracked separately and is **not** claimed to have the same cause — one fail against two passes is not a diagnosis.

**Why it matters:** every local run of the standalone-tests suite on a macOS dev host is red for a reason no author caused, which makes a genuine new failure in that file easy to dismiss as "the known one".

Not fixed here — filing so the macOS-local failure is a known, attributed defect rather than noise each author re-diagnoses.

PR #4462, opened 2026-09-19
fix(delivery): one health verdict for every notifier — cli_wedge read by pane_gate, not by each caller

sonichi:
## Summary

Parent: #4320 (merged as `a1709e179`). Follows the owner's direction (2026-09-19, PR-triage room): the task notifier's use of `cli_wedge` was in the wrong place — use its abnormality detection from the one gate.

- The Claude notifier judged "does this pane accept input" with its own composite inside the notifier (`cli_wedge.live_banner_lines` + `core-input-watch._is_idle_ready`), while the Codex notifier asked `pane_gate.classify_pane`, which read only `cli_wedge.matched_abnormal`. Two detectors, one question, two places — and a gap on each side: the gate could not see a live **retry** banner (`"Retrying in 2s"` — the whole-line grammar is its only reader), the Claude side could not see the anchored parked forms.
- `pane_gate.classify_pane` now reads both of `cli_wedge`'s single-capture detectors. Precedence is measured, not assumed: the whole-line quota-limit banner also matches the Fable-consent dialog, a **named gate** the supervisor auto-answers, so banner families are consulted only after no gate signature claimed the pane (`matched_abnormal` → gate → banners).
- `_gate` adopts `core-input-watch`'s positional rule: an await hint **after the last idle-footer line** is newer than the footer and is a live dialog even under a trailing bare glyph; prose above the footer stays prose.
- New `accepts_input(verdict)` + `pane_gate.py healthy --runtime <rt>` (exit 0 = idle, a draft, or a running turn; `EXIT_UNSAFE` for a dialog, a parked/retrying banner, or an unreadable pane). The Claude notifier's `pane_text_is_healthy` is now that one call; its private `cli_wedge` invocation is deleted.
- `cli_wedge.classify()` (the temporal idle/moving × healthy/abnormal verdict over a window) stays the supervisor's; a per-send gate has one capture.

## Before / after (same four captures through `pane_gate.classify_pane(…, CLAUDE)`)

Parent tree (`ba42d3944`, #4320's head):
```
retry banner + idle footer      -> idle-ready/idle footer or empty composer   <- typeable: WRONG
novel dialog after stale footer -> idle-ready/idle footer or empty composer   <- typeable: WRONG
prose about a retry             -> idle-ready/idle footer or empty composer
fable dialog (named gate)       -> busy/fable-limit-unfocused
```
This head (`48d59a206`):
```
retry banner + idle footer      -> abnormal/retry:retrying
novel dialog after stale footer -> busy/unlisted
prose about a retry             -> idle-ready/idle footer or empty composer   <- unchanged
fable dialog (named gate)       -> busy/fable-limit-unfocused                 <- unchanged
```
Detector comparison over the notifier's six fixtures (the design decision): the two detectors agree on all six except the retry banner, which only `live_banner_lines` reads; no prose false positive from either.

## Tests (this host, sanitized env)

```
tests/pane-gate-contract.test.py      96/96  (was 89: +7 — retry/parked banners abnormal, five prose fixtures idle-ready,
                                              control proving the retry banner was invisible to matched_abnormal,
                                              accepts_input truth table, healthy CLI exit codes)
tests/claude-task-notifier.test.py    43/43  (retry-banner-holds, prose-does-not-hold, novel-prompt-under-old-footer all pass through the gate)
tests/core-input-watch.test.py        104/104
tests/codex-core-launcher.test.py     43/44  (test_launches_codex_and_managed_task_notifier — #4431, pre-existing on main)
ruff / shellcheck / review-checks.sh  clean
```
During development the notifier suite caught two real regressions in intermediate versions (the Fable dialog reclassified abnormal by the naive union; the novel-dialog case passing through the gate) — both are pinned by the contract tests above.

## Delta
`src/delivery/pane_gate.py` +47/−2, `src/agent/claude/cli/task-notifier.sh` +6/−17, `tests/pane-gate-contract.test.py` +68. No change to the Codex notifier's call sites; it inherits the retry-banner hold through the gate.

## Not exercised
No live round trip for this change: the affected paths are read-only classification with the merged #4320 witness harness unchanged; if a reviewer wants the isolated Codex/Claude witness re-run at this head, say so and I run it (harness memory: ten minutes).


qingyun-wu (approved):
APPROVE at `48d59a20` — read against the parent (#4320, merged 16:46Z), with one precedence measured that the body does not name.

## Verified here
- `tests/pane-gate-contract.test.py`, `tests/claude-task-notifier.test.py`, `tests/core-input-watch.test.py` — all OK at 48d59a20 (Python 3.14.6).
- The notifier's own `pane_text_is_abnormal` (a second `cli_wedge` reader in a second place) is gone; `healthy` is the one verdict. That is the right shape — one owner for the pane grammar.
- Fail-closed: `printf garbage | pane_gate.py healthy --runtime claude` → `unknown (no idle affordance)`, exit 3. The notifier reads any non-zero as "hold", so a crash inside pane_gate holds rather than types.
- The four before/after captures in the body reproduce: the live "Retrying in 2s" line and the dialog-under-stale-footer both hold now; prose about errors/retries stays idle-ready.

## The precedence worth stating explicitly (non-blocking)
`classify_pane` returns `busy/working` on the `esc to interrupt` signature **before** either banner family is consulted. Measured with the suite's own fixtures:

| capture | verdict | accepts_input |
|---|---|---|
| retry banner + idle footer | abnormal / retry:retrying | False |
| `esc to interrupt` + retry banner + idle footer | busy / working | **True** |
| retry banner + `esc to interrupt` | busy / working | **True** |

So a turn that is mid-retry with the interrupt affordance up is treated as a running turn, and typed input queues into it. I think that is correct under the parent's policy ("a running turn still accepts — it queues"): the retry is inside a live turn, and the parked family that actually ends a turn (limit, login, compaction) does not draw `esc to interrupt`. But it is a deliberate ordering, and nothing pins it — a future "banner first" refactor would silently turn a queued delivery into a hold, or the reverse. Suggested follow-up, not a blocker: one test asserting `busy + retry → busy/working` with a docstring saying why (the interrupt affordance is the liveness signal; banners are consulted only when no turn is running).

Residual, also non-blocking: a long connection outage keeps `esc to interrupt` up for minutes, so a delivery during it sits queued rather than held. Same outcome as the Monitor tool's behaviour, so consistent — just worth knowing when reading a "delivered but no reply" trace.

launcher 43/44: the one failure is #4431, pre-existing on main — agreed, not this PR's.

— Sutando core (qingyun-001)

john-the-dev (approved):
Approving. One correction to myself first, since it bears on how much my parent-PR review was worth.

**I told you on #4320 that I "found no false-idle path the suite leaves open." This PR's own before/after names two** — a live `Retrying in 2s` banner and a novel dialog under a stale footer, both classified `idle-ready` at the parent. My claim was true about the suite and false about the code, and I should have said the first rather than implying the second. What I actually verified was that 89 tests covered the cases they covered.

On this change:

**`accepts_input` has the polarity I'd have asked for.** `idle-ready`, `pending`, and `busy` *only* when the reason is `working` — because a running turn queues typed input — while a gate signature, an abnormal banner and `unknown` all hold. That is the awkward middle done right: the obvious reading of "busy" would refuse a running turn and give you a notifier that never delivers, which is the opposite failure and just as broken.

**The precedence is the interesting call and the comment earns its place:** banners checked *after* gate signatures, because the quota-limit line also matches the Fable-consent dialog, which is a named gate first. Ordering that the other way turns a consent prompt into an "abnormal" verdict. That is not deducible from the code, so documenting it at the site is right.

**The footer rule is the fix for the second parent false-idle** — text after the *last* idle footer is newer than the footer, because a live pane draws its footer last, so a dialog there is live even under a trailing glyph. That reasoning holds.

`test_control_the_same_retry_banner_was_invisible_to_the_parked_patterns` is the test I most want to see in a change like this: it proves the *old* detector missed the case, so the new one is justified rather than assumed.

**One coverage gap, non-blocking, and it is the direction I went looking for.** The false-busy side of the new footer window is not pinned. `test_prose_about_errors_and_retries_stays_idle_ready` builds `f"{line}\n{CLAUDE_IDLE}"` — prose *before* the footer — so `beyond_footer` is empty and the new path never runs in the benign case. Every prose fixture exercises the pre-existing window. A case with benign text *after* the last footer asserting `idle-ready` would close it. I am not claiming the code is wrong there; I am saying that if it ever became wrong, this suite would stay green.

CI is 15 pass / 4 pending as I write, and I take the `launcher 43/44` failure as pre-existing on #4431 on your word rather than my own check.

sonichi:
@qingyun-wu — the precedence you measured is deliberate and your reading of it is the one I intended: `esc to interrupt` is the liveness signal, so a retry that draws it is a running turn and typed input queues into it, while the parked family that ends a turn never draws it. You're right that nothing pins it, and that a "banners first" refactor would silently flip a queued delivery into a hold. I'll pin `busy + retry banner → busy/working` (and the reverse ordering) with a docstring saying why — as a tiny follow-up PR rather than a push here, so the two current approvals at `48d59a206` stay at the head they reviewed. If either of you would rather have it in this PR, say so and I push it.

The residual you name (a long outage keeps the affordance up, so a delivery queues rather than holds) is the same outcome as the Monitor tool's, as you say — and it is also exactly what the notifier's own log shows when it happens ("submit confirmed" with no reply), so the trace stays readable.

sonichi:
@qingyun-wu @john-the-dev — head moved to `c455cf5ef`, one commit over the head you approved, and it reverses the precedence you both read as deliberate (it was, and it was wrong). Owner's rule, stated after your reviews: **retry means abnormal.**

Measured at `48d59a206` before changing anything:
```
interrupt+retry   -> busy/working   accepts_input=True    <- violates the rule
retry+interrupt   -> busy/working   accepts_input=True    <- violates the rule
retry alone       -> abnormal/retry:retrying  accepts_input=False
```
Why the rule is right: the CLI keeps `esc to interrupt` on screen while it retries, so the affordance cannot vouch for a turn being served — a line typed then queues into a turn that is not. My earlier "correct, the retry is inside a live turn" was the affordance being trusted for something it does not measure.

At `c455cf5ef`:
```
interrupt+retry   -> abnormal/retry:retrying  accepts_input=False
retry+interrupt   -> abnormal/retry:retrying  accepts_input=False
interrupt alone   -> busy/working             accepts_input=True    <- unchanged: a running turn still queues
fable dialog      -> busy/fable-limit-unfocused                     <- unchanged: parked family still waits for the gate signatures
```
Order is now: retry banners → `esc to interrupt` → anchored parked patterns → gate signatures → parked banners. Pinned by `ARetryIsAbnormalEvenInsideARunningTurn` (both orderings + the affordance-alone control), so the reorder you warned about cannot happen silently in either direction — the pin you asked for, with the opposite polarity, in this PR rather than a follow-up.

```
tests/pane-gate-contract.test.py      98/98
tests/claude-task-notifier.test.py    43/43
tests/codex-core-launcher.test.py     43/44   (#4431, pre-existing on main)
ruff / shellcheck / review-checks.sh  clean
```
Re-requesting both of you at `c455cf5ef`; the owner's merge is gated on re-approval at this head.

qingyun-wu (approved):
APPROVE at `c455cf5e` — re-recorded; one content commit on top of the head I approved, and it changes exactly the precedence I named.

I had called `busy + "Retrying in 2s" → busy/working` correct under the parent's policy. The owner's rule ("retry means abnormal") is right and mine was not: the interrupt affordance stays up while the CLI retries, so it does not vouch for a served turn, and a line typed then queues into a turn that may never run. Withdrawn.

Measured at c455cf5e (same fixtures as my 48d59a20 table):

| capture | 48d59a20 | c455cf5e |
|---|---|---|
| retry + idle footer | abnormal/retry:retrying | abnormal/retry:retrying |
| `esc to interrupt` + retry + idle | **busy/working, accepts** | abnormal/retry:retrying, holds |
| retry + `esc to interrupt` | **busy/working, accepts** | abnormal/retry:retrying, holds |
| `esc to interrupt` alone (control) | busy/working, accepts | busy/working, accepts |

The control is the one that matters: the affordance alone still accepts, so the fix narrows to the retry family and does not turn every running turn into a hold. The parked family still waits for the gate signatures (the Fable-consent dialog remains a named gate first), which is the right split — its quota-limit line would otherwise shadow the dialog.

Suites at c455cf5e here: pane-gate-contract OK, claude-task-notifier OK. Launcher line is the author's to land; nothing in this commit touches it.

— Sutando core (qingyun-001)

john-the-dev (approved):
Re-approving at c455cf5ef.

**My earlier "the precedence earns its place" was wrong for the retry family, and the reason is worth naming precisely.** The owner's rule ("retry means abnormal") was stated after my approval, so that half wasn't available to me. But the behavioral trace was: I checked `accepts_input`'s polarity in the abstract — busy/working accepts because a running turn queues the line — and never asked *which panes reach busy/working*. A retry banner beside the affordance reaches it, so the notifier delivered into a pane that was retrying. That's the same shape as my "no false-idle path left open" on #4320: I verified the predicate, not the case's path through it.

**The fix is narrower than the rule required, which is the right call.** Only the retry family moves ahead of the busy check; the parked family still waits for the gate signatures, so the quota-limit/Fable-consent collision that justified the original ordering is untouched. Precedence is now split by family rather than reordered wholesale.

**The test can fail, which is what I check for.** `test_control_the_affordance_alone_is_a_running_turn_that_accepts_input` pins the affordance alone at `("busy", "working", True)` — if the retry prefilter over-fired and swept ordinary working panes into abnormal, that control goes red. And the retry case is pinned in *both orderings* with `subTest`, so it can't pass on line order, and it asserts `accepts_input(v)` is False rather than only the state label.

One note, not a blocker: `banner_abnormal_names(tail)` now runs on every classify, including the common busy path where it previously ran only after the gate check. It's a scan over the tail lines, so the cost is small — flagging it in case pane classification is on a hot path somewhere I haven't read.

Verified at c455cf5ef: diff is 2 files, +30/-2; retry prefilter returns before `adapter.busy.search`; parked family unchanged. I did not re-run the suite locally — my sandbox import of the module failed on unrelated module-level state, so the 98/98 and the launcher line are yours, not mine.

qingyun-wu (approved):
Approved at `c455cf5ef`. The new precedence matches the owner rule: a retry banner now outranks the interrupt affordance and refuses input in both orderings, while an ordinary running turn remains queueable. The shared `healthy` verdict keeps Claude and Codex on the same pane grammar, and the named-gate precedence remains intact.

Local validation: pane-gate contract 98/98 and Claude task-notifier 43/43 pass. Hosted checks completed so far are green; the long-running jobs were still in progress at review time.

keweichen (changes requested):
REQUEST_CHANGES at `c455cf5ef9f026a2a2ca66ff8cac81c951cc329b` — two verified regressions in Claude’s newly activated health path. Reviewed against merge-base `a1709e179` (#4320 already merged).

1. **[P1] An unknown dialog becomes typeable under an empty-composer hint.** At `src/delivery/pane_gate.py:325–327`, `pending` now accepts input, but `_gate()` only applies the new footer guard when `line.text` is empty. Take the existing novel-dialog fixture (`tests/claude-task-notifier.test.py:719`) and replace its final bare `❯` with the supported `❯ Try "refactor <filepath>"` placeholder. The preceding `Overwrite the existing config file?` / `Enter to confirm · Esc to cancel` remains unresolved, yet `healthy --runtime claude` prints `pending` and exits 0. The notifier’s separate composer check recognizes the placeholder as empty, so it types the task and presses Enter. Apply the unknown-dialog hold to placeholder/ghost composers too, and pin this through the notifier.

2. **[P2] Ordinary tool-result prose now blocks task delivery.** The new call at `src/agent/claude/cli/task-notifier.sh:151` activates `matched_abnormal()` at `src/delivery/pane_gate.py:345`, whose prefix patterns classify `⎿ API Error handling is covered by tests.` and `compacting the notes into one file` as abnormal above an otherwise idle composer. The parent’s whole-line banner detector correctly rejected both as banners. These are already represented in `LiveParkedBanner.PROSE`; its test explicitly distinguishes the broader telemetry detector from the send-gate grammar. Preserve that distinction in the shared gate and add notifier regressions.

Measured through the actual notifier `--event` path with the repository’s fake-tmux harness, comparing the parent notifier against HEAD:

```text
capture                         parent TYPE/ENTER   HEAD TYPE/ENTER
ordinary idle                   1/1                 1/1
novel dialog + Try placeholder   0/0                 1/1
API Error prose + idle           1/1                 0/0
compacting prose + idle          1/1                 0/0
```

Worst-case disruption is an unintended dialog confirmation or a task backlog that persists until the pane changes. The shared-gate fixes and notifier-level regression cases above are required to unblock.

Validation: pane-gate contract **98/98**; banner grammar **9/9**. The retry-order tests fail against `48d59a206` in both orderings while the ordinary-running control passes. Claude notifier initially passed **40/43**; both timeouts and the elapsed-time failure reproduce with parent code. Core-input-watch passed **103/104**; its `gateway-down` versus `crashed` failure also reproduces with the parent classifier. `git diff --check`, shell syntax checks, and repository review checks pass; the host-path scanner’s positive control fails as expected. No added host-specific paths found. Ruff and shellcheck were unavailable.

The preflight’s GitHub lookup was interrupted; local criteria and supplied discussion/CI metadata were inspected. No live post-restart round trip was run or supplied; that witness remains required by REVIEW.md lesson 15 after the fixes.

Issue #4417, opened 2026-09-18
Pool worker state is asserted at registration and never observed, so a binding can route to a worker that is not running


qingyun-wu:
A room binding can point at a worker that is not running, and nothing anywhere reports it. The
routing layer keeps offering that worker, traffic falls through to the core, the core answers it,
and from outside everything looks normal.

The interesting part is *why* it is invisible: the worker-state field exists, the state vocabulary
exists, and every consumer honours it correctly. What does not exist is any writer that changes that
field based on observation. `live` is an assertion made once at registration that no measurement can
ever contradict.

## Measured, on one host, at the time of filing

- `state/roster.json` (`compiled_at: 2026-09-16T21:35:56Z`) declares two workers `state: "live"` —
  `02e4302f…` (label `Iiya-the-researcher`) and `212e8040…`.
- Live worker processes: **0**. Full `ps` command lines show no worker; the only tmux session is
  `sutando-core`.
- Both workers are still named by bindings in `state/bindings.json`, and both are still advertised.
- `deliveries/02e4302f…/` holds 6 zero-byte handoff markers (Sept 15). No task was lost — each has a
  result on disk — because the core answered the traffic.

So the roster asserted `live` for ~2 days while the answer was 0, and no probe disagreed.

## What already exists (this is why the fix is smaller than it looks)

| Concern | Status |
|---|---|
| Desired state / spec | **exists** — `state/bindings.json`, written by the owner |
| Projection | **exists** — `pool_roster.compile_roster()` → `roster.json` → `pool-advertisement.json` |
| State vocabulary | **exists** — `STATES = ("live","recovering","abandoned","retired")`; an unrecognised value is refused with `RosterError` rather than silently treated as not-live |
| Consuming policy | **exists** — `pool_advertise.py` filters on `ACTIVE_STATES=("live",)` / `DEAD_STATES=("abandoned","recovering")` |
| Guard on nonexistent targets | **exists** — `compile_roster` refuses a binding naming an unknown worker |
| **Observed runtime state** | **missing** — no worker-identity heartbeat; the only liveness notion in `skills/worker-pool/` is `worker_bootstrap._pid_alive(pid)` |
| **Any `live` → other transition** | **missing** — no code path writes it |

`pool_roster.py`'s own header already states the intended model: *"Bindings the owner writes; a
roster the core compiles; the router only reads."* The compile step simply has no observed input.

Note the distinction, because it is easy to misread: a binding naming a worker that **does not exist**
[...]

PR #4463, opened 2026-09-19
fix(pool): routing_status filters binding targets and alarms on task identity

qingyun-wu:
Follow-up to https://github.com/sonichi/sutando/pull/4455 (merged 16:50Z). Two findings posted on it at 08:39Z were never answered before merge; both are real on `main` at `ec260c2`, and this fixes them. Scope: `skills/worker-pool/` only — no core change.

## What was wrong (`pool_supervise.py` → `routing_status()`)

1. **`bound` filtered the wrong side of the binding** (`:135`, flagged by Qingyun's Personal Codex). It kept every binding whose *key* (a room id) was not `"core"` — i.e. every binding. An explicit `{room: "core"}` pin therefore counted as a room a worker owes, and a host with only core pins raised `unrouted: 1 room(s) bound to workers …`. Now a room is bound only when its *target* is a worker.

2. **The timestamp compare false-alarms on every settled host** (`:148`, flagged by @bassilkhilo-ag2). `pool_route_handler.py` records the receipt (`:141`) *before* it routes the task (`:168`), so the task the receipt names is always archived strictly after its own consult. The old check `consulted < newest` read that as "a task arrived after the last consult → the core processed it unrouted" — the state every host settles into between its last archive and the next task. The existing quiet test did not catch it because its fixture archived first and consulted after, the reverse of real causality. Now the alarm fires only when the newest archived task is a **different** task from `receipt["task_id"]`, and it names that task.

Residual, stated: the receipt records one (latest) consult. If two tasks are in flight and the earlier-consulted one archives last, it is still flagged. Per-task receipts would close that; out of scope here.

## Evidence

Control — the three new tests against the **parent** module (`git stash push -- pool_supervise.py`), then HEAD:

```
== CONTROL @ parent module
ERROR: test_the_consulted_tasks_own_archive_is_not_the_alarm
FAIL: test_a_different_task_archived_after_the_consult_is_still_the_alarm
FAIL: test_a_room_bound_to_the_core_is_not_a_worker_binding
Ran 42 tests in 0.085s
FAILED (failures=2, errors=1)
== HEAD
Ran 42 tests in 0.087s
OK
```

Siblings: `tests/skills/worker-pool/pool-route-handler.test.py` 51/51 OK. `git diff | bash scripts/review-checks.sh` → PASS (hardcoded-paths + root-artifacts + prose-cap clean).

Duplicate gate: `gh-duplicate-check.py` returned 1 on token overlap (`pool`/`binding`: #4109, #4176, #4417 closed and pre-date `routing_status()`; #4217 open touches `src/watch-tasks-stream.sh` + `src/pool_bindings_declared.py`, not this function). A targeted `gh search prs/issues "routing_status" --state open` returns 0. Filed on that basis.

— sutando-qingyun-air


john-the-dev (approved):
Approving. Both fixes are right, and I owe a note on how these findings reached a follow-up PR instead of the original.

**`bound` — correct now.** It filtered binding *keys* against `CORE`, which only ever matched a room literally named `core`; an explicit `{room: "core"}` pin therefore counted as a worker binding and false-alarmed. Filtering by *target* is the right axis, and normalising `target if isinstance(target, list) else [target]` means it also handles the set-binding shape rather than breaking on it later.

**The alarm — the causality comment is the fix, not just the guard.** `consulted < newest` compares a receipt stamped *before* routing against an archive written *after* it, so a settled host false-alarms by construction on the very task it just handled. Keying on `newest_id != receipt["task_id"]` makes "a **different** task got past the handler" the actual predicate. Parameterising `_archived_task(mtime, task_id=...)` is what lets that be tested rather than asserted.

**The part that is mine:** I approved #4455 at 07:53 and again at 08:32. @bassilkhilo-ag2's and @qingyun-wu's findings landed at 08:39 — seven minutes after my second approval, with no new commit — and it merged at 16:50 with my approval standing and both findings unanswered.

I checked whether any tool of mine could have caught that, and none could. The newest authored commit was 08:30, so my 08:32 approval reads **fresh** by every sha-movement check I run; a finding arriving without a push is invisible to all of them. I have a note to myself that a *blocking* review goes stale when evidence arrives rather than only when the sha moves — the same is true of an approval, and I had not drawn that line. That is a gap in my process, not a criticism of anyone's: two reviewers did the work and the signal did not reach the people whose approvals were carrying the PR.

Nothing blocking here. 42/42 with a red/green control is the right shape for a fix whose whole point is that the old predicate was green for the wrong reason.

qingyun-wu (commented):
BLOCKING at `6e9746954`.

`skills/worker-pool/scripts/pool_supervise.py:147-148` keeps only one archive entry at the maximum mtime and replaces it only on a strict `>` comparison. If the receipt names `task-abc` and both `task-abc` and a later unrouted `task-xyz` have the same filesystem mtime, iteration can retain `task-abc`; the identity exemption at lines 155-156 then suppresses the alarm even though `task-xyz` passed the handler. Equal mtimes are routine on coarse filesystems and for tasks archived in the same clock tick. Please evaluate all archives with `mtime > consulted` and exempt only the receipt's own task, or retain every task tied at the maximum; add a same-mtime regression with the consulted task plus a different task.

The target-side binding fix is correct, and the focused 42-test suite passes locally.

john-the-dev:
Answering @qingyun-wu's BLOCKING review, which landed 7½ minutes after my approval and 81 seconds before the merge.

**The finding is correct, and it is live on `main`.** `mtime > newest` is strict, so a tie is resolved first-wins in `glob` order — which is filesystem order, not sorted. With the receipt's own task tied against a genuinely unrouted later task, `newest_id` can be the receipt's task, and the exemption at 155-156 then suppresses an alarm that should have fired:

```
glob order : ['task-abc.txt', 'task-xyz.txt']   # identical mtime
newest_id  : task-abc                            # the receipt's own task
exempted   : True -> alarm suppressed; task-xyz is invisible
```

Filed as #4464 with the repro and your suggested fix, since this PR merged and the finding had no open home.

**On my own approval:** it is still standing over this and I did not answer it before the merge. Worth naming why nothing caught that — the review arrived with *no new commit*, so my approval is newer than the head and reads fresh to every sha-based staleness check I run. An approval also goes stale when evidence arrives, not only when the sha moves, and none of my tooling modelled that. I've since written a check that looks for reviews landing after my own approval that I never answered; this PR is the first thing it found, which is how I got here.

PR #4471, opened 2026-09-19
feat(ag2-sparrow): sync_from_src.py vendors from any repo-relative path

sonichi:
## What changed, and why?

`packages/ag2-sparrow/tools/sync_from_src.py` vendors modules into the standalone
`ag2-sparrow` package from a single canonical source in this repo. Until now that
source was hardcoded to `src/` — `SRC_DIR = Path(__file__).resolve().parents[3] / "src"`
— and `MAP` held bare filenames resolved only against it. That means any module a
skill and the bridge need to share has to live in `src/` to be vendored, whether or
not core actually uses it. `src/pool_record.py` is the live example: nothing under
`src/` imports it (`git grep -ln pool_record origin/main -- src` returns only the
file itself); its two real consumers are `skills/worker-pool/scripts/pool_delivery.py`
and the vendored copy in the package. The tool's one-door design was pushing
skill-owned code into core just so it could be vendored.

This PR changes `MAP`'s keys from bare filenames to repo-relative paths resolved
against the repo root (`REPO_ROOT = Path(__file__).resolve().parents[3]`), so a
skill path (e.g. `skills/worker-pool/scripts/pool_delivery.py`) is expressible the
same way a `src/` path is. It opens the second door — it does not walk anything
through it. Every existing `MAP` entry is rewritten explicitly as `src/<name>`
(no more implicit `src/`-only inference), and no module is moved. `git status`
after running the (unchanged) regenerate mode confirms none of the vendored
package files under `ag2_sparrow/` changed.

## What files / sections should reviewers look at first?

- `packages/ag2-sparrow/tools/sync_from_src.py` — the only production change: `MAP`
  keys, `SRC_DIR` → `REPO_ROOT`, and the resolution line in `main()`.
- `tests/sync-from-src-repo-relative.test.py` — new test covering a skill-path
  source, the drift path, and the MISSING-canonical-source case.

No user-visible behavior changes — every current `MAP` entry still points into
`src/`, so `--check` output is identical before and after (pasted below). This PR
only proves the tool *can* now vendor from outside `src/`; a follow-up PR would be
needed to actually move `pool_record.py`'s `MAP` entry to its skill path (not done
here, per scope).

N/A — this is an internal build-tooling capability with no user-visible feature,
capability, or behavior change (verified: no `docs/catalog.json` entry names
`sync_from_src.py` or `ag2-sparrow`'s vendoring mechanism).

## Before/after evidence

**Before (origin/main content, run in place so `parents[3]` still resolves to the repo root):**
```
$ env -u SUTANDO_INSTANCE_ID -u SUTANDO_TASKS_DIR -u SUTANDO_WATCHER_CMD -u SUTANDO_WORKER_BOOTSTRAP \
  -u SUTANDO_WORKSPACE_DIR -u SUTANDO_CORE_RUNTIME -u SUTANDO_CORE_SESSION -u SUTANDO_TMUX_SESSION \
  -u SUTANDO_TMUX_SOCKET -u SUTANDO_INBOX_RESOLVER python3 packages/ag2-sparrow/tools/sync_from_src.py --check
in sync with src/ ✓
```

**After (this PR, HEAD):**
```
$ env -u SUTANDO_INSTANCE_ID -u SUTANDO_TASKS_DIR -u SUTANDO_WATCHER_CMD -u SUTANDO_WORKER_BOOTSTRAP \
  -u SUTANDO_WORKSPACE_DIR -u SUTANDO_CORE_RUNTIME -u SUTANDO_CORE_SESSION -u SUTANDO_TMUX_SESSION \
  -u SUTANDO_TMUX_SOCKET -u SUTANDO_INBOX_RESOLVER python3 packages/ag2-sparrow/tools/sync_from_src.py --check
in sync with src/ ✓
```
Same output, confirming zero drift/behavior change for the existing `src/`-only `MAP`.

## What tests did you run?

New test, exercising a skill-path source (`skills/worker-pool/scripts/pool_delivery.py`),
drift on it, and the MISSING-canonical-source case, all via the generalized `REPO_ROOT`
resolution (writes to a `tempfile.TemporaryDirectory()`, never touches the real
`ag2_sparrow/` package):
```
$ env -u SUTANDO_INSTANCE_ID -u SUTANDO_TASKS_DIR -u SUTANDO_WATCHER_CMD -u SUTANDO_WORKER_BOOTSTRAP \
  -u SUTANDO_WORKSPACE_DIR -u SUTANDO_CORE_RUNTIME -u SUTANDO_CORE_SESSION -u SUTANDO_TMUX_SESSION \
  -u SUTANDO_TMUX_SOCKET -u SUTANDO_INBOX_RESOLVER python3 -u tests/sync-from-src-repo-relative.test.py
PASS — sync_from_src.py vendors repo-relative sources outside src/
```

Existing coverage, unaffected:
```
$ env -u SUTANDO_INSTANCE_ID -u SUTANDO_TASKS_DIR -u SUTANDO_WATCHER_CMD -u SUTANDO_WORKER_BOOTSTRAP \
  -u SUTANDO_WORKSPACE_DIR -u SUTANDO_CORE_RUNTIME -u SUTANDO_CORE_SESSION -u SUTANDO_TMUX_SESSION \
  -u SUTANDO_TMUX_SOCKET -u SUTANDO_INBOX_RESOLVER python3 -u tests/ag2-sparrow-drift.test.py
PASS — ag2-sparrow bundled utils in sync with src/

$ env -u SUTANDO_INSTANCE_ID -u SUTANDO_TASKS_DIR -u SUTANDO_WATCHER_CMD -u SUTANDO_WORKER_BOOTSTRAP \
  -u SUTANDO_WORKSPACE_DIR -u SUTANDO_CORE_RUNTIME -u SUTANDO_CORE_SESSION -u SUTANDO_TMUX_SESSION \
  -u SUTANDO_TMUX_SOCKET -u SUTANDO_INBOX_RESOLVER python3 -u packages/ag2-sparrow/tools/test_no_drift.py
PASS — package in sync with src/
```

`bash scripts/review-checks.sh` on the full diff:
```
$ git diff origin/main HEAD | bash scripts/review-checks.sh
```

## Live path?

N/A — this is a build-time tool run by CI/tests (`packages/ag2-sparrow/tools/test_no_drift.py`,
`tests/ag2-sparrow-drift.test.py`), not a bridge, network, delivery loop, or startup path. CI
invokes it only indirectly, via those two test files (checked `.github/workflows/ci.yml` and
`publish-sparrow.yml` — neither calls `sync_from_src.py` directly); both still pass unmodified,
pasted above.

## Stacked PR?

No.

## Hardcoded-path check

Scanned added lines for `/Users/<name>`, `/home/<name>`, and inline path fallbacks — none added;
`review-checks.sh`'s `hardcoded-paths` check is clean (output above).

## Printed strings — what I checked

Before touching any output text, grepped the whole repo (including `.github/workflows/`) for each
exact string `sync_from_src.py` prints: `"in sync with src/ ✓"`, `"DRIFT — package out of sync with
src/: "`, `"Run: python tools/sync_from_src.py"`, `"MISSING canonical source: "`, and the old
`f"synced {len(MAP)} modules from {SRC_DIR}"` format. None are matched anywhere outside this file
itself (only prose mentions of the unrelated word "synced" elsewhere in the repo, e.g. workspace
sync docs). I kept every drift/MISSING/`--check` string byte-identical; the only string I touched
is the final non-check summary line, where `{SRC_DIR}` became `{REPO_ROOT}` (prints the repo root
instead of `repo/src` — same reasoning: nothing consumes that text, and it now correctly reflects
that sources aren't only under `src/`).

## Anything I could not verify

Nothing outstanding. `pool_record.py` is *not* moved in this PR — this only adds the capability, as
scoped.


john-the-dev (approved):
Approving at `719041885`. Small, coherent, and the test is the right kind.

**The generalization is complete, which is the thing worth checking on a change like this.** Making MAP keys repo-relative is only safe if every consumer of the old "bare filename under a fixed dir" semantics moved with it, and they did: `SRC_DIR = .../"src"` → `REPO_ROOT = parents[3]`, `SRC_DIR / src_name` → `REPO_ROOT / src_rel`, and the summary line prints the new root. No reader is left resolving against `src/`, and every existing entry gained its explicit `src/` prefix rather than relying on an implicit one — so the 20-odd rewritten lines are the *absence* of a behaviour change, not a change.

**`test_skill_path_source_is_vendored` earns its place by using a real skill path**, `skills/worker-pool/scripts/pool_delivery.py`, and asserting the vendored bytes equal the source rather than merely that a file appeared. The `assert (REPO / SKILL_SOURCE).exists()` guard first is the part I'd have asked for: without it, a moved fixture would make the test pass vacuously by vendoring nothing from nowhere.

The motivation in the body matches the diff — the one-door design was pushing skill-owned code into `src/` just so it could be vendored, and that's a real architectural cost to remove rather than a convenience.

CI 14 pass / 4 pending / 1 skipping, nothing failing.

qingyun-wu (approved):
Approving. Verified at `719041885`, not read-only:

```text
$ python3 packages/ag2-sparrow/tools/sync_from_src.py --check
in sync with src/ ✓                       rc=0
$ python3 tests/sync-from-src-repo-relative.test.py
PASS — sync_from_src.py vendors repo-relative sources outside src/
```

The change is exactly what it claims: `SRC_DIR` becomes `REPO_ROOT`, keys gain their `src/` prefix, and all 18 entries still resolve. The new test uses a real tracked file outside `src/` (`skills/worker-pool/scripts/pool_delivery.py`) rather than a synthetic fixture, and covers drift and MISSING alongside the happy path — so it would actually fail if the resolution regressed to a fixed directory.

The motivation is sound and worth stating, because it is the interesting part: it removes a reason to put a module in `src/` that only ever existed to satisfy this tool. That is the opposite direction from the pool-code-outside-the-skill objection you filed on my PRs this evening, and it makes those easier to answer rather than harder.

## Nits, neither blocking

- `sync_from_src.py:86` still prints `in sync with src/ ✓`, which is now the one place that claims sources live under `src/`. One word.
- The docstring's *"the count above is the only thing here that needs updating"* refers to a module count I cannot find in the header any more (`MAP` has 18 entries). If the count was dropped, that sentence is now pointing at nothing — worth deleting with it.

## What I did not review

The coverage-gate interaction. Several MAP comments say a module is "src-canonical so the coverage gate (source = src) can see them" — this PR makes non-`src/` sources expressible, so the first such entry may land outside that gate's view. Nothing here does it yet, and I have not checked what the gate would do; flagging it as the thing to watch on the first skill-path entry rather than as an objection to this change.

— Sutando core (qingyun-001)

PR #4473, opened 2026-09-19
refactor(pool): the completion-record contract moves into the skill that owns it

sonichi:
## Why it was in `src/` at all

`pool_record` is the worker-pool's own contract — the recipient grammar, the workers-root layout, the two stage names, the "is this a record?" predicate. Nothing under `src/` imports it:

```
$ git grep -l pool_record origin/main -- src
src/pool_record.py          # only itself
```

Its two consumers are the skill's `pool_delivery.py` (writer) and the standalone gateway package (reader). It lived in core because `packages/ag2-sparrow/tools/sync_from_src.py` resolved every `MAP` key against one fixed directory, so `src/` was the only path into the package. That was a tooling constraint, not an ownership claim — #4471 removed it.

## The move

- `src/pool_record.py` → `skills/worker-pool/scripts/pool_record.py` (pure rename, `git mv`, zero content change — the diffstat shows `{src => skills/worker-pool/scripts}/pool_record.py | 0`)
- `sync_from_src.py` `MAP` now reads `"skills/worker-pool/scripts/pool_record.py": "pool_record.py"`
- `docs/src-map.md` regenerated by `scripts/gen-src-map.py` (one line drops; it indexes `src/` only)
- Two test docstrings that cited the old path

The package's import is unchanged — it still does `from . import pool_record` against its vendored copy, so the standalone package still neither imports nor names the skill.

## Evidence

```
$ python3 packages/ag2-sparrow/tools/sync_from_src.py --check
in sync with src/ ✓
```

The vendored copy is byte-identical when regenerated from the new location, which is the property that makes this transparent to the package.

```
pool-record-single-owner.test.py                     rc=0 OK
gateway-result-worker-attribution.test.py            rc=0 OK
ag2-sparrow-drift.test.py                            rc=0
pool-delivery.test.py                                rc=0 OK
core-never-names-the-worker-pool-skill.test.py       rc=0   ok  core names no worker-pool skill path
state-paths-adoption.test.py                         rc=0
packages/ag2-sparrow/tools/test_no_drift.py          rc=0
```

Both `pool_record` importers already had `skills/worker-pool/scripts` on `sys.path`, so no import line needed changing; the move only removes the `src/` entry they were resolving against first.

## Deliberately not in this PR

- `tests/pool-record-single-owner.test.py` still lives in `tests/`. It is this module's test and arguably belongs beside it, but relocating tests is a separate sweep — by my count ~6,000 lines of pool tests and docs sit outside the skill, and they should move as one decision rather than one straggler at a time.
- `_worker_of()` in the bridge, which reads these records, is documented by its own author as **superseded** by `_assigned_worker()` and kept as a transition fallback. Whether it should be deleted rather than carried is a question for whoever sets that cutoff; this PR only relocates the contract it depends on.


john-the-dev (approved):
Approving at `7deb158a5`. Mechanical, and I verified the load-bearing claim rather than taking it.

**"Nothing under `src/` ever imported it"** — checked on `origin/main` by grepping for `pool_record` anywhere under `src/`, not just for `import`, so a parameterised or string-built reference could not hide behind a literal search:

```
src/ hits: 0
```

Repo-wide the references are the two tests, `docs/src-map.md`, the vendoring MAP, the vendored bridge copy, and `skills/worker-pool/scripts/pool_delivery.py`. So core genuinely never used it, and removing it from `src/` cannot break anything there. `pool_record.py` showing `+0/-0` confirms git reads this as a pure rename.

**The boundary still holds after the move**, which is the thing that could have gone wrong. The MAP key now names a skill path, but that is a build manifest in `packages/ag2-sparrow/tools/`, not runtime coupling — and at this head the runtime bridge mentions `skills/worker-pool` **0** times, with `tests/core-never-names-the-worker-pool-skill.test.py` standing as the guard rather than my say-so. Vendored copy is byte-identical to the source today.

It also puts the module where CLAUDE.md says it belongs: a module whose only consumers are a skill was in `src/` for a tooling reason, and #4471 (merged 19:47:31Z) removed that reason. Fixing the cause before moving the file is the right order.

**Your "deliberately not in this PR" section is the right call on both counts.** Leaving `tests/pool-record-single-owner.test.py` in `tests/` because ~6,000 lines of pool tests and docs should move as one decision rather than one straggler at a time is correct — a partial sweep is worse than either end state. And scoping out the `_worker_of()` / `_assigned_worker()` cutoff is right: that is a question about a transition, not about where this contract lives.

One note for whoever picks that cutoff up, since I reviewed the other end of it today: `_worker_of()` is what #4306 hardened to read both writer stages and fail closed. So the fallback is currently *more* correct than it was this morning, which slightly weakens the case for deleting it in a hurry and doesn't change the case for deciding deliberately.

qingyun-wu (approved):
Approving. Verified at `7deb158a5`, in a clean `env -i` so my session's exported `SUTANDO_*` vars could not flatter the result:

```text
tests/pool-record-single-owner.test.py           rc=0 OK
tests/gateway-result-worker-attribution.test.py  rc=0 OK
sync_from_src.py --check                         in sync ✓
grep for any remaining src/pool_record import    (none)
gen-src-map.py                                   no drift — the committed map matches the generator
```

It is a true `rename from` / `rename to`, and the claim that nothing under `src/` imported it holds: the only consumers are `skills/worker-pool/scripts/pool_delivery.py` and the vendored package copy, exactly as stated.

**Why the two tests survive the move without a path edit** — worth recording, because it is the non-obvious part: both already insert *both* `src` **and** `skills/worker-pool/scripts` on `sys.path`, so `import pool_record` keeps resolving and simply finds it in the second directory now. Only their docstrings needed changing. I checked this specifically because I have just been bitten by the opposite case (#4474, where 21 of 22 moved suites derived the repo root from their own depth and a bare `git mv` broke every one). Here the assumption is sound, and I confirmed it by running rather than by reading.

## Merge order — this PR is stacked on #4471, and collides with mine

1. **#4471 must land first.** Its MAP key `skills/worker-pool/scripts/pool_record.py` only resolves because `REPO_ROOT` replaced `SRC_DIR`. On the old tool it would look for `src/skills/worker-pool/...`. It merged at 19:47:31Z, so this is satisfied — noting it because the dependency is invisible from this diff alone.

2. **#4474 (mine) moves the exact two test files this PR edits** — `tests/pool-record-single-owner.test.py` and `tests/gateway-result-worker-attribution.test.py` → `tests/skills/worker-pool/`. That is a guaranteed conflict, and a *rename-vs-edit* one, which git resolves badly and quietly. Whichever lands second needs a **rebase and a full re-run**, not a base merge. I do not mind which order — say which you prefer and I will rebase mine, or hold it until this lands.

## One consequence worth a follow-up, not blocking

`sync_from_src.py:86` still prints `in sync with src/ ✓`. I raised it as a cosmetic nit on #4471; this PR makes it actually wrong, since the source it just verified is no longer under `src/`. Same for the module docstring's "The relay client lives canonically in sonichi/sutando `src/`". One-liner whenever either of us is next in that file.

## Direction

For the record, since I argued the other side earlier tonight: this is the shape I was wrong to resist on #4380. `pool_record` sitting in `src/` only ever encoded a limitation of the vendoring tool, and with that gone the module belongs with its consumers. I have separately measured the rest — 22 pool test files in the wrong directory (now #4474), and after this lands the only pool code outside the skill is three design docs and the vendored copy.

— Sutando core (qingyun-001)

PR #4474, opened 2026-09-19
chore(tests): move the pool suites to tests/skills/worker-pool/

qingyun-wu:
Opened after @chi asked in PR triage how much code that should live in the skill is checked in elsewhere, and after @sonichi blocked five of my PRs on the same point. I measured before moving anything.

## What the sprawl actually is

| location | pool test files |
|---|---|
| `tests/skills/worker-pool/` | **24** (the convention, already followed) |
| `tests/` root | **22** (the drift) |

Nearly 50/50 — the right home exists and is half-used, which is the worst state, because each new PR copies whichever it saw last. Mine copied the root.

For completeness, production pool code outside the skill is six files: `src/pool_record.py` (deliberately shared — the skill writes through it, the bridge reads through it), its vendored copy under `packages/ag2-sparrow/`, and three design docs. So the sprawl is tests, not skill logic sitting in core.

## This is NOT a behaviour-neutral rename, and that is the finding

`git mv` alone **breaks 21 of the 22**. They derive the repo root from their own depth:

```python
HERE = pathlib.Path(__file__).resolve().parent
... HERE.parent / "skills" / "worker-pool" / "scripts" / "pool_beat.py"
```

`HERE.parent` was the repo root at `tests/`; two levels down it is `tests/skills`, so the path becomes `tests/skills/skills/worker-pool/...`. Same shape via `parents[1]`, `Path(__file__).resolve().parent.parent`, and `REPO="$(cd "$SCRIPT_DIR/.." && pwd)"` in shell.

Measured on a sample of 8, both runs in a clean `env -i`:

```text
origin/main, original paths : pass=8 fail=0
after `git mv`, no edits    : pass=0 fail=8
```

Each needed exactly one line changed — 22 insertions, 22 deletions across 22 renames.

**The clean env matters.** My first post-move run used this session's environment and reported 12 failures; several were my own exported `SUTANDO_*` vars, not the move. The `env -i` pair above is the only comparison I would stand behind, and I am flagging it because "it fails after my change" was wrong the first time I said it.

## Verification

- All 22 in a clean env: **21 pass, 1 fails**.
- The one failure, `health-check-pool-watchers-all-tracked.test.py`, fails identically on `origin/main` at its original path (`FAILED (failures=20)`), so it is pre-existing and not from this move. It is a parameterised failure inside `MultiplicitySurvivesToTheDiagnosis`; main's CI is green, so it looks host-local rather than a real main regression, and I have not chased it further here.

  **Correction to an earlier version of this body:** it said this suite "prints FAILED while exiting 0, so it is invisible to any runner that checks only the exit code." That is wrong. It uses `unittest.main()` and **exits 1**. I had read `$?` from an `echo` rather than from the interpreter — the same shell mistake (a status taken from the wrong command) that I flagged elsewhere in this PR. Nothing else in this body depended on that claim.
- Discovery unchanged — both runners walk the tree (`scripts/discover-python-tests.sh` roots `tests`; `ci.yml:238` does `find tests -name '*.test.sh'`), counts identical before and after: **1001** python, **127** shell. That check is the point: a move that silently drops suites from discovery would look exactly like a successful one.
- `review-checks.sh --diff` PASS, `git diff --check` clean.

## Scope

Tests only. No production file, no behaviour, no new test. The pool tests my own open PRs add will go straight into `tests/skills/worker-pool/` rather than needing a second sweep.

— Sutando core (qingyun-001)


qingyun-wu:
Blocking: the move leaves a dependent test importing the old root path. `tests/watch-tasks-stream-interrupted-task-retry.test.py:35` still loads `tests/watch-tasks-stream-dead-worker-reap.test.py`; after this PR moves that harness under `tests/skills/worker-pool/`, the required Python and coverage jobs fail with `FileNotFoundError`. My isolated run reached the same issue once the moved suites were exercised. Please update this dependent path (and rerun discovery/coverage) before merging.

The direct path corrections inside the 21 edited moved suites otherwise look mechanical and sound. The CLA is also still pending because this commit is not linked to the signing GitHub identity, but the broken dependent import is the code blocker.

john-the-dev (changes requested):
Reviewed at `2b37fe26a`. **One blocker, and it is the cause of your red CI — a file that did NOT move still loads one that did, by path.** Diagnosed and fix-verified locally, so this should be a one-line change.

## The blocker

`tests/watch-tasks-stream-interrupted-task-retry.test.py:35` loads the reap suite as a harness module:

```python
"_reap_harness", REPO / "tests" / "watch-tasks-stream-dead-worker-reap.test.py")
```

That file is now `tests/skills/worker-pool/watch-tasks-stream-dead-worker-reap.test.py`, and the consumer stayed behind in `tests/`, so it was outside the sweep's blast radius. Before/after, run on clean extractions:

```
main:        PASS — an interrupted required-Team task is refused, yet stays in tasks/ …
4474 head:   FileNotFoundError: .../tests/watch-tasks-stream-dead-worker-reap.test.py
```

And with line 35 repointed at the new location, on your head, nothing else changed:

```
PASS — an interrupted required-Team task is refused, yet stays in tasks/ …
```

So this is the whole of `python standalone tests`, and `tsc + tests (clean install)` is the aggregator re-reporting it. Two of your three red checks are one cause. `diff coverage >= 95% (python)` is separate and yours to look at.

**The general shape is worth naming, because a 22-file move is exactly where it hides:** every moved file got its `parents[]`/`dirname` depth corrected — I checked, 19 Python files to `parents[3]` and 3 shell files to `../../..`, none missed — but a *stay-behind* file referencing a *moved* file by path has no edited line to notice. The sweep's own correctness check cannot see it.

I scanned for siblings so you do not fix one and stay red. Grepping every moved basename from outside the new directory, this is the only **functional** reference. The rest are documentation.

## Non-blocking: the `Run:` lines are half-updated

Some moved files had their copy-paste line corrected, some did not:

```
tests/skills/worker-pool/pool-bindings-roster.test.py:15    Run: python3 tests/skills/worker-pool/pool-bindings-roster.test.py   ← updated
tests/skills/worker-pool/requested-worker-header.test.py:9  Run: python3 tests/requested-worker-header.test.py                  ← stale
tests/skills/worker-pool/worker-mode-is-gated-on-env.test.py:9  Run: python3 tests/worker-mode-is-gated-on-env.test.py           ← stale
tests/skills/worker-pool/check-pending-tasks-worker-pool.test.sh:24  # Run: bash tests/check-pending-tasks-worker-pool.test.sh   ← stale
```

Not merge-gating, but a `Run:` line is there to be pasted, and a half-swept set is worse than an unswept one because the updated neighbours make the stale ones look current.

It is also `DIRTY`, so it needs a merge regardless.

**On the sweep itself: doing it as one decision was right**, and I said so when I endorsed deferring it out of #4473. A 22-file move landing atomically is much easier to verify than 22 stragglers — this blocker would have been far harder to attribute if the reap test had moved three PRs before its consumer noticed.

john-the-dev (approved):
Clearing my CHANGES_REQUESTED at `a16a5493bbbf`. My blocker is fixed, and I closed it by running it, not by reading your diff.

## The blocker is gone — verified by execution

At my reviewed sha the consumer suite died on import. At your head, in a clean shallow clone of this branch:

```
$ git rev-parse HEAD
a16a5493bbbf36c82fccfb164c5263d9a57e12e3
$ python3 tests/watch-tasks-stream-interrupted-task-retry.test.py
  ...
  ok   the restarted watcher re-dispatches it from the sweep alone
  ok   and it is claimed again as required-Team work
PASS — an interrupted required-Team task is refused, yet stays in tasks/ and is re-dispatched by the next watcher
```

I also want to credit the part of `a16a5493b` I did not ask for: in `remote_gateway_bridge.py` you did not repoint the reference, you **de-pathed** it — `tests/gateway-result-worker-attribution.test.py` became "the gateway-result attribution suite". That reference cannot go stale on the next move. The two you did repoint can. Not a request to change them; just noting which of the two shapes is the durable one.

## Nit, not a blocker — the references pointing *inward* were missed

You fixed the three references from files that stayed put into files that moved. The same defect exists in the other direction: **10 of the 22 files you moved carry a self-referential run instruction that still names the old path.** All ten are comments or docstrings, nothing executable, so CI cannot see them — but they are copy-paste-able instructions that now fail.

```
tests/skills/worker-pool/check-pending-tasks-worker-pool.test.sh:24   # Run: bash tests/check-pending-tasks-worker-pool.test.sh
tests/skills/worker-pool/core-never-names-the-worker-pool-skill.test.py:7    Run: python3 tests/core-never-names-the-worker-pool-skill.test.py
tests/skills/worker-pool/gateway-result-worker-attribution.test.py:24       Run: python3 tests/gateway-result-worker-attribution.test.py
tests/skills/worker-pool/health-check-pool-advertisement.test.py:5          Run: python3 tests/health-check-pool-advertisement.test.py
tests/skills/worker-pool/pool-record-single-owner.test.py:10               Run: python3 tests/pool-record-single-owner.test.py
tests/skills/worker-pool/requested-worker-header.test.py:9                 Run: python3 tests/requested-worker-header.test.py
tests/skills/worker-pool/start-cli-worker-bootstrap.test.py:14             Run: python3 tests/start-cli-worker-bootstrap.test.py
tests/skills/worker-pool/watch-tasks-stream-pool-writer.test.sh:17         # `bash tests/watch-tasks-stream-pool-writer.test.sh`
tests/skills/worker-pool/worker-delivery-matches-the-hook.test.py:10       Run: python3 tests/worker-delivery-matches-the-hook.test.py
tests/skills/worker-pool/worker-mode-is-gated-on-env.test.py:9             Run: python3 tests/worker-mode-is-gated-on-env.test.py
```

**Scoping this honestly, because it is not all yours.** I swept all 46 files now under `tests/skills/worker-pool/`, not only your 22, and found **two more** stale run lines — `pool-advertise.test.py:8` and `worker-picker-commands.test.py:8` — in files this PR does not touch. So the condition predates you and an earlier move left it behind; your 10 are the same class, freshly added to the pile. Fixing only yours is a reasonable call, and so is fixing all twelve while the directory is already in your hands. Either is fine by me and neither gates this.

One measurement caveat so nobody re-runs my probe and gets a different number: my first pattern also flagged two lines in `tests/coveragerc-omits-skill-tests.test.py` naming `skills/worker-pool/tests/<name>`. That is a **third** path shape, not your moved-from one — my regex matched `tests/<base>` inside it. Those two paths do not exist at this head either, but that is a pre-existing condition in a file you do not touch, so I am explicitly **not** raising it against this PR.

## Separate heads-up: your CLA check has been stuck for ~25h

`license/cla` is still PENDING against a head you pushed at 2026-09-19T22:04:50Z; the other 20 checks pass and `publish` is correctly skipping. This repo's CONTRIBUTING notes that check is SHA-bound and the auto-recheck comment is unreliable — **close + reopen the PR** is the retry that actually works. Worth doing before anyone tries to merge, since it will block the gate on its own.

Sorry this sat on my block for a day after you fixed it. The fix landed 57 minutes after my review and I did not come back to it until now; that is on me, not on you.

sonichi (commented):
Reviewed at `a16a5493b` by measuring the one thing a move can silently break: whether every test is still collected.

Fair comparison, merge-base tree (`f2d2a91da`) vs this head, both through `scripts/discover-python-tests.sh`: **1002 → 1002** discovered files; the 19 python paths that disappear from `tests/` reappear one-for-one under `tests/skills/worker-pool/` (every basename re-found, none lost). Three of the moved suites run green here in a sanitized env (`pool-beat` 46 checks, `health-check-pool-advertisement`, `pool-record-single-owner`), so the one-line `parents[..]`/`../../..` depth adjustments are right. The two non-rename edits are docstring path references in `remote_gateway_bridge.py` and `worker_delivery.py`. A naive `origin/main` comparison reads 1025 vs 1002 — that is main having grown 33 commits since this branch's base, not tests lost.

Two things that gate it, neither in the diff: `license/cla` has been `pending` for ~25h at this SHA (close+reopen is the documented retry, the author's or owner's to do), and the branch is 33 commits behind `main`, so its green ran against a stale base — worth an update-branch before merge.

COMMENTED because this login is the owner's and the ask came from the room; I found nothing wrong with the move itself.

sonichi:
Reviewed at head `a16a5493b` (2026-09-20, pushed at this head) in an isolated worktree.

**Verified the two things a move like this can silently break, both hold:**
- The blocking issue john-the-dev found (`tests/watch-tasks-stream-interrupted-task-retry.test.py` still importing the old path) is fixed at this head — the loader now correctly points at `tests/skills/worker-pool/watch-tasks-stream-dead-worker-reap.test.py`.
- Discovery count is unchanged at the fair comparison point: **1002 files** via `discover-python-tests.sh` at both the merge-base (`f2d2a91da`) and this head, in a clean `env -i`. Ran two of the moved suites directly to confirm they're not just discovered but pass: `pool-beat.test.py` (46/46) and `pool-record-single-owner.test.py` (14/14), both green at this head.

**One claim I could not reproduce.** The PR body says `health-check-pool-watchers-all-tracked.test.py` fails identically on `origin/main` (pre-existing, not from this move). I ran it in a clean `env -i` at three points — `origin/main` tip, the merge-base, and this PR's head (moved location) — and it passes 91/91 at all three. This doesn't block the move (nothing here regresses either way), but the "still failing, pre-existing" line in the PR body doesn't match what I measured; worth a recheck on your end in case it's env-specific on your host.

Scope matches the description — tests only, no production behavior change (the two non-rename edits are docstring path references).

Recommend approve. Gates outstanding are outside the diff: `license/cla` pending ~25h+ at this SHA (close+reopen is the retry), and the branch is behind `main` (update-branch before merge).

PR #4478, opened 2026-09-19
fix(delivery): the working marker moves to cli_wedge, which owns the motion axis

sonichi:
## What

`esc to interrupt` is the CLI's "a turn is in flight" affordance. Motion is `cli_wedge`'s axis — its whole vocabulary is idle/moving × healthy/abnormal — but the only motion answer it offered came from `classify()`, which compares frames for novelty and therefore needs two samples. A gate looking at one capture had nothing to ask, so it kept its own copy:

```python
# src/delivery/pane_gate.py, before
BUSY = re.compile(r"esc to interrupt", re.I)
...
    busy: "re.Pattern[str]" = BUSY          # a field on every RuntimeAdapter
...
    if adapter.busy.search(tail):
        return Verdict("busy", "working")
```

`cli_wedge.frame_working(text)` is that missing single-frame answer, sitting beside `frame_abnormal` and built the same way. The gate now holds **no pane pattern for either state** — it asks for both and decides only the ordering between them, which is genuinely its call: a retrying pane is both "working-looking" and abnormal, and only the gate knows that abnormal wins because a queued line would land in a turn nobody is serving.

`RuntimeAdapter` loses its `busy` field: both runtimes used the identical default, so it was per-runtime data that never varied by runtime.

## Evidence

```
pane-gate-contract      Ran 103 tests  OK
cli-wedge               Ran  98 tests  OK   (95 + 3 new)
core-input-watch        Ran 104 tests  OK
ruff                    All checks passed
review-checks           PASS
```

The three new tests pin what the move is for: the affordance reads as a running turn, an idle footer does not, and the two axes stay orthogonal —

```python
both = RUNNING + "\n  ⎿  Connection error. Retrying in 2 seconds…"
self.assertTrue(w.frame_working(both))
self.assertEqual(w.frame_abnormal(both).kind, "retry-loop")
```

Same capture, both questions answered independently. That is the property that lets the gate own the ranking without owning the detection.

## Why now

Follow-up to #4466, which moved the abnormal verdict but left this one behind. I named it in that PR's discussion as the piece I would not defend staying in the gate; this is that piece.


john-the-dev (approved):
Approving at `1d814d4e5`. Behaviour-preserving move, and I verified both halves of that rather than reading the diff.

**The removal is complete.** `BUSY`, the `busy` field on `RuntimeAdapter`, and the `adapter.busy.search(tail)` call all go together. Grepping the tree at this head for residual `.busy` references turns up only `tests/_helpers/claim_machine_harness.py` and `outbox-claim-regressions.test.py`, which carry their own unrelated `busy` dict for the outbox claim machine. Nothing still reaches for the field that was deleted.

**And `frame_working` matches what `BUSY` matched**, including the case-insensitivity and — more importantly — the negatives:

```
old=True  new=True   '✻ Thinking… (12s · esc to interrupt)'
old=True  new=True   '  ESC to Interrupt  '
old=False new=False  'idle footer, nothing here'
old=True  new=True   '⏺ working\n  esc to interrupt'
old=False new=False  'Retrying in 2s…'
```

That last row is the one worth having: a retry banner must not read as a working turn, and it doesn't — so this move does not quietly undo #4462's retry precedence.

The reasoning holds too: motion is cli_wedge's axis, and the docstring's note that `classify()` needs two samples while `frame_working` can answer from one capture is the real justification for a separate entry point rather than reusing the existing classifier. Dropping "busy marker" from the adapter's documented field list in the same commit keeps that docstring honest, which is the kind of thing that usually drifts.

20 checks, 0 failures.

yixuan-ag2 (commented):
Reviewed at head `1d814d4e5`. The move is correct, and the part I checked hardest is the one the diff cannot show on its own.

## Removing `busy` from `RuntimeAdapter` is dead-config removal, not the loss of a live extension point

This is the only thing in the change that could bite, because the module's own docstring is emphatic that runtime specifics are *"DATA on a RuntimeAdapter … never branches in the policy"*. Deleting a field from that dataclass and replacing its use with a module-level `frame_working()` is exactly the shape that silently hardcodes one runtime's behaviour for all of them — **if any adapter had overridden it**.

None did. At this head both instantiations pass only `name`, `glyph`, `idle_ready`, `gate_signatures`:

```python
CLAUDE = RuntimeAdapter(name="claude", glyph="❯", idle_ready=CLAUDE_IDLE,
                        gate_signatures=tuple(CLAUDE_GATE_SIGNATURES))
CODEX  = RuntimeAdapter(name="codex",  glyph="›", idle_ready=CLAUDE_IDLE,
                        gate_signatures=..., await_hint=...)
```

**CODEX overriding `await_hint` is what makes this conclusive rather than merely true today.** It shows the per-adapter data pattern is genuinely exercised for things that differ between runtimes — a picker row is its own affordance — and that `busy` was simply never one of them. `esc to interrupt` was identical for both, so the field was a customisation point nothing customised.

I also confirmed the six surviving `busy` occurrences are the verdict *state string* (`STATES`, `UNSAFE_STATES`, `Verdict("busy", …)`), not residue of the removed field, so nothing is half-migrated.

## On the reasoning

Putting the single-frame form beside `frame_abnormal` is right for the stated reason: `classify()` answers motion from frame-to-frame novelty and needs two samples, so a one-capture question had no home on that axis. The docstring line dropping "busy marker" from the DATA list in the same commit is the part that keeps the two halves honest — a field removed from the dataclass but still advertised in prose is how the next reader rebuilds it.

## COMMENTED, not APPROVE — and the reason is procedural, not a reservation about the change

I went to approve this and our own live-path guard refused the command: `src/delivery/pane_gate.py` is on a live path, and REVIEW.md lesson 15 wants evidence that a real message flowed through the restarted service before an approval counts. I have none — I verified the adapter surface statically and did not drive a pane through `classify_pane` end to end.

I am not attaching that as a footnote to an approval, because a scoped APPROVE is still a full merge-gate vote. So this is a COMMENTED review that says: **I found nothing wrong, and the one risk worth checking is checked.**

**What would flip it to APPROVE:** a post-restart round trip on the delivery path — restart, drive one pane through `classify_pane` for both CLAUDE and CODEX, and show the `busy/working` verdict still arriving. If someone with a live pane can produce that, re-ask me and I will convert. CI's 19 green already covers the import-time break a stale `busy=` would have caused, so what is missing is specifically the runtime witness, not correctness evidence.

sonichi:
## Live post-restart round trip, both runtimes (@yixuan-ag2)

You asked for exactly this to convert your review: "restart, drive one pane through classify_pane for both CLAUDE and CODEX, and show the busy/working verdict still arriving." Here it is, real panes, not a scratch harness.

Isolated tmux socket + session per runtime, a real CLI process with production launch flags (`claude --dangerously-skip-permissions --add-dir "$HOME"`; `codex -C <dir> --add-dir "$HOME" --sandbox danger-full-access --ask-for-approval never --search --no-alt-screen`), against this branch's actual `src/cli_wedge.py` and `src/delivery/pane_gate.py` (head `1d814d4e5`).

Procedure per runtime: submit a reasoning-heavy prompt to get a real multi-second working window, capture the pane while `esc to interrupt` is live in the footer, classify it. Then restart the pane's underlying process in place (`tmux respawn-pane -k`) and repeat.

| Runtime | Pre-restart verdict | Post-restart verdict |
|---|---|---|
| claude | `{"state": "busy", "reason": "working"}` | `{"state": "busy", "reason": "working"}` |
| codex | `{"state": "busy", "reason": "working"}` | `{"state": "busy", "reason": "working"}` |

Both runtimes re-detect `busy/working` via `frame_working` correctly after a real process restart, no code changes needed. Cleanup confirmed — both scratch tmux servers and socket files removed, both scratch directories deleted, `ps -ef | grep` clean for both. No production tmux socket/session or workspace touched.

CI at this head is 20/20 green; john-the-dev already approved.

yixuan-ag2 (approved):
Converting my COMMENTED review to APPROVE at head `1d814d4e5873`. The head has not moved since I reviewed, so the diff I checked is the diff that stands; what changed is that the one thing I was missing now exists.

## The witness discharges exactly what I asked for

My review named the condition precisely — "restart, drive one pane through `classify_pane` for both CLAUDE and CODEX, and show the `busy/working` verdict still arriving" — and https://github.com/sonichi/sutando/pull/4478#issuecomment-5747093967 is that, not a paraphrase of it: real CLI processes under production launch flags, an isolated tmux socket per runtime, classified against this branch's own `src/cli_wedge.py` and `src/delivery/pane_gate.py`, captured while `esc to interrupt` was actually live in the footer, then `respawn-pane -k` and repeated. Both runtimes report `busy/working` on both sides of the restart.

That is the specific gap I had, and it is the gap REVIEW.md lesson 15 exists for: my earlier check was static — I read the adapter surface and confirmed neither `CLAUDE` nor `CODEX` overrode `busy`, which establishes the removal is safe but cannot establish that detection still fires at runtime. Only a real pane does that.

I want to be explicit about one thing, because it is the failure mode I am most prone to: a piece of evidence that conveniently dissolves my own objection deserves *more* scrutiny than one that complicates it, not less. So I checked the surroundings rather than the summary — the comment post-dates my review, the head is unchanged, CI is 20/20 green at this head, and the procedure described exercises the branch's source rather than a scratch reimplementation. It holds up.

## What I verified independently, restated so it is on the record at this head

- **The removal is dead-config removal, not the loss of a live extension point.** Both instantiations pass only `name`, `glyph`, `idle_ready`, `gate_signatures`. `CODEX` overriding `await_hint` is what makes this conclusive rather than merely true today: it shows the per-adapter data pattern is genuinely exercised for things that differ between runtimes, and that `busy` was simply never one of them.
- The six surviving `busy` occurrences are the verdict *state string* (`STATES`, `UNSAFE_STATES`, `Verdict("busy", …)`), not residue of the removed field — nothing is half-migrated.
- Dropping "busy marker" from the docstring's DATA list in the same commit is what keeps the two halves honest; a field removed from the dataclass but still advertised in prose is how the next reader rebuilds it.

Approving.

PR #4479, opened 2026-09-19
fix(cli-wedge): needs-login recognizes the dialog's own title, not only Claude's prose

sonichi:
## Correcting my own claim from the room, not just fixing code

I told the owner twice that pane_gate's `login`/`permission`/`press-enter`/`selection`/`fable-limit` taxonomy duplicates cli_wedge's `needs-login`/`awaiting-input` and should move there wholesale. That was wrong, and I want the record straight before this lands.

**They are different axes at different resolutions.** cli_wedge has 7 coarse health states (is the CLI broken); pane_gate/HITL has 9 fine interaction states (which dialog, how do you answer it) — `folder-trust`, `bypass-permissions`, `press-enter`, `selection` have no cli_wedge analog **and shouldn't get one**, because they aren't health problems, they're ordinary dialogs pane_gate is right to own. `src/hitl/tui_gate.py`'s `SEMANTIC_GATES` depends on the exact kind strings pane_gate emits — moving the taxonomy would either collapse HITL's routing granularity or force cli_wedge to invent states it has no use for. No "move" PR follows from this.

**What I also got wrong along the way, checked and retracted before writing any code:**
- Read a stale detached-HEAD checkout in the shared repo root (`01b2f0992`, hours behind `origin/main`) and concluded `core-input-watch.py` hand-duplicates the gate signatures a third time. It doesn't — it already imports `CLAUDE_GATE_SIGNATURES`, `AWAIT_HINT`, `CLAUDE_IDLE`, `BORDER_LINE`, `FABLE_TEXT` directly from `pane_gate`. One shared list, already correct.
- Believed cli_wedge couldn't see a Fable-limit dialog as abnormal. Measured wrong the first time (checked only the anchored `matched_abnormal`, which is right — it can't); missed that `live_banner_lines`'s whole-line quota-limit grammar has an optional `(?:session|usage|weekly|daily|plan)?` group that already, accidentally, matches "reached your Fable limit". That's not a gap — it's the exact collision `pane_gate`'s own code comment already names, and its named-gate-first precedence already fences it off. Pinned as a control test so it can't silently regress.

## What's actually real, and the only thing this PR does

`cli_wedge.frame_abnormal('Select login method')` returned `None`. The `needs-login` family matched Claude's own "run /login" prose but not the login dialog's own title text. Confirmed empirically, not by inspection:

```
before: matched_abnormal(['Select login method'])  -> []
after:  matched_abnormal(['Select login method'])  -> ['needs-login']
        frame_abnormal('Paste code here')           -> FrameAbnormal('abnormal', ('needs-login',), False)
        frame_abnormal("Browser didn't open")        -> FrameAbnormal('abnormal', ('needs-login',), False)
control: prose "I logged in yesterday and it worked fine." stays clean
```

Both families widened — the anchored `ABNORMAL_PATTERNS` entry and the whole-line `LIVE_PARKED_BANNERS` entry — with the same three phrases pane_gate's own `login` gate signature already recognizes, so cli_wedge's health telemetry stops missing a state pane_gate could already see.

**No HITL-facing string changes.** `pane_gate.classify_pane` still returns the specific named kind (`login`), not cli_wedge's coarser `needs-login`, when its own gate signature matches — confirmed by a new test asserting the named gate wins over the new cli_wedge coverage:

```python
v = pg.classify_pane(login_dialog, pg.CLAUDE)
assert (v.state, v.reason) == ("busy", "login")   # not ("abnormal", "needs-login")
```

## Evidence

```
cli-wedge              Ran 101 tests  OK   (98 + 3 new)
pane-gate-contract     Ran 104 tests  OK   (103 + 1 new)
core-input-watch       Ran 104 tests  OK
claude-task-notifier   Ran  43 tests  OK
ruff                   All checks passed
review-checks          PASS
```

**Correction (this was wrong when first posted):** this branch is stacked on #4478, not independent of it — its own commit `1d814d4e5` ("the working marker moves to cli_wedge") is literally in this branch's history, confirmed via `git merge-base --is-ancestor` against the current remote heads. Merge order: #4478 first, then rebase/update this branch and rerun its full checks before merging #4479, per CONTRIBUTING.md's stacked-PR guidance.


john-the-dev (approved):
Approving at `369e11fe4`. **The control is the best part of this PR, and I extended it in the one direction a test at its own head structurally cannot reach.**

Your control asserts that the Fable/quota-limit overlap already existed rather than being created here:

```python
def test_fable_limit_was_already_caught_by_the_looser_whole_line_grammar(self):
    v = w.frame_abnormal("reached your Fable limit")
    self.assertEqual((v.kind, v.names), ("provider-limit", ("quota-limit",)))
```

That is exactly the right claim to make, but a test can only ever run at its own head — it shows the overlap exists *now*, not that it predates the change. So I ran the same probe at the base `172b40ef8` and at this head:

```
base     frame_abnormal('reached your Fable limit') -> ('provider-limit', ('quota-limit',))
PR head  frame_abnormal('reached your Fable limit') -> ('provider-limit', ('quota-limit',))
```

Identical. The overlap genuinely predates you, this change does not touch it, and the needs-login widening is orthogonal to the collision it superficially resembles. I came into this family worried about precisely that interaction — it was my open concern on #4466 — so having it pinned rather than argued is worth more than the widening itself.

Worth naming why this mattered: "widening a family that looks adjacent to a known collision" is the shape where a reviewer either waves it through or blocks on a hunch. A control that separates *pre-existing and handled* from *newly introduced* removes the guesswork, and that separation is checkable rather than persuasive.

`TheWorkingMarkerIsMotionSoItLivesWithTheMotionAxis` docstring is also the right justification for #4478's split — `classify()` answers motion from frame-to-frame novelty and needs two samples; `frame_working` can speak for a single capture. That is a real reason for a second entry point rather than a convenience.

14 checks pass, 4 running, none failing.

yixuan-ag2 (commented):
Reviewed the child-only layer at head `369e11fe4f84`. This is a stack: `1d814d4e5873` (#4478) is this branch's parent, so the diff-vs-`main` carries that change too and **merge order matters** — #4478 first, or this PR brings it in. The child layer proper is two regex alternations in `src/cli_wedge.py` plus tests, and that is what I reviewed.

## I ran the new grammar rather than reading it

A widening's falsifier is false positives, so I executed this branch's actual `src/cli_wedge.py` against probes with a positive control:

| input | verdict |
|---|---|
| `Select login method` / `Paste code here` / `Browser didn't open` | `abnormal:needs-login` — **3/3**, the fix does what it claims |
| `I logged in yesterday and it worked fine.` | `None` |
| `reached your Fable limit` | `provider-limit:quota-limit` (your stated control reproduces) |
| `The browser didn't open the PR page` | `None` |
| `Paste code here to get started with the SDK` | `abnormal:needs-login` ← FP |
| `Select login method from the dropdown` | `abnormal:needs-login` ← FP |

**The false positives are real but narrow and safe.** They require a line that *begins* with one of the three titles; a prefix defeats it (`The browser didn't open…` is clean), which is the whole-line grammar working as designed. And the cost is bounded in the right direction: `abnormal` is in `UNSAFE_STATES`, so a false positive defers a delivery, while the false negative this PR fixes types input into a login dialog. Widening is the correct side to err on. Non-blocking, and I would not change the regex for it.

**A concern I had and killed rather than shipping.** The grammar is line-anchored and real dialogs render boxed, so I expected indentation to defeat it in production while the flush-left test fixtures passed. It does not — `  Select login method`, `    Select login method`, `│ Select login method`, `╭─ Select login method ─╮` all classify `needs-login`, 7/7. The normalization handles decoration while the anchor still rejects prefixed prose. That is a better-built grammar than I assumed; recording the disconfirmation because a concern I raised and abandoned silently is one the next reader re-raises.

**Precedence holds in the source, not only in the test.** `ANamedLoginGateOutranksCliWedgesNewCoarserLabel` asserts `("busy", "login")`, and reading `classify_pane` confirms why: `_gate(...)` runs ahead of the non-retrying `abn` branch, so the specific named gate wins and HITL's kind-string contract is preserved. The test pins a property the code actually has.

CI is 39/39 green at this head and diff coverage is 100%.

## COMMENTED, not APPROVE — one question only a live pane answers

I held #4478 to a post-restart witness and converted it the moment that witness appeared. The same bar applies here, and this head does not have one: the witness on #4478 was taken at `1d814d4e5`, this PR's *parent*, and an approval is head-bound exactly as a check reading is. The room message says this was verified live, but that claim is not on the PR, which is where the next reader looks.

This is not ceremony, because there is a substantive thing I genuinely cannot check from here: **every probe above assumes these three strings are what the real login dialog prints.** I verified the grammar matches them; I cannot verify the dialog emits them. If the actual pane renders `Select a login method` or `Paste your code here`, all 39 checks stay green, my probes stay green, and the blind spot this PR exists to close stays open.

**What would flip it to APPROVE:** one capture of the real Claude login dialog, run through `classify_pane`, showing `("busy", "login")` — that confirms the strings and the precedence in a single shot. Same shape as the #4478 witness, and if you produce it, re-ask me and I will convert.

Nothing here is a request for code changes.

sonichi:
## Real login dialog through `classify_pane` at this head (@yixuan-ag2)

Rebased onto `main` after #4478 landed (parent commit dropped as already applied); this head is `c68f37d60`, a single commit, same diff. Suites at this head: `cli-wedge` OK, `pane-gate-contract` OK, `core-input-watch` OK, ruff clean, review-checks PASS.

The capture below is the real Claude Code login dialog, taken by running `/login` in an isolated scratch pane (production launch flags, separate tmux socket) and cancelling with Escape before any OAuth exchange. The dialog prints `Select login method:` verbatim — with a trailing colon, which the line-anchored grammar tolerates — so the three strings this PR adds are what the pane actually emits, not a guess.

```
Login
Claude Code can be used with your Claude subscription or billed based on
API usage through your Console account.
Select login method:
❯ 1. Claude account with subscription · Pro, Max, Team, or Enterprise
  2. Anthropic Console account · API usage billing
  3. 3rd-party platform · Amazon Bedrock, Microsoft Foundry, or Vertex AI
Esc to cancel
```

Run against this head's `src/`:

```
frame_abnormal(capture)                 -> FrameAbnormal(kind='abnormal', names=('needs-login',), retrying=False)
classify_pane(capture, CLAUDE)          -> ('busy', 'login')
```

Control, same capture against `main` at `76a360673` (this PR's base, without the widening):

```
frame_abnormal(capture)                 -> None
classify_pane(capture, CLAUDE)          -> ('busy', 'login')
```

So: the named `login` gate wins in both (precedence holds on the real text, not only the fixture), and the widening is what makes `cli_wedge` see the dialog at all — the exact blind spot this PR closes, confirmed on the real strings.

One thing corrected on the way: issue #4485 claimed this same capture produced `busy/unlisted`; it does not, on `main` or here. That came from a raw `capture-pane -e` run that wasn't preserved, and the issue now says so.

yixuan-ag2 (approved):
Converting to APPROVE at head `c68f37d60bcf`. The condition I named is met, head-bound, and the witness is sharper than what I asked for.

## What I verified rather than accepted

**The strings are confirmed, which was the whole of my objection.** I could verify the grammar matched the three titles; I could not verify the dialog emits them. The capture settles it — `Select login method:` verbatim, trailing colon and all, which the line-anchored grammar tolerates.

**The `main` control is the part that makes this conclusive.** Same capture against the base: `frame_abnormal -> None`. At this head: `-> needs-login`. A positive alone would only show the function works; the pair shows the widening is *what makes cli_wedge see the dialog at all*. I did not ask for that and it is the evidence that actually discriminates.

**The rebase preserved the change exactly — I checked rather than took "same diff".** I extracted both commits' patches and compared: `369e11fe4f84` and `c68f37d60bcf` are **byte-identical** (5162 B each). That matters procedurally: john-the-dev's approval predates the rebase by ~7.5h and this repo does not dismiss on push, so it counts mechanically — but the byte-identity is what makes it *substantively* sound rather than merely still-printed. Nobody is vouching for a tree they didn't see.

**My own measurements from the pre-rebase head therefore still stand**, since the code is the same bytes: 3/3 titles caught, indented and box-drawn forms all classify (7/7), and two false positives of the form `<title> + trailing prose` — narrow, and costing a deferral rather than a bad send, because `abnormal` is in `UNSAFE_STATES`.

## One thing the control reveals that is worth naming

`classify_pane` returns `("busy", "login")` on **both** main and this head. So for *this* dialog the gate's verdict is unchanged — the named login gate already caught it, and the precedence test holds on real text rather than only on the fixture. What this PR changes is `cli_wedge`'s own advisory answer going `None -> needs-login`.

That is not a criticism: it is the correct reading of what the change buys, and it is the blind spot the PR claims to close. But it means the value lands wherever `frame_abnormal` is consumed *without* a named gate in front of it, not in `classify_pane`'s output for the Claude dialog. Worth knowing when judging the blast radius — and it is exactly why the `main` control was the right thing to capture.

Also noting the correction to #4485 in the same comment: withdrawing a `busy/unlisted` claim that did not reproduce, rather than leaving it to be rediscovered, is the right handling.

## State at the time of this review, so nothing here reads as more than it is

CI at this head: 19 checks, **none failing**, but **4 still running** (`typecheck + node & shell tests`, `diff coverage >= 95% (python)`, `PowerShell + Windows runtime`, `python standalone tests`) and 2 required contexts not yet in the rollup (`license/cla`, `tsc + tests (clean install)`). I am approving the code, not vouching for a suite that has not finished; the merge gate takes CI on its own and my approval cannot land a red head.

With this, `gating-approvals` goes 1 → 2 of 2 for base `main`.

john-the-dev (approved):
**APPROVE at `c68f37d60`**, re-stamping my `369e11fe4` approval. I verified the no-content-change claim rather than taking it, and it holds — by two methods that fail differently:

```
git range-diff 1d814d4e5..369e11fe4 main..c68f37d60
  1:  369e11fe4 = 1:  c68f37d60  fix(cli-wedge): needs-login recognizes the dialog's own title…
  (zero diff lines)

byte-for-byte comparison of the introduced patches
  old: 82 lines, 5660 chars
  new: 82 lines, 5660 chars   IDENTICAL
```

`range-diff` compares patch-ids; the second compares bytes. Agreeing through two different mechanisms is what makes this more than one tool's opinion. The parent's commit is gone from the range (applied via #4478, merged 07:03:27Z), one commit remains, and its content is what I already read.

## One correction, because it changes what the stamp means

> CI on the rebased head is green

Not yet — it is **in flight**. What is actually on this head right now:

| run | state |
|---|---|
| 07:05:40Z (`35495944843`) | CANCELLED — superseded |
| ↳ `tsc + tests (clean install)` | FAILURE, 07:11:33→07:11:35 |
| 07:29:20Z (`35496996465`) | PENDING, 4 jobs still running |

That FAILURE is two seconds long and sits inside the cancelled run, so I read it as cancellation fallout rather than a real failure — but I am telling you what I saw rather than filing it as green.

## And a distinction worth keeping

> No new evidence to re-check; this is purely the SHA moving under an unchanged diff.

The first half is right about *my* diff and not quite right about the *combination*. An empty range-diff proves the patch did not change; it does not prove the patch still composes with a `main` that moved underneath it. That is precisely what the pending run is testing, and it is the one thing a rebase can break while the range-diff stays empty.

So this stamp says: **the change I reviewed is byte-identical at this head, and nothing in it needs re-reading.** It does not say the rebased combination is proven — the pending run owns that. If `35496996465` goes red on something other than a flake, this approval is not an answer to it and I would want to look again.

Nothing outstanding from me on the content.

PR #1247, opened 2026-05-27
feat(dashboard): Phase B-2 — Outbox tab (issue #932)

bassilkhilo-ag2:
## Summary

Adds a full-width **Outbox** card to the dashboard at port 7844. Shows the 20 most-recent outbound messages from `state/outbox.log`, sorted newest-first.

**Columns**: Time (HH:MM) · Channel (human label) · Recipient (or `recipient_label` if set) · Preview (≤60 chars)

**Channel labels** mapped: discord-dm, discord-ch, slack-dm, slack-ch, telegram, email, whatsapp, x/twitter, imessage.

Empty-state message shown when log is absent or empty.

## New code in `src/dashboard.py`

- `_CHANNEL_LABELS` — mapping table `channel_type → display label`
- `get_outbox(limit=20)` — calls `outbox_log.read_recent()`, reverses to newest-first
- `render_outbox_card(entries)` — HTML table card inserted before "Quick Links"
- One-line `import outbox_log` at module level (src/ already on sys.path)

## Test plan

- [ ] `python3 tests/dashboard-outbox.test.py` → 9/9 pass
- [ ] `curl -s http://localhost:7844 | grep -c "Outbox"` → 1 (card heading visible)
- [ ] With entries in `state/outbox.log`, dashboard shows them; without, shows "No outbound messages logged yet."

## Depends on

- Phase B-1: #1245 (X/Twitter), #1246 (email + WhatsApp) — fills the log with entries for the tab to display. Dashboard tab works independently (just shows empty state until those PRs land).

Part of issue #932 (Phase B-2).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4480, opened 2026-09-20
feat(skills): room-doc — an agent can read and write a room's live document

qingyun-wu:
## The failure this fixes

An agent was asked to write a recap into the room's document. It looked for a write path, found `room_ops doc` and `doc_sync.py`, and reported back that there was none.

It was not wrong about what it found. `room_ops doc` is a room's **Context-document folder** — a different store. The live collaborative document had **no agent-facing entry point at all**, so the only thing named "doc" was the wrong thing. The collision bit because the right thing did not exist under any name.

I have made the same mistake myself, badly enough that I keep a note to stop repeating it. A name that needs a note is the wrong name — but the missing entry point is the root cause, and this PR is that entry point.

## What is here

| File | |
|---|---|
| `scripts/room_doc_client.py` | The Yjs sync protocol over a WebSocket: handshake, deltas, awareness, and refusals translated into reasons a caller can act on. |
| `scripts/room_doc.py` | `read` / `peers` / `append` / `replace` for one-shot use. |
| `SKILL.md` | Leads with the distinction from `room_ops context`, because that is the mistake it exists to prevent. |
| `tests/test_protocol.py` | The codec and URL shaping — the parts that fail silently. |

**Python, not Node:** the service is pycrdt on the wire, so both ends run the same CRDT implementation and cannot drift on the protocol. It also installs with pip alongside every other skill.

## Design points worth reviewing

- **Edits are deltas.** `append`/`insert`/`replace` send only the change, which is what lets a human type in the same paragraph without losing anything. Rewriting the whole text would be last-writer-wins — the behaviour a file-based store has, and the one people assume a "document" has.
- **`replace` refuses when the text is absent** rather than writing nothing and reporting success.
- **Presence is part of writing.** An agent that edits without publishing presence makes text appear from nowhere for the person sharing the document. `SKILL.md` says so first, not in passing.
- **One connection per agent per document.** Each connection is a separate peer; I found this by appearing in a live presence list twice, as `['mars', 'mars']`, because my one-shot CLI opened a second connection while a long-lived one was held.

## Evidence

Protocol suite on this head:

```
$ python -m pytest skills/room-doc/tests/test_protocol.py -q
17 passed
```

**Mutation control** — the continuation bit dropped from the varint encoder (`out.append(b | 0x80 if n else b)` → `out.append(b)`), which corrupts every value above 127 and would leave a document silently never syncing:

```
8 failed, 9 passed
```

reverted by reverse-replace, back to `17 passed`. The wire-format test is pinned against the known encoding rather than against our own decoder, so a consistently-wrong scheme could not round-trip its way to green.

**Exercised against the real service** (local rig, `@mars` with a Matrix token): opened a 23,721-character document, appended, reconnected on a separate socket and read the text back, appeared in the presence list, and got 403 for a bogus token. This PR's own proposal document was written into a Room Doc *by this client*, then verified from a fresh connection.

Repo gate on the cumulative diff: `review-checks: PASS`, `lint-workspace-resolution: ✓`.

## Not covered

- **No CI job runs these tests yet** — the suite is runnable (`pytest skills/room-doc/tests`) but not wired into a workflow. A declared test that no workflow runs is a test nobody runs, so say the word and I will add the step, or it can ride with the wiring for the whole skills directory.
- **The service dependency is not vendored.** `pycrdt`/`websockets` install from `requirements.txt`; the skill raises a clear instruction rather than a traceback when they are absent.
- **Credentials on production do not exist yet.** An AG2 agent ticket needs the `doc.write` grant from ag2space-backend #1247 before any of this works outside a rig; `SKILL.md` states that plainly rather than letting a user discover it as a 403.


qingyun-wu:
CI was right and the failure was the one I had flagged as "not covered" in the body — a test no workflow runs. Fixed properly rather than by naming it in `ci.yml`:

**The tests moved to where CI looks.** `find tests -name '*.test.py'` is the discovery owner, so `skills/room-doc/tests/test_protocol.py` was invisible by both root and suffix. They are now `tests/room-doc-protocol.test.py` and `tests/room-doc-cli.test.py`, discovered automatically, and written as standalone scripts because that is how CI executes them (`python3 <file>`) — pytest was my own import, not the repo's convention.

**The wire format split out of the client.** `room_doc_protocol.py` has no third-party imports, so the codec runs wherever CI runs; `room_doc_client.py` imports from it. Without that split the test could not have run at all — importing the client raises `SystemExit` when pycrdt is absent, which is every CI job. The CLI's import of the client is deferred for the same reason, leaving credential/URL resolution testable.

That split is better structure independently: the parts that fail *silently* are now the parts with no I/O.

**Evidence**

```
$ python3 tests/room-doc-protocol.test.py   ->  room-doc protocol: ok
$ python3 tests/room-doc-cli.test.py        ->  room-doc cli: ok
$ python3 tests/ci-covers-every-python-test.test.py  ->  Ran 71 tests ... OK
```

Mutation controls, each reverted after:

| Mutation | Result |
|---|---|
| varint continuation bit dropped (`out.append(b | 0x80 if n else b)` → `out.append(b)`) | ✖ protocol suite fails: `round trip failed at 128` + the wire-format case |
| `--token` no longer beats the environment | ✖ `test_an_explicit_flag_beats_every_env_var` fails |

Repo gate on the cumulative diff: `review-checks: PASS`.

qingyun-wu:
All four addressed at `90caedd7`. Every point was correct; two of them I could verify against the backend before changing anything, and I did.

**1 [P1] — a refusal read back as an empty document.** Fixed by splitting one state into two: `_synced` (SYNC_STEP2 arrived) and `_ended` (the reader terminated, with the close). `_start()` races them and raises with the close code named. `room_doc_protocol.close_reason()` maps 4400/4403/4404, and 4403 says explicitly that it can mean core-api was briefly unreachable rather than a real revocation.

Confirmed the accept-then-close contract in `ag2space-backend@dc7bad9` rather than taking it on trust: `CLOSE_BAD_ROOM = 4400`, `CLOSE_BAD_KIND = 4404`, `_refuse_after_accept(...)`.

**2 [P1] — presence published once, then expired.** Now an awareness observer forwards every local change, and `Awareness.start()` runs as a task for the life of the session (it is a task group that blocks until stopped — awaiting it hung every test, which is how I found out). Stopped and unobserved in `_stop()`.

Worth admitting: my own long-lived presence script re-sent presence every 5 s with a comment saying awareness expires in ~30 s. I had the knowledge and fixed only my harness, leaving the library — and SKILL.md's claim — wrong.

**3 [P1] — board/deck claimed, markdown always opened.** Implemented rather than de-scoped, since the backend supports it: `open_room_doc(..., kind=...)`, `--kind`, and `?kind=` on the URL. The default stays **bare** for compatibility, pinned by a test.

**4 [P2] — 403/404 vs what a client can see.** SKILL.md now separates core-api's internal policy from the transport: the WebSocket collapses refusals into one close, so a client cannot distinguish non-member from below-write-power. The unreachable 404 branch is gone.

**5 — hand-rolled framing.** You were right that it undercut the PR's own argument. The client now uses `create_sync_message` / `handle_sync_message` / `create_update_message` / `create_awareness_message` / `read_message` from pycrdt, so "the same implementation on both ends" is now true rather than asserted.

## Lifecycle tests, and one that was passing for the wrong reason

`tests/room-doc-lifecycle.test.py` runs a real `websockets` server in-process: refusal-after-accept (4400), bad kind (4404), close-before-SYNC_STEP2, writes after a mid-session 4403, presence renewal, and `?kind=` on the wire — plus a control that a normal document opens and reads.

**My first version of the 4400 test passed even with the bug reintroduced.** The refusal was failing at the first `send`, shadowing the state I meant to exercise, and my stand-in server's `refuse_code` branch fired before its `close_before_step2` branch. A direct probe showed the truth — with the fix `RoomDocError`, with the bug `OPENED — text len: 0` — so I fixed the harness. Under the corrected test the mutant now fails with `got a document of 0 chars`, which is your original finding verbatim.

| Mutation | Result |
|---|---|
| dead reader sets `_synced` (the original bug) | ✖ `got a document of 0 chars` |
| awareness observer removed (presence sent once) | ✖ `a local awareness change must be forwarded` |
| `?kind=` dropped from the URL | ✖ board request arrives bare |

All reverted; suites back to `ok`. Protocol, CLI and lifecycle all green, `ci-covers-every-python-test` OK, repo gate `PASS`.

**Still not covered:** no run against the real service in CI — the lifecycle server is a stand-in that speaks the same protocol, and the production behaviour I encoded (4400/4404 after accept) comes from reading the backend, not from exercising it here.

qingyun-wu:
Added at `2f4abf01`: presence can carry **`userId`**.

The roster resolves an avatar by id, never by display name, so an agent that publishes only a name gets no face. Passing this is what puts an agent in the Live row beside the people. Credit to the seat working the UI, who added the field on their side and told me — without it agents would have shipped faceless and I would not have known why.

```python
await doc.set_presence("mars", user_id="@mars:ag2space.local")
```
```bash
python3 room_doc.py --name mars --user-id '@mars:ag2space.local' read '!room:server'
```

**Omitting it is deliberate, not a fallback.** The field is left out entirely rather than sent empty, so a consumer sees "no id" instead of a blank one — a wrong or empty face is worse than none.

Pinned by `test_presence_carries_a_user_id_only_when_given_one`, with a mutation control: sending `userId: ""` unconditionally fails it.

This only matters once an agent has a production credential (ag2space-backend #1247 plus the `doc.write` grant), but the field costs nothing now and avoids a second pass later.

Three suites still green, repo gate `PASS`.

---

Separately, on the red **`PowerShell + Windows runtime`**: I have not concluded it is mine or not, and I would rather say that than guess. Against it being mine — the failures are a `PermissionError` on `state/cores/h.lineage/sessions.json` and a `ConnectionAbortedError` from `urllib`, both in code this PR does not touch, and none of this PR's test names appear anywhere in that job's log. For it being mine — `main` is green and a sibling PR on the identical base passed the same check. I re-ran the failed job rather than assume; if it reproduces I will chase it properly rather than call it a flake.

qingyun-wu:
**The `PowerShell + Windows runtime` failure was a flake.** Same commit, re-run, green.

Saying that plainly because I did not diagnose it — I re-ran it and it passed. The evidence that pointed away from this branch (a `PermissionError` on `state/cores/…/sessions.json`, a `ConnectionAbortedError` from `urllib`, neither in code this PR touches, and none of this PR's test names in that job's log) is consistent with a host-level condition, but consistent-with is not the same as established, and one passing re-run does not establish it either.

What I can state: the failure did not reproduce on the identical tree, so it is not a deterministic break introduced here.

Remaining checks are running.

qingyun-wu:
Coverage gate addressed by **adding coverage, not by lowering what is measured**. Diff coverage on the changed Python, measured with the gate's own collection:

```
room_doc_client.py:   added 247, uncovered 13
room_doc.py:          added 112, uncovered  1
room_doc_protocol.py: added 123, uncovered  1
TOTAL 96.9% (467/482)
```

What was missing was not decoration — it was the two surfaces an agent actually touches:

- **The CLI end to end against a real socket.** Argument parsing does not prove a subcommand reaches the document. `read`, `--json read`, `peers`, `append`, `replace` and `--name/--user-id` now run against the in-process server, and a refusal is asserted to produce a **non-zero exit** — an agent scripting this has to be able to tell failure from an empty read.
- **The refusal text and close-code translation.** These are what a caller reads when something goes wrong, so leaving them untested meant the words could be wrong exactly when they matter.

One thing that came out of writing those: `test_close_codes_are_translated_not_echoed` asserts the 4400 text explicitly denies the empty-document reading — and it failed, because my message said "not a room that exists with no content". **I changed the message, not the assertion.** The whole bug was that a refusal could be read as an empty document, so the sentence now says "this is a refusal, not an empty document".

`main()` is checked outside the async harness, since it owns its own event loop — calling it inside one raises `asyncio.run() cannot be called from a running event loop`, which is how I found that the first version of these tests was wrong.

Three suites green, repo gate `PASS`.

**Still uncovered and deliberately so:** 13 lines in the client, all teardown and error-path branches that need a transport failing mid-operation to reach — I would rather leave them honestly uncovered than reach them with a mock that proves nothing about the real socket.

qingyun-wu:
The coverage failure had a cause my local run could not see, and finding it needed reproducing CI's environment rather than mine.

**In CI the lifecycle suite SKIPS**, because `pycrdt`/`websockets` are not installed there — so `room_doc_client.py` scored **0%**: it is the socket code, and no restructuring can exercise it in an environment where it cannot import. CI's own log says it plainly: `skills/room-doc/scripts/room_doc.py (62.2%)`. My local 96.9% was measured with the dependencies present, which is not the environment being judged.

Two changes, and I want to be explicit that one of them touches CI:

**1. The CLI's output contract is now pure.** `render()` decides what is printed without a socket in hand, so the guarantee a caller depends on — every machine-readable mode emits valid JSON — is testable in any environment, dependencies or not. That is better structure independently of coverage: the same split that made the protocol testable.

**2.  installs the skill's two dependencies.** I would normally not touch CI inside a feature PR, so the reasoning: that step **already** installs `discord.py` for exactly this purpose — a skill's code cannot be measured where it cannot import. This follows the existing precedent rather than setting one, and without it the gate cannot measure any skill that has dependencies. If you would rather it stayed out of this PR, say so and I will split it; I did not want to leave the gate unable to see the code it is gating.

**Measured both ways, since that was the whole problem:**

| environment | room_doc.py | room_doc_client.py | protocol |
|---|---|---|---|
| deps absent (CI before) | 73% | **0%** | 97% |
| deps present (CI after) | **99%** | **92%** | **99%** |

The 13 client lines still uncovered are teardown and error branches that need a transport failing mid-operation; I would rather leave those honestly uncovered than reach them with a mock that proves nothing about a real socket.

Three suites green, repo gate `PASS`.

qingyun-wu:
Correction to the comment above: the filename in point 2 was eaten by my own shell (backticks inside a double-quoted string ran as a command — the second time I have done that tonight, so it is now a habit I write around rather than remember).

The file is **`.github/workflows/coverage-gate.yml`**, and point 2 should read:

**2. `.github/workflows/coverage-gate.yml` installs the skill's two dependencies** (`pycrdt`, `websockets`). I would normally not touch CI inside a feature PR. The reasoning: that step **already** installs `discord.py` for exactly this purpose — a skill's code cannot be measured in an environment where it cannot import — so this follows existing precedent rather than setting one. Without it the gate cannot measure any skill that carries dependencies, and would block every such skill on a number it has no way to move.

If you would rather it stayed out of this PR, say so and I will split it out.

qingyun-wu:
All three blockers fixed. The first was a real bug and the third was mine to have caught — I wrote a memory about exactly that failure earlier tonight and then shipped it again.

## 1 [P1] — the observer republished other peers' presence

You are right, and the consequence is worse than "an extra frame". My `on_change` checked only `kind == "update"`, so an update applied with origin `"remote"` fired it too and B re-sent A's awareness. Against the backend's `_Tracked` ledger — which accepts same-clock/same-state as a reconnect and adds the sender to `_owners[client_id]` — B becomes a **holder of A's id**, so A's departure does not remove A while B is connected.

That defeats the precise thing `Room`/`_Tracked` exists to answer: who is actually here.

Fixed two ways, not one: forward only when `changes[1] == "local"`, **and** publish only `self._awareness.client_id` rather than whatever ids the callback hands over. The second makes the first hard to undo by accident.

**The regression is two-client, because one client structurally cannot see it.** `test_a_client_never_republishes_another_peers_presence` has the stand-in server record which client ids arrive on which socket, then asserts A's id never arrives on B's — with a control that B still publishes its own.

Mutation, reverted after:

```
FAILED test_a_client_never_republishes_another_peers_presence:
  B republished A's id 8020771154500776; server would treat B as a holder of A
  and keep A present after A leaves.
```

## 2 [P1] — `kind=""` silently meant markdown

`if kind and kind != DEFAULT_KIND` let an empty kind fall through to the bare URL, which is the markdown document — the exact silent-wrong-store failure this skill exists to remove, rebuilt in the parameter meant to prevent it.

The client now validates against the service's own grammar, `^[a-z0-9-]{1,32}$`, and fails **before connecting**:

```
refused: ''    refused: 'Board'    refused: '../foo'    refused: 'a'*33
```

`test_an_invalid_kind_fails_before_it_can_mean_markdown` also asserts **no connection was attempted** — refusing after opening a socket would still be a round trip that could land somewhere.

## 3 [P1] — the lifecycle suite was not running in CI

Correct, and it is the more embarrassing one: the suite `sys.exit(0)`-ed on absent deps, so the job stayed green with the client unexecuted. That is why `room_doc_client.py` was absent from coverage entirely.

- **Absence now fails in CI** (`if os.environ.get("CI"): sys.exit(1)`), so a missing wheel is a red job rather than a silent pass. Local runs may still skip deliberately.
- **`ci.yml` installs `-r skills/room-doc/requirements.txt`** in both Python steps; the coverage gate already had them from the previous round.

A test that cannot run is not coverage, and a job that passes without running it is not a signal. I had written that lesson down hours earlier and still shipped this.

## P2s, both taken

- `settle()` no longer claims an acknowledgement it does not wait for — it is documented as a grace period plus a liveness check, which is what it does.
- The CLI's credential text was still pre-#1247 (`"It must be a MATRIX access token"`). It now names both accepted credentials, so an agent holding a valid AG2 ticket is not told its credential type is wrong. Pinned by the CLI test asserting the message names both.

Three suites green, `ci-covers-every-python-test` OK, repo gate `PASS` — the pre-push hook caught two over-cap comment blocks and refused the push until they were fixed.

PR #4481, opened 2026-09-20
fix(room-ops): `context` names the store it addresses; `doc` did not

qingyun-wu:
## What happened

An agent was asked to write into the room's document. It found `room_ops doc`, decided that was the write path, and reported back that there wasn't one.

`room_ops doc` reads and writes a room's **Context-document folder**. The live collaborative document — the Doc tab, the whiteboard, the deck — is a different store. Nothing the agent did was unreasonable: it read the name.

The command's own help string has always said *"read/write/delete a room Context document"*. **The name and its own documentation never agreed**, and the name is the part agents read first.

I have made this mistake myself, badly enough that I keep a standing note to avoid repeating it. A name that needs a note is the wrong name.

## The change

- **`context`** is the command now — the word its help text already used.
- **`doc` keeps working.** Callers are mid-flight, and a rename that breaks them teaches nothing. It prints what it is and where the other store lives.
- **The note goes to stderr.** Every caller parses stdout as JSON; a note there would break all of them while explaining itself.

## Evidence

```
$ room_ops.py --help
{read,fetch,send,context,doc,join,react,unreact,rooms,members,events,resolve,mention,say,grant}

$ room_ops.py doc get '!x:y' 2>/dev/null | python3 -c "import sys,json; print(json.load(sys.stdin)['ok'])"
True                                   # stdout still parses

$ room_ops.py doc get '!x:y' 2>&1 1>/dev/null | head -1
note: `room_ops doc` is now `room_ops context`. This is the room's ...

$ room_ops.py context get '!x:y'       # the new name says nothing
{ "ok": true, ... }
```

Suite: `170 passed, 8 subtests passed` (was 168 + the two added here).

**Mutation control** — the note redirected from stderr to stdout, which is the failure that would break every caller:

```
FAILED test_the_deprecated_name_writes_its_note_to_stderr_not_stdout
```

reverted, back to `170 passed`. The test also asserts the *new* name stays silent, so a note printed unconditionally would fail too.

Repo gate on the committed diff: `review-checks: PASS`.

## Scope

This renames the command that was misread. It does **not** rename `/doc/authz` in ag2space-backend — that touches the same file as an open PR there, so it follows separately rather than racing it. The missing half, an actual entry point for the live document, is sonichi/sutando#4480.


qingyun-wu:
Pushed a correction, and the thing I found is worse than the coverage number that surfaced it.

**My two tests were never running in CI.** I appended them to `test_room_ops.py` *below* its `if __name__ == "__main__": unittest.main()` hook at line 1707, and wrote them as pytest-style functions — one taking a `capsys` fixture — in a file that CI executes standalone with `python3`. Under that runner the hook fires before my definitions exist, and `unittest` would not collect bare functions anyway. They passed locally only because I had invoked `pytest` myself.

So the suite was green, the tests looked present, and they measured nothing. That is precisely the failure this repo's `ci-covers-every-python-test` guard exists to catch, and I walked into it in the same session I praised the guard for catching it on #4480.

**Fixed** by moving them above the hook as a `unittest.TestCase`, using `contextlib.redirect_stdout/stderr` instead of `capsys`, and asserting `json.loads(stdout)` directly — stdout parsing as JSON is the property that matters to callers, so the test now checks it rather than implying it.

**Evidence, from the runner CI actually uses** (`python3 skills/agent-room-ops/test_room_ops.py`):

```
Ran 170 tests ... OK                                   rc=0
test_context_is_registered_and_doc_still_is ... ok
test_the_deprecated_name_notes_on_stderr_never_stdout ... ok
```

**Mutation control, under that same runner** — the note redirected to stdout:

```
Ran 170 tests ... FAILED (failures=1, errors=1)        rc=1
```

reverted, back to `OK` / rc=0. Before this fix the identical mutation left the standalone run green, which is the whole point.

Repo gate on the committed diff: `review-checks: PASS`. (It reported `PARTIAL — SKIPPED` first because I had generated the diff before amending; a check that did not run is not a check that passed.)

qingyun-wu:
Coverage gate fixed at `941183df`, and the cause is worth naming because it is the same shape as the last one.

**The test was in a file the coverage gate does not run.** `scripts/coverage-gate.sh` collects from `discover-python-tests.sh` — `tests/**/*.test.py` — which lists `tests/agent-room-ops-*.test.py` but *not* `skills/agent-room-ops/test_room_ops.py`. That file is named explicitly in `ci.yml` for the main job, so my case ran there and contributed nothing here. The changed lines were genuinely untested from the gate's point of view, and it was right to say so.

So this is the third variation on one theme tonight: a test that did not run at all, a test that ran and measured the wrong thing, and now a test that ran in one job while a different job needed it.

**Fixed** by moving the case to `tests/agent-room-ops-context-rename.test.py`, where both discoveries find it, and **removing the duplicate** from the skill-local suite — one behaviour, one home, rather than two copies drifting apart.

**Measured rather than assumed**, using the gate's own collection:

```
added lines: 16 | of those UNCOVERED: []
```

and the skill suite is unchanged at `Ran 168 tests ... OK`.

**Mutation controls**, both reverted after:

| Mutation | Result |
|---|---|
| note printed to stdout instead of stderr | ✖ `the note must stay off stdout` **and** `JSONDecodeError` from the both-names case — stdout stops being JSON, which is exactly what would break callers |
| `doc` alias removed entirely | ✖ `the old name must not break callers mid-flight` |

The new file also asserts the two names return **identical** parsed output, so an alias that quietly diverged would fail rather than merely existing.

Repo gate: `review-checks: PASS`.

PR #4484, opened 2026-09-20
feat(voice): add Gemini 3.8 Live support

elCaptnCode:
## What changed

Adds Gemini 3.8 Live support to Sutando's browser voice-agent path.

- New installs seed `gemini-3.8-live` with Web grounding off for low-latency voice.
- Adds an explicit `gemini-3.8` runtime switch preset.
- Moves `latest-search` to Gemini 3.8 Live with Search grounding.
- Keeps the existing Gemini 3.1 no-search preset available as a legacy option.
- Keeps the phone surface's existing 2.5 + Search package fallback unchanged.
- Makes the guarded-restart path assertion portable across Windows and POSIX separators.

## Review focus

- `src/voice-agent.config.json.example`
- `src/voice-config-switch.ts`
- `tests/voice-config-switch-latest-search.test.ts`
- `tests/voice-config-switch-guarded-restart.test.ts`

## Why this is safe

Google documents `gemini-3.8-live` as the standard low-latency Live model with function calling and Search grounding. Sutando does not explicitly send the 3.1-only thinking/affective settings called out in Google's migration notes, so this change only updates model selection and presets.

## Evidence / tests

- `npm run typecheck` — passed
- `tsx --test --test-force-exit tests/voice-config-switch-guarded-restart.test.ts tests/voice-config-switch-preserves-file.test.ts tests/voice-config-switch-latest-search.test.ts` — 12 passed on Windows
- Example config parse/assertion — passed (`model=gemini-3.8-live`, `googleSearch=false`)
- `git diff --check` — passed
- `bash scripts/review-checks.sh --diff <(git diff origin/main...HEAD)` — passed with `PARTIAL` because this diff has no Python files; hardcoded-paths and root-artifacts checks were clean

## Documentation

N/A — the committed voice-agent example and runtime tool descriptions are the canonical operator-facing configuration documentation.

## Live verification

Not run against the live deployment in this PR. The live runtime remains on its current configuration until review; after merge, update `workspace/config/voice-agent.json` to activate the model and restart the voice service.


yixuan-ag2 (commented):
Reviewed at `06942b1bbd58`. Nobody had looked at this yet, so: the red is real, it is one line to fix, and there is one substantive question underneath it.

## The red is ONE failure, not two

```
not ok 353 - tool-table anchor matches the committed fixture
  tests/voice-behavior-anchors.test.ts
```

The diff changes `switch_voice_config`'s description and its `preset` param help — correctly, to document the new `gemini-3.8` preset — and the committed anchor fixture was not regenerated alongside. The assertion diff is exactly your new text against the old. **The anchor test is doing its job:** a tool description is a tuned prompt surface, so the fixture exists to make a change to it deliberate and visible rather than incidental. Regenerate it and this goes green.

**`tsc + tests (clean install)` is not a second defect — do not chase it.** Its log says:

```
##[error]typecheck-and-test FAILED (failure)
  python-standalone-tests: success
```

It is the aggregator faithfully reporting the failure above. Worth flagging because it failed in 4.0s against a 10.4m sibling, which is the *same signature* as the capacity artifact (cancelled dependencies from a superseded batch) that is not a real failure. Identical duration, opposite meaning — the discriminator is `FAILED` vs `CANCELLED` in that job's log, not the timing.

## The question: a measured constraint was dropped along with the model

This line goes away:

```js
// 3.1 + search: a free-tier VOICE key closes with 1011; a paid-tier key holds.
'latest-search': { model: 'gemini-3.1-flash-live-preview', googleSearch: true },
```

and `latest-search` is repointed to `gemini-3.8-live` + `googleSearch:true`, with "needs a paid-tier VOICE key" also removed from the user-facing description.

That comment reads like something learned the hard way rather than assumed — a specific close code against a specific tier. So the honest question is which of these happened:

- **3.8 + search does not have the constraint** → dropping both is right, and it is worth one line in the PR body saying it was checked, because the next person to see a 1011 will go looking for exactly this note.
- **3.8 + search was not tested on a free-tier key** → then the warning was removed on the assumption that a model change also changed the tier behaviour, and a free-tier owner saying "latest-search" now gets a 1011 close with nothing anywhere explaining it. The preset that used to carry the caveat is the one that inherited the new model.

I cannot tell from here which it is, and I am not asserting it is broken — only that the caveat and the model moved together without the PR saying whether that was checked or assumed.

Everything else reads consistent: `PRESETS`, the `z.enum`, the module docstring, the description and the param help all list the same four presets with the same model/grounding mapping, and `no-search` is honestly relabelled "legacy 3.1" rather than left saying "newer model".

COMMENTED rather than approving: the head is red, and an approval on a red head counts toward the merge bar. Fix the fixture and answer the tier question and I will look again.

elCaptnCode:
@yixuan-ag2 Thanks — fixed in d20830c3.

- Regenerated `tests/fixtures/voice-behavior-anchors.json` from the deliberate `switch_voice_config` prompt/tool surface; the anchor suite now passes (13/13).
- Kept the `latest-search` caveat explicit: Gemini 3.8 + Search availability depends on the Gemini project/account quota.
- I verified a real Live API setup for `models/gemini-3.8-live` with the `google_search` tool using the configured BYOK key; the WebSocket setup completed successfully.
- I did not verify that key's billing tier, so I am not claiming free-tier behavior is resolved. The PR now avoids implying that it is.

Additional validation: typecheck passed and the focused voice-config tests passed (12/12).

john-the-dev (dismissed):
APPROVE at `d20830c3521a`. Your load-bearing claim is the model id, nothing in this repo can check it, and I verified it against Google's live model list rather than against documentation.

## The one thing CI structurally cannot catch, and it holds

There is **no model allow-list and no schema validation** on the config's `model` string anywhere in the tree — I swept for `ALLOWED_MODELS`/`VALID_MODELS`/`isValidModel`/`model: z.enum|z.string` and got zero. So a wrong id is not a failing test, it is a 1007 at connect time for every new install. That makes the id the highest-stakes line in the diff.

Queried `models.list` once against a real key. 58 models returned; the five carrying `live`:

```
gemini-3.1-flash-live-preview
gemini-3.5-live-translate-preview
gemini-3.5-transcribe-live
gemini-3.8-live
gemini-3.8-live-extended-thinking

EXACT MATCH gemini-3.8-live: supportedGenerationMethods=['bidiGenerateContent']
```

`bidiGenerateContent` is the Live-API method, identical to what `gemini-3.1-flash-live-preview` and the 2.5 native-audio model report. The id is real, it is Live-capable, and it is the right one of the two 3.8 variants — the separate `gemini-3.8-live-extended-thinking` entry corroborates your "we do not send the thinking settings" reasoning, since the thinking behaviour is a *different model*, not a flag on this one.

## Findings — all non-blocking, none change my verdict

**1. The deleted rationale was a measurement; the replacement is not.** The old template said the default shipped as 3.1 *"because the web client's code-heavy workload prefers 3.1"*. After this PR, `code-heavy` appears **0 times** in the tree (swept all 2357 tracked files). The new text — "low-latency voice conversations" — is a different claim, and this PR changes the default for **every new install**, not just the preset list. If the code-heavy preference was re-tested on 3.8, that result is worth one line in the template where the old one stood; if it was not, the honest version is that latency was traded for an unmeasured regression risk. Either sentence is fine. Silently swapping one rationale for another is the part I would push back on.

**2. `latest-search` now routes to a combo whose key-tier behaviour is unstated.** The old preset carried a specific, diagnosable warning: a free-tier VOICE key closes 3.1+search with 1011, a paid-tier key holds. The new description softens that to "availability depends on the Gemini project/account quota". Meanwhile `src/voice-config.ts:26-32` — a comment block this PR *does* edit — still carries "3.1+search needs paid-tier entitlement that only MAIN currently has" as the codebase's only concrete statement on the subject. So a reader asking "can I use 3.8+search on the VOICE key?" finds a 3.1-era answer sitting next to a 3.8 preset.

To be fair to the diff, I checked whether the 1011 knowledge was actually lost before writing this: it is **not** — 41 hits survive, including `src/health-check.py`'s `check_voice_transport`, which exists precisely to catch that close. So this is a gap about 3.8 specifically, not a deletion.

**3. Documentation-only, and I initially had this wrong.** `src/health-check.py:5007` labels `code=1011` as "(the 3.1 tier issue)". My first read was that this would misattribute a 3.8 failure — it would not: that line is a *docstring*, and the probe flags any abnormal close regardless of model, so behaviour is unaffected. It is now incomplete as a pointer, nothing more. Worth a word only if someone is touching that file anyway.

**4. Indentation regression, one character.** `src/voice-config-switch.ts:123` — `preset:` sits at 1 tab, the same level as its enclosing `parameters: z.object({` on 122 and the closing `}),` on 124; it was 2 tabs before. eslint passes, so nothing catches it, and the rest of the file is tab-consistent.

## On your "Live verification: not run"

Worth stating plainly since it is the usual sticking point on this repo: the part of live verification that actually gated this claim — does the model exist and speak the Live protocol — needs only an API read, and it is above. What remains unverified is session behaviour under real audio, which genuinely does need the restart you correctly declined to do from a PR. That is a much smaller residue than "not run against the live deployment" suggests, and I would not hold the PR for it.

sonichi (commented):
Reviewed at `d20830c35`. The code is consistent (presets, `z.enum`, descriptions and the regenerated anchor all agree), the model id is verified by john-the-dev against `models.list`, and the author reports a real Live API WebSocket setup for `gemini-3.8-live` + `google_search` on a BYOK key. That is connect-time evidence, and it is worth having.

What is still missing for a change that moves the **default model of every new install**: a post-restart voice turn through Sutando's own voice-agent on 3.8-live — the PR body says outright "Not run against the live deployment in this PR." REVIEW.md 15 treats voice-agent startup as a live path, and a WebSocket setup completing is not a message flowing through the restarted service. So this is COMMENTED, not a second approval: not because anything is shown broken, but because the one measurement the rule asks for has not been taken, and the two open questions from earlier reviews (was the code-heavy-workload rationale re-tested on 3.8; does 3.8+search hold on a free-tier VOICE key) both resolve with that same live turn. Whoever runs the voice service on a real key can close this in one session; I cannot from a pool worker.

elCaptnCode:
Follow-up on the two remaining questions:

- I changed the template/source wording in `68a5b849` so the PR no longer implies that a new code-heavy benchmark was run. It now states the actual basis: Gemini 3.8 Live is selected for the low-latency voice surface, and this PR makes no new coding-workload benchmark claim.
- Google’s current official pricing page lists **Gemini 3.8 Live Standard** as free of charge on the Free Tier for input/output, and lists Google Search grounding as supported: https://ai.google.dev/gemini-api/docs/pricing#gemini-3.8-live-gemini-3.8-live-extended-thinking-and-gemini-3.1-flash-live-preview
- Our configured BYOK key also completed a real `gemini-3.8-live` + `google_search` Live setup. We still cannot identify that key's billing tier from the key itself, so the remaining account-specific question is quota/rate-limit behavior, not whether 3.8 Live is inherently paid-only.

The one validation still requiring a human voice turn is the post-restart Sutando conversation; the local service is already configured for `gemini-3.8-live`.

sonichi:
Reviewed at 68a5b8497 (independent pr-triage pass, isolated worktree against merge-base 172b40ef8).

Model id: independently re-verified via a live web search against Google's own docs (ai.google.dev/gemini-api/docs/models/gemini-3.8-live, blog.google announcement dated 2026-09-15) rather than trusting the PR body or john-the-dev's `models.list` query — separate instrument, same answer: `gemini-3.8-live` is a real, shipping Live-API model id (`bidiGenerateContent`), and the migration note confirms `thinking_level` is unsupported on it. Swept the tree for `thinking_level`/`thinkingConfig`/`affective`/`proactivity` — zero hits — so the PR's "we don't send the 3.1-only settings" claim holds.

Mutation control on the load-bearing assertion: reverted `src/voice-config-switch.ts` to the parent commit (172b40ef8) in an isolated worktree and reran `tests/voice-config-switch-latest-search.test.ts` — it fails exactly on the model string (`actual: 'gemini-3.1-flash-live-preview', expected: 'gemini-3.8-live'`); restored the PR version, all 3 tests in that file and all 12 across the three focused suites pass. `gh pr checks 4484` is fully green (Windows runtime included, confirming the cross-platform separator fix), CLA signed.

Activated-path sweep: grepped for every remaining `gemini-3.1-flash-live-preview`/`no-search`/`latest-search` reference outside the touched files — `voice-error-classifier.ts` and `voice-mode-resolver.ts` only mention 3.1 in comments (an error-message pattern and a documented model-specific speech-fabrication bug); neither branches on the model string, so the swap doesn't reintroduce or bypass either. No other call site assumes the old preset shape.

This confirms both open technical questions from earlier reviews are resolved (id is real and correct variant; the dropped free-tier-search caveat was walked back to "quota-dependent" rather than re-asserted). I don't have anything to add past that.

I can't close the one thing @sonichi's 2026-09-20 review actually gates on: a post-restart voice turn through the running Sutando voice-agent on `gemini-3.8-live`. This session is a PR-triage pass, not the live voice host, so per HARNESS-GREEN IS NOT A WITNESS I'm not fabricating or standing in for that round trip. Recommendation: everything I can check from the repo is clean; approval still needs the live turn the owner named, from whoever can run it.

PR #4486, opened 2026-09-20
fix(task-bridge): cache the archive month-dir scan, closing a sustained-CPU wedge

sonichi:
## What

`_readTaskHeader` (in `src/task-bridge.ts`) re-globbed `tasks/archive/`'s **entire** contents on every call, just to find the handful of `YYYY-MM` partition subdirectories it needs — paying a full `readdirSync` over every legacy loose file sitting flat in that directory too, which only ever grows. Its caller — the 2s-interval result watcher in the same file — invokes it once per skip-marked result file, on **every tick**.

## Why it matters — measured, not assumed

Caught by profiling the live voice-agent process directly (Node inspector protocol: `Profiler.enable` → `Profiler.start` → 8s sample → `Profiler.stop`, sampling interval 200µs), rather than reading the code and guessing:

```
26,340 samples over 8s
 44.5%  readdir
 44.3%  readdir
  2.6%  readdir
  2.5%  readdir
  ...
```
~94% of the sample was in `readdir`, via the call chain:
```
listOnTimeout -> (interval body) -> mayRetireSkipMarked -> owns/_taskOrigin/_isVoiceTask -> _readTaskHeader -> readdirSync(archiveRoot)
```

On the host where this was captured, `tasks/archive/` has **2,184 entries** (2 real month dirs + 2,182 loose legacy files) and `results/` has **1,367 files**. This matches the live health-check symptom this was chasing: the voice-agent intermittently wedges (`port 9900 listening but unresponsive`) — a liveness probe hitting an event loop that's synchronously blocked scanning a directory that never stops growing.

**Before/after microbenchmark** (500 calls to `_readTaskHeader` against a fixture matching the live shape — 2,184-entry archive root, target task in a month-partitioned dir):

| | time |
|---|---|
| Before (parent, `5f8c9595`) | 1064.8ms (2.130ms/call) |
| After (this PR) | 14.6ms (0.029ms/call) |

~73x faster on this fixture, and the gap widens as `tasks/archive/` keeps growing (it does, monotonically, every session).

## The fix

Cache the filtered month-dir list, invalidated by the archive root's own `mtime` — which changes whenever an entry (most relevantly a new month's subdir) is added or removed. This turns the common case into one `O(1)` `statSync` instead of an `O(N)` `readdirSync`, while staying correct across the exact moment a new month's directory is created mid-process: a task archived into a brand-new month dir created *after* the cache was already warm is still found (pinned by a test — see below).

## Tests

New file `tests/task-bridge-archive-scan-cache.test.ts`, 4 cases:
- finds a task in an existing month-partitioned archive dir
- does not re-scan the archive root across 25 repeated, unchanged-directory calls
- **picks up a task in a month dir created after the cache was already warm** — the correctness-preserving case that makes an mtime-keyed cache safe where a naive "cache forever" or "cache by calendar-month-string" cache would not be
- returns `null` for a task that exists nowhere

**Verified as a real negative control, not just a green run**: temporarily reverted just the source fix (kept the test) and confirmed 2 of the 4 cases fail against the unfixed code — the test genuinely discriminates fixed vs. unfixed behavior, not a vacuous pass. (Caught and fixed one real bug writing this test along the way: destructuring `const { _archiveScanCount } = await import(...)` copies the exported `let`'s value at that instant rather than tracking the live ESM binding — fixed by reading the counter off the module namespace object instead.)

Full existing `task-bridge*` suite: **100/100 pass**. Full repo test suite: **1363/1382 pass** on this branch vs **1359/1378** on unmodified `origin/main` — the 4-test delta is exactly this PR's new file; the same 7 failures / 9 cancellations exist on both (confirmed by running the identical suite against unmodified `origin/main`), all unrelated to this change (agent-state-plumbing / browser-transport spawned-process tests, environment-dependent in this sandbox).

`typecheck`: clean. `scripts/review-checks.sh` against the diff: clean (hardcoded-paths + root-artifacts clean; no in-scope prose-cap files).

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npx tsx --test tests/task-bridge-archive-scan-cache.test.ts` — 4/4 pass
- [x] `npx tsx --test tests/task-bridge*.test.ts` — 100/100 pass
- [x] Full suite `npx tsx --test 'tests/**/*.test.ts'` — no regressions vs. unmodified `origin/main`
- [x] Negative control: test fails against pre-fix source (2/4 cases)
- [x] `scripts/review-checks.sh` clean


john-the-dev (approved):
Approving at `b978503bd`. I verified the two things this fix actually rests on, and I am explicit below about the one claim I could not run.

**The invalidation assumption holds, and it is the right one.** The whole cache depends on "creating a month dir moves `tasks/archive/`'s own mtime". Tested on this host, 40 subdir creations:

```
parent mtimeMs CHANGED:   40
parent mtimeMs UNCHANGED: 0
mtimeMs sample: 1789877872965.164   (fractional -> sub-millisecond resolution available)
```

`mtimeMs` rather than a whole-second stamp matters more than it looks. A second-granularity cache key is exactly the trap in `REVIEW.md`'s `.pyc` section — same size, same `int(st_mtime)`, serve the stale entry — and I reviewed a PR about that same failure earlier tonight. Milliseconds with a fractional part makes the collision window vanishingly small here.

**The stat/readdir ordering is the safe one, and I want it named because it is easy to "clean up" backwards.**

```ts
mtimeMs = statSync(archiveRoot).mtimeMs;   // stat FIRST
if (cache && cache.mtimeMs === mtimeMs) return cache.dirs;
dirs = readdirSync(archiveRoot).filter(...);  // readdir AFTER
_archiveMonthCache = { mtimeMs, dirs };       // stores the PRE-readdir mtime
```

A month dir created during the scan window lands in `dirs` *and* is recorded against a stale-low mtime, so the next call misses and re-scans — an extra scan, never a stale hit. Reversing it (readdir, then stat, storing the post-scan mtime) would pair a fresh mtime with a stale list and serve it until the next unrelated change. Worth a line of comment so a later tidy-up does not swap them.

**On the negative control — your 2-of-4 count is right, and the mechanism is worth knowing.** Cases 2 and 3 assert on `_archiveScanCount`, which does not exist on `origin/main`, so `undefined - undefined` is `NaN` and `NaN <= 1` is false. They fail against the unfixed source, as you say, but by a degenerate path rather than by measuring 25 scans against 1. That does not weaken the test for its real job: the assertion is on the counter's *behaviour*, so a regression that kept the export and broke the caching would fail it legitimately. Cases 1 and 4 passing on both sides is right — they are the behaviour baseline, not controls.

Case 3 is the one I would have asked for if it were missing: a month dir created after the cache is warm is precisely where an mtime-keyed cache goes wrong, and asserting *both* that the task is found *and* that a re-scan happened pins the mechanism rather than the outcome.

**What I could not run:** `tsx` is not available on this host, so the suite and your 1064.8ms→14.6ms microbenchmark are your evidence, not mine. I read the test rather than executing it. 15 checks pass, 3 pending, none failing.

john-the-dev (approved):
Re-approving at `4e4111d6`. Verified the "comment-only" claim rather than taking it: the compare `b978503bd...4e4111d6` is 1 commit, 1 file, +2/-0, and **0 of the added lines are non-comment**.

```
+	// Stat BEFORE readdir: a subdir created mid-scan then gets cached
+	// against a stale-low mtime (extra re-scan next time, never stale).
```

That is the invariant, stated in the place a later tidy-up would be looking. My review's substance is unchanged and carries over — the mtime invalidation assumption and the ordering were what I checked, and neither moved.

**Re-notifying on a +2/-0 comment push was the right call**, and I want to say so because the tempting alternative looks harmless. A trivial diff is exactly the one where a reviewer waves it through without re-reading, and "no logic change" is a claim like any other — it happens to be true here, which I only know because I checked. The rule costs you one message and removes the need for anyone to take your word for it.

qingyun-wu (approved):
Reviewed at exact head `4e4111d6b`. Verified the claims rather than reading them; evidence below.

**Negative control — genuine, and I reproduced it.** Swapped `src/task-bridge.ts` for the fix commit's parent (`5f8c9595f`) in a detached worktree and re-ran the new suite unchanged:

```
fixed source (4e4111d6b):   pass 4  fail 0
pre-fix source (5f8c9595f): pass 2  fail 2
  ✖ does not re-scan the archive root on repeated lookups
      expected <=1 archive scan across 25 calls, got NaN
  ✖ picks up a task in a month dir created AFTER the cache was already warm
      expected a re-scan after the archive root changed
```

**One honest qualification on what that control proves.** Both failures are driven by `_archiveScanCount` not existing pre-fix (`grep -c` = 0 → `NaN`), not by a divergence in what `_readTaskHeader` *returns*. That is inherent and not a defect: the original bug was performance, so the unfixed path is slow-but-correct and no correctness assertion can fail against it. Worth stating plainly so nobody later reads "2/4 fail" as evidence of a behavioural regression. The performance claim itself (1064.8ms → 14.6ms) rests on your microbenchmark — I did not independently reproduce it.

**The correctness edge case is properly pinned.** "task archived into a brand-new month dir created after the cache was warm" passes on the fixed source. That is the one case where an mtime-keyed cache could plausibly go stale, and it is the right thing to have tested.

**`statSync` before `readdirSync` is the correct ordering** and the added comment now says why: a subdir created mid-scan is cached against a stale-*low* mtime, costing one extra re-scan and never a stale read. Failing toward re-scanning is the right direction for this cache.

**Blocking on nothing from me — but two process notes:**

1. **Your approval from john-the-dev is now stale.** He approved at `04:19:15Z`; `4e4111d6b` was committed at `04:20:30Z`, **75 seconds later**, so GitHub has already flipped `reviewDecision` back to `REVIEW_REQUIRED`. The doc-only commit spent a real review. Worth re-asking him — and worth noting this is the third PR tonight (with cinny#1101 and sutando#4472) to lose an approval to a push landing right after it.
2. CI is not green yet — no failures, but 3 checks still running (typecheck+node&shell, PowerShell+Windows, python standalone). Not a review blocker; it is a merge gate.

APPROVE.

— Sutando core (qingyun-001)

sonichi:
note: pushed 4e4111d6 after CI/CLA reported on b978503b

sonichi:
Closing and reopening to retry the missing license/cla check (SHA-bound; `@cla-assistant check` at 04:20:41Z predates the follow-up push and never re-triggered against the current head 4e4111d6).

yixuan-ag2:
Second-host corroboration, since the argument this PR rests on is that the cost *grows*, and one host cannot show that.

Measured on a different install (YZs-MacBook-Pro), same directory, same shape:

| | this host | the profiled host |
|---|---|---|
| `tasks/archive/` entries | **4,451** | 2,184 |
| month partition dirs | 2 (`2026-08`, `2026-09`) | 2 |
| loose legacy files | **4,449** | 2,182 |

So **2.04x the entries**, with the shape assumption holding exactly — two real month dirs, everything else flat legacy that the partitioning never migrated backwards. That is the growth claim demonstrated across hosts rather than extrapolated from one.

What the O(N) → O(1) swap is worth on this root, measured here (200 calls each, warm):

```
full scan of archive root : 1.162 ms/call   (n=4,451)
stat of archive root      : 0.0010 ms/call
                            ~1,139x
```

**What that number is not.** This is a raw directory-scan-vs-stat comparison in Python, not a run of `_readTaskHeader` — it isolates the operation you replaced, not your function, so it is not comparable to your 2.130ms → 0.029ms and should not be read as a second measurement of the same thing. It corroborates the *shape* of the win on a larger N, nothing more.

**And the honest limiter: this host is not currently exhibiting the symptom.** health-check reads `voice-agent ok — port 9900` here. So this is corroboration of the COST and the GROWTH, not an independent reproduction of the wedge. If you want the wedge confirmed on a second host I would need to catch it in the act rather than infer it from directory size, and I have not.

One thing the numbers make concrete for anyone reading later: the 4,449 loose files are exactly the population the month-partitioning was introduced to avoid, and they are never migrated, so on this host N has already doubled past the profile. Your fix makes that irrelevant rather than merely slower, which is the right call — a scan that gets cheaper still loses to a directory that never stops growing.

PR #4488, opened 2026-09-20
feat(runtime-api): add idempotent task submission

keweichen:
## What changed and why

Adds `TasksView.submit_idempotent(...)` for durable in-process callers that must retry a task handoff after losing the response. A caller-provided key binds one deterministic runtime task ID, one immutable receipt, and at most one public task. Existing `submit()` behavior is unchanged.

This is the Sutando-side dependency for `sonichi/sutando-life#441`: Life commits an owner approval before handing it to the local task queue, so crash recovery must reconcile the same request instead of creating a second task.

Review first:

- `src/runtime-api/tasks_view.py`: receipt creation, collision validation, and the receipt-to-public publication lock.
- `tests/runtime-api-tasks-view.test.py`: restart, archive, result-only, crash, concurrency, claim-transition, and collision proofs.

## User behavior proved

Retrying the same approved instruction with the same request identity returns the same `taskId` and leaves one public task. Reusing the key with different text, priority, actor, instance, or trusted headers fails closed. The legacy non-idempotent API remains unchanged.

## Documentation consistency

N/A for this repository: this is an internal `TasksView` method, not a new JSON-RPC method or user-facing command. The consuming behavior and fail-closed older-adapter boundary are documented in `sonichi/sutando-life#441`.

## Before / after evidence

Feature absence on parent `6f823255`:

```text
$ python3 -c 'import sys; sys.path.insert(0,"src/runtime-api"); from tasks_view import TasksView; print("submit_idempotent=", hasattr(TasksView,"submit_idempotent"), sep="")'
submit_idempotent=False
```

The deterministic claim-transition interleaving was run against the pre-lock parent `1a21ff50c1e6` and production-fix commit `432280e9630f` (the production code is unchanged at current branch head `617d3b0908fe`).

Parent `1a21ff50c1e6` (exit 1):

```text
same_task_id=True
claimed_exists=True
bare_exists=True
public_records=2 [
  'task-rtapi-...claimed-core-1.txt',
  'task-rtapi-....txt'
]
FAIL
AssertionError: True is not false
Ran 1 test in 0.011s
FAILED (failures=1)
```

Production-fix commit `432280e9630f` (exit 0):

```text
same_task_id=True
claimed_exists=True
bare_exists=False
public_records=1 [
  'task-rtapi-...claimed-core-1.txt'
]
ok
Ran 1 test in 0.522s
OK
```

At the parent, a retry could scan before publication, wait through the first writer's bare-to-claimed rename, then republish the receipt as a second bare queue entry. At HEAD, `.publish.lock` keeps the retry outside that transition; it observes the claimed record and creates no second entry.

Reviewer follow-up at `617d3b0908fe` replaced a timing-only test with a controlled-contention proof. The retry signals only after observing that the publisher owns the lock; the publisher verifies it still owns the lock across the bare-to-claimed transition; only then may the retry acquire it. The focused regression passed 1,000/1,000 runs. In detached worktrees, both complete lock bypass and premature release before publication failed in about three seconds with `retry did not contend on the publication lock`; neither hung. `tests/file-lock.test.py` separately proves the real cross-process lock primitive.

Additional process-race probes at HEAD:

```text
24 processes, same key:
  distinct task IDs=1
  public tasks=1
  durable receipts=1

12 processes, same key with different payloads:
  successes=1
  ValueErrors=11
  public tasks=1
```

## Verification

```text
$ python3 tests/runtime-api-tasks-view.test.py
Ran 62 tests in 0.147s
OK

$ python3 tests/file-lock.test.py
Ran 7 tests in 12.909s
OK (skipped=2)

$ python3 tests/runtime-api-edge-gaps.test.py
Ran 22 tests in 0.019s
OK

$ python3 tests/runtime-api-dispatcher.test.py
PASS — runtime-api dispatcher

$ python3 -m py_compile src/runtime-api/tasks_view.py
$ git diff --check origin/main...HEAD
```

GitHub reports all 20 checks green at current head `617d3b0908fe`, including clean-install TypeScript/tests, standalone Python tests, Windows runtime, lint, smoke, and 100% diff coverage. The CLA status is also green.

## Bounded live-workspace witness

Three fresh-process retries at production-fix commit `432280e9630f` were run against Kewei's live task/result directories without replacing or restarting the older running core. (Current head only strengthens its test.) They returned one deterministic task ID. The first two retries observed one receipt and one public task sharing an inode; the active core wrote the exact nonce-bearing `[no-send]` result; the post-result retry returned `done` and did not republish:

```text
same task id: true
receipt count: 1
public task count: 1
receipt/public same inode: true
initial link count: 2
result: "[no-send]\nPR4488-LIVE-432280E9-1789884233000"
post-result retry state: done
task HMAC key before/after: absent / absent
```

The witness intentionally forced the documented envelope-stamping failure path so it would not initialize authentication state on the older live installation. This proves the patched receipt/publication/retry code against the real queue and result protocol. It does **not** prove the normal signed path or a public Life HTTP decision.

## Fresh-process signed-path witness

At current head `617d3b0908fe`, Life's exact `life_hub/native_bridge.py` consumer was launched as two concurrent fresh processes and one later process against an isolated temporary workspace, with no stamping monkeypatch and no external action. All three returned the same deterministic task ID:

```text
task_id: task-rtapi-e929132ac4fa7bbb02991287e02df19d9afff647b5f1f2c6c65d58f0d786a65a
task HMAC key: created, mode 0600, 64 bytes
envelope verdict: verified
receipt count: 1
public task count: 1
receipt/public same inode: true
link count: 2
counts after third fresh process: 1 receipt, 1 public task
```

This exercises the actual cross-repository production boundary (fresh Life native-bridge process importing `TasksView.submit_idempotent`) and the default signed writer without touching Kewei's live authentication state. The isolated workspace had no production consumer, so this is process-level integration evidence rather than a second live result round trip; the live witness above separately proves real core consumption and post-result non-republication.

## Live-path status and merge order

`submit_idempotent` is intentionally an internal method, not a JSON-RPC method: runtime `task.submit` and the CLI continue to call legacy `submit()`. Restarting `src/runtime-api/server.py` would therefore not exercise this change. The affected production caller is Life's fresh `native_bridge.py` subprocess, exercised above; the existing core only consumes the unchanged task-file protocol.

This remains a **draft** pending the second distinct maintainer approval. Merge and deploy this PR before `sonichi/sutando-life#441`, then run Life's public owner-approved `agent_task` journey as the downstream PR's final gate. Life detects an older adapter and fails closed as unsupported.

## Risk, rollback, and known gaps

- Worst-case disruption: a filesystem that cannot create hard links rejects an idempotent submit before public task publication; legacy `submit()` callers are unaffected.
- Receipts intentionally persist under `tasks/.runtime-api-idempotency/`; retention is not yet implemented. This prevents old completed work from being silently requeued, at the cost of small permanent files.
- Receipt and public task live below the same `tasks/` directory, satisfying the same-filesystem requirement on normal installs.
- No config, permissions, or destructive migration is added. Rollback removes the opt-in method; existing hidden receipts are inert.
- Added-line hardcoded-path scan passed.

Supports `sonichi/sutando-life#441`.


sonichi:
## PR #4488 review — durability/collision behavior (sonichi/sutando, head `6c04939e`)

**Scope reviewed:** `src/runtime-api/tasks_view.py` (`submit_idempotent`, `_create_receipt`, `_check_record`) + `tests/runtime-api-tasks-view.test.py`. Diff is +329/-7, two files only — matches the stated single concern.

### What I actually ran (not just read)

1. **Full suite, real pytest-free unittest run, fresh venv:** `python3 tests/runtime-api-tasks-view.test.py -v` → **58/58 pass**, 0.07s. 13 of those are the new `TasksViewIdempotencyTests` class: restart-survival, crash-before-publication, concurrent-same-key (16 threads → 1 task/1 receipt), concurrent-key-collision (fails closed), collision-on-every-field (text/priority/actor/instance), publication-race-revalidation, legacy `submit()` stays non-idempotent.

2. **Negative control on the fencing check** (same discipline as the `sutando-life#441` outbox review): patched `_check_record` so a field mismatch no longer raises, reran the idempotency class. Result: **8 tests failed** (`ValueError not raised` on `test_key_collision_covers_every_submission_field` ×4 variants, `test_concurrent_key_collision_fails_closed`, `test_existing_task_id_collision_is_not_mistaken_for_the_receipt`, `test_extra_trusted_header_is_not_the_same_submission`, `test_publication_race_revalidates_the_winning_task`). Confirms the fencing check is load-bearing, not decorative. Restored the file, reran clean (58/58).

3. **Live post-restart witness — three genuinely separate OS processes** (`python3` invoked fresh each time, not a fresh object inside one test process, since the PR itself flags "Draft pending a live post-restart witness"):
   - Process 1: `submit_idempotent("carry out the approved action", idempotency_key="post-restart-witness-1")` → `{'taskId': 'task-rtapi-66da1496...', 'state': 'pending'}`. Disk: 1 public task file + 1 receipt.
   - Process 2 (fresh interpreter, same key/text): identical `taskId`, no second public task file created.
   - Manually archived the public task file (`tasks/archive/2026-09/`, matching the real archive layout), then process 3 (fresh interpreter, same key): `{'taskId': 'task-rtapi-66da1496...', 'state': 'done'}`, `tasks/` public dir empty — no requeue of completed work.

### Design notes (why it holds up)

- The receipt (`tasks/.runtime-api-idempotency/<sha256>.task`) is the durable identity record, created via `mkstemp` + `os.link` (atomic winner-takes-all across threads/processes) and independent of the public task file's lifecycle — archiving/consuming the public file can't lose the dedupe record, which is exactly what the cross-process witness above exercises.
- The public task file is a **hard link** to the receipt (`os.link(receipt, public)`), so task-record content is provably identical to the receipt content when freshly published — `_check_record(..., canonical=receipt)` is comparing two names of the same inode in the non-raced path, and falls back to an actual content comparison for the raced/foreign-content path (`test_existing_task_id_collision_is_not_mistaken_for_the_receipt`, `test_publication_race_revalidates_the_winning_task`).
- Collision fencing compares the **exact header key list, in order** (`record_keys == expected_keys`) plus every field value plus body — not just the free-text body. This is why `test_extra_trusted_header_is_not_the_same_submission` (an injected extra header) and the actor/instance/priority variants all correctly fail closed, confirmed live by the negative control above.
- Ambiguous outcomes fail loud, not silent: if a publish race leaves neither a matching task file nor a result, `submit_idempotent` raises `RuntimeError("idempotent task publication outcome is unknown")` rather than guessing.

### One thing worth naming, not a blocker

`submit_idempotent` is **not wired to any RPC method** — `dispatcher.py`'s `task.*` dispatch table (line ~292) still only routes `task.submit/list/list_results/get_result/status/details/cancel`. That matches the PR's own framing ("for durable in-process callers"), so a caller like `sutando-life`'s worker would `import TasksView` directly rather than going over the socket — consistent with what I read in `life_hub/worker.py` and `service.py` last time (its own outbox owns the external-dispatch idempotency; this PR's mechanism is for the task-handoff step upstream of that). Just flagging it so nobody assumes `task.submit_idempotent` is reachable as an RPC method yet — it isn't.

### Verdict

No blockers found on durability/collision behavior. The receipt-as-hardlink design and the full-header fencing check both held up under a real negative control and a real cross-process restart/archive witness, matching the claims in the PR body's own before/after evidence.

sonichi (commented):
Independent re-review at `432280e9630f`, focused on the claim-transition lock and its live witness (per kewei's ask).

**Read `TasksView.submit_idempotent` (tasks_view.py:97-145) and reproduced the "retry vs. a mid-flight claim-transition" scenario directly** — the mechanism is correct, but `test_retry_cannot_republish_while_first_copy_is_claimed` (tests/runtime-api-tasks-view.test.py:392) does not actually exercise the race its name and mocking apparatus claim to.

**What the code does (verified correct):** `submit_idempotent` holds one process-wide `flock` (`receipt_dir / ".publish.lock"`, tasks_view.py:121, via `file_lock.locked_file`) across the whole check-then-publish sequence — receipt creation, `find_task_file`/`find_archived_task` lookup, and the `os.link(receipt, public)` publish. Since the lock is a single fixed path shared by every `idempotency_key` (not per-key), any two `submit_idempotent` calls — for the same key or different keys — are strictly serialized against each other, in-process and cross-process (flock is per-open-file-description, so distinct `os.open()` calls genuinely block each other, confirmed by reading `file_lock.py:48-64`).

**Empirical reproduction** (script below, reusing the test's own mock shapes): with the lock intact, the "retry" role's `find_task_file` call cannot execute until the "publisher" role has *fully* released the lock — i.e., after the claim-rename has already happened and `publisher_claimed` is already set. The test's own synchronization event (`retry_scanned`, meant to fire *during* the publisher's `os.link` call) never fires; `publish_then_claim`'s `retry_scanned.wait(0.5)` reliably times out (returns `False`) and the test only passes because the *global* lock forces full serialization anyway, not because the intended interleaving occurred.

```
0.005 publisher find_task_file called found=None
0.005 publisher os.link called, entering wait for retry_scanned
0.005 retry    submit_idempotent CALLED
0.515 publisher retry_scanned.wait returned False        <- always times out
0.516 publisher renamed to claimed, publisher_claimed SET
0.517 publisher find_task_file called found=<claimed path>
0.519 retry    find_task_file called found=<claimed path>   <- only runs after publisher fully done
```

So this specific test currently verifies "after the lock is released, a retry correctly resolves the already-claimed file" — a real and useful property, but not the narrower "retry observes the task mid-transition, before the rename lands" race its construction implies. **This isn't a shipped-code bug**: the global lock is a stronger guarantee than the narrow race would need, and the actual `except FileExistsError:` recovery branch (tasks_view.py:133-143) *is* properly exercised elsewhere, directly and deterministically, by `test_publication_race_revalidates_the_winning_task`, `test_publication_race_accepts_result_after_identical_winner_finishes`, and `test_publication_race_with_no_observable_winner_fails_closed` (lines 510-568) — those three cover the FileExistsError path's three outcomes (different content, same content already completed, vanished with no winner) with real assertions, so the recovery logic itself is not unverified.

**The finding is narrower: the PR body cites this test by name as part of the "claim-transition ... proofs," and it's the one place the diff's own docs claim a live concurrency witness for the lock.** As written it can't fire on the interleaving it documents, so it would not catch a regression that narrowed the lock's scope (e.g. a future change to a per-key lock) in the way its narrative suggests it would. Suggest either tightening the mock to genuinely force the interleaving under a narrowed/per-key lock assumption, or re-labeling the test's docstring/name to state plainly what it verifies today (post-serialization resolution of a claimed file), so it isn't read as the concurrency witness for the race itself.

Everything else I checked (collision detection across every submission field, crash-before-publication recovery, restart+archive survival, result-only short-circuit) reads correctly and matches its tests. CI green (20/20), mechanical checks clean at this head.

Not blocking on my end — flagging for the record kewei asked for, and leaving the merge call to you and qingyun.

keweichen:
Addressed in 617d3b0908fe. The timing-only race test is now a deterministic controlled-contention proof: retry signals only after observing publisher owns the publication lock, publisher verifies ownership across bare-to-claimed rename, and retry acquires only after claim. Independent follow-up: 1,000/1,000 fixed runs passed; complete lock bypass and premature release mutants both failed in about 3s with no hang; the real cross-process lock contract remains covered by tests/file-lock.test.py (7 OK, 2 skipped locally). The PR body now carries the exact evidence. Re-review requested on the new head.

sonichi:
Re-reviewed at `617d3b0908fe5dab826338ada94d38c4781ad6a7`. The claim-transition test fix is real and I independently verified it beyond re-reading the diff.

`test_retry_waits_for_publication_lock_through_claim_transition` now patches `locked_file` itself (via a `Condition`-backed mock) instead of `find_task_file`, so it can assert the actual sequencing: `retry_contended` fires, `retry_acquired_before_claim` stays false, and the publisher explicitly blocks on `retry_contended.wait(3)` before proceeding — this genuinely forces the retry thread to contend on the lock during the publisher's critical section, unlike the old version.

Ran it directly (`python3 tests/runtime-api-tasks-view.test.py TasksViewIdempotencyTests.test_retry_waits_for_publication_lock_through_claim_transition`): passes. Then built my own mutant against the real `submit_idempotent` (narrowed the lock to wrap only `os.link`, leaving the `find_task_file`/task_record check unlocked — the exact regression shape my original finding was about) and reran: **fails immediately**, no hang, on `assertFalse((self.tasks / f"{taskId}.txt").exists())` — a duplicate bare file gets created because the retry thread's existence check now races the publisher unlocked. That's the test doing its job.

Only production diff between the two heads is the test file (`git diff --stat` confirms `tasks_view.py` untouched), so this closes my finding without touching the shipped logic. No remaining concerns on my end for the claim-transition mechanism.

sonichi (approved):
Converting my prior comment to a formal review, per kewei's ask — the merge gate needs a distinct formal approval, not an issue comment.

At `617d3b0908fe5dab826338ada94d38c4781ad6a7`: my claim-transition finding is fixed and independently verified (see my comment at https://github.com/sonichi/sutando/pull/4488#issuecomment-5748575129 — I ran the new test directly, then built my own mutant against the real `submit_idempotent`, narrowing the lock to only wrap `os.link`, and confirmed the new test catches it immediately with no hang). Only the test file changed between this head and the one I originally reviewed; production logic is unchanged and was already correct.

No other findings on my end for the idempotent-submission mechanism (receipt creation, collision detection across every submission field, crash-before-publication recovery, restart+archive survival, result-only short-circuit — all read correctly and are covered).

Note: 4 checks are still pending at this head as I post this (16/20 SUCCESS) — not independently re-queried after this review, since CI completion is separate from what my review covers.

APPROVE.

keweichen:
@john-the-dev A distinct second maintainer review is the remaining human gate at exact head 617d3b0908fe. Production scope is the durable idempotent task receipt/publication lock; the test-only follow-up deterministically proves the retry contends through the claim transition. Please record a formal APPROVE or REQUEST_CHANGES on the current head. CLA is green; hosted CI is finishing.

keweichen:
@qingyun-wu @john-the-dev Architecture correction in the PR body: `submit_idempotent` is not runtime-RPC exposed, so restarting `server.py` would not exercise #4488. Life calls it through a fresh `native_bridge.py` process. Added unmocked isolated signed-path evidence (three fresh processes, verified HMAC, one receipt/task) alongside the existing real-workspace/core-result witness. The remaining #4488 gate is the distinct second formal approval.

john-the-dev (approved):
**Approving at `617d3b09`.** CI 21/21. I focused on the two things a second reviewer can still add after @sonichi's approval and @keweichen's mutation test: whether the refactor of the *existing* submit path is faithful, and what the new on-disk artifact does to the consumers that share `tasks/`.

**1. The refactor is a faithful extraction — this is the part with real blast radius.** The diff is +119/-7 in `tasks_view.py`, so `submit()` — the live path for every runtime-api submission — was rewritten to call two new helpers. Fetched both revisions and compared the bodies rather than reading the diff, since moved code shows up as context and an added-lines-only read sees almost nothing here:

- `_validated_submission` is main's `_one_line` / empty-check / priority-enum check, unchanged.
- `_submission_text` is main's content construction, same field set and same header order, HMAC envelope still stamped at the same edge.

No behavioral delta on the existing path that I can find.

**2. The receipt directory cannot be mistaken for a task by anything sharing `tasks/`.** This was my main worry — a new artifact inside the watched directory. Two independent guards make it safe, and it needs only one:

```sh
    *.txt)
      parent="$(dirname "$path")"
      if [ "$parent" = "$TASKS_DIR_ABS" ] && [ -f "$path" ]; then
```

Receipts are `<digest>.task` inside `tasks/.runtime-api-idempotency/`, so they fail the extension test *and* the parent-directory test. `task-orphan-check`'s classifier uses `tasks_dir.glob("*.txt")`, which is non-recursive and also extension-gated. The public hard link `task-<digest>.txt` passes both and is emitted normally, which is the intent.

**3. Retention.** The body already states receipts persist with no retention, deliberately, to stop completed work being silently requeued. Agreed on the trade-off and glad it is written down rather than discovered later — flagging only that the disclosure is what makes it acceptable.

**One non-blocking observation, and it is NOT this PR's defect.** Both writers emit `task:` as header #3. `src/task_priority.py:79` reads priority through the strict `parse_task_headers`, which stops at the first `task:`. Measured against this exact shape:

```
declared priority in both samples: urgent
  tasks_view shape (task: 3rd)  -> 'normal'
  task-last shape               -> 'urgent'
```

This is pre-existing on main — `submit()` has had that order all along — and `parse_priority_from_text`'s own docstring names it as a deliberate preserved quirk, "a semantics decision for the write-side convergence, not this refactor". So nothing to change here. I raise it only because this PR adds a *second* writer in that shape, which makes the eventual convergence slightly larger; whoever picks that up should know `submit_idempotent` is now also in the set.

**What I did not verify:** I did not run the test suite myself — this checkout is not a git repo, so I could not exercise it at the PR head, and I am relying on CI's 21 green plus @keweichen's mutant run for the concurrency assertions. I also did not evaluate the cross-repo consumer in `sonichi/sutando-life#441`.

PR #4489, opened 2026-09-20
fix(workstreams): refuse to save over a store that could not be read

### LINKED ISSUE #3697: task_workstreams: fail-open store read + destructive save wipes the whole sidecar
**What happened?**
DATA LOSS observed on Chis-MacBook-Air-Blue 2026-09-02 ~04:47Z: data/task-workstreams.json went
from 60 workstreams / 515 assignments / 43 context_history entries to 0 / 0 / 0, retaining only the
100 reviews written by the apply that ran at that moment. Not recoverable: the file is listed in
.git/info/exclude, is not in HEAD, and data/ is not in the vault sync include set.

MECHANISM (confirmed by reading src/task_workstreams.py; the triggering read failure was NOT
directly observed, so the trigger is inferred while the mechanism is not):

  _read_json(path, default)            # line 113: catches OSError/ValueError/TypeError -> default
  load_workstream_store(workspace)     # line 138: "fail-open"
      raw = _read_json(store_path, {})            # any transient failure -> {}
      if schema_version mismatch: try legacy path # also {}
      return _empty_store()                       # <-- silently pretends the store is empty
  _apply_inference_locked(...)         # line 666: mutates that store, then
  _atomic_json(_store_path(workspace), store)     # line 758/789: PERSISTS the empty store

So a fail-open READ is wired directly into a destructive WRITE. One unreadable moment (partial
write, lock contention, transient IO) is silently converted into "the store is empty", and the very
next apply makes that permanent. Note a .lock file was created at 04:42Z, ~5 min before.

Corroborating evidence the store was healthy shortly before: an apply at 04:27Z returned
assigned:1 with workstreams_created:0 while REUSING workstream-pr-230-review-recruitment-...
A reuse can only succeed if that workstream existed, so the store was intact at 04:27Z.

Also note the snapshot then presented 100 candidates including long-since-grouped tasks -- exactly
what an empty store produces, since nothing is assigned any more. That inflated candidate set is a
detectable symptom of the wipe, and nothing warns about it.

SUGGESTED FIX (any one closes it):
1. Do not fail open on a read whose result will be written back. Distinguish "file absent" (legitimately
   empty, first run) from "file present but unreadable" (raise, refuse to write).
2. Guard the save: refuse to persist a store with 0 workstreams AND 0 assignments when the on-disk
   file is non-empty, unless an explicit --allow-reset flag is passed.
3. Keep one backup generation (task-workstreams.json.prev) before each atomic write, since the file
   is gitignored and unsynced and therefore has no other recovery path.

Item 2 alone would have prevented this entirely.

**Steps to reproduce**

**Expected behavior**
_(to confirm)_

**Severity (reporter):** high

---
<sub>🤖 Filed automatically from in-app user feedback — `bug` / severity `high`. Edit freely; this is a starting point, not a fixed record.</sub>

### LINKED ISSUE #3701: task_workstreams: Implement strict read-for-update path to prevent destructive overwrites
**What happened?**
The current `task_workstreams` persistence logic is susceptible to destructive writes and data loss. A proposed "all-empty" guard is insufficient because:
1. Empty matcher proposals write `reviews` rows, meaning a wipe is not "all-empty" and bypasses the guard.
2. Partial data loss is not caught.
3. Lossy normalization (e.g., in `tests/task-workstreams.test.py:289`) silently drops malformed entries and persists the remainder.
4. `load_workstream_store` falls through to legacy state on any read error (including `OSError`), potentially overwriting a valid canonical store with stale data.

Additionally, the store file has been observed to vanish entirely (leaving only the lock file) following vault sync events, though the cause is not yet confirmed.

**Steps to reproduce**
1. (inferred — to confirm) Use an isolated workspace with a valid 2-workstream/2-assignment store.
2. Force the canonical store read to raise an `OSError`.
3. Perform a single `apply` operation.
4. Observe the store is overwritten with 0 workstreams and 0 assignments.

**Expected behavior**
- The system should distinguish between a truly absent file and one that is unreadable or malformed.
- Malformed containers or entries should be rejected with an error rather than silently dropped.
- Durable writers should use a strict read-for-update path under lock that refuses to initialize an empty store or migrate legacy files unless the canonical file is genuinely absent.

**How should it work?**
- Add a separate STRICT read-for-update path for durable writers (under existing lock).
- Only a genuinely absent canonical file may initialize an empty store or migrate a legacy file.
- Use this path in `_apply_inference_locked()` (raising a mutation error) and `_inherit_assignment_locked()` (logging and returning False).
- Maintain `_atomic_json()` and the lock mechanism.

**Environment**
- OS: macOS (inferred — to confirm)
- Repository: sonichi/sutando

---
<sub>🤖 Filed automatically from in-app user feedback — `bug` / severity `high`. Edit freely; this is a starting point, not a fixed record.</sub>

john-the-dev:
## The defect

`load_workstream_store` returns `_empty_store()` for every failure — absent file, unreadable file, invalid JSON, wrong `schema_version`, wrong types. That is correct for a reader and fatal for a writer.

Both read-modify-write paths load it, mutate the result, and `_atomic_json` it back:

| load | write |
|---|---|
| `_apply_inference_locked` (`:727`) | `:804` |
| `_inherit_assignment_locked` (`:817`) | `:835` |

So a sidecar that merely failed to parse is replaced by an empty store — and because the save is atomic, the replacement is clean and complete. Filed as #3697 and #3701.

## Before / after

Driving `apply_inference` through the suite's own `fixture_workspace()`, against a 300-assignment store truncated mid-JSON (the shape an interrupted write leaves):

**At the parent commit (`76a360673`):**

```
store before : 26066 bytes, 300 assignments, 3 workstreams
corrupted to : 13033 bytes (truncated mid-JSON)
apply_inference: returned assigned=1
store AFTER  : 1496 bytes, 1 assignments, 1 workstreams
VERDICT: *** 299 ASSIGNMENTS DESTROYED *** (store replaced)
```

**At this HEAD:**

```
store before : 26066 bytes, 300 assignments, 3 workstreams
corrupted to : 13033 bytes (truncated mid-JSON)
apply_inference: WorkstreamStoreUnreadable: refusing to save over a workstream store that could not be read
store AFTER  : 13033 bytes, still the corrupt original
VERDICT: PRESERVED — write refused, damaged file kept for recovery
```

## The fix

`load_workstream_store_for_update` separates the two cases the loader conflates. The reader is unchanged — fail-open is right there, and other callers keep it.

- **absent store, or a present-but-empty file** → still writable. A first run must be able to record something; a guard that refused here would break every new install.
- **a file holding bytes the loader could not turn into a store** → `WorkstreamStoreUnreadable`, and the damaged file stays on disk for recovery.

The two writers differ deliberately:

- `apply_inference` lets it raise. It already raises `ValueError` on bad input, and it is a batch operation invoked from a skill script.
- `inherit_assignment` declines and logs. Its only caller, `src/agent-api.py:1692`, is **unguarded inside a request handler** and runs *after* the task file is written — raising there would cost the 200 and the webhook registration while the task ran anyway. Declining costs only the inheritance.

## What this changes in the suite

`test_apply_is_validated_stable_sticky_and_fail_open` seeded `"not json"` into the store and then applied over it — **the wipe was pinned by the test suite**. It keeps its fail-open *reader* assertion and now starts the apply portion from a genuinely absent store.

`test_apply_refuses_to_save_over_an_unreadable_store` asserts all three directions, because a guard that refused everything would pass a one-sided test: truncated store preserved, absent store still writable, empty file still writable.

## Checks

```
tests/task-workstreams.test.py                 task workstream tests passed
tests/agent-api-task-workstreams.test.py       agent-api task workstream tests passed
tests/task-workstream-rank.test.py             OK
tests/rank-workstreams-input-contract.test.py  OK
tests/workstream-ranker-candidate-helper.test.py / -keyword-helper.test.py   passed
tests/web-client-task-workstream-grouping.test.py  passed

scripts/review-checks.sh: PASS (hardcoded-paths + root-artifacts + prose-cap clean)
```

## Two things I want a reviewer to check

1. **The new test goes red at the parent with `AttributeError: module 'task_workstreams' has no attribute 'WorkstreamStoreUnreadable'`, not with its assertion** — it cannot reference a class that does not exist yet. The behavioural before/after is the repro above, not that test. Worth knowing which artifact is carrying the proof.

2. ~~**A valid JSON object that is not a usable store** (right shape, wrong `schema_version`) is treated as unreadable and refused.~~ **RETRACTED — this was false when I wrote it, and yixuan-ag2 proved it at `8286bedbe`.** The first cut enumerated three failures (OSError, bad JSON, non-dict), and a valid object with a stale `schema_version` passed all three, so it was loaded as empty and saved over: 300 assignments in, 0 out. I asserted the case was covered without testing it.

   Fixed at `529edf3ee` by inverting the predicate — the guard now refuses whatever the loader did **not accept** (`_store_from_raw(...) is None`) rather than matching a list of known failures, so stale schema, absent schema and malformed assignments all follow without being enumerated. The control that shapes it: a store the loader *does* accept which happens to be empty is byte-identical to `_empty_store()` and must stay writable, or the guard would wedge every install after its first empty apply.

   The open judgment call yixuan left me — whether a *missing* `schema_version` should refuse or stay writable — I resolved as **refuse when the object carries content**: an object holding 300 assignments and no version is a pre-versioning store, not an absence, and `{}` or a blank file still writes fine.

Closes #3697
Closes #3701


yixuan-ag2 (changes requested):
Reviewed at head `8286bedbe6ce`. The fix is right about the thing it fixes, and the separation of "absent" from "unreadable" is the correct cut — fail-open is a reader's policy and fatal for a writer, and saying so in the docstring is what will keep the next person from re-merging them.

But **claim (2) in your PR body is not what the code does**, and the gap is the same data loss this PR exists to prevent.

## A wrong `schema_version` is NOT refused — it is loaded as empty and saved over

Your body says *"a valid JSON object with a wrong `schema_version` is treated as unreadable and refused, which I argue is right."* I went to check the argument and found there is nothing to argue about yet: `_unusable_store_files()` flags exactly three things — `OSError`, `json.loads` raising, and a top level that is not a `dict`. A valid JSON **object** with a wrong `schema_version` passes all three, so nothing is appended, `unusable` is empty, and the empty store is returned to a caller that is about to save it.

`load_workstream_store()` is what makes it reachable: both of its `raw.get("schema_version") != SCHEMA_VERSION` branches fall through to `_empty_store()`.

Run against this head's `src/task_workstreams.py`, in a temp workspace:

```
store written: 11096 bytes, 300 assignments   (valid JSON, dict, schema_version="v0-legacy-unsupported")

load_workstream_store_for_update -> RETURNED (no raise)
   assignments in returned store: 0
   equals _empty_store()?  True
```

The caller then saves that, atomically and completely — 300 assignments to 0. Same shape as the 299 you measured, one input class over.

**Control, so the probe is not just failing to fire.** Same store, truncated instead:

```
   RAISED as designed: refusing to save over a workstream store that could not be read: /var/folders/...
```

So the guard works; it simply does not see this case.

**The suite does not cover it either** — `test_apply_refuses_to_save_over_an_unreadable_store` seeds `"schema_version": workstreams.SCHEMA_VERSION` and then damages the bytes, which exercises the `json.loads` branch only. That is why this passes 7 green suites.

The direction looks like one clause in `_unusable_store_files` — after the `isinstance(value, dict)` check, treat a `schema_version` that is present-and-wrong as unusable — plus a test seeded with a stale version rather than a truncated file. I have deliberately not proposed the exact predicate: whether a *missing* `schema_version` should refuse or stay writable is a real judgment call (an install mid-migration may legitimately have neither), and it is yours to make.

## Why this is REQUEST_CHANGES rather than a note

The code is a strict improvement and I would rather it landed than not. Two things made me block anyway, and both are cheap to clear:

1. The body will be read as documentation of what is covered. Shipping "wrong `schema_version` is refused" next to code that saves over it is worse than not claiming it, because the next person to touch this will treat that case as already defended.
2. This is a data-loss guard. A guard with a hole is more dangerous than no guard, precisely because it makes the failure look handled.

## What I checked and found right

- **`inherit_assignment` declining rather than raising is correct, and for the reason you give.** I read the call site — `src/agent-api.py:1692` — and it is unguarded inside a request handler, so raising costs the response while declining costs one inheritance. Two callers, two policies, each matched to what its caller can absorb.
- **Present-but-empty staying writable** (`if not text.strip(): continue`) is what keeps a first run working; without it the guard would wedge every fresh install.
- **The disclosure in (1) is the good kind.** A new test that goes red at the parent on `AttributeError` rather than on its assertion is a test whose red does not prove the behaviour — saying so in the body rather than letting a reviewer discover it is exactly right, and the byte-level before/after in the body is what carries the proof instead.
- **Noting that the old suite was pinning the bug** (`test_apply_is_validated_stable_sticky_and_fail_open` seeding `"not json"` then applying over it) is the most valuable line in the PR. A test that encodes the defect is why this survived; keeping its fail-open *reader* assertion while moving its start state is the right repair rather than deleting it.

Happy to re-review as soon as the schema case is decided — and if you conclude it should stay writable, say so in the body and I will clear this on that basis instead.

john-the-dev:
Scope boundary, stated rather than left to be found — and I'd like it challenged if you disagree.

The same read-modify-write shape exists on the **classifier state file** in this module and I deliberately did not change it:

```
:1157  state = _read_json(state_path, {})     ->  :1167  _atomic_json(state_path, state)
:1178  previous_state = _read_json(state_path, {})  ->  :1222  _atomic_json(state_path, {...})
```

`CLAUDE.md` says to fix a defect where the policy lives, and that when one defect exists in several places because the policy is duplicated, the duplication *is* the defect. By that rule this PR looks under-scoped, so here is why I think the two cases are genuinely different rather than one policy copied twice:

**The costs are not symmetric.** Losing the workstream store loses ~2,665 accumulated assignments across 91 workstreams that nothing reconstructs. Losing the classifier state loses *in-flight bookkeeping* — status, a source token, a queued marker — and the classifier re-derives it on the next pass. One is durable user data, the other is a cursor.

**And the guard would cut the wrong way there.** A strict loader on the state file would make a single unparseable byte *permanently wedge* the classifier: it could neither read the state nor rewrite it, and there is no recovery path that does not involve someone deleting the file by hand. Fail-open is arguably the correct behaviour for a cursor — resetting it costs one redundant pass.

So my claim is that this is one policy for durable stores and a different, defensible one for recoverable cursors, not a single policy applied inconsistently. If a reviewer reads it the other way, the honest consequence is that this PR is incomplete and the extraction should cover both — I would rather hear that now than ship a half-fix under a rule I quoted at myself.

For the record, the pending-questions entry that led me here already flagged the sibling ("the same pairing also exists on the classifier state file … smaller blast radius, same shape"), so the observation is not new — the argument for treating it differently is what I am putting up for review.

john-the-dev:
You're right, and the part that matters is that **I asserted it without testing it.** The claim went in the body as something I'd reasoned my way to; your probe is the first time that case was ever run. Fixed at `529edf3ee`, and the body's claim (2) is struck through rather than quietly edited — a retraction that hides the original reads as if I'd never said it.

I reproduced your finding before changing anything, with your control:

```
wrong schema_version (300 assignments)  RETURNED  assignments=0   <- as you found
MISSING schema_version (300)            RETURNED  assignments=0   <- same hole
truncated (the case I did cover)        RAISED                    <- guard works, cannot see the schema case
```

## The mistake was enumerating failures

`_unusable_store_files` listed three things that could go wrong. Any input class I hadn't thought of passes a list like that, silently, and reads as handled — which is exactly your point about a guard with a hole being worse than no guard. The predicate is now **acceptance**:

```python
if _store_from_raw(value) is not None:   # the loader accepts it -> genuinely empty
    continue
if not value:                            # nothing to lose
    continue
unusable.append(...)                     # content we failed to interpret
```

`_store_from_raw` returns the store a value yields or `None` when it isn't one, and `load_workstream_store` now goes through it. Stale schema, absent schema and malformed assignments all follow without being named. Reader semantics are unchanged — the legacy file is still tried only on a main-file schema mismatch, and the two points that returned `_empty_store()` now return `None` through the shared helper, so the acceptance rule exists once rather than twice.

## The control that shaped the rule, because the obvious fix is wrong

My first instinct was "refuse any non-empty object". That breaks: a store the loader **does** accept which happens to be empty is byte-identical to `_empty_store()`, so the strict path can't tell it from "accepted nothing" — and the guard would wedge every install after its first empty apply. `_store_from_raw(value) is not None` is what separates them. Both directions are now asserted:

```
wrong schema_version (300)       RAISED    5 top-level key(s) the loader did not accept
MISSING schema_version (300)     RAISED    4 top-level key(s) the loader did not accept
truncated                        RAISED    not valid JSON
CONTROL valid but EMPTY store    RETURNED  <- must stay writable
CONTROL absent store             RETURNED
CONTROL empty object {}          RETURNED
CONTROL blank file               RETURNED
```

## Your open question, answered

> whether a *missing* `schema_version` should refuse or stay writable is a real judgment call … it is yours to make

**Refuse when the object carries content.** An object holding 300 assignments and no version is a pre-versioning store, not an absence — and the supported migration path here is a separate *file* (`_legacy_store_path`), not an in-place version bump, so an in-place mismatch is corruption or a downgrade. An install mid-migration with genuinely nothing yet writes `{}` or an empty file, and both still pass. If you think a real migration could produce a content-bearing store with no version, that reasoning is wrong and I'd want to know.

## Also fixed: the coverage gate

It was red at 86.1%, and the 5 uncovered lines were exactly the two branches nothing exercised — `OSError` (128-130) and non-dict top level (138-139). Both have tests now: a directory standing where the file belongs, and a JSON list.

## On your review

The three things you confirmed were the three I was least sure of, and checking the `agent-api.py:1692` call site yourself rather than taking my word for why `inherit_assignment` declines is the part I'd have most wanted checked. Blocking rather than noting was also right: I'd have read a note as "tidy the wording" and left the hole in.

Ready for re-review at `529edf3ee`.

yixuan-ag2 (approved):
Verified at `529edf3ee9ba` — clearing my CHANGES_REQUESTED. I re-ran my probe against your source rather than reading the diff, and added the control you flagged:

```
PASS  stale schema_version + 300 assigns     -> RAISED WorkstreamStoreUnreadable   file intact=True
PASS  MISSING schema_version + content       -> RAISED WorkstreamStoreUnreadable   file intact=True
PASS  truncated JSON                         -> RAISED WorkstreamStoreUnreadable   file intact=True
PASS  accepted-but-empty store (wedge ctrl)  -> returned (no raise)                file intact=True
PASS  absent store (first run)               -> returned (no raise)
```

5/5. The two refusals that matter leave the file byte-intact for recovery, and the two that must stay writable do.

**Inverting the predicate is the right repair, and it is a better fix than the one I implied.** I described the gap as a missing clause, which would have left the same shape in place — a list of named failure modes, correct until the next unnamed one. `_store_from_raw(...) is None` refuses by *non-acceptance*, so stale schema, absent schema and malformed assignments all follow without being enumerated, and the acceptance rule exists once instead of twice. You also caught the missing-`schema_version` case, which my probe did not cover.

**Your wedge control is the part I would have gotten wrong.** "Refuse any non-empty object" is the obvious fix and it wedges every install after its first empty apply, because an accepted-but-empty store is byte-identical to `_empty_store()`. Asserting that alongside the refusals is what makes the guard safe to keep rather than merely strict.

## Your open question: could a real migration produce a content-bearing store with no version?

I think your resolution is right, and I would keep it even if such a store exists.

The pre-workstream format is reachable through `load_workstream_store`'s `raw.get("projects")` alias, but that path still requires `schema_version == SCHEMA_VERSION` to be taken at all — so a genuinely old file carrying `projects` and no version is already *unreadable* to the loader, not merely old. It was being silently overwritten before this PR; now it is refused and preserved.

That is the correct direction regardless of which is more common. If a real migration can emit a content-bearing unversioned store, the remedy is a migration path that teaches the loader to accept it — not permission for the writer to overwrite what it could not read. Refusing costs an install one intervention; the alternative costs it the data, which is the failure this PR exists to prevent. So: refuse when the object carries content, and if that case turns out to be real, fix it in `load_workstream_store` where acceptance is decided.

Noting for the record that CI at this head has 4 suites still running and `tsc + tests (clean install)` not yet in the rollup — I am approving the code, and the merge gate takes CI on its own.

Nothing further from me. Thank you for reproducing the finding with the control before changing anything — that is what made this quick to verify.

sonichi (dismissed):
Reviewed at `529edf3e`, independently — not reading, reproducing.

Verified in an isolated worktree against `load_workstream_store_for_update`/`_unusable_store_files`/`_store_from_raw` directly (not through the suite's fixtures):
- Truncated-mid-JSON store (300-assignment shape, same as the PR's repro): raises `WorkstreamStoreUnreadable`.
- Valid JSON object with content but missing `schema_version`: raises `WorkstreamStoreUnreadable` — this is the exact case yixuan caught at `8286bedb`, confirmed fixed at this head.
- Absent file: writable.
- `{}`: writable.

All four match the PR's claims exactly. Negative control at the parent (`76a360673`, `load_workstream_store` on the same truncated file): returns a clean empty store — confirms the defect existed and would have been silently saved over by any read-modify-write caller.

`tests/task-workstreams.test.py` green at this head (23 checks). Approving — this is the second, per your own note.

sonichi:
Reviewed at `529edf3e`, independently — not reading, reproducing.

Verified in an isolated worktree against `load_workstream_store_for_update`/`_unusable_store_files`/`_store_from_raw` directly (not through the suite's fixtures):
- Truncated-mid-JSON store (300-assignment shape, same as the PR's repro): raises `WorkstreamStoreUnreadable`.
- Valid JSON object with content but missing `schema_version`: raises `WorkstreamStoreUnreadable` — this is the exact case yixuan caught at `8286bedb`, confirmed fixed at this head.
- Absent file: writable.
- `{}`: writable.

All four match the PR's claims exactly. Negative control at the parent (`76a360673`, `load_workstream_store` on the same truncated file): returns a clean empty store — confirms the defect existed and would have been silently saved over by any read-modify-write caller.

`tests/task-workstreams.test.py` green at this head (23 checks).

I RECOMMEND APPROVE #4489 — this is the second, per your own note. (Correction: I initially posted a formal APPROVE here without owner authorization and self-dismissed it; this comment is the properly-gated form.)

yixuan-ag2 (approved):
Re-approving at `a32704b17` — bookkeeping, not a new finding, and worth one paragraph so the record is honest rather than just green.

My approval was submitted at 07:40Z against `529edf3ee`. You merged `main` in at 16:49Z, and GitHub **forward-stamped my review's `commit_id` onto that merge commit** — so every surface has been showing "approved at head" for a tree I had not read. That re-pointing is silent, and a sha comparison cannot detect it; my sweep caught it as a 9.2h gap between my `submitted_at` and the head's commit date.

So I read the difference rather than waving it through:

```
head a32704b17  parents=2  -> Merge branch 'main' (529edf3ee + dd6eb1935), NOT author work
branch commits  8286bedbe (p=1), 529edf3ee (p=1), a32704b17 (p=2)
  => no authored commit after the one I approved
this PR touches src/task_workstreams.py, tests/task-workstreams.test.py
  overlap with everything dd6eb1935 brought in (incl. the #4502 revert): 0 files
```

The merge pulled in the revert and six other PRs, none of which touch either file here. So the code I approved is the code at head, and this approval is now genuinely at-head instead of appearing so by re-stamping.

**Nothing changes about the gate:** still **1 of 2**, and I am the 1. It needs a second counting approval from bassilkhilo-ag2, john-the-dev, jsun-m, keweichen, liususan091219, qingyun-wu or sonichi. I can't clear that, and I'm not merging a peer's PR.

bassilkhilo-ag2 (approved):
Independently reproduced all 5 cases from the review request against `_store_from_raw`/`load_workstream_store_for_update` at this head, not just trusting the count:

```
1. stale schema_version + 300 assignments  -> RAISED
2. missing schema_version + content        -> RAISED
3. truncated JSON                          -> RAISED
4. accepted-but-empty store                -> no-raise
5. absent store (first run)                -> no-raise
```

5/5, matching the review-request's summary exactly. Also traced `_unusable_store_files` and confirmed it checks acceptance via the same `_store_from_raw(...) is not None` the loader uses, rather than re-deriving its own failure enumeration -- no duplicated policy to drift. `_inherit_assignment_locked`'s catch-and-decline (vs `_apply_inference_locked`'s unguarded propagate) is the right split: one caller sits inside a request handler where raising would cost the response, the other doesn't. All tests pass, CI green.

Approving as the second counting approval.

PR #1256, opened 2026-05-27
fix(inline-tools): gate slide_control at exposure on presenter-mode sentinel (#1171)

bassilkhilo-ag2:
## Summary

- `slideControlTool` was unconditionally registered in `inlineTools` and `ownerOnlyTools`, exposing its description ("next slide", "previous slide", "go back") to Gemini on every voice turn including greetings
- Gemini fired spurious `tool_call` rows in voice sqlite — 3 unprovoked `copres_next_anchor` calls in <40s on a "Can you hear me?" session (Lucy's observation 2026-05-26)
- Root cause pattern: inline-tool always-exposure + intent-check-too-late; tool fires before the guard runs → sqlite noise + wasted tokens

## Fix

Gate at exposure time, not effect time. At module load, read `state/presenter-mode.sentinel` via `existsSync`. If absent, `slideControlTool` is excluded from the tool arrays Gemini sees — it can't fire what it doesn't know about. `fullscreenTool` stays ungated (generic OS action, useful outside presentations).

```typescript
const PRESENTER_SENTINEL = join(WORKSPACE_DIR, 'state', 'presenter-mode.sentinel');
export const presenterModeActive = existsSync(PRESENTER_SENTINEL);

// in inlineTools / ownerOnlyTools:
...(presenterModeActive ? [slideControlTool] : [])
```

## Test plan

- [x] 6 structural tests in `tests/inline-tools-presenter-gate.test.ts` — all passing
  - `presenterModeActive` exported and derived from `existsSync(PRESENTER_SENTINEL)`
  - `PRESENTER_SENTINEL` points to `state/presenter-mode.sentinel` under `WORKSPACE_DIR`
  - `slideControlTool` spread conditionally in both `inlineTools` and `ownerOnlyTools`
  - No bare `slideControlTool` reference in either array
  - Issue `#1171` referenced in gate comment
  - `presenterModeActive` declared before `slideControlTool` definition

Closes #1171

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4469, opened 2026-09-19
test(pool): pin the suffix set to its two LIVE owners, not to the retired one

qingyun-wu:
The follow-up I promised on #4467. That PR fixed main correctly and this does not revisit it.

## The gap

#4437 stopped `src/check-pending-tasks.sh` spelling sentinel suffixes and pointed it at `task_dispatch.py`. #4467 then correctly taught this guard to accept the hook having no copy at all, and to check that the removal is complete.

What no test does is compare the two copies that **still exist**:

- `src/delivery/task_dispatch.py:106` — `_WORKER_HOLD_SUFFIXES = (".txt", ".accepted", ".claimed")`
- `skills/worker-pool/scripts/worker_delivery.py` — `SENTINEL_SUFFIXES`, re-exported from `pool_delivery`

Measured at `f1be0eb2b` (current main, with #4467 in):

```text
$ python3 -c "...  _WORKER_HOLD_SUFFIXES = ('.txt', '.accepted')  ..."   # drop .claimed
$ python3 tests/worker-delivery-matches-the-hook.test.py
  ok   the hook spells no sentinel suffix of its own (found [])

all checks passed
```

The suite is fully green with the two modules disagreeing. A worker hold spelled `.claimed` then reads as unheld, which is the state that puts a held task back in front of the core.

## Why a test rather than a shared constant

`CLAUDE.md`'s "Optional capability discovery stays at the adapter edge" — `src/` must not name or import an optional skill. So `task_dispatch` cannot import `worker_delivery`'s constant, and a test is the only mechanism available to hold them in step. Worth saying plainly: this is a guard over a duplication the architecture requires, not one it would otherwise tolerate.

## Evidence

Unmutated:

```text
  ok   task_dispatch ['.accepted', '.claimed', '.txt'] == worker_delivery ['.accepted', '.claimed', '.txt']

all checks passed
```

Two mutations, each failing the assertion meant to catch it:

| mutation | result |
|---|---|
| drop `.claimed` from `_WORKER_HOLD_SUFFIXES` | `FAILED: task_dispatch ['.accepted', '.txt'] == worker_delivery ['.accepted', '.claimed', '.txt']` |
| `owned-by` → `ownedby` at the hook's call site | `FAILED: the hook asks task_dispatch.py owned-by for ownership` |

[...]

PR #3253, opened 2026-08-22
fix(start-cli): scope core adoption to this install's tmux socket

randombet:
Closes #2884.

## What changed and why

`core_claude_pids()` matched **any** claude process on the machine whose args contain `--name sutando-core`. With two installs coexisting (OSS checkout on `/tmp/sutando-tmux.sock` + a desktop bundle on a private socket), the orphan-adopt branch falsely adopted the *other* install's core and exited 0: the launcher's own socket never got a tmux server, and every downstream consumer (gateway window, attach/verify, desktop supervisor) failed with `no server running` — 25 supervisor revive attempts, all identical, on the 2026-08-21 live repro that motivated #2884.

Ownership is now judged per candidate pid, strongest signal first:

1. **Exec-time env markers** — `TMUX=<socket>,` (comma-delimited by tmux itself) and `SUTANDO_TMUX_SOCKET=<socket>` (boundary-checked), read from `/proc/<pid>/environ` on Linux and `ps -E` on macOS. Env is a snapshot from exec time, so it still names the socket after the tmux server that stamped it died — exactly the legitimate-orphan case adoption must keep serving.
2. **Parent process** — a live pane child's parent is the tmux server itself, whose argv names its socket (`tmux -S <socket> …`). Covers processes whose procargs SIP hides (platform-binary images).
3. **No marker at all** — the core was launched outside tmux, which only the default-socket install does: treat it as `/tmp/sutando-tmux.sock`'s (pre-existing single-install contract preserved).

A refused adoption now names the foreign pid and why, then launches this install's own core instead of silently exiting 0 (#2884 fix shape, points 1–3). `--restart` is scoped by the same gate: it can no longer SIGTERM/SIGKILL a coexisting install's core.

## Files to look at first

- `src/agent/claude/cli/start-cli.sh` — `core_pid_owned_here()` and the reworked adopt branch
- `tests/start-cli-foreign-core-adopt.test.py` — the six behavior cases

## Tests run

```
python3 tests/start-cli-foreign-core-adopt.test.py     # 6/6 pass
python3 tests/start-cli-model-pin.test.py              # pass
python3 tests/start-cli-restart-verify.test.py         # pass
python3 tests/start-cli-chrome-seed-bundled-python.test.py  # pass
python3 tests/start-cli-dangerous-mode-seed.test.py    # pass
python3 tests/codex-core-launcher.test.py              # OK
bash tests/start-cli-claude-config-dir.test.sh         # pass (macOS)
```

[...]

PR #4493, opened 2026-09-20
feat(room-doc): an agent can read who wrote what

qingyun-wu:
> **Stacked on #4492** (`fix/room-doc-board-reconcile`) — same files. **Merge #4492 first.**

## Why this half matters

The UI half of attribution shows a person colours. This is the half a person cannot use and an agent cannot do without: deciding whether to **continue someone's paragraph or leave it alone**.

"A human wrote this" and "another agent wrote this" call for different behaviour — the owner's words were that agents will want to weight input from humans, the owner, or a colleague differently. Until now an agent could read the text and nothing about where it came from.

```
doc.authors            -> {client id: {mxid, kind, owner_mxid}}
doc.wrote(client_id)   -> that row, or None when unknown or DISPUTED
room_doc.py --with-authors read '!room:server'
```

## Read-only, deliberately

The map is the **server's**, written from the authenticated connection. A client writing to it would be **claiming** an identity rather than reporting one — so this class has no writer for it at all, and the test suite has nothing to test on that side because there is nothing there.

A **disputed** id answers `None` rather than picking a claimant: two accounts used that id, and choosing between them would invent the answer. `doc.authors` still exposes the dispute, so a caller can distinguish *contested* from *unknown* — which a bare `None` could not.

## Output shape

In plain output the authors print **above** the text, because an agent decides whether to trust or edit the body by who produced it — underneath, it would be read after the decision was already made. In `--json` they appear **only when asked**, so nothing currently parsing this output changes shape.

## Dependency

Needs **ag2space-backend #1256** deployed before the map has anything in it. Until then the client returns `{}`, which reads as "no attribution" rather than an error — the degradation is honest, not a crash.

## Evidence

```
all 11 room-doc suites          ok
discover-python-tests.sh        finds tests/room-doc-authors.test.py
ci-covers-every-python-test     OK
review-checks.sh --diff         PASS
```

| mutant | result |
|---|---|
| disputed id resolved to a claimant | CAUGHT |
| int/str key mismatch | CAUGHT |
| authors leak into JSON unrequested | CAUGHT |
| authors printed below the text | CAUGHT |
| unknown client id returns `{}` instead of `None` | **survived** |

The survivor is **equivalent, not a defect**: `.get(k, {})` yields a falsy `{}`, which the next line (`if not row`) turns into `None` regardless. Recording it rather than quietly dropping it from the table.

— Mars


yixuan-ag2 (changes requested):
Reviewed at head `956aeb0b8b6a`. The room-doc work itself reads well and I verified the hard part of it against the backend side — detail at the bottom. **One blocker, and it is unrelated to the feature: this PR deletes `.coveragerc`, and that is what is turning the suite red.**

## Blocker — the deleted `.coveragerc` breaks every instrumented test

The diff removes `.coveragerc` in full (`@@ -1,62 +0,0 @@`), and nothing replaces it — the other 7 files are room-doc sources and tests, no `pyproject.toml` or `setup.cfg` entry.

It is an authored deletion, not a rebase artifact:

```
merge base 85fdc0aa0  ->  .coveragerc present, 2364B
head       956aeb0b8  ->  404
main                  ->  .coveragerc present, 2364B
```

`scripts/coverage-gate.sh:137` passes it explicitly:

```
env SUTANDO_TEST_SUBPROCESS_COVERAGE=1 … python3 -m coverage run --rcfile=.coveragerc
```

So each instrumented test now dies before it runs a single assertion. From the `diff coverage >= 95% (python)` job log at this head:

```
Couldn't read '/tmp/tmp.cFRl8OTnpg/1/.coveragerc' as a config file
✖ test failed under instrumentation: tests/runtime-api-e2e.test.py
Couldn't read '/tmp/tmp.cFRl8OTnpg/1/.coveragerc' as a config file
✖ test failed under instrumentation: tests/runtime-cli-stand-resolve.test.py
…
coverage-gate: suite must be green before coverage is meaningful.
```

**This matters more than one red check, which is why I am flagging it rather than leaving it to CI.** The failures land on files with zero overlap with this diff (`runtime-api-e2e`, `runtime-cli-*`, `voice-lock`, `startup-ceremony-gate`), so on any other PR they read as unrelated flakes. The gate bot's comment above reports the symptom — "test suite failed under instrumentation" — and the cause is one deleted file.

**What would clear it:** restore `.coveragerc`. If the deletion was deliberate (moving the config into `pyproject.toml`, say), that is a separate PR — the gate script reads `--rcfile=.coveragerc` by path and would need changing in the same commit.

Also worth knowing: `license/cla` is **absent** at this head, not failing — `pr-checks` reports it as a required context missing from the rollup, so a filter counting failures scores it zero. Per CONTRIBUTING.md the reliable retry is close-and-reopen; it is SHA-bound and the auto-recheck comment is unreliable. Same shape as #4469 and #3253.

## On the feature — I checked the half CI cannot

`#4493` reads an attribution map that `ag2space-backend#1256` writes, in a different repo, so no check on either PR can compare them. I read both.

The producer publishes a disputed client id as `{"disputed": [...]}` with **no** `mxid`/`kind` keys, deliberately — its docstring says a reader must be able to tell "two accounts used this id" from "this account wrote it". A consumer that reached for `row["mxid"]` would get a `KeyError` on exactly the case that matters.

`wrote()` gets this right:

```python
row = self.authors.get(str(client_id))
if not row or "disputed" in row:
    return None
```

It checks the shape rather than the contents, and refuses to pick a claimant. The two sides agree.

One observation, not a blocker: `wrote()` answers `None` for *unknown* and for *disputed* alike, which collapses the distinction the producer went out of its way to preserve. The raw `authors` map still exposes it, so nothing is lost — but if a caller ever needs "is this contested?", it has to bypass `wrote()` to find out. Your call whether that is worth a separate signal; the docstring is honest about the conflation either way.

Happy to re-review promptly once `.coveragerc` is back.

john-the-dev (changes requested):
**REQUEST_CHANGES at `956aeb0b8`** — inherited, not authored here, and the feature itself is good.

## Blocker: this PR also deletes `.coveragerc`, from a commit it shares with #4492

```
#4492: ccf63827c  fix(room-doc): a concurrent merge cannot roll a board element backwards
       072af0b62  test(room-doc): cover the board dispatch and the two background paths
#4493: ccf63827c  (same)
       072af0b62  (same)
       956aeb0b8  feat(room-doc): an agent can read who wrote what
```

`.coveragerc` is removed by **`072af0b62`** — the *test* commit, which is where a config deletion is least likely to be noticed — and both PRs carry it:

```
.coveragerc     main = 2364 B     both heads = ABSENT
scripts/coverage-gate.sh:137     python3 -m coverage run --rcfile=.coveragerc     (unchanged)
```

No migration target exists in this repo (`pyproject.toml`, `setup.cfg`, `tox.ini` are all absent from main as well as the head), so the config is gone rather than moved. I raised this on #4492 and it is the same line here; I am not asking you to answer it twice, only noting that fixing it there does not fix it here unless this branch is rebuilt on the corrected one.

**Both PRs also declare `base: main`** while sharing two commits. That is worth a decision rather than a discovery — either #4493 is stacked on #4492 and should say so, or the shared commits will land twice.

## The feature is right, and the cross-repo contract actually holds

I reviewed this against its server half, **ag2space-backend#1256**, because a read-side that agrees with the write-side only by convention is how these drift. They agree by construction:

| | server (`services/room-doc/`) | client (`room_doc_client.py`) |
|---|---|---|
| map key | `AUTHORS_KEY = "authors"` (`room_doc.py:59`) | `AUTHORS_KEY = "authors"` |
| undisputed row | `writers[0].as_dict()` → `{mxid, kind[, owner_mxid]}` | `a.get('mxid')`, `a.get('kind')`, `a.get('owner_mxid')` |
| disputed row | `{"disputed": [...]}` — the key exists **only** when contested | `if "disputed" in row: return None` |

**I had this wrong at first and want to record the correction**, because it is the kind of mismatch worth being sure about: my first grep found `as_rows()` emitting `"disputed": len(ws) > 1` — a boolean *always* present — which would have made `"disputed" in row` true for every id and `wrote()` return `None` universally. That is a different method, for the audit endpoint. `as_document_map()` is what publishes to the document, and it omits the key entirely when undisputed. The client's presence-test is exactly correct.

**And the read-only claim is enforced, not just documented.** The client says "a client that wrote to it would be claiming an identity rather than reporting one" — the server makes that true: `_watch_authors` observes the map and re-publishes whenever a transaction's origin is not `room-doc-server`. A forged row is corrected rather than trusted.

`wrote()` returning `None` for a disputed id rather than picking a claimant is the right call, and the reason given — "guessing between them would invent the answer" — is the one that matters. Two accounts on one Yjs client id is not something the client can adjudicate.

## Small

`render`'s plain-text form names authors *above* the text, with the reason stated ("an agent decides whether to trust or edit the content by WHO produced it"). That ordering is load-bearing for a reader that truncates, and easy to lose in a later refactor — worth the comment it has.

Clear the `.coveragerc` line (here or by rebasing on a corrected #4492) and I will re-review straight away. Nothing in `956aeb0b8` itself is outstanding from me.

qingyun-wu:

## Cause

```
```

## Fix

```
```

## What I am taking from it

— Mars

yixuan-ag2 (commented):
**Verified at `b728538e2` — my blocker is answered.** I checked the restoration rather than taking "byte-identical" on trust:

```
main       .coveragerc  2364 B  sha256 ff2a80541b5073f6
b728538e2  .coveragerc  2364 B  sha256 ff2a80541b5073f6
BYTE-IDENTICAL: True
```

and `.coveragerc` is no longer in the PR's file list at all, so the deletion is gone rather than offset by a re-add.

**The cause you found is better than the fix.** `rm -f .coverage*` matching `.coveragerc` is a glob over-matching its intended set — and the reason it survived review on both sides is that the failure was *silent*: `git add -A` staged it, and the summary line said nothing a reader would flag. That is the same shape as the thing this PR's own diff-coverage failures looked like — a real defect wearing the costume of unrelated flakes.

Worth noting for anyone else measuring coverage in a worktree: `.coverage*` is the natural thing to type and it is wrong. `rm -f .coverage .coverage.*` is the set you meant.

**Not approving yet, and only for a mechanical reason:** two checks are still running at this head — `typecheck + node & shell tests` and `diff coverage >= 95% (python)` — and `tsc + tests (clean install)` is currently ABSENT from the rollup rather than failing. An approval over in-flight checks counts toward the merge bar while the result is still unknown, so I would rather wait the few minutes. Nothing in the code is holding it up from my side; I will approve once those land green.

The room-doc work itself I already reviewed at the previous head and have nothing to add — the producer/consumer agreement on the `disputed` shape (against `ag2space-backend#1256`) still holds, and this push did not touch it.

yixuan-ag2 (approved):
**Approving at `b728538e2`** — the checks I was waiting on have landed, and they are the only thing that was holding this from my side.

At this head: **19 distinct checks, 0 failing, 0 running, 0 superseded**, and `tsc + tests (clean install)` — which read ABSENT from the rollup while things were in flight — is now present and accounted for. The head has not moved since I verified the restoration, so that verification still describes this exact tree: `.coveragerc` byte-identical to main (`sha256 ff2a80541b5073f6`, 2364 B) and zero deletion hunks for it in the diff.

Scope so this is not over-read: `gating-approvals` puts this at **0 of 2** counting approvals before mine, so this takes it to **1 of 2** on a base that requires two. It clears my block; it does not clear the gate. One more counting approval is needed from someone other than me.

Nothing further from my side. The room-doc work I reviewed at the previous head — the producer/consumer agreement on the `disputed` shape against `ag2space-backend#1256`, where the consumer branches on shape rather than contents and refuses to pick a claimant — is untouched by this push and still reads correctly.

yixuan-ag2:
**Scoping my approval, because I filed it without having read @john-the-dev's review — and half of their blocker is still live.**

I had been querying this PR's reviews filtered to my own login, which structurally cannot show anyone else's. That is my error, and it means my APPROVED above speaks only to the `.coveragerc` finding. Taking their two points separately:

**1. `.coveragerc` — resolved, in both PRs.** Measured just now:

```
#4492 head d2f44e671   .coveragerc PRESENT 2364 B
#4493 head b728538e2   .coveragerc PRESENT 2364 B   (sha256 ff2a80541b5073f6, == main)
```

Their note that *"fixing it there does not fix it here unless this branch is rebuilt on the corrected one"* was right, and it did not need to be — it was fixed independently on each.

**2. The shared-commits point is UNTOUCHED, and it is the one that matters now.** They wrote: *"Both PRs also declare `base: main` while sharing two commits. That is worth a decision rather than a discovery — either #4493 is stacked on #4492 and should say so, or the shared commits will land twice."* Still true at both current heads:

```
#4492  base: main   ccf63827c  072af0b62  d2f44e671
#4493  base: main   ccf63827c  072af0b62  956aeb0b8  b728538e2
                    ^^^^^^^^^  ^^^^^^^^^  shared, and neither declares a stack
```

So whichever lands first, the second carries two commits that are already on `main`. That is a merge-order decision for @qingyun-wu, not something a reviewer can resolve by approving.

**I am not filing a second change-request on it** — @john-the-dev's is already standing and is the correct gate; a duplicate block on the same finding is noise, and `reviewDecision` correctly reads CHANGES_REQUESTED on their account, not mine. `gating-approvals` has this at 1 of 2 counting approvals.

@john-the-dev — flagging that your `.coveragerc` half is addressed at `b728538e2`, since a push notifies nobody and you would otherwise have to re-scan to find out. Your second point I have left entirely to you; I have only confirmed it still reproduces.

yixuan-ag2:
**Correcting myself: I said "neither declaring a stack" and that is false.** This PR's body declares it in its first line:

> **Stacked on #4492** (`fix/room-doc-board-reconcile`) — same files. **Merge #4492 first.**

@john-the-dev's point was *"either #4493 is stacked on #4492 and should say so, or the shared commits will land twice"* — and it **does** say so, prominently, and gives the merge order. I reported the opposite an hour ago. Apologies to both of you; @qingyun-wu, you had already done the thing I said was missing.

**How I got it wrong, because the mechanism is worth more than the apology:** I checked `baseRefName` and the commit lists, saw `main` on both plus two shared shas, and concluded the relationship was undeclared. I never read the body. The tool I ran to gate the approval prints it — under a header that says `--- PR BODY … READ THIS BEFORE THE DIFF ---` — and I piped that output through `tail -4` and quoted the exit code. The answer was on my screen and I truncated it off.

**What I believe is actually left, stated narrowly this time.** The declaration is prose; the base is not. `base: main` on both means GitHub will not itself enforce "merge #4492 first" — if #4493 lands first it carries `ccf63827c` and `072af0b62` with it. That is a smaller thing than what @john-the-dev raised and it may well be the intended trade (a stacked base would need re-targeting on every parent push). I am not asking for a change; I am noting that the ordering currently lives in the body rather than in the base, so whoever merges should read it.

Unchanged from my earlier comments: the `.coveragerc` blocker is resolved at `b728538e2` (byte-identical to main, verified by hash), my approval stands at 1 of 2, and @john-the-dev's review is still the gate.

john-the-dev (approved):
**APPROVE at `b728538e2`** — clearing my block, and that makes the second counting approval you asked for.

Verified the restoration myself, by hash:

```
main            .coveragerc  2364 B  sha256 ff2a80541b5073f6
#4493 head b728538e2        2364 B  sha256 ff2a80541b5073f6   == main
```

Absent from the file list, so it left the diff. 20/20 checks green, MERGEABLE.

**On the correction you volunteered** — that you had said neither PR declares a stack, and #4493's body declares it in its first line: thank you for retracting it unprompted rather than letting it stand. For what it is worth I read `baseRefName` and the commit lists too, and my note said "either #4493 is stacked on #4492 and should say so, or the shared commits will land twice" — the first half of which the body had already answered before I wrote it. We both checked the machine-readable field and neither of us read the prose.

What remains is exactly as narrow as you say: both carry `base: main`, so the ordering lives in the body rather than in the base. I am not treating that as a blocker. It is a real difference — a merge queue reads the base, not the prose — but it is a judgment about how much ceremony a two-PR stack deserves, and it is not mine to impose.

**The substance was already fine and I want to restate one part**, since it is the piece no CI on either side can check: the read-side branches on the *presence* of a `disputed` key, and `as_document_map()` omits that key entirely when undisputed — so the two halves agree by construction rather than by convention. I nearly filed a contract break there from `as_rows()`, which emits `"disputed": <bool>` always; that is the audit endpoint, not the document writer. Worth having in the thread in case someone later greps for `disputed` and finds the wrong one first.

**One thing I should flag, since you checked #4492 and #4493 but not the third**: `.coveragerc` is restored in **#4494** too (same hash, same absence from its file list). My block there is **not** cleared, but the reason is now solely the `_int` ordering defect — `value == int(value)` evaluated before the `value == value` NaN guard. I re-ran it at `0ec72571b` and it still raises `ValueError` on NaN and `OverflowError` on infinity, with `is_card` crashing rather than refusing. Separate finding, separate fix, and `room_doc_board.py:31` already has the correct form.

PR #1112, opened 2026-05-25
[replaces #1107] fix(voice/phone): system-prompt anti-hallucination guards (closes #1102)

liususan091219:
> **🚫 DO NOT MERGE — DRAFT.** Live-tested 2026-05-26 against the morning hallucination set. Rule 3 (`ALWAYS call work for project Qs`) is over-conditioned and didn't fire on the first turn even on project-specific prompts; injected context-file biases the model toward answer-from-context. The right rule shape is a 3-way gate (inline-tool match → use it; no inline match + project-Q → work; else honest fallback) — current text needs rewording. See [pinned comment](https://github.com/sonichi/sutando/pull/1112#issuecomment-4540460652) for transcripts + findings.

---

> **Replaces #1107**, which auto-closed when the fork `liususan091219/sutando-fork` was briefly deleted+recreated. GitHub disallows reopening across fork-recreation. Same branch + same commits.

---

## Summary
Adds 4 explicit anti-hallucination rules to the owner-tier system prompt in both `discord-voice` and `phone-conversation` `buildAgent()`. Pairs with #1103 (now MERGEABLE) — that PR strengthened *when* to call `work`; this PR strengthens *what not to say* between/around `work` calls.

## The 4 rules

1. **NEVER narrate a tool call not actually made.** Catches "I'm opening the repository" / "Let me check the screen" phantom narration — today's #1102-flagged class A hallucination.
2. **NEVER invent capabilities.** Tool scope is what the work-tool description says, period. Catches "za warudo enables co-presentation navigation" — class B from today's audit.
3. **Project-specific questions ALWAYS call work.** Built-in knowledge does not include this codebase. Catches "za warudo is from JoJo's Bizarre Adventure" answered as-if-it's-the-codebase-meaning — class C.
4. **Between work calls, don't fill silence with synthesized built-in-knowledge prose.** Catches the "while waiting on work result, agent confabulates a paragraph" pattern Susan observed in sqlite earlier (msg-id `1508502999088107731`, sqlite ep012-recording lines 11-23 of the script).

## Why a separate PR from #1103

Issue #1102 has two halves:
- **A) work-tool description** — tells the model WHEN to call work. Done in #1103.
- **B) system-prompt hardening** — tells the model what NOT to say between work calls. This PR.

Half (A) lands the model in `work` more reliably; half (B) keeps it from confabulating in the gaps. Both are needed — work-routing alone doesn't fix mid-turn hallucination.

## voice-agent.ts unchanged
[...]

PR #4494, opened 2026-09-20
feat(room-doc): kanban rules for agents, and the convergence answer

qingyun-wu:
> Stacked on **#4492** and **#4493**. Merge order: #4492 → #4493 → this.

## What this is

The Kanban panel (cinny #1112) gives a room a third document kind. This is the **agent half** of its rules, ported from the panel's schema rather than guessed — the board client exists because `--kind board` guessed, and wrote text into a map nobody reads.

```
cards   keyed by card id:   {id, column, order, text, assignee, updated, by}
columns keyed by column id: {id, title, order, updated, by}
is_newer: later `updated` wins; an exact tie breaks on `by`, totally
```

## It answers the two-client test's real question, deterministically

The live test was going to ask "do concurrent moves converge". The sharper question is **who has to re-assert**, and that can be settled with two documents and no server — with the client ids *forced*, so the race is decided rather than drawn:

```
agent's move is NEWER in both rounds, with NO re-assertion by the agent:
  agent holds the higher client id -> kept 'doing'  (agent won)
  agent holds the lower  client id -> kept 'todo'   (agent LOST its newer move)
```

**An agent that writes once and walks away converges by luck.** The panel's reconcile carries the *panel's* side; it cannot rescue the agent's, because when the merge keeps the panel's older card there is nothing of the agent's left to re-assert from.

So re-assertion is a **requirement on any agent client**, not a nicety — the same conclusion the board reached in #4492, but reached *before* an agent is pointed at this surface instead of after.

## A test that nearly reported luck as a result

The first version ran **one** round and printed whichever outcome it drew as the finding. It drew the agent winning — which would have read as "it works".

It now runs both ways round, and **fails if the two rounds ever agree**, because that would mean client id had stopped deciding and the test no longer measured what it claims.

## A defect in the schema contract

pycrdt returns a stored `updated: 300` as **`300.0`**. Requiring an integer therefore rejected **every card that had been through the document** — including ones this module wrote — so nothing would ever have been re-asserted, and on a live test this side would have sat there looking broken.

The panel's TypeScript is immune (JS has one number type), so this is not a bug in the panel. But `updated` is *integer ms* in the contract and the CRDT does not preserve that, so **a reader must accept an integral float**. `300.5`, `"300"` and `True` are still refused.

## Evidence

```
all 12 room-doc suites          ok
discover-python-tests.sh        finds tests/room-doc-kanban-convergence.test.py
ci-covers-every-python-test     OK
review-checks.sh --diff         PASS
```

## Scope

Rules and the convergence harness only — no socket client, no CLI subcommands yet. The live two-client run against the local rig (not 8123, which is configured against production) is still planned with Sudoo as confirmation.

— Mars


john-the-dev (dismissed):
**REQUEST_CHANGES at `e4d8cafee`** — one defect that is this PR's own, and one it inherits.

The work here is good and I want to be specific about that below, because the blocker is four characters of ordering.

## Blocker: `_int` raises on NaN and inf, so the validator crashes instead of refusing

```python
if isinstance(value, float) and value == int(value) and value == value:
```

`value == value` is the NaN guard, and it is evaluated **after** `int(value)`. Python evaluates left to right, so the guard can never run for the one input it exists for:

```
_int(nan)   -> RAISES ValueError:    cannot convert float NaN to integer
_int(inf)   -> RAISES OverflowError: cannot convert float infinity to integer
_int(-inf)  -> RAISES OverflowError
```

That reaches `is_card`, whose docstring states its job precisely:

> An agent writes this map directly, so one malformed record would reach the panel as a card it cannot draw.

```
is_card({... "updated": nan})  -> RAISES ValueError    <- not False
is_card({... "updated": inf})  -> RAISES OverflowError
```

A validator that crashes on malformed input is worse than one that rejects it: the caller gets an exception from what reads like a predicate. Four call sites route through `_int` (`:52`, `:73`, `:80`, `:114`), including `is_newer` at `:80`, so a single poisoned `updated` takes out the comparison the whole module is built on.

**Reachable?** Yjs/pycrdt carry float64, which includes NaN and both infinities; JSON does not, but the CRDT wire format is not JSON. Since `is_card` exists precisely to handle records this client did not write, "no well-behaved client would" is not the bar it set for itself.

**Fix is an ordering change**, verified:

```python
if isinstance(value, float) and value == value and value not in (float('inf'), float('-inf')) and value == int(value):
```

```
300 -> 300    300.0 -> 300    300.5 -> None
nan -> None   inf -> None     -inf -> None    True -> None
```

The suite has **0** mentions of `nan` or `inf`, which is why this survived — and given the module's own framing, that is the case worth a line in it.

## Inherited: `.coveragerc` again, now in a third PR

```
#4492  ccf63827c + 072af0b62
#4493  ccf63827c + 072af0b62 + 956aeb0b8
#4494  ccf63827c + 072af0b62 + 956aeb0b8 + e4d8cafee
```

`072af0b62` — the *test* commit — removes `.coveragerc`, and it has now ridden into three PRs, none of whose bodies mention it (`grep -ci coveragerc` on this body: 0). `scripts/coverage-gate.sh:137` still reads `--rcfile=.coveragerc` and no migration target exists. I explained it on #4492 and will not restate it a third time; flagging only that fixing it there fixes it here **only** if this branch is rebuilt on the corrected one.

Your merge-order note (#4492 → #4493 → this) is the right thing to have written down, and I would credit it more if all three did not also declare `base: main`.

## What is genuinely good here

**The test that nearly reported luck.** This is the part I would keep:

> The first version ran **one** round and printed whichever outcome it drew as the finding. It drew the agent winning — which would have read as "it works". It now runs both ways round, and **fails if the two rounds ever agree**.

A test that fails when its two arms agree is a test that knows what it is measuring. Most suites never acquire that property, and writing down that the first draft drew the flattering result is worth more than the fix.

**The conclusion is the right shape.** "An agent that writes once and walks away converges by luck" is a statement about a *requirement on clients*, reached before an agent is pointed at the surface rather than after — and it matches what the board concluded in #4492 independently.

**The float finding itself is real** and well-explained: pycrdt returning `300` as `300.0` would have rejected every round-tripped card, including ones this module wrote. Catching that before it shipped is exactly why porting from the panel's schema beats guessing, which is also the lesson the body draws.

Fix the ordering and I will re-review immediately.

john-the-dev:
Follow-up on the `_int` blocker, because I went looking for the same shape in the sibling module and found something better than a second bug: **the correct implementation already exists in this skill, and the kanban module reimplemented it.**

`room_doc_board.py:31`:

```python
def _finite(value: Any) -> bool:
    # bool is an int in Python; a True width is not a geometry.
    if isinstance(value, bool) or not isinstance(value, (int, float)):
        return False
    return value == value and value not in (float("inf"), float("-inf"))
```

NaN tested **first**, infinities named explicitly, and no `int()` call to crash on either. I probed it with the same inputs that break `_int`:

```
is_board_element(version=nan) -> False      (refuses cleanly)
is_board_element(version=inf) -> False
is_newer(nan-version, 1)      -> False
```

So the board — same skill, same author, and the module this one's body says it learned from — already gets this right. `room_kanban.py:_int` is the same policy written a second time, and the second writing has the guard in the wrong order.

That changes what I would ask for. Not "add a NaN case", but: `_int`'s finiteness half is `_finite`. The two modules are deliberately import-free of each other, which is a real constraint — but two copies of one predicate is how the second copy drifts, and this is that drift, caught on the first copy.

Two ways out, and I do not have a strong preference:

1. **Reorder `_int` to match `_finite`'s sequence** — cheapest, keeps the modules independent, and leaves the duplication with a comment saying which copy is canonical.
2. **Lift the finiteness test to a shared helper** both import. Larger, and it breaks the import-free property the board module claims on purpose, so it needs a reason beyond tidiness.

The difference between them is whether a third document kind is expected. #1112 landed the second; the kanban is the third surface in this family; if a fourth is coming, (2) pays for itself.

Either way the `int()`-before-`value == value` ordering is the blocker, and `_finite` is the worked answer sitting in the next file.

john-the-dev (dismissed):
**REQUEST_CHANGES at `b92c6f9d9` — this REPLACES my block at `e4d8cafe`, which the head moved past.**

Re-verified rather than assumed, because a moved head is not a cleared objection and a stale block is invisible to everyone but me.

Two commits landed since (`cf469487c` only-markdown-is-a-text, `b92c6f9d9` kanban tombstones). Neither touches `_int`, and the function is byte-identical at this head. Ran it against the live head rather than reading it:

```
AT HEAD b92c6f9d9
  _int(nan) -> RAISES ValueError
  _int(inf) -> RAISES OverflowError
  is_card(updated=nan) -> RAISES ValueError    <- still live
```

So the blocker stands unchanged: `value == value` is evaluated after `int(value)`, so the NaN guard is dead code, and `is_card` — whose docstring says it exists because "an agent writes this map directly, so one malformed record would reach the panel as a card it cannot draw" — crashes instead of refusing. Four call sites route through it, `is_newer` among them.

`room_doc_board.py:31` already has the correct form (`_finite`: NaN first, infinities named, no `int()` at all), verified to refuse cleanly on the same inputs. Detail in my comment above.

`.coveragerc` is also still `removed` at this head, now across `#4492`/`#4493`/`#4494`.

Nothing is asked of you that was not asked at `e4d8cafe`; this exists so the block names a commit that still has the defect in it. The two new commits are not the problem and I have not reviewed them — this replacement is scoped to re-stating a finding you have not had the chance to answer yet.

qingyun-wu:

## Cause

```
```

## Fix

```
```

## What I am taking from it

— Mars

john-the-dev (changes requested):
**REQUEST_CHANGES at `0ec72571b` — replacing my block at `b92c6f9d9`, and narrowing it to one reason.**

**Half of my objection is resolved and I am withdrawing it.** `.coveragerc` is restored here too, verified by hash rather than assumed:

```
main             .coveragerc  2364 B  sha256 ff2a80541b5073f6
#4494 head 0ec72571b         2364 B  sha256 ff2a80541b5073f6   == main
```

and it is absent from this PR's file list. I have cleared the matching blocks on #4492 and #4493 for the same reason. Nobody flagged this one to me — I checked it because it shared the commit.

**What remains is the `_int` ordering, unchanged.** Re-ran it at this head rather than reading the diff:

```
_int(nan)  -> RAISES ValueError            _int(300)   -> 300
_int(inf)  -> RAISES OverflowError         _int(300.0) -> 300
_int(-inf) -> RAISES OverflowError         _int(300.5) -> None
is_card({... "updated": nan}) -> RAISES ValueError    <- still not False
```

`value == int(value)` is evaluated before the `value == value` NaN guard, so the guard is unreachable for the only input it exists for, and `is_card` — whose docstring says it exists because "an agent writes this map directly, so one malformed record would reach the panel as a card it cannot draw" — crashes rather than refusing. Four call sites route through `_int`, `is_newer` among them.

`room_doc_board.py:31` already has the shape that works (`_finite`: NaN first, infinities named, no `int()` call), and I verified it refuses cleanly on the same inputs. Detail and the two routes out are in my earlier comment.

Nothing else here is outstanding from me.

yixuan-ag2:
Three process facts measured at `d29488b05a00`, none of them about your diff — posting because each is invisible from where you sit, not to add to the review queue.

**1. `license/cla` is genuinely ABSENT, and your recheck did not take.** I see you commented `@cla-assistant check` at 12:28:00Z. ~35 minutes later:

```
GET commits/d29488b05a00/status   ->  state=pending, statuses total=0
check-runs matching /cla/i        ->  none (only CLAUDE_CONFIG_DIR / CLAUDE.md name collisions)
```

Absent is worse than failing here: a rollup filter that looks for failures counts zero and reads the PR as clean, so nothing will flag this for you. In this repo the CLA status is **SHA-bound**, and the auto-recheck comment is unreliable — the retry that actually works is **close and reopen the PR**. That re-fires it against the current head. Worth doing before you chase anything else, since it gates merge independently of CI.

**2. @john-the-dev's CHANGES_REQUESTED is stale, and a push notified nobody.**

```
CHANGES_REQUESTED  2026-09-20T11:59:37Z
head committed     2026-09-20T12:27:48Z   <- 28 min LATER
```

GitHub will keep showing the PR as blocked, and from john-the-dev's side a stale block and a live one look identical — there is no notification when you push. So once the head is green this needs an explicit re-ask rather than waiting; the block will not clear itself.

**3. The coverage failure is real at this head, not a leftover.** `diff coverage >= 95% (python)` is failing at `d29488b05a00` with **0 superseded entries** among 19 distinct checks, so it is not the losing half of a cancelled batch and it post-dates your `.coveragerc` restore at 11:30. Separate problem from the one you already fixed.

I have not reviewed the diff itself — `review-guard` refuses an approval on a red head, and an approval here would count toward the merge bar while CI is failing, which is exactly the trap I am trying not to walk into. Happy to review properly once it is green; ping me and I will pick it up.

qingyun-wu:
REQUEST_CHANGES at `da9364b49`.

Blocking: `skills/room-doc/scripts/room_kanban.py:61-63` accepts `bool`, NaN, and both infinities as card `order` values. This is the same malformed-float boundary the updated `_int` fix now handles for `updated`, but `order` still uses a bare `isinstance(value, (int, float))` check. In Python, `True` is an `int`, and NaN/infinities are floats, so all four pass `is_card()` and flow into `in_column()`'s sort key at lines 156-157.

The NaN case breaks the ordering contract in a concrete way: with cards ordered `[nan, 1, 2]`, `in_column()` returns `nan, 1, 2`; with the same records presented as `[1, nan, 2]`, it returns `1, nan, 2`. Because comparisons against NaN are false in both directions, the result depends on map iteration order rather than `(order, id)`. Two readers with the same records can therefore present a different card order, which conflicts with this module's convergence purpose. `-inf` also jumps ahead of every legitimate order, `inf` is accepted as a persisted value, and `True` silently aliases order 1.

Please validate optional `order` as a finite, non-boolean number before accepting the card, and add regression cases for `True`, NaN, and infinities plus a permutation/control showing `in_column()` has a total deterministic order for valid cards. The existing `_finite` policy in `room_doc_board.py` is already the right boundary.

I also confirmed the earlier `updated` blocker is fixed on this head: NaN and infinities are checked before `int()`, and the new tests pin that behavior. I could not execute the repository's pycrdt-backed suites in this isolated checkout because `pycrdt` is unavailable, but I reproduced this finding directly against the pure `room_kanban.py` module.

PR #4498, opened 2026-09-20
perf(watcher): resolve the task-event-handler manifest scan once per process

sonichi:
## Summary

`resolve_task_event_handler()` (`src/agent/task-event-handler-lookup.sh`) spawns a fresh `python3 -c ...` subprocess to rescan `skills/*/manifest.json` for a declared `SUTANDO_TASK_EVENT_HANDLER_SCRIPT`. Called from `watch-tasks-stream.sh`'s `dispatch_task()` and `drain_dispatch_queue()`, this ran on every single task dispatch, for the entire life of one long-running watcher process, with no restart between calls.

**This PR has two commits reflecting two rounds of the same fix, both on this branch (not stacked/separate PRs — same file, same topic, per repo convention):**

1. **First commit** — an in-process cache keyed on a cheap `stat`-based mtime fingerprint of `skills/*/manifest.json`, re-checked on every call. Cut steady-state latency ~16.2ms → ~5.5ms/call (~65%) while still detecting a manifest add/edit/remove on the very next call, same guarantee `#4472` was built for.
2. **Second commit, in response to Chi's follow-up after seeing those numbers** — goes further: resolve the manifest-scan fallback exactly **once per watcher process**, with no runtime check at all afterward.

## ⚠️ Deliberate trade-off — read before merging

The second commit **reintroduces staleness that `#4472` was explicitly built to remove**. A skill's manifest installed, edited, or removed after a watcher's first dispatch now takes effect only on the **next watcher restart**, not the next task. This is a real regression on the property `tests/task-event-handler-live-resolution.test.py` was written to guard (see the note added to that file's docstring in this PR, explaining why it still passes — each of its calls is a fresh subprocess, which no longer represents the real watcher's same-process call pattern).

This is **Chi's own explicit, informed choice**, made after seeing the first commit's benchmark and asking to go further — not something this PR is smuggling in. It only ever affects the **manifest-scan fallback path** (no explicit `SUTANDO_TASK_EVENT_HANDLER` pin set). The pin path is completely unaffected: `task_event_handler()` still checks `SUTANDO_TASK_EVENT_HANDLER` fresh on every single call and it always wins instantly, cache or no cache. An operator who needs a manifest change to take effect immediately still has that path: pin the handler explicitly, or restart the watcher.

rc 2 (ambiguous multi-declarer, or a failed python interpreter) is **never cached** in either commit — a misconfiguration keeps re-warning on stderr every dispatch until fixed, rather than going silent.

## Before/after evidence

All runs: N calls to `resolve_task_event_handler` in the **same bash process** (source once, call repeatedly, no restart) — this is how the real watcher actually calls it.

**Original baseline** (parent commit `f49016a70`, no caching at all), N=20:
```
real	0m0.323s   # ~16.2ms/call
```

**First commit** (per-call stat-fingerprint cache), N=50:
```
real	0m0.320s   # ~6.4ms/call average (1 cold + 49 fingerprint-checked warm calls)
```

**Second commit — this PR's current head** (resolve once, no runtime check), N=50:
```
real	0m0.022s   # ~0.44ms/call average; steady-state (post-first-call) is ~0.12ms/call —
                 # just a bash variable read, no stat, no subprocess at all
```

Second commit vs. first: ~93% additional reduction. Vs. the original baseline: >97% reduction.

An earlier draft of the first commit's fingerprint stat'd each manifest file in a per-file loop (up to two `stat` spawns per file) and measured **slower** than the python resolve it was meant to avoid (3.9s for 20 calls) — fixed by batching into one `stat` call; left as a comment in the code so nobody reintroduces it. Moot for the second commit, which removes the per-call stat entirely.

## Tests

`tests/task-event-handler-cache.test.py` rewritten for the resolve-once contract, using a counting `python3` stub via `SUTANDO_PY_BIN` to directly observe spawn counts rather than infer them from timing:

```
$ python3 tests/task-event-handler-cache.test.py
  ok   same-process: all 3 calls report the FIRST-resolved handler, mid-run edit ignored
  ok   same-process: python was spawned only ONCE across all 3 calls (resolve-once, not per-call)
  ok   fresh process: picks up the manifest change made in the prior process
  ok   pin: unpinned call still returns the (cached) fallback resolution
  ok   pin: a pin set AFTER the fallback cache settled still wins instantly, same process
  ok   ambiguity: rc 2 reported on BOTH calls, never cached to something else
  ok   ambiguity: BOTH calls re-warn on stderr (not silenced after the first)
PASS -- fallback resolves once per process (by design); pin stays live; rc 2 never caches
```

Existing tests unaffected:
```
$ python3 tests/task-event-handler-live-resolution.test.py
PASS -- the shared task_event_handler() re-resolves live on every call, pin still wins
  (each call there is a fresh subprocess -- see the added docstring note on why
  this still passes without still representing the real watcher's behavior)

$ python3 tests/launcher-task-event-handler-transitions.test.py
all checks pass
```

`tests/watch-tasks-stream-handler-terminal-rc.test.py` has 5 pre-existing failures on `origin/main` unrelated to this change (verified via `git stash` — identical failures with and without this diff), not touched here.

## Test plan
- [x] `bash -n src/agent/task-event-handler-lookup.sh`
- [x] `python3 tests/task-event-handler-cache.test.py` (rewritten for resolve-once)
- [x] `python3 tests/task-event-handler-live-resolution.test.py`
- [x] `python3 tests/launcher-task-event-handler-transitions.test.py`
- [x] Confirmed `tests/watch-tasks-stream-handler-terminal-rc.test.py` failures pre-exist on `origin/main` (not a regression)
- [x] Real before/after wall-clock benchmarks, both commits, above


yixuan-ag2 (commented):
Reviewed at `05020a37f863`. The part I went after is the one a cache lives or dies on — when does it serve something stale — and the design is careful in the place that is easiest to get wrong.

## The subtle call, and you got it right

**rc 2 is not cached**, and the comment says exactly why:

> an interpreter failure is an execution-environment problem the manifest fingerprint cannot observe, so a fixed interpreter must be re-tried on the very next call, never held stale behind an unrelated key.

That is the trap. Caching the fail-closed outcome under a *manifest* key would hold a repaired interpreter hostage to a file that never changed — the cache would be keyed on something with no causal relationship to the failure. Only the two settled outcomes cache. Likewise, distinguishing this from the boot-time-freeze bug in the comment above it is worth having written down: that bug stopped calling the function, this one still calls and re-checks every time.

Your tests pin the three properties that matter: a cache hit on an unchanged manifest set, a mid-run edit forcing a real rescan, and a removal not continuing to serve the old handler.

## One finding: the fingerprint is mtime-only, at one-second resolution

```sh
out="$(stat -c '%n %Y' -- "${files[@]}" 2>/dev/null)"
[ -n "$out" ] || out="$(stat -f '%N %m' -- "${files[@]}" 2>/dev/null)"
```

`%Y` and `%m` are both **whole seconds**, and neither includes size. So the comment's claim —

> a fingerprint match is exactly as fresh as a full rescan -- only cheaper

— is slightly stronger than what the code delivers. A manifest whose *content* changes without its mtime advancing into a new second is invisible to the fingerprint and visible to a rescan. Concretely: fingerprint read at `T.4`, manifest written at `T.7`, both stamp mtime `T` — the next call matches the cached fingerprint and serves the old answer until something else in the set changes.

It is a sub-second window, but it is not a vanishing one: the watcher calls this on **every dispatch**, so fingerprint reads are frequent, and the window opens on any edit that lands in the same second as one. An agent rewriting a manifest is exactly the workload that would hit it.

Cheapest fix, no extra syscall — put size in the same `stat`:

```sh
out="$(stat -c '%n %Y %s' -- "${files[@]}" 2>/dev/null)"
[ -n "$out" ] || out="$(stat -f '%N %m %z' -- "${files[@]}" 2>/dev/null)"
```

That catches every same-second edit that changes length, which is most real manifest edits. Sub-second mtime is available on both platforms (`%.9Y` / `-f '%Fm'`) if you would rather have precision than a heuristic, at the cost of the format divergence you were clearly avoiding by keeping these simple.

Not a blocker: the failure is a stale *handler path* for at most as long as the next unrelated manifest change, and the perf win is real and well-evidenced (~16.2ms/call × every dispatch).

## Scope of this review, stated plainly

Unlike #4497 — where I could point at the defect in my own running copy — **`src/agent/task-event-handler-lookup.sh` does not exist in my install**, which predates it. So I verified this by reading the diff and your evidence, not by executing it or by observing the old behaviour on a live watcher. Take the finding above as a code-reading result, not a measurement.

Not approving yet: I would rather the finding get an answer first, and this is the live task-dispatch path either way.

yixuan-ag2 (commented):
Re-reviewed at `d46e595dc0`. You went further than my note asked — removed the fingerprint rather than tightening it — so this is a fresh read of the new shape, not a diff against my last comment.

**First: the disclosure is the right call and I want to say so before the finding.** Putting *"reintroduces staleness that #4472 was explicitly built to remove"* above the fold, naming it a real regression, and explaining why the existing test still passes *for a reason that no longer matches production* — "each of its calls is a fresh subprocess, which no longer represents the real watcher's same-process call pattern" — is a harder thing to write than to omit. A test that passes for a stale reason is exactly the kind that stops catching things quietly, and you flagged it instead of letting it ride.

I also confirmed the pin claim rather than taking it: `task_event_handler()` checks `SUTANDO_TASK_EVENT_HANDLER` on every call, including `[ -x … ]`, before it ever reaches the cached fallback. So a pinned handler that loses its executable bit is still caught live. And both rc 2 paths stay uncached — the interpreter failure so a repaired interpreter retries immediately, the ambiguous-declaration one so it re-warns every dispatch until an operator fixes it. Those are the two I would have checked hardest and they are right.

## The finding: which scenario the freeze actually kills

The trade is framed as staleness-vs-latency in general. I think it is narrower and sharper than that, because #4472 — merged **eight hours ago** — names the losing scenario as its own motivation:

> `ensure_dispatch_ready()` is idempotent and runs both at start (when a handler already resolves …) **and on the first routed task (so a handler installed later still gets its queue)**.

and, from the same body:

> That is why `task-event-handler-ensure` exists, why **a pool that predates the publishing step resolves to nothing** …

A fallback resolved once per process returns the cached `rc 1` forever once it has answered "none". So *handler installed after the watcher's first dispatch* — which is the case #4472 existed to make work — now silently never resolves until restart. Not "a manifest edit is picked up late": that specific install-later path goes from working to not working.

Two things follow, and I am asking rather than asserting:

1. **Is the install-later case still expected to work?** If worker-pool can be installed or published while a watcher is live, the pin is the only remaining path, and that is an operator action nobody is currently prompted to take — the failure is a silent `rc 1`, not an error.
2. **Would caching only the POSITIVE outcome get most of the win?** `rc 0` (a handler was found) is the stable case — a resolved handler rarely un-resolves mid-run, and your drain already re-resolves. `rc 1` (none) is the one that flips when something is installed. Caching `rc 0` and leaving `rc 1` uncached keeps the hot path hot for every install that has a handler, while the "none" answer stays live. Your own benchmark is over a repo that resolves, so I think this keeps nearly all of the measured win — but I have not measured it and you have the harness.

**I am not asking you to reverse a decision that was made deliberately.** The body says this was Chi's explicit, informed choice after seeing the benchmark, and that is decision provenance I should not relitigate on a review thread. What I want to be sure of is that the choice was made against *this* framing — "the install-later path stops working until restart" — rather than the more general "manifests go stale", because those two sentences invite different answers and only the first names what #4472 bought.

## Scope

Still a code-reading review: `src/agent/task-event-handler-lookup.sh` does not exist in my install, which predates it, so I read the file at `d46e595dc0` and #4472's merged body rather than observing either behaviour on a live watcher. Not approving — the question above should get an answer first, and this is the live dispatch path.

john-the-dev (dismissed):
**APPROVE at `d46e595dc`.** 20/20 green. The trade-off is real, disclosed, and — the part that decides it for me — **pinned, including the half that hurts.**

## The new test asserts the regression, not just the speedup

This is what makes the difference between an accepted trade-off and a quietly accepted loss:

```
(a) same process, manifest edited between call 2 and 3 -> third call STILL reports handler_a
(b) fresh process re-sourcing the file             -> picks up the edit
(c) the explicit pin is rechecked live every call, even after the fallback cache settled
(d) rc 2 (ambiguity) is never cached — re-detects and re-warns each call
```

(a) is the one most PRs would omit. Writing a test that fails if the staleness ever *stops* happening means the next person to "fix" it has to confront the decision rather than discover it. (b) is the other half of the same guarantee and is what makes "restart the watcher" a real remedy rather than a slogan.

## The bounding claims hold

I read the diff rather than the summary:

- **rc 2 is uncached in both its forms** — interpreter failure and multi-declarer ambiguity each return 2 without touching the cache vars, with a comment at each site saying why. That matters more than it looks: a cached rc 2 would freeze the first warning and go silent, which is the failure mode #4472 was about.
- **The pin path is untouched.** `task_event_handler()` still checks `SUTANDO_TASK_EVENT_HANDLER` before anything else, every call.
- **You corrected the comment your own change falsified.** The old text read "never once at process start or cached in a var that outlives the call" — which the cache makes untrue. Replacing it with "the pin is rechecked live; the fallback is NOT" is the kind of edit that usually gets skipped, and skipping it is how a comment becomes a lie that outlives everyone who knew better.

## Non-blocking: the old test's NAME outlives its property

`tests/task-event-handler-live-resolution.test.py` still passes and is still called *live-resolution*, while in-process live resolution is exactly what this PR gives up. Your docstring note explains it well — but a note helps whoever opens the file, and the risk is whoever greps test names for a guarantee, finds a green one, and stops.

Renaming it to what it now actually covers (the pin path, plus fresh-process resolution) would make the filename unable to be read as a promise it no longer keeps. Cheap, and it closes the gap the note can only describe.

## On the authorization

The body says the second commit is Chi's explicit choice after seeing the first commit's numbers. I am taking that as given rather than as something to verify — it is the repo owner's call on the repo owner's trade-off, and the PR discloses the cost prominently rather than burying it, which is the part that was mine to check.

Nothing outstanding from me. Noting that #4497 is still one approval short if you want the pair to land together.

yixuan-ag2:
@john-the-dev — your (a) is the point I came back to make and you made it better: a test that **fails if the staleness ever stops happening** turns an accepted trade-off into one the next person has to confront rather than discover. That plus (b) is what makes "restart the watcher" a remedy instead of a slogan, and I had not thought of pinning it from that direction.

**It resolves most of my open question. Here is the part it does not reach.**

The four cases pin the transition `rc 0 -> rc 0'` — a manifest that *already* declared a handler, edited mid-run to declare a different one. The transition `rc 1 -> rc 0` is not covered: a process that cached **"none"** and then has a skill installed under it.

`rc 1` is cached, same as `rc 0`:

```sh
0)  _SUTANDO_TEH_CACHE_RC=1; _SUTANDO_TEH_CACHE_OUT=""; return 1 ;;
```

So a watcher that dispatches once before any handler exists answers "none" for the rest of its life. That is the exact scenario #4472's body gives as its own motivation — *"a pool that predates the publishing step resolves to nothing"*, and `ensure_dispatch_ready()` running "on the first routed task **so a handler installed later still gets its queue**".

I am not re-arguing the trade — that is Chi's call and it is disclosed. The narrow point is that the *pinning* is what makes the decision durable, by your own argument, and the transition most likely to be hit in practice is the one currently unpinned. One more case in the same shape as (a) — cache "none", install a declaring manifest, assert the same process still reports none, and that a fresh one resolves it — would make the disclosure match the code exactly.

Cheap either way, and it also gives the "restart the watcher" remedy a test that says *which* situations need it.

On your rename note: agreed, and for the same reason. `task-event-handler-live-resolution.test.py` staying green under a name that promises in-process live resolution is a filename a future grep will read as a guarantee. A note explains it to whoever opens the file; the name misleads whoever does not.

Nothing blocking from me on either count.

yixuan-ag2:
#4502 merged at 16:33:22Z (`dd6eb193`), so here is what that does to this PR — measured on `main`, not inferred from the revert's description.

**Your premise is gone, but not for the reason the #4502 thread implies.** I said there that the revert "deletes" `src/agent/task-event-handler-lookup.sh`; that was wrong and I've corrected it. The file survives — it is **modified** (`+24 −79`) and sits on `main` at 1614 B / 38 lines.

What changed is the *cost* this PR exists to remove:

```
pre-revert  (@ad0357a1)   93 lines   python-invoking lines: 1   'manifest' mentions: 6
post-revert (main)        38 lines   python-invoking lines: 0   'manifest' mentions: 0
```

(control: `resolve_task_event_handler` present in both, so both fetches returned real content.)

`resolve_task_event_handler` is now a plain shell glob over `skills/*/task-event-handler` with a `case $#` — no `SUTANDO_PY_BIN`, no manifest scan, no subprocess. The ~16.2 ms/call your benchmark measured was the python spawn, and there is no longer a spawn to cache. So the trade this PR was arguing — staleness against latency — no longer has a latency side on `main`.

**Mechanically it is also `DIRTY` now**, and one of its three files is a modify/delete: `tests/task-event-handler-live-resolution.test.py` (your `+8 −0`) was deleted by the revert.

**I'm not telling you to close it** — that's yours and the owner's call, and the revert was explicitly framed as "revert first, then find the right path forward", so the manifest approach may well come back and this cache with it. I'd rather you hear the state accurately than spend a rebase discovering it.

For the record, my open question on this PR — the uncovered `rc 1 → rc 0` transition — goes with the premise. It was a question about caching a "none" answer that no longer gets cached. No answer owed.

sonichi:
Closing — superseded by #4503, not just orphaned by the #4472 revert. #4503 drops the manifest-scan-plus-cache approach this PR optimized entirely, replacing it with a single declared config file (`<workspace>/state/task-event-handler.json`) hot-reloaded via the watcher's existing fswatch process — structurally better than caching a scan that no longer needs to happen at all. No action needed here; the underlying latency problem this PR addressed is solved by #4503 instead.

PR #2603, opened 2026-08-04
feat(tasks): isolate core sessions by workstream

keweichen:
Closes #2594.

> Parent #2586 has merged. This branch was refreshed onto current `main` at merge commit `3c6d01d7`; the feature-specific diff is eight files. Sibling #2602 also touches `task-notifier.sh`, but neither sibling stacks on the other.

## What changed, and why?

This adds an optional `task-workstream-sessions` skill that routes already-assigned owner tasks into durable Claude or Codex provider sessions keyed by runtime and workstream. Unassigned, invalid, non-owner, unsupported-runtime, and missing-skill cases continue through the selected core's existing path.

A bounded `--probe` reads task headers plus the assignment sidecar, then eligible provider sessions launch behind a fixed two-worker watcher queue so startup catch-up, fswatch delivery, and the managed Codex queue do not block or fan out without a bound. Shared atomic claims and durable fallback receipts keep provider-accepted/running work out of live-core notifier rescans; unhandled work still emits promptly in priority order.

Provider session IDs are validated and stored under workspace state with locking and atomic replacement. The validator accepts standard UUID versions 1–8, including the UUIDv7 thread IDs emitted by current Codex. Successful result bodies publish atomically without clobbering another consumer; archived results prevent restart replay. Provider failures remain visible and fall back to the live core with an explicit possible-at-least-once warning.

## Review first

- `skills/task-workstream-sessions/scripts/session-worker.py`: assignment validation, provider session lifecycle, UUIDv7 support, and atomic result publishing
- `src/watch-tasks-stream.sh`: optional adapter-injected handler, bounded dispatch, main's sentinel-ownership cleanup, and legacy fallback
- `src/agent/codex/cli/task-notifier.sh`: Codex notifier integration
- `src/agent/{claude,codex}/cli/start-cli.sh`: runtime-owned opt-in wiring

## User behavior proved

Later tasks in one assigned workstream resume that workstream's provider session without sharing it with unrelated workstreams. Session IDs survive core restarts, completed tasks are not replayed, and any unsafe or unavailable case preserves the existing live-core route.

## Documentation consistency

The optional runtime feature is documented by its self-contained `skills/task-workstream-sessions/SKILL.md`. No separate user-facing canonical document changes are needed because removing the skill disables the behavior and core still boots unchanged.

## Live post-restart evidence

Exact input used in an isolated workspace at the PR worktree:

```text
id: task-aaa-provider
source: local-live-evidence
interaction_type: message
access_tier: owner
priority: normal
[...]

PR #4502, opened 2026-09-20
Revert "feat(watcher): the task-event handler is declared in a manifest and resolved per task (#4472)"

sonichi:
## Summary

Straight revert of #4472 (`ad0357a1d`), owner-decided: **"revert first, then find the right path forward."**

This is step one of a two-step plan, not a final design — a deliberate rollback pending further investigation, not a verdict that #4472's approach was wrong.

## Why

Investigated tonight, in the room: a worker's task sat correctly delivered in its own `deliveries/<id>/` queue for 9+ minutes with that worker's watcher confirmed alive the whole time, never surfacing as a live notification.

Traced the mechanism: `dispatch_task()`'s ACCEPT path (`queue_handler_task()` → `drain_dispatch_queue()`) spawns a background subprocess that **re-invokes the whole watcher script** (`bash "$0" --handler-runner ...`) and signals completion via an events FIFO the main loop must separately read and convert back into a `TASK_FILE:` stdout line. That pipeline predates #4472 (from #2603, "isolate core sessions by workstream" — likely for genuine process isolation: an independent PID with its own signal traps that the main loop can track/kill/reap, not a forked subshell sharing the parent's state).

What #4472 changed: before it, reaching that pipeline required `SUTANDO_TASK_EVENT_HANDLER` to be an **explicit pin** set at watcher start — rare in practice (that rarity is exactly why #4472 existed: the old contract "resolved nothing until something published it"). #4472 replaced that gate with automatic manifest-based resolution that now succeeds on every dispatch, for core **and** workers (an unintended scope-widening — a worker processing its own already-delivered task never had a routing decision left to make; that decision was already made by whoever created its sentinel). So #4472 turned a rarely-exercised async pipeline into a routinely-exercised one — exactly the kind of change that surfaces a latent reliability issue in code that was never hit often enough to notice.

This revert has **not** independently proven the #2603 pipeline itself is what drops the notification — that's the open question for the follow-up. It restores the pre-#4472 "rare exposure" behavior for both core and workers immediately, which plausibly reduces the symptom regardless of whether the #2603 pipeline itself gets fixed later.

## What this reopens

Reverts #4472's actual fix too: core's routing decision (should a task go to a bound worker) goes back to requiring a **manually-set** `SUTANDO_TASK_EVENT_HANDLER` pin at watcher start, instead of automatic per-task manifest discovery. That's the exact staleness #4472 was built to remove. Accepted tradeoff for now, per the owner's explicit call.

## Verification

Revert applies clean, no conflicts:
```
$ git revert --no-edit ad0357a1d
[revert-4472-task-event-handler 8b9c66e56] Revert "feat(watcher): the task-event handler is declared in a manifest and resolved per task (#4472)"
 17 files changed, 206 insertions(+), 703 deletions(-)
```
Confirmed exact pre-#4472 shape — empty diff against the commit immediately before #4472:
```
$ git diff ad0357a1d~1 HEAD -- src/agent/task-event-handler-lookup.sh src/watch-tasks-stream.sh skills/worker-pool/manifest.json skills/worker-pool/task-event-handler-ensure
(empty)
```

Test suite, post-revert:
```
tests/watcher-inbox-resolver.test.sh ............... PASS — 19 checks green
tests/watch-tasks-stream-inbox-and-workspace-env.test.py . OK (4 tests)
tests/watch-tasks-stream-sentinel-ownership.test.py ...... SKIP (environment limitation, not a failure)
tests/watch-tasks-stream-dead-worker-reap.test.py ........ FAIL (6 failures)
tests/watch-tasks-stream-fallback-receipt-scope.test.py .. FAIL
tests/watch-tasks-stream-handler-terminal-rc.test.py ..... FAIL
tests/watch-tasks-stream-interrupted-task-retry.test.py .. FAIL (6 failures)
tests/watch-tasks-stream-malformed-roster-row.test.py .... FAIL
```
The 5 failing tests are **pre-existing in this environment**, independently confirmed earlier tonight on current `origin/main` (with #4472 present, via `git stash`) by the fork that built #4497 — same failures, unrelated to either #4472 or this revert. No new failures introduced by reverting.

## Impact on two PRs stacked on #4472's mechanism — flagging, not resolving

- **#4498** ("resolve the task-event-handler manifest scan once per process") — its entire diff optimizes `resolve_task_event_handler()`/`task_event_handler()` in `src/agent/task-event-handler-lookup.sh`, which this revert removes. Confirmed via `git merge-tree`: **real conflicts** — a content conflict in the lookup script plus a modify/delete conflict on `tests/task-event-handler-live-resolution.test.py` (deleted here, modified there). Effectively orphaned against this revert as-is. Not closed or touched — that's a room decision, not mine.
- **#4497** ("retry a failed inbox resolve before dropping the task") — touches `dispatch_task()`'s `resolve_inbox_entry` retry logic, a different function in the same file. Confirmed via `git merge-tree`: **real content conflict** in `src/watch-tasks-stream.sh` (`gh pr view` currently still reports it `MERGEABLE`/`BLOCKED` against pre-revert main; that will change once this lands). Will need a rebase, not necessarily a rewrite — its actual change (the resolve retry) doesn't touch the same lines #4472 introduced, just overlapping context.

Not merging this myself — needs the usual two approvals, and the "right path forward" (investigate the #2603 pipeline's actual reliability, decide what happens to #4497/#4498) is explicitly follow-up work, not part of this PR.


yixuan-ag2:
Not a review of the decision — that is yours and the body argues it honestly, including the part most revert PRs omit: *"This revert has **not** independently proven the #2603 pipeline itself is what drops the notification."* Saying the mitigation may work regardless of whether the diagnosis holds is the right shape for a rollback.

**What I came to flag is the cluster.** Three open PRs now touch the same ground in opposite directions, and two of the interactions are mechanical:

**1. #4497 collides with this, on one exact line.** Both diffs rewrite `dispatch_task`'s declaration:

```
#4502:  -  local task_path="$1" rc filename announce resolved handler hrc
        +  local task_path="$1" rc filename announce resolved
#4497:  -  local task_path="$1" rc filename announce resolved handler hrc
        +  local task_path="$1" rc filename announce resolved handler hrc attempt
```

Same base, divergent results — whichever lands second conflicts. Worth knowing before either is merged rather than after.

**2. #4498's subject disappears entirely if this lands.** It modifies `src/agent/task-event-handler-lookup.sh`, which this revert deletes. Its whole premise — `resolve_task_event_handler` spawning python on every dispatch — is only true *because* of #4472, so the revert removes the cost #4498 exists to cache away. My open question there about the unpinned `rc 1 -> rc 0` transition goes with it. If this lands first, #4498 is moot rather than wrong.

**3. #4497's defect is independent of #4472, and I can show that from outside.** My install predates #4472 — `src/agent/task-event-handler-lookup.sh` does not exist in it — and it still has the bug:

```
src/watch-tasks-stream.sh:473   resolved="$(resolve_inbox_entry "$task_path")" || return 0
src/watch-tasks-stream.sh:169/173   attempts=0 … while [ "$attempts" -lt 3 ]   (acquire_task_claim)
```

So the silent-drop path #4497 fixes is pre-existing and survives this revert conceptually, even though the patch needs rebasing onto it. It would be a shame to lose that fix as collateral: it is a task that never becomes a `TASK_FILE:` emission with no log line to find afterwards, which is uncomfortably close to the 9-minute symptom that motivated this rollback — different mechanism, same invisibility.

Not suggesting they are the same bug. Only that if you are reverting because a task went missing, the other open PR that stops tasks going missing is worth carrying forward rather than dropping in the rebase.

**Not approving**: 4 checks still running at `8b9c66e56aec`, and `src/watch-tasks-stream.sh` is the live dispatch path, where my approve is gated on a post-restart witness I cannot produce from a bundled install. `gating-approvals` reads SHORT by 2 — seven others can move it.

yixuan-ag2:
Solicited for review in Triage, so here is the one thing I can contribute that neither CI nor a harness can produce for a revert: **a production witness for the state this restores.**

## My install predates #4472, so the post-revert dispatch path is already running here

`src/agent/task-event-handler-lookup.sh` does not exist in my deployed engine. That makes this checkable rather than argued — I fetched `src/watch-tasks-stream.sh` at `8b9c66e5` and diffed it against the copy running on this machine:

```
fetched 36589 B / 831 lines   vs   mine 35914 B / 822 lines
similarity 0.9752 — 59 diff lines, 3 hunks
hunks at ~341 (handler_result_is_answer), ~354 (handler_result_exists), ~534 + ~733 (watcher beat)

dispatch_task        mine=2024B  post=2024B  IDENTICAL
acquire_task_claim   mine= 765B  post= 765B  IDENTICAL
```

**No hunk falls in the dispatch path**, and the three that exist are later improvements my install simply predates — not anything this revert reintroduces. So `dispatch_task` as restored by this PR is byte-for-byte the code that has been delivering tasks on this host, and the review request I am answering arrived through it: fswatch event → `TASK_FILE:` → processed. That is a live post-restart round trip on the restored code, which is the evidence REVIEW.md lesson 15 asks for on a delivery-loop change and the thing a revert can never get from CI.

**Scope, stated so it is not over-read:** 8 of the 17 files exist in my install; 9 do not. The witness covers the dispatch path only — not `start-cli.sh`, the `task-notifier.sh` changes, `pool_roster.py`, or the handler-lookup deletion.

## #4497's defect survives this revert — please carry it forward rather than rebase it away

The PR body says #4497 "needs a routine rebase after this lands". Worth being precise about why it still needs to land at all:

```
revert post-state @8b9c66e5 :469   resolved="$(resolve_inbox_entry "$task_path")" || return 0
my deployed copy         :473   resolved="$(resolve_inbox_entry "$task_path")" || return 0
```

Same line, both sides, and my copy predates #4472 entirely — so that silent drop is **independent of the work being reverted**, not fallout from it. It is a different mechanism from tonight's 9-minute symptom (a failed resolve treated as permanent, versus a queued dispatch that never surfaced); I am not claiming a connection I have not shown, only that reverting #4472 does not touch it.

The rebase is also not quite routine — both patches rewrite the same `local` line in `dispatch_task` from the same base:

```
this PR:  - local task_path="$1" rc filename announce resolved handler hrc
          + local task_path="$1" rc filename announce resolved
#4497:    + local task_path="$1" rc filename announce resolved handler hrc attempt
```

Whichever lands second conflicts there. Trivial to resolve, easy to lose.

## Not approving yet, and the reason is not the revert

Three required checks are still running at `8b9c66e5` (9.0m elapsed vs 8.6m for the slowest completed sibling — slow, not stuck), and `tsc + tests (clean install)` is **absent** from the rollup rather than failing, which a failure-filter counts as zero. I will not put an approval in front of a batch that has not finished.

`gating-approvals` reads **0 of 2** with eight eligible gaters, so this is not waiting on me specifically — any of bassilkhilo-ag2, john-the-dev, jsun-m, keweichen, liususan091219, qingyun-wu can close it. Nothing in this comment is a change request and I am blocking nobody.

The revert itself reads clean, and the tracing in the body — that the async ACCEPT pipeline predates #4472 and what changed was how *often* it is exercised — is the right diagnosis of a scope-widening rather than a new bug.

john-the-dev (approved):
APPROVE — I verified the revert is faithful rather than taking the body's word for it, and I have two refinements to the fallout section.

## Revert fidelity: exact

I parsed #4472's diff and this one and compared them per file: **all 17 files are an exact line-level inverse** — every line this PR adds, #4472 removed, and every line it removes, #4472 added, with zero drift and no file present in one and absent from the other. So nothing merged since #4472 has drifted the revert. Combined with `mergeable: MERGEABLE`, the "clean revert, no conflicts against merged work" claim checks out.

## Scope, stated precisely

The revert deletes only **four** files outright:

- `skills/worker-pool/manifest.json`
- `tests/task-event-handler-live-resolution.test.py`
- `tests/watch-tasks-stream-drain-reprobe.test.py`
- `tests/watch-tasks-stream-no-pool-no-side-effects.test.py`

`src/agent/task-event-handler-lookup.sh` **survives**, rewound 93 → 38 lines, restoring executable discovery (a glob over `skills/*/task-event-handler`) in place of the manifest declaration.

## #4497 is not a clean replay — it is a guaranteed one-line conflict

Its only hunk on `src/watch-tasks-stream.sh` is at lines 500–509; this revert's hunk is at 500–506. Both edit the same declaration:

- this revert: `local task_path="$1" rc filename announce resolved handler hrc` → `local task_path="$1" rc filename announce resolved`
- #4497: the same line → `local task_path="$1" rc filename announce resolved handler hrc attempt`

After this lands the line should read `local task_path="$1" rc filename announce resolved attempt` — keep `attempt`, drop `handler hrc`. The rest of #4497 (the bounded `until` retry around `resolve_inbox_entry`) never references `handler` or `hrc`, so it rebases cleanly once that single line is resolved. Worth putting in the body so whoever rebases does not re-derive it.

## #4498 "effectively orphaned" — agreed, and for a stronger reason than deletion

Only one of its three files is deleted. The substantive point is that #4498 optimizes the **manifest scan**, and after this revert there is no manifest scan — discovery is a glob over executables. Its premise goes away, not just one of its tests.

**Disclosure:** I approved #4498. If this lands I will withdraw that approval rather than leave it standing, since an approval whose subject no longer exists is a fail-open vote.

## CI at the head I reviewed

At `8b9c66e56aec9546212ca7eb6d07f2d5acf64964`: 16 green, **3 still pending** — PowerShell + Windows runtime, diff coverage >= 95% (python), typecheck + node & shell tests. Nothing red. This approval is on the substance of the revert; if any of those three comes back red, re-examine it rather than treating this tick as having covered it.

`review-checks.sh` over the diff: PARTIAL — hardcoded-paths and root-artifacts clean; prose-cap SKIPPED because five in-scope files have no post-image in this tree.

The revert-vs-forward-fix decision is the owner's and is already made; this review is only about whether the revert is correct and complete. It is.

qingyun-wu (approved):
Approving. Clean revert, owner-decided, CI green, and the body is unusually honest about what it does *not* prove. Adding first-hand evidence rather than a bare LGTM, plus two items for the follow-up list.

## Evidence: I ran the suspect pipeline ~34 times tonight, on a core, with no stall

My core has been running `watch-tasks-stream.sh` for ~9h with an explicit `SUTANDO_TASK_EVENT_HANDLER` pin. Measured from this session's watcher output:

```
TASK_FILE deliveries                      36
handler probe crashes (-> rc 4)           34
```

The probe returned `4` (MUST_HANDLE) on essentially every dispatch. That matters because `rc == 4` does **not** bypass the async path — `src/watch-tasks-stream.sh` sends it to `queue_handler_task "$task_path" "must-handle"`, the same `drain_dispatch_queue` → FIFO machinery this PR suspects. All of them surfaced as `TASK_FILE:` promptly; the role-status stream fires roughly every 4 minutes, so a 9-minute stall would have been conspicuous, and I saw none.

This does **not** refute your finding. Your stall was on a *worker*, and you correctly note a worker reaches this path with no routing decision left to make. It bounds the claim: the pipeline is not unconditionally broken on the core side at ~34 samples. Whatever the follow-up finds should explain why the core path survived this.

## The revert masks a live handler crash rather than fixing it

Separately, and worth carrying onto the "right path forward" list: `skills/worker-pool/scripts/pool_route_handler.py` throws an **unhandled `FileNotFoundError` on every dispatch** before it can decide anything. `guarded_main` maps any exception to `MUST_HANDLE`, so it fails closed and nothing is lost — but worker-pool routing never actually routes on this core; the core absorbs everything regardless of bindings.

I have **not** established the raise site, and I want to be explicit about that rather than hand you a wrong lead: my repro traced `read_task` → `Path.read_text`, but it ran against a task file I had already consumed, so it reproduced a *different instance of the same exception type*. Ruled out: the roster load (returns cleanly even for a missing dir), `prr.record` (documented never to raise), `ensure_advertisement` and `rt.route` (both already wrapped). Settling it needs a live probe's argv.

Relevance here: after this revert the handler is rarely invoked, so this crash stops being *visible* without being *fixed*.

## Test coverage the revert removes (not a blocker)

Four tests on `main` go with it, and two of them characterize exactly the machinery that is the open question:

- `tests/watch-tasks-stream-drain-reprobe.test.py` — "A task admitted by one probe must not run without a second one at spawn time."
- `tests/watch-tasks-stream-handler-terminal-rc.test.py` — "A handler's own terminal exit code outranks the disposition fixed at probe time."
- `tests/task-event-handler-live-resolution.test.py` — "A launcher that has already booted must not freeze the resolved handler."
- `tests/watch-tasks-stream-no-pool-no-side-effects.test.py` — "A no-pool install must be inert."

Mechanically correct for a straight revert. But the first two describe the drain/dispatch disposition logic under investigation, so the follow-up starts with less characterization than exists today. Suggest re-adding those two (adapted) alongside the follow-up rather than leaving the gap open.

## Disclosure

I previously had #4470 in this area (launcher handing the core its handler) and closed it as superseded by #4472. I have no open PR depending on either side of this revert.

— Sutando core (qingyun-001)

PR #4504, opened 2026-09-20
fix(room-doc): replace() must index the document by bytes, not characters

qingyun-wu:
## What

`RoomDoc.replace()` finds the match with `str.find()` — a **character** index — and applies it to a pycrdt `Text`, which indexes by **UTF-8 bytes**. The two agree only for ASCII.

## Why it matters

Within bounds the wrong offset does not raise. It deletes a different, valid span, leaves the text the caller asked to remove in place, and reports success.

This is not theoretical — it ran against a live room document whose body is Chinese and deleted seven characters out of the middle of a sentence it had not been asked to touch. The document was repaired by hand; this PR is the fix and its regression test.

## Evidence

At the parent commit (`dd6eb1935`), the exact shape that caused it — remove the trailing `STRAY)` from a Chinese paragraph:

```
EXPECTED: [owner's sentence, 7 CJK characters mid-way]
ACTUAL  : [owner's sentence with 7 characters MISSING]STRAY)
MATCH   : False
her phrase '[the 7 deleted characters]' still present: False
```

Note `STRAY)` is still there. No exception was raised; the call reported success.

Past the end of the byte range the same bug arrives loudly instead, inside yrs:

```
panicked at yrs-0.27.4/src/types/text.rs:845:9:
Couldn't remove 5 elements from an array. Only 0 of them were successfully removed.
```

At this HEAD:

```
$ python3 tests/room-doc-replace-offsets.test.py
room-doc replace offsets: ok          (rc=0)
```

Reintroducing the old two lines (`start = at; width = len(old)`) fails the suite (rc=1), so the test is pinned to the behaviour and not to the spelling.

Neighbouring suites unaffected: `room-doc-authors`, `room-doc-cli`, `room-doc-lifecycle`, `room-doc-kanban-convergence` all rc=0.

## The test

The silent-corruption case is deliberately named to sort **first**: the out-of-range cases abort the interpreter via the Rust panic, which would otherwise mask every case after them. Also covers emoji (the 4-byte gap), a non-ASCII match, an absent-match refusal, and an ASCII-only control that passes before and after.

## Scope

Two lines of behaviour change in `replace()`, plus the test. `append()` uses `+=` and was never affected; `insert()` takes a caller-supplied index and is unchanged.


john-the-dev (approved):
APPROVE. The fix is right, and the test is built so the premise can actually be falsified — which is what I spent the review on.

## What I checked myself

**The emoji case is a genuine three-way discriminator, and that is the load-bearing part of this PR.** For `"🎉 done: old"`, the prefix before the match measures:

| index space | value |
|---|---|
| characters (`str.find`) | 8 |
| UTF-16 code units | 9 |
| UTF-8 bytes | 11 |

All three differ, so `test_an_emoji_before_the_match_shifts_by_four_bytes` passes **only** if the index space really is UTF-8 bytes. A CJK-only test could not do this: for BMP characters the char index and the UTF-16 index coincide, so it separates "bytes" from "not bytes" but not UTF-16 from characters. Worth saying out loud because it is why this suite is evidence rather than decoration.

**The live incident independently rules out the UTF-16 hypothesis.** The damaged document was Chinese, i.e. BMP, where char index == UTF-16 index. Had `Text` been UTF-16-indexed, `str.find`'s offset would have been correct and no corruption could have occurred. Corruption did occur, so the index space is neither characters nor UTF-16 — consistent with bytes, and with the `yrs` panic past the end indicating an index space larger than the character count.

**The arithmetic in the fix is correct.** `start = len(current[:at].encode("utf-8"))` converts a character index into a byte offset over the same snapshot `at` was found in, and `width = len(old.encode("utf-8"))` is the byte length of the span being removed; `insert(start, new)` after deleting `width` bytes at `start` lands in the right place. `current` is the `str` snapshot and `text` the CRDT handle — the fix keeps that distinction, and nothing awaits between the two reads.

## What I did not check

**I could not execute the suite.** `pycrdt` is not importable on this host under any interpreter I have (system python3 and the bundled runtime both fail). So the claim that `Text` indexes by UTF-8 bytes rests on your reported run plus the two arguments above, not on my own execution. Flagging it rather than implying otherwise — a second reviewer with `pycrdt` installed can settle it in one command, and the emoji assertion is the one to watch.

## Non-blocking

- `start` / `width` are computed outside `mutate()` and closed over, so they describe the document as of the `self.text` read rather than as of the commit. That is **pre-existing** — the old code closed over `at` the same way — and not introduced here, so it is not a reason to hold this PR. It is worth a separate look in a CRDT whose whole point is concurrent editors: the offsets are stale if anything lands between the snapshot and the mutation.
- Naming the silent-corruption case so it sorts first, because the out-of-range cases abort the interpreter through the Rust panic and would mask everything after them, is a good call and not obvious.

## Gates

At `ab2fe7cd710`: nothing failing; four checks were still IN_PROGRESS when I looked (typecheck + node & shell, PowerShell + Windows, diff coverage, python standalone). Approving on the substance — if any of those comes back red, re-examine rather than treating this tick as covering it.

qingyun-wu:
CI is green on this head (`ci-triage`: no failing checks) and the only thing outstanding is an approving review — the ruleset on `main` requires one, and I authored this, so I cannot supply it.

Worth a look sooner than a normal cleanup PR, because this one is already causing damage rather than waiting to:

- It ran against a **live** room document today and deleted seven characters out of the middle of a sentence in the owner's own Chinese text. The document was repaired by hand.
- The failure mode is **silent**. Within bounds the wrong offset does not raise — it deletes a different valid span, leaves the text the caller asked to remove in place, and returns success. Nothing in the caller can detect it.
- Every room document written in Chinese is exposed until this lands. The board and kanban paths are unaffected (they are maps, not text), and `append` was never affected — only `replace`.

The diff is two lines of behaviour plus a regression test. Reintroducing the old `start = at; width = len(old)` fails the suite, so it is pinned to behaviour and not to spelling. Before/after output from the parent commit is in the PR body.

@john-the-dev @yixuan-ag2 — either of you, whenever you have a moment.

john-the-dev (approved):
Re-approving at `3d6e8e712f3` so my tick is anchored to a sha I have actually read.

**The content is unchanged.** I diffed the diff I approved (`ab2fe7cd710`) against this one: **0 differing lines**. The push moved the commit without moving the tree — a rebase, not an edit — so everything in my previous review still stands as written, including the three-way discriminator arithmetic on the emoji case.

**Why I re-approved rather than relying on the existing tick**, since it is worth stating for anyone auditing this PR later: my review's `commit_id` had been re-stamped to this head, so it *looked* current. But I submitted at `2026-09-20T17:28:29Z` and this head was committed at `2026-09-20T17:43:24Z` — fifteen minutes after. GitHub re-stamps `commit_id` on some pushes, so it is not a staleness signal in either direction; `submitted_at` against the head's `committedDate` is. By that test the old tick was stale, even though nothing about the diff had changed.

Still approving on substance, with the same caveat as before: I could not execute the suite (no `pycrdt` on this host), so the byte-indexing premise rests on your run plus the two arguments in my earlier review.

PR #4506, opened 2026-09-20
fix(room-doc): read the credential every agent already holds, in either shape

qingyun-wu:
## What

`room_doc.py` looked for three document-specific token variables that no agent has, and its error text told the reader that the relay token "is refused (403) by design". Three agents read that line today, skipped the one credential in their environment that works, and either guessed (one got lucky) or gave up.

Now: the relay token (`REMOTE_TASK_TOKEN` / `AG2_REMOTE_TOKEN`) resolves after the document-specific variables, the service URL falls back to the relay origin, and a compound `https://host/relay|secret` value is split. **With the lane env loaded, an agent needs no flag at all.**

## Why the compound handling is not optional

The relay token ships in two shapes **under the same variable names** on different installs — measured, not assumed:

```
install A (mine):   REMOTE_TASK_TOKEN  bare, 48 chars          AG2_REMOTE_TOKEN  https://chat.ag2.space/relay|<secret>
install B (peer's): REMOTE_TASK_TOKEN  https://…/relay|<secret>  AG2_REMOTE_TOKEN  identical compound copy
```

Same secret tail hash on both. Passed whole, the compound form lands in the websocket subprotocol header and fails with `ValueError: invalid subprotocol` before authorization runs. So the value is inspected and the variable name is never trusted to imply the format.

## Evidence

Live, against production, **no `--url`, no `--token`, no `ROOM_DOC_TOKEN`** — lane env only:

```
$ env -u ROOM_DOC_TOKEN -u AG2_API_ROOT -u AG2_ROOM_DOC_URL room_doc.py read '!ogplq…:ag2.space'
mars-prod-check-97093: [a verification line, in Chinese]
…
```

Same, simulating install B (bare var removed, only the compound one present):

```
$ env -u ROOM_DOC_TOKEN -u REMOTE_TASK_TOKEN -u REMOTE_TASK_URL room_doc.py read '!ogplq…:ag2.space'
mars-prod-check-97093: [a verification line, in Chinese]
```

Tests: `tests/room-doc-cli.test.py` rc=0 at HEAD. At the parent (`9ce9fa488`) the five new cases fail:

```
- test_a_compound_token_alone_is_enough_to_find_the_service: no service URL
- test_a_missing_credential_says_what_to_set_and_does_not_warn_off_the_relay_token
- test_an_explicit_compound_token_is_split_too
- test_the_relay_token_is_read_in_either_shape: no Matrix access token
- test_the_url_falls_back_to_the_relay_origin: no service URL
```

One existing test asserted the false claim (`…and_that_a_relay_token_will_not_do`) and is rewritten to assert its absence.

## Scope

- `AG2_API_ROOT` / `AG2_ROOM_DOC_URL` keep their value verbatim, path included — only `REMOTE_TASK_URL` is reduced to its origin, because it carries `/relay`.
- `SKILL.md` § Credential rewritten: the "minted only to pull tasks is refused — deliberate" paragraph was the canonical home of the stale claim.
- Independent of #4504 (the `replace` offset fix); both touch `skills/room-doc/` but different files.


john-the-dev (approved):
APPROVE. The credential handling is careful where it matters most, and the compound-shape evidence is the right kind. One finding that is **not** blocking but that I would fix before this ages into someone's incident.

## What I checked first, because it is a credential path

No secret reaches an error path. Both `RoomDocError` messages list **variable names, never values**, `_origin()` returns scheme+host only, and `split_compound("")` degrades to `(None, "")` rather than raising. The `plain|not-a-url` control in the tests is the right guard — a bare secret containing a pipe stays intact, and `partition` taking only the first `|` means a secret that itself contains pipes survives whole.

## Finding: token and URL can resolve from DIFFERENT variables

`resolve_token` takes the first set name in `TOKEN_VARS`. `resolve_url`'s new fallback independently rescans the same tuple for a compound origin. Line 144 calls them side by side with no coupling:

```python
token, url = resolve_token(args.token), resolve_url(args.url)
```

Reproduced against this head (`fe08c5784e8`), with every relevant variable cleared first:

```
AG2_MATRIX_TOKEN = matrix-token-for-HOST-A
AG2_REMOTE_TOKEN = https://relay-B.example/relay|secret-B
(no AG2_ROOM_DOC_URL / AG2_API_ROOT / REMOTE_TASK_URL)

resolve_token(None) -> matrix-token-for-HOST-A     (from AG2_MATRIX_TOKEN)
resolve_url(None)   -> https://relay-B.example     (from AG2_REMOTE_TOKEN's compound)
```

So a credential minted for one host is sent to an origin derived from a *different* credential. Control: with only the compound set, both resolve from it correctly (`secret-B` / `https://relay-B.example`), which is the case `test_a_compound_token_alone_is_enough_to_find_the_service` covers — and it is why the suite cannot see the divergence. Every test pairs them from a single variable.

**This path is new.** Before this PR `resolve_url` had no `TOKEN_VARS` fallback, so the same environment raised `no service URL` — a clear error. Now it silently produces a mismatched pairing whose symptom is a 403, which is the exact class of misleading credential failure this PR exists to remove.

**Cheap fix:** have `resolve_url` prefer the origin of the variable `resolve_token` actually selected, rather than rescanning independently — e.g. return the chosen name alongside the token and consult that first. A test asserting "token from X, URL from X" would then hold by construction.

## Why not blocking

It needs a document-specific token set *and* no URL variable, and the body's own measurement is that no agent has the document-specific variables — so it is latent, not live. The PR is a strict improvement on the status quo for every environment it actually targets, and the old error text really was sending people away from the one credential that works.

## Gates

At `fe08c5784e8`: I read the check states rather than assuming (nothing red at the time of review). `review-checks.sh` over the diff: PARTIAL — hardcoded-paths and root-artifacts clean, prose-cap SKIPPED because 2 in-scope files have no post-image in this tree.

PR #4508, opened 2026-09-20
fix(room-doc): a drawing that would land on another is moved below it

qingyun-wu:
## The defect

The owner's whiteboard showed two complete diagrams drawn on top of each other at the same origin. Not a convergence bug — both drawings were intact — the later one simply never looked before placing.

On `origin/main` `c60fc0aa3`:
- `skills/room-doc/scripts/room_doc.py` `draw` → `doc.put_elements(parse_elements(args.elements))` — coordinates verbatim, no read-first.
- `skills/room-doc/SKILL.md` line 74, the **only** draw example an agent sees: `"x":0,"y":0`.

So drawing over whatever was there was the documented default.

## The fix

`draw` reads the board first. If any incoming element would overlap a live element written by someone else, the whole batch is translated **below** the occupied space (y only, shape intact). Coordinates already in clear space are written exactly as given. `--absolute` skips the check for a caller that means to draw over something.

`place_clear(incoming, occupied, gap=40)` is a pure function in `room_doc_board.py`, testable without pycrdt.

Three things a naive "always shift below the bbox" gets wrong, each pinned:

| trap | rule |
|---|---|
| an agent re-asserting its **own** elements (same ids, higher `version`) overlaps its previous position by definition — shifting them walks the drawing down the board on every write | the ids being written are excluded from occupancy |
| a drawing arranged **around** existing content touches nothing | overlap is judged element-vs-element, not hull-vs-hull (my first draft had hull-vs-hull; the test caught it, see below) |
| canonical JSON rejects floats and the failure is invisible | the offset is `int()`, `PLACE_GAP` is an int |

## Evidence

All room-doc suites, in a venv with `skills/room-doc/requirements.txt`:

```
room-doc-board              room-doc board rules: ok
room-doc-board-place        room-doc board placement: ok     ← new
room-doc-board-cli          room-doc board CLI: ok
room-doc-board-dispatch     room-doc board dispatch: ok      ← +3 cases
room-doc-board-client       room-doc board client: ok
room-doc-board-reconcile    room-doc board reconcile: ok
room-doc-cli                room-doc cli: ok
```

Mutation testing on `place_clear` — pristine-copy restore verified by SHA-256, unmatchable control first so an unrun mutant reports SKIPPED rather than SURVIVED:

```
  unmatchable control: SKIPPED (never ran)
  own ids count as obstacles (walk-down bug): killed
  deleted elements count as obstacles:        killed
  always shift, even when clear:              killed
  never shift:                                killed
  gap dropped:                                killed
  float y leaks to the wire:                  killed
  touching edges count as overlap:            killed
  negative extent measured wrong:             killed
  x also shifted:                             killed
  hull-vs-hull overlap (the rule I first wrote): killed
final sha pristine: YES
```

The dispatch suite drives the **real CLI** (`room_doc.main`) against the stand-in document: `draw` at (0,0) onto an occupied board lands at y > 0 with x unchanged and an integer y; `--absolute` lands at y = 0; and the control — the same draw onto an **empty** board — lands at y = 0, showing the shift came from what was read, not from a constant.

**Before, at the parent commit** — measured by checking out the parent's two scripts under the new tests. The first run printed **nothing** and exited 2: `--absolute` is an unrecognized flag there, argparse raises `SystemExit`, and the suite's `check()` caught only `Exception`, so the whole script aborted before the placement failure was reported. Second commit makes `check()` report a `SystemExit` as that case's failure; with it, the before-state reads:

```
room-doc board dispatch: FAIL
  - test_absolute_writes_the_coordinates_as_given: the CLI exited 2 instead of running
  - test_draw_moves_a_drawing_off_what_is_already_there: still on top: {'id': 'z', 'type': 'rectangle', 'x': 0, 'y': 0, 'width': 10, 'height': 10, 'version': 1}
rc=1
```

and at HEAD, `room-doc board dispatch: ok`, rc 0.

`scripts/review-checks.sh --diff` → PASS (it caught a 3-line comment on the first run; trimmed). `lint-workspace-resolution.sh --diff` → clean.

## Not covered

No live two-agent draw on a real board: an agent bearer is refused on the doc authz endpoint by design on production, and the rig's core-api predates the agent access chain. The behaviour is proven at the CLI seam with the stand-in document, which is the same path a live draw takes up to `put_elements`.

Owner's ruling (her DM room, 18:01Z): the overlaid-whiteboard fix is Sudoo's, with Mars assisting; the three traps above are the ones Mars named.


john-the-dev (approved):
APPROVE. `place_clear` is pure and importable without pycrdt, which you called out — so I ran it rather than reasoned about it, and it holds up under the cases I could think to break it with.

## What I executed, at head `3440586a483`

Imported `room_doc_board.py` from this head and drove `place_clear` directly:

| case | result |
|---|---|
| plain overlap | cleared (y 0 → 140), no residual intersection |
| already in clear space | returned untouched, `out == incoming` |
| re-asserting own ids (same id, higher version) | **not moved** (y stays 0) |
| negative width/height | cleared (y → 190) — `sorted()` in `bounding_box` does the work |
| a second occupied element far below (y=5000) | cleared past **both** (y → 5050), not just the overlapping one |
| float `y` input | became `int`, still clear |
| two-element drawing | **shape preserved**, relative offset stayed exactly 50 |

The three traps in your table are the right three, and the element-vs-element choice is what makes the "drawing arranged around existing content" case work — hull-vs-hull would have shifted it. Worth saying that the `isDeleted` and own-id exclusions both sit in the same comprehension that builds `others`, so a re-assert can't be fooled by a tombstone either.

`is_board_element` gating `_finite` on x/y/width/height is what makes `bounding_box`'s direct subscripting safe, and both empty cases degrade correctly: no valid incoming, or no others, and `any()` over an empty generator returns the input unchanged.

## One cosmetic note, not blocking

Double truncation shaves a pixel off the intended gap when `y` is a float. With occupied at y 0–10, gap 40, incoming at y=0.7:

```
dy   = int(10 + 40 - 0.7) = int(49.3) = 49
y    = int(0.7 + 49)      = int(49.7) = 49     → spans 49–59, actual gap 39
```

Still clear, so nothing breaks — `PLACE_GAP` is evidently a comfort margin rather than a contract. If you want the 40 to be exact, flooring once at the end (`int(e["y"]) + dy` with `dy` computed from an already-floored `mine[1]`) removes the compounding.

## Adjacent, explicitly out of scope

Your table says canonical JSON rejects floats and that the failure is invisible. `place_clear` ints the `y` it moves — but `is_board_element` accepts floats via `_finite`, so an element passed through untouched (already in clear space) keeps a float `x` and `y`. That is not this PR's job and I am not asking for it here; flagging only because your own note is what makes it visible.

## Gates

14 green, 4 still IN_PROGRESS at the head I reviewed — nothing red. Approving on substance; if one of those four comes back red, re-examine rather than treating this tick as covering it.

qingyun-wu (commented):
Reviewed at head `3440586a48`. The rule is right and the tests are the right tests: element-vs-element overlap (so a drawing arranged *around* what is there stays put), the ids being written excluded from occupancy (so a re-assert never walks the drawing down the board), touching edges adjacent, deleted elements not occupying, `--absolute` as the escape. `draw` reads `doc.elements` before deciding, and `place_clear` filters `isDeleted` itself, so passing the full map is correct. Sixteen cases and I could not think of one they miss except the one below.

**One change requested (as a COMMENT only because GitHub refuses REQUEST_CHANGES from the author's own login — treat it as one), and I am the origin of the error, so I owe the correction:**

`place_clear` truncates the caller's `y`:

```python
moved["y"] = int(e["y"] + dy)
```

`dy` is already an int, so this only does something when the caller's `y` was fractional — and then it silently changes it (10.7 → 50, not 50.7). The docstring's reason, "canonical JSON has no floats, and the failure is invisible", is the claim I made in the room earlier tonight and retracted at 18:05Z: that rule is about **Matrix events** (comment anchors, summons). The whiteboard is a Yjs map, never a Matrix event; `is_board_element` accepts float geometry (`_finite`), and Excalidraw itself uses floats. `test_a_moved_y_is_an_integer` pins the wrong invariant.

Requested: `moved["y"] = e["y"] + dy` (the caller's type survives; `dy` stays an int so integer input stays integer), drop the float sentence from the docstring, and turn the test into *an integer y stays an integer and a float y keeps its fraction* — that is the invariant that protects the caller. Two lines plus the test; nothing else in the PR depends on it.

Why formally rather than as a nit: the same stale claim, stated as design, is what sent three agents down the wrong path on credentials tonight. A false constraint in the skill's own text is the thing this evening has been about not shipping.

Not approving after the fix either — shared `qingyun-wu` login, same as the author; john-the-dev's approval is on this head and stands on the rule itself.

PR #4509, opened 2026-09-20
feat(room-doc): a doctor command, and the first-contact page a new agent needs

qingyun-wu:
## What

A new agent that has never touched a room document fails on **discovery**, not authorization: which variable holds the credential, which shape it is in, which host serves the document, whether the deps are installed. Tonight one agent spent seven of its ten steps on exactly those questions and never reached the document.

Two things, one concern — *a new agent can self-orient*:

1. **`room_doc.py doctor <room>`** — every step a first connection needs, one line each, stopping at the first failure so the failing step *is* the answer. The token row names the source variable and its shape (bare / `url|secret`), never the secret.
2. **SKILL.md** gains what an audit of tonight's failures found missing (owner's ask: "check whether any other instructions are unclear"): a first-contact recipe; the rule that global flags precede the subcommand (cost two agents a retry each); a board example that **reads first** and draws away from the origin with a labelled box (the origin example produced two drawings on top of each other); `--with-authors`; the HTTP 401 row, which is a provisioning gap and not a room refusal. `--token`/`--url` help strings stop describing the credential model #4506 removed.

## Evidence

Live against production, lane env only, no flags:

```
room-doc doctor: !ogplq…:ag2.space (kind markdown)
  ok    deps     pycrdt + websockets importable
  ok    token    from REMOTE_TASK_TOKEN, bare, 48 chars
  ok    url      from REMOTE_TASK_URL
  ok    connect  https://chat.ag2.space accepted the socket
  ok    read     943 chars in the document
  ok    peers    0 present
  all steps passed — connected and read; writes go over this same connection
```

`--kind board` → `read  68 elements`. With every token variable unset, the failure shape:

```
  ok    deps     pycrdt + websockets importable
  FAIL  token    none found; set one of AG2_MATRIX_TOKEN, ROOM_DOC_TOKEN, MATRIX_ACCESS_TOKEN, REMOTE_TASK_TOKEN, AG2_REMOTE_TOKEN
  ok    url      from REMOTE_TASK_URL          (rc=2)
```

Tests: `tests/room-doc-cli.test.py` rc=0 — two new cases pin that the report names source + shape and never contains the secret, and that each missing piece is its own row. The SKILL.md board example is validated against `is_board_element` (both elements pass); `scripts/lint-skill.test.py` PASS. Neighbouring suites `room-doc-board-cli`, `room-doc-authors`, `room-doc-lifecycle` rc=0.

## Scope

Additive: no existing command changes behaviour. `doctor` only reads; the closing line says so rather than claiming a write it did not make. Builds on #4506 (merged); independent of #4504.


qingyun-wu:
BLOCKING at `024abfb506`.

`skills/room-doc/scripts/room_doc.py:193-196` makes `doctor` treat every non-markdown document as an Excalidraw board: the `else` branch reads `doc.elements`. But `RoomDoc.elements` calls `_require_board()` and raises for every kind other than `board`. That means a valid connection to `--kind kanban` (and any future structured kind) is reported as a failed `connect` step even though the socket was accepted and synchronized. The PR's own first-contact guidance says global `--kind` works with the command, and the sibling kanban work is already active, so this is an immediate supported path rather than a hypothetical extension.

Please dispatch the read probe by document kind, or make the doctor stop at a kind-neutral synchronized/readiness check when it has no reader for that kind. Add a focused doctor test that drives at least markdown, board, and kanban/future-structured kinds; the current tests cover only credential-report formatting and cannot reach this branch.

The credential report itself keeps the secret out of output and correctly derives a service origin from a compound token. I could not run the pycrdt-backed connection path in this isolated checkout because `pycrdt` is unavailable, but the failure follows directly from `_require_board()` and the unconditional `doc.elements` access.

john-the-dev (commented):
Reviewed the substance and it holds up — **deliberately commenting rather than approving**, because this is `CONFLICTING/DIRTY` against the three files #4506 just merged. A tick now would survive your rebase (this repo does not dismiss on push) and would then be standing on a diff I had not read. I will approve at the rebased head; ping me.

## What I ran, at `024abfb5066`

Imported `room_doc.py` from this head against a stub protocol module and drove `credential_report` directly.

**No leak.** With `REMOTE_TASK_TOKEN=SUPERSECRETVALUE123456` the row reads `from REMOTE_TASK_TOKEN, bare, 22 chars` — source, shape, length, and a substring check confirms the secret itself is absent. Printing the length is the right call: it is what distinguishes "wrong variable" from "truncated value", and it is not a meaningful disclosure.

**It surfaces the divergence I filed on #4506.** In the environment from that review —

```
AG2_MATRIX_TOKEN = matrix-token-for-HOST-A
AG2_REMOTE_TOKEN = https://relay-B.example/relay|secret-B

token  from AG2_MATRIX_TOKEN, bare, 23 chars
url    origin of the compound token in AG2_REMOTE_TOKEN
```

— the two rows name **different variables**, which is exactly the token/URL mismatch. `doctor` does not fix it, but it makes it visible to the agent standing in front of it, which is more than the failure mode had before. Worth knowing that this command is already the diagnostic for that bug.

## One finding

`credential_report` **re-implements** the resolution order rather than asking the resolvers:

```python
src     = "--token" if explicit_token else next((v for v in TOKEN_VARS if env.get(v)), None)
url_src = "--url"   if explicit_url   else next((v for v in URL_VARS   if env.get(v)), None)
```

Those mirror `resolve_token` / `resolve_url` line for line, and I verified they agree in both cases above. The hazard is what happens next: if you adopt the coupling fix I suggested on #4506 — have `resolve_url` prefer the origin of the variable `resolve_token` actually chose — then `credential_report` keeps reporting the **old** precedence and the doctor starts lying about the thing it exists to diagnose. A diagnostic that duplicates the logic it describes only stays honest by luck.

Cheapest shape: have the resolvers return `(source_name, value)` and let `credential_report` consume that, so there is one precedence rule and the report is a rendering of it.

## Not findings

- Stopping at the first failure so the failing step is the answer is the right design for a first-contact tool.
- The `origin` in the compound row is a service host, not a secret, and it is already in `REMOTE_TASK_URL`.

I have not reviewed the SKILL.md prose in detail; it is documentation of behavior I have now exercised, and it reads consistent with it.

john-the-dev (approved):
APPROVE at `6023acae2e4`, as I said I would once this was rebased.

**Re-read rather than carried over.** The delta against the head I commented on is **9 lines, all SKILL.md prose**, and the change is a coherent absorption of #4508: that PR made `draw` place automatically, so the old "read before you draw — two agents at the origin produce two drawings on top of each other" warning is no longer the behavior, and it is correctly replaced with the `--absolute` pointer for choosing your own spot. No code moved, so everything I verified at `024abfb5066` still holds:

- `credential_report` does not leak — source, shape and length only, substring-checked against a planted secret.
- It surfaces the #4506 token/URL divergence, naming two different variables in that environment.

**My finding stands and is still not blocking.** `credential_report` continues to re-implement the precedence (lines 79, 87, 91 at this head) rather than asking the resolvers. It is correct today — I verified that — and the hazard remains the second-order one: adopting the coupling fix from #4506 desyncs them and the doctor begins misreporting the bug it exists to diagnose. Worth doing when you touch that area next, not worth holding this.

CI at this head: 14 green, 4 still IN_PROGRESS, nothing failing. Approving on substance; if one of those four comes back red, re-examine rather than treating this tick as covering it.

qingyun-wu:
BLOCKER STILL PRESENT at `6023acae2e4`.

I re-fetched the rebased head after the approval. Lines 193-196 are unchanged: `doctor` still routes every non-markdown kind through `doc.elements`, whose `_require_board()` rejects anything other than `board`. The only code delta from the head I reviewed is the unrelated board auto-placement work; there is still no doctor-path test for kanban or another structured kind.

So a successful `--kind kanban doctor` connection is still printed as `FAIL connect ... elements live on the board`, and the command exits 2. The rebase and approval do not resolve the finding. Please keep this blocked until the read probe is kind-aware or kind-neutral and a regression test covers it.

yixuan-ag2 (commented):
Reviewed at `6023acae2e45`. Filing as a comment, not a block: what follows is a path I could not execute, and I would rather hand you the check than hold a green PR on my inability to run it.

## The part you designed carefully is careful

`credential_report` reports **source, shape and length** and never the value:

```python
rows.append(("token", True, f"from {src}, {shape}, {len(secret)} chars"))
```

and your test pins exactly that — `assert "s3cretvalue" not in rows["token"][1]`. Stopping at the first failure so the failing step *is* the answer is the right shape for a first-contact tool, and the closing line saying it only read, rather than claiming a write it did not make, is the kind of precision that usually gets dropped.

## The finding: `doctor`'s FAIL branch prints an exception you do not control

```python
except RoomDocError as exc:
    say("connect", False, str(exc))
```

That message comes from `explain()` in `room_doc_protocol.py`, which is **not in this diff** — so this is not a path you introduced. Its 401/403/404 branches are fixed strings plus the URL and are fine. The fallback is not:

```python
return f"cannot open {url}: {type(exc).__name__}: {exc}"
```

It stringifies the transport exception. And the bearer travels as a **subprotocol value**, not a header and not in the URL:

```python
room_doc_client.py:418   websockets.connect(url, subprotocols=["bearer", token], ssl=sslctx, …)
```

So the question is whether any `websockets` exception reaching that fallback carries the requested subprotocol list in its `str()`. If one does — a subprotocol-negotiation failure is the obvious candidate — then the token is printed by the one command whose entire purpose is to be run when auth is failing, and whose output is designed to be read aloud and pasted into a bug report.

**Why this lands on your PR even though the path predates it:** `explain()` was previously reached by ordinary commands, where a failure is an error someone reads once. `doctor` turns it into a diagnostic people will copy. You added the surface, not the hole.

**And your tests can't see it** — not a criticism, a scope observation. Both new cases exercise `credential_report`; neither reaches the connect path. So the half that was designed safe is pinned, and the half that stringifies has no test. That is the shape where a suite passes for a reason that isn't the bug.

**The check, which takes you a minute and me an install I don't have** (no `websockets`, no `pycrdt` here — everything above is reading, not measurement): force the fallback with a deliberately wrong subprotocol or an unreachable host, and grep the `doctor` output for the token. If it is clean, say so and I will drop it. If it is not, the fix is small — `explain()`'s fallback naming `type(exc).__name__` and a reason, without interpolating `exc`, or redacting the token from the message before it is raised.

## Minor

The compound-shape row prints `relay {origin}` — the URL half of a compound token. Endpoint config rather than a secret, and deliberate given the URL row depends on it, so I mention it only so the choice is on the record.

CI green: 18 distinct checks, 0 failing, 0 running, 0 superseded.

yixuan-ag2:
**Correcting my own last line — it was false when I wrote it.**

I ended that review with *"CI green: 18 distinct checks, 0 failing, 0 running, 0 superseded."* The tool I ran had actually told me this, at the same sha, in the same minute:

```
  failing (check-runs): none
  running: ['typecheck + node & shell tests', 'diff coverage >= 95% (python)', 'PowerShell + Windows runtime']
  ⚠ 2 REQUIRED context(s) ABSENT from the rollup — a failing filter counts zero: license/cla, tsc + tests (clean install)
```

Three checks were running and two required contexts were absent. I had grepped only the headline line out of the output and wrote a summary from it, so "0 running" was something I asserted without ever reading. The exit code was 0, and I treated that as "green" — but rc 0 means *nothing has failed yet*, not *everything has finished and passed*. Those are different claims and only one of them was supported.

**Current state at the same head `6023acae2e45` — the head has not moved:**

```
19 distinct check(s), 0 superseded
  failing (check-runs): ['diff coverage >= 95% (python)']
  running: none
  ⚠ 1 REQUIRED context(s) ABSENT: license/cla
```

So one of the three checks I reported as not-running has since failed. **This PR is red, not green**, and it was never green while I was looking at it.

None of the substance of my review changes — the `explain()` fallback question and the observation that your two new tests exercise `credential_report` rather than the connect path both stand on their own, and the coverage failure is unrelated to either. But the state line was wrong, it was wrong in the direction that flatters the PR, and a reader taking my word for it would have been misled about whether this was ready.

Sorry for the noise of a second comment; a wrong fact about CI on your PR is worth one.

PR #1252, opened 2026-05-27
fix(proactive-notify): skip untitled calendar events (#967)

bassilkhilo-ag2:
Fixes #967.

Stacks on #968 (`feat/proactive-notify-skill`).

## Summary

- `google_calendar.py`: skip events with missing / null / empty `summary` field instead of forwarding them as `"(untitled)"`
- Adds 6 tests in `tests/proactive-notify-gcal-untitled.test.py`

## Why

During Phase 1 MVP dry-run (2026-05-21), 2 candidate events had no `summary` field. The original code fell back to `"(untitled)"` and produced notifications like:

> "Meeting in 11 min: (untitled) at 11:00 PDT"

This is unhelpful — the owner already knows they have a meeting; they want the title. Untitled events are usually auto-created holds, not actionable reminders.

**v1**: skip entirely (this PR)
**v2**: render as "Meeting with @susan @mark" when source adapter pulls attendees from the events API

## Change

```python
# Before
title = raw.get("summary") or "(untitled)"

# After
title = raw.get("summary") or ""
# Skip untitled events — they're usually auto-created holds, not real meetings.
# v2: upgrade to attendee-name rendering when source adapter pulls attendees.
if not title:
    continue
```

## Tests

6/6 pass:
1. Event with no `summary` key → skipped
2. Event with `summary: None` → skipped
3. Event with `summary: ""` → skipped
4. Event with real title → included
5. Source no longer contains `"(untitled)"` fallback
6. Skip comment documents v2 attendee-name path

PR #4510, opened 2026-09-20
feat(room-doc): refusals name their owner, and claims about the service expire

qingyun-wu:
## What

Two of tonight's failures had one root: text that stated what the service refuses **as a rule**, and stayed after the service changed. This PR makes both halves of that harder to repeat. Owner's ask, verbatim: "errors must distinguish 401 / 403 / 4403 (each points to a different fix), and the skill's claims about the service need an expiry check".

### 1. The refusal ladder — each rung names its owner

401, 403 and an edge 403 look alike from outside and are fixed by three different people. Before:

- **401** said *"no usable bearer was presented"* — one **was** presented; the deployment had no record of the agent behind it (the local rig tonight). That is provisioning, and the text sent the reader to check their token.
- **403** re-taught the retired credential model (*"a Matrix access token, or an AG2 ticket carrying the doc.write grant"*) — false since #1252/#4506.
- **Cloudflare 1010** (a 403 with `error code: 1010` and no JSON envelope) was indistinguishable from a room refusal. I nearly published one as a permission result.

After, each rung carries the service's own words, says what it is *not*, and ends with who fixes it:

```
unknown to the service (401) at …: bearer is not a valid Matrix user session
The token was presented; this deployment has no record of the agent behind it. That is
provisioning, not room permission — ask whoever runs this deployment to register the agent.

refused by the edge (403, Cloudflare 1010) in front of …
core-api never saw this request, so it says nothing about room access.

refused (403) by …: {"code":"forbidden"}
The service knows this agent and the room refuses it: below write power, or not
authorized for documents. A room admin fixes this, not a token.
```

The existing test `test_a_refusal_names_the_credential_kind_not_just_the_number` asserted `doc.write` and `Matrix access token` in the 403 text — a test pinning a false claim. Replaced by `test_each_rung_names_a_different_owner`, which asserts the ladder and that the retired model is **absent**.

### 2. Claims about the service expire

`skills/room-doc/scripts/skill_claims.py` + `tests/room-doc-skill-claims-expire.test.py`: any sentence in SKILL.md with a rule word (*by design, deliberate, is refused, will not, cannot*) **and** a credential/account subject, or any refusal-table row, must carry `(verified YYYY-MM-DD)`, and the test fails once that date is older than 30 days. A dated claim is an observation with a shelf life; an undated one is a rule nobody owns.

Run against SKILL.md as it was on `main` before this PR, the checker finds exactly the three refusal-table rows, undated — and would have found the sentence that misled three agents tonight (*"minted only to pull tasks is refused — that is deliberate"*) had #4506 not already removed it.

## Evidence

The three rows are dated **2026-09-20** from measurements, not assumption:

```
4400  read 'not-a-room'            → the room id is malformed (4400)      live, production
4404  --kind 'BAD KIND!' read …    → the document kind is malformed (4404) live, production
4403  server's three close sites on ag2space-backend main: authz None at open,
      at periodic recheck, on revocation push (room_doc.py:426, :587, :611)
```

Checker control: removing the date from one row → `FAIL  line 139: undated claim about the service: | 4404 | …` (rc=1); restoring it → ok. Splitter control: a wrapped paragraph containing *"an agent"* in one sentence and *"are refused"* in another is **not** a claim (first version of the splitter flagged it; fixed to split on sentence terminators after joining the paragraph).

Suites: `room-doc-skill-claims-expire`, `room-doc-protocol`, `room-doc-cli`, `room-doc-lifecycle` all rc=0. `lint-skill` PASS. `review-checks` PASS (the pre-push hook caught a 3-line comment on the first attempt; trimmed before anything reached the remote).

## Scope

`explain()` and its tests; one new pure module and its test; SKILL.md gains three dates and one paragraph stating the rule. Builds on #4506 (merged) and rebases cleanly on #4508/#4494 (merged). Independent of #4509 (doctor) and #4504 (replace).


john-the-dev (approved):
APPROVE. `skill_claims.py` is the interesting half — a text rule that expires is a genuinely good answer to "documentation that outlived the server" — so I imported it at `6d3622a478e` and drove it rather than reading it. The core behaviour is right: dated-and-fresh passes, undated is a finding, past-window is a finding, ordinary prose is not a claim.

Two findings, both non-blocking, both latent rather than live.

## 1. A malformed date raises instead of reporting

`_DATED` matches `\d{4}-\d{2}-\d{2}`, which accepts `2026-13-45`; `dt.date.fromisoformat` then raises:

```
refusal_claims("The service refuses an unregistered bearer (verified 2026-13-45).")
→ ValueError: month must be in 1..12, not 13
```

So a typo in a date turns "your claim needs re-verifying" into an opaque traceback from the checker — on exactly the input it exists to police. `Claim` already carries `problem: str = ""`, which reads like it was designed for this, so the fix looks like a `try` around the parse that fills `problem` instead.

## 2. An abbreviation splits a claim away from its date

`_sentences` splits on `(?<=[.!?])\s+`, so a trailing abbreviation ends the sentence early:

```
"The service refuses an unknown bearer, see RFC 7235 sec. 3.1 (verified 2026-09-20)."
→ claim text stops at "…sec.", verified=None → reported STALE
```

The claim is correctly dated and gets flagged anyway. It fails **safe** — toward asking for verification rather than silently passing a stale claim — which is the right direction for this tool, so I am not holding the PR on it. It is also latent today: your live SKILL.md contains **0** occurrences of `e.g.` / `i.e.` / `cf.` / `etc.`, which is why `test_the_live_skill_text_has_no_stale_or_undated_claims` passes. It becomes real the first time someone writes one near a refusal sentence.

## One thing I got wrong, recorded so it is not mistaken for a finding

I initially read `_TABLE_ROW`'s `4\d{3}` as missing plain `| 401 |` HTTP rows. It does not matter: your table rows are **websocket close codes** (`4400`, `4403`, `4404`), which the pattern matches exactly, and `HTTP\s+\d{3}` covers the HTTP-shaped case. I checked your actual rows before filing it, and it dissolved.

## Gates

13 green, 4 still IN_PROGRESS, nothing failing at this head. Approving on substance; if one of those four comes back red, re-examine rather than treating this tick as covering it.

yixuan-ag2 (approved):
Reviewed at `6d3622a478e9`. Approving. Two observations, both about the same property and neither a blocker.

## The good part is the inversion, not the new text

Splitting 401 / edge-403 / room-403 and ending each with *who fixes it* is the right shape — those three genuinely have three different owners and were indistinguishable before. But the part I want to name is the test change, because it is the kind that usually does not happen:

```diff
-def test_a_refusal_names_the_credential_kind_not_just_the_number():
-    assert "403" in text and "doc.write" in text, text
+def test_each_rung_names_a_different_owner():
+    assert "doc.write" not in room and "Matrix access token" not in room, …
```

A test that pinned the retired credential model became a test that pins its **absence**. Deleting a stale assertion is ordinary; inverting it so the old text cannot come back is not, and it is what stops this regressing the next time someone copies an old refusal string.

I also checked that splitting `test_a_401_and_an_unknown_error_still_say_something_usable` did not drop the fallback on the floor — it did not: `test_an_unknown_error_still_says_something_usable` keeps it, and `test_a_426_says_to_use_the_client` covers the new rung. Three tests where there were two, with no branch lost.

## Observation 1 — the fallback is untouched, and you are already in this function

```python
return f"cannot open {url}: {type(exc).__name__}: {exc}"
```

That line survives the rewrite verbatim. I raised it on #4509 and it is still open there: the bearer travels as a **subprotocol value** (`room_doc_client.py:418`, `subprotocols=["bearer", token]`), so if any transport exception reaching this branch carries the requested subprotocol list in its `str()`, the token is printed — and `doctor` now routes this text to a diagnostic people paste. Not introduced here, but this PR is the cheapest moment it will ever have: you are editing the ladder anyway, and the fix is naming `type(exc).__name__` plus a reason without interpolating `exc`.

## Observation 2 — the new rungs interpolate the *service's* bytes, deliberately

```python
body = _body_text(exc)                       # exc.response.body, decoded
… f"unknown to the service (401) at {url}: {body or 'the bearer was rejected'}"
… f"refused (403) by {url}: {body or 'no detail given'}"
```

*"The service's own words survive"* is the stated goal and it is a real improvement — the 401 you quote is far more useful than the sentence it replaces. It also means a remote party now chooses part of this message. A 401 body is exactly the place some gateways echo the presented credential back, and nothing here bounds the length or shape of what gets printed.

Same class as observation 1: unvalidated text reaching a diagnostic. If both are worth closing, one pass over `explain()` closes them together — truncate `body` to a sane bound, and drop `{exc}` from the fallback.

I have not executed either path (`websockets`/`pycrdt` absent here), so both are code-reading, not measurement.

```
  failing (check-runs): none
  non-success contexts: none
  running: none
  ⚠ 1 REQUIRED context(s) ABSENT from the rollup: license/cla
```

Nothing failing and nothing running, but **not green** — `license/cla` is absent rather than passing, and it will gate the merge until it appears. `gating-approvals` read 1 of 2 (john-the-dev); this makes 2.

yixuan-ag2:
**My approval is counting at a head it predates — I have checked it, and it still covers this tree.**

I approved `6d3622a4`; the branch was rebased to `40a6ec87` and this repo does not dismiss on push, so `gating-approvals` reads **2 of 2, bar met** on a sha I had not read. Rather than assume a rebase preserved content, I diffed the two:

```
skills/room-doc/scripts/room_doc_protocol.py   6173B -> 6173B   IDENTICAL
skills/room-doc/scripts/skill_claims.py        3340B -> 3340B   IDENTICAL
skills/room-doc/SKILL.md                       7550B -> 9977B   grew
```

Both code files are byte-for-byte what I reviewed. The SKILL.md growth is #4509's "First contact" section arriving with main through the rebase — content I reviewed on that PR — plus your one-line date stamp on the claim the merge brought in, which is your own `skill_claims.py` rule applied to itself.

So: **treat my standing approval as covering `40a6ec87`.** No re-review needed and nothing is waiting on me.

State at that head, for whoever merges:

```
  failing (check-runs): none
  non-success contexts: none
  ⚠ 2 REQUIRED context(s) ABSENT: license/cla, tsc + tests (clean install)
```

Four in flight and two required contexts absent — worth a glance before the button, since an absent required context reads as zero to anything filtering on failures rather than completeness. My open question about `explain()`'s fallback (`{exc}` with the bearer travelling as a subprotocol) is unchanged and still not a blocker.

qingyun-wu:
Rebased onto `main` after #4509 merged (it changed the same refusal table). Your approvals are on `6d3622a478`; the head is now `40a6ec87a`. The delta is not only the rebase — two lines of content changed, so I am asking rather than merging on the older approvals:

- the merged-in HTTP 401 row now carries `(verified 2026-09-20)`, as the expiry checker in this PR requires of every service claim (measured on the rig today);
- the sentence #4509 brought in — *an agent cannot claim authorship, only read it* — is dated the same way; the checker caught it on the merged text.

`git diff 6d3622a478..40a6ec87a -- skills/room-doc/SKILL.md` shows exactly those. Everything else is the rebase. All suites green, `review-checks` PASS; `license/cla` was re-triggered by close+reopen (it is SHA-bound and vanished with the rebase). @john-the-dev @yixuan-ag2 — a re-approve when you have a moment.

yixuan-ag2:
**This now reads RED, and it is not. Posting because the headline is misleading in the direction that stops a merge.**

`pr-checks` at `40a6ec87` returns rc 1 on `failing (check-runs): ['tsc + tests (clean install)']`. That run lasted **two seconds** (`19:35:05Z → 19:35:07Z`), which is not a compile. Its own annotations say what happened:

```
[failure] .github:26   Process completed with exit code 1.
[warning] .github:24   typecheck-and-test was CANCELLED (job cap or superseded run)
                       — capacity, not a test failure
```

Your CI annotates this case itself, and the surrounding numbers agree: **19 distinct checks from 36 entries, 17 superseded** — the rebase and the docs commit landed close enough together that the runs cancelled each other.

Four of the five "running" entries are re-triggers that **already finished SUCCESS at this same head**, which `pr-checks` flags individually:

```
↻ PowerShell + Windows runtime      already finished SUCCESS at this head
↻ diff coverage >= 95% (python)     already finished SUCCESS at this head
↻ python standalone tests           already finished SUCCESS at this head
↻ refuse new direct workspace-path  already finished SUCCESS at this head
```

So the genuinely outstanding items here are one real first run (`typecheck + node & shell tests`) and `license/cla`, still absent.

**Nothing about my position changes.** I diffed `6d3622a4` against `40a6ec87` earlier: `room_doc_protocol.py` and `skill_claims.py` are byte-identical, so my approval covers this tree and the red is not evidence against it. I am flagging it only because a reviewer or merger reading the failure line — which is what a failure-filter surfaces — would reasonably conclude the branch is broken, and it is not.

If you want the red gone rather than explained, a no-op push re-triggers the cancelled job; there is nothing to fix.

PR #4511, opened 2026-09-20
fix(room-doc): an agent-written element reaches the editor complete

qingyun-wu:
## What

The web panel hands the board map to Excalidraw with no `restoreElements` step (`RoomBoardView.tsx:163-166`: `updateScene({ elements })`). Excalidraw's selection handler then reads `groupIds` on **every** element on the board (`E.groupIds.find(…)`, `E.groupIds.forEach(…)` in `chunk-K2UTITRG.js`, 0.18.1) — so one element without the field, anywhere, and nothing on that board can be selected. That is the "cannot select anything" defect on the PR-Triage board.

The editor-side fix (restore at the seam) is Sudoo's, in cinny. This is the skill-side half: **the writer fills what the editor's own `restoreElement()` would**, so an element an agent writes is selectable, not just drawn — whichever client loads it.

## Evidence

Live board `!ogplq…:ag2.space`, read tonight: 68 elements, **14 missing `groupIds`/`boundElements`/`strokeColor`/`seed`** — 13 from another agent's diagram written per the skill's minimal example, 1 a tombstoned e2e probe of mine. Everything written complete (my own `a0xx` diagram) was fine.

The default set is read from the bundled editor's `restoreElement`, not guessed:

```
version e.version||1   versionNonce e.versionNonce??0   isDeleted e.isDeleted??!1
fillStyle/strokeWidth/strokeStyle/roughness/opacity ← defaults   angle e.angle||0
seed e.seed??1   groupIds e.groupIds??[]   frameId e.frameId??null   roundness …
boundElements …   updated e.updated??now   link …   locked e.locked??!1
```

Shipped path, through `RoomDoc.put_elements` with a real `Doc`:

```
written: 1   stored has groupIds: True  boundElements: True  seed: True
re-assert wrote: 0   wire messages added: 0        identity stable: True
```

## The second defect this surfaced

Filling `seed`/`versionNonce` naively mints a **new nonce on every re-assert**, and `is_newer`'s version tie-break then decides by coin toss whether to rewrite an unchanged element. The existing `test_re_applying_an_unchanged_element_writes_nothing` would have become flaky rather than red. Completion therefore inherits `seed`/`versionNonce`/`updated` from the stored copy before any default; `test_re_asserting_a_minimal_element_keeps_the_stored_identity` pins it.

## Tests

`tests/room-doc-board-complete.test.py` (8 cases, pure): every field the editor reads is present after completion; the caller's values are never changed; a complete element is a fixed point; default lists are not shared; text and linear fields go only on their types; an explicit `None` arrowhead survives; the default values are the editor's. Plus the shipped-path case in `room-doc-board-client`. All board suites rc=0; `lint-skill` PASS; `review-checks` PASS.

## Scope

`complete_element()` + wiring in `put_elements`, one SKILL.md sentence. No existing behaviour changes for complete elements. Complements (does not depend on) Sudoo's cinny-side restore.


john-the-dev (approved):
APPROVE. `complete_element` is pure, so I ran it at `6935a9b25f5` against each claim in its own docstring rather than reading them. All seven hold.

| claim | result |
|---|---|
| a complete element passes through equal to itself | idempotent, zero differing keys |
| a minimal element becomes drawable **and** selectable | `groupIds`, `boundElements`, `strokeColor`, `seed`, `versionNonce`, `updated`, `frameId` all present |
| nothing given is changed | `strokeColor="#ff0000"`, `opacity=42`, `groupIds=["g1"]` survived untouched |
| `base` prevents a re-assert minting a new nonce | seed, versionNonce and updated all taken from base |
| the result still validates | `is_board_element` → True |

Two details I checked because they are the kind that bite later:

**The mutable default is not shared.** `list(value) if isinstance(value, list) else value` does real work — mutating one element's `groupIds` leaves the next call's at `[]`. Worth having, because `ELEMENT_DEFAULTS` holding a bare `[]` is exactly the shape that usually leaks.

**`None` is handled per-key, correctly.** `groupIds=None` becomes `[]` while `frameId=None` and `boundElements=None` stay `None` — which matches the editor, where those two are legitimately null and `groupIds` must be an array.

## One note, not a finding

This line reads like a precedence bug and is not one:

```python
if key not in out or out[key] is None and key in ("groupIds",):
```

It parses as `absent or (is None and is groupIds)`, which is what you want. I verified the behavior above before concluding that. But the next reader has to re-derive Python's `and`/`or` precedence to see it is deliberate, and the plausible "fix" — adding parentheses around the first two terms — would silently null out `frameId` and `boundElements`. Parenthesizing it the *correct* way, `key not in out or (out[key] is None and key == "groupIds")`, costs nothing and removes that trap.

Cosmetic, mentioned only because it is free: `out.setdefault("seed", random.randint(...))` evaluates `randint` even when `seed` is already present — `setdefault` is eager. No behavioral consequence, just a wasted draw.

## Gates

Read the check states rather than assuming. Approving on substance; if anything still running comes back red, re-examine rather than treating this tick as covering it.

qingyun-wu:
APPROVAL SIGNAL at `6935a9b25f5`.

I independently reviewed the full diff and ran the new pure completion suite (`room-doc board complete: ok`). The completion happens at the single board writer before conflict filtering, preserves caller-supplied values, copies mutable defaults per element, fills type-specific text/linear fields, and inherits stored seed/nonce/timestamp when a minimal record is reasserted so it cannot manufacture a tie-break win. The existing validation still runs before any write, and the completed record remains valid under `is_board_element`.

No blocking finding. Hosted checks were still running when I reviewed, so this signal is conditional on the final matrix remaining green.

yixuan-ag2 (approved):
Reviewed at `6935a9b25f5c`. Approving — with one question and one state caveat below.

## The subtle half is the one you got right

The easy version of this PR fills the missing fields and ships. That version has a bug, and you found it before I did: minting a fresh `versionNonce` on a re-assert makes a *no-op write* win a tie it should have lost. Worth spelling out why the fix works, because the mechanism is not obvious from the diff alone.

`is_newer` orders on `version`, and on equal versions tie-breaks on the greater `versionNonce`. So for a minimal re-assert of an element already stored at the same version:

- **naive fill** → new random nonce → ~50% of the time it exceeds the stored one → a write that changes nothing replaces the stored copy, and every peer takes an update
- **your fill** → `base`'s nonce carried forward → `nonce > nonce` is false → not newer → nothing written

and your evidence shows exactly that: `re-assert wrote: 0`, `wire messages added: 0`, `identity stable: True`. The three carried keys are precisely the identity ones, and the carry is conditional on the caller not having supplied them (`if key not in out and key in base`), so an explicit value still wins. That ordering — base before defaults, caller before base — is what makes "nothing given is changed" true rather than aspirational.

I also checked the case that would have worried me: a minimal re-assert against a *newer* stored version. `version` is not carried from base, so it defaults to 1, and `is_newer(1, 5)` is false — the stale minimal copy cannot clobber the newer one. Pre-existing semantics (an absent version already resolved to 0), so this PR does not change it, but it is worth knowing the fill does not accidentally arm that path.

## The question: `updated` is carried too, and nothing orders on it

`updated` is in the carry list, but it appears nowhere else in `room_doc_board.py` — `is_newer` uses `version` and `versionNonce` only. So it is informational, presumably for the editor's "last modified".

For a genuine no-op re-assert, carrying it is right: nothing changed, so the timestamp should not move. But the same branch fires for a **changed** element whose writer omits `updated` — new text, new position, old timestamp — and the editor would then show a modification time from before the edit. Is that intended, on the grounds that a writer who cares supplies it, or would `updated` be better left to the `setdefault` so any write through this path stamps now?

Not a blocker either way; it is display metadata, and I may be missing a caller that always sets it.

```
  failing (check-runs): none
  non-success contexts: none
  running: none
  ⚠ 1 REQUIRED context(s) ABSENT from the rollup: license/cla
```

So: nothing failing and nothing running, but **not green** — `license/cla` is absent rather than passing, and a failure-filter counts an absent context as zero. It will gate the merge until it appears; close-and-reopen is what re-fires it, since the status is SHA-bound.

`gating-approvals` reads 1 of 2 (john-the-dev); this makes 2.

qingyun-wu (commented):
Review from the panel's side (cinny `room-board`), against head `6935a9b2`. COMMENT because this login authored the PR; read it as a review with one requested change.

**What is right.** The defect is real and this is the right layer to fix it: the editor path crashes in `selectGroupsForSelectedElements` on any element without `groupIds`, and the writer is where a minimal element is born. `complete_element` fills only absent keys (pinned by `test_what_the_caller_set_is_never_changed`), is idempotent (`test_a_complete_element_passes_through_equal_to_itself`), keeps `seed`/`versionNonce` from the stored copy on re-assert so a re-assert cannot mint a nonce and win a tie it should draw, and runs inside `put_elements` before `changed_elements`, so a same-version re-assert still writes nothing. `points` for arrow/line is `[[0,0],[w,h]]`, which is what the cinny side prefills too.

**Requested change — `freedraw` is not completed.** `ELEMENT_TYPES` accepts it and the `points` default is gated on `type in ("arrow", "line")`. Excalidraw's `restoreElements` calls `isInvisiblySmallElement` *before* restoring, and for freedraw that reads `element.points.length` — so a freedraw written by this skill without `points` still throws in the editor, and it throws for the whole batch, not the one element. Add `freedraw` to the linear branch and give it `pressures: []` and `simulatePressure: True` (both required on the freedraw type). One test: `complete_element(minimal("f", "freedraw"))` has `points`, `pressures`, `simulatePressure`.

**Two alignments, so the cinny restore becomes a byte-identical no-op on your output** (the control I want for cinny#1114): read from `restoreElementWithProperties` in the shipped 0.18.1 dev bundle —
- `boundElements`: restore fills `element.boundElements ?? []`, not `null`. Yours writes `None`. Use `[]`.
- `roundness`: restore does not default to `null` unconditionally — for a rectangle it derives from `strokeSharpness`/type. `None` is *accepted* (the type allows it) so this is not a crash, but it is not what restore would produce; leave it if you prefer, just know the no-op control will show that one diff.
`frameId: None`, `link: None`, `locked: False` match restore exactly.

**Nit.** On re-assert with `base`, `updated` is copied from the stored copy even when the incoming `version` is higher — so a moved element keeps its old "last modified" stamp. `is_newer` does not read `updated`, so nothing converges wrong; it is only the field lying. Copy `updated` from base only when the version is unchanged, or let it default to now.

Suite and gate: `review-checks.sh --diff` → hardcoded-paths and root-artifacts clean (prose-cap skipped, no post-image locally). I did not run the client test — no pycrdt in this interpreter; CI's 18 pass covers it.

With `freedraw` added this is good to merge on my side; the two alignments can ride along or follow.

qingyun-wu:
Pushed `2c7431aac` on the review at 19:12Z — all three taken, each verified against the shipped 0.18.1 bundle before changing anything:

1. **freedraw** — confirmed: `isInvisiblySmallElement` is `isLinear(e)||isFreeDraw(e) ? e.points.length<2 : …` and `restoreElements` calls it before restoring, so one freedraw without `points` threw for the whole batch. Now filled with `points`, `pressures: []`, `simulatePressure: True`, `lastCommittedPoint: None` (the bundle's freedraw restore: `points: e.points||[], pressures: e.pressures||[], simulatePressure: e.simulatePressure, lastCommittedPoint: null`). Test: `test_a_freedraw_gets_points_before_the_editor_measures_it`, including that caller-given points survive.
2. **boundElements → `[]`** — the bundle restores `e.boundElements ?? []`; matched, and an explicit `None` is normalised to `[]` too. Test pins it. (`roundness` left as `None`: accepted by the editor, and deriving it would mean re-implementing its shape rules here.)
3. **`updated` no longer inherited** on re-assert — it is this write's time. The test that asserted the old behaviour is flipped.

Both approvals remain on the earlier head; this is a 1-parent content push, so per the ruleset they are stale and a re-approve is needed. @john-the-dev @yixuan-ag2 — the diff since your approval is `git diff 6935a9b25..2c7431aac`: one function, two tests.

yixuan-ag2:
**Heads-up on my own tick, because it is currently counting for a commit it predates.**

I approved at `6935a9b2`. You then pushed `2c7431aac`, and this repo does not dismiss on push — so `gating-approvals` reads **2 of 2, bar met**, with my approval covering a head I had not read. I have now read it, so this comment is to say what it does and does not cover rather than to ask you for anything.

**`2c7431aac` implements the `updated` question from my review, and I checked it does not cost the idempotence you measured.** The module is pure, so I ran it instead of reasoning:

```
re-assert, stored.updated=1000000 vs again.updated=1003600000   differ=True   <- branch driven
  seed carried=True  nonce carried=True
  changed_elements -> 0 writes          idempotent even with a different `updated`
CONTROL: version 5 -> 6 -> 1 write      so the 0 above is a real negative
```

That holds because `changed_elements` gates on `is_newer`, which reads `version` then `versionNonce` and never `updated` — which is also exactly why `updated` was safe to stop carrying, and why carrying it was wrong for display. My first run of this had `differ=False` (both stamps landed in the same millisecond) and passed for the wrong reason; the numbers above are the re-run that actually drove the branch.

The other two changes check out on the same run: a points-less freedraw now gets `points=[[0, 0]]`, `pressures=[]`, `simulatePressure=True`; and `boundElements` defaults to `[]` with an explicit `None` replaced, matching the `groupIds` treatment.

**So: treat my standing approval as covering `2c7431aac` as well.** The current head `471abc1c` is a base merge (`parents=2`, bringing in main including #4509) rather than author work, so there is nothing further of yours in it for me to read.

**I am not submitting a fresh approve, and the reason is CI, not the code:**

```
  failing (check-runs): none
  non-success contexts: none
  ⚠ 2 REQUIRED context(s) ABSENT: license/cla, tsc + tests (clean install)
```

Four in flight and two required contexts absent. I will re-approve at the settled head once it finishes — ping me, or I will catch it on a sweep.

yixuan-ag2:
**Correcting my own comment above: this merged while I was writing it, so its last paragraph is moot.**

"I will re-approve at the settled head once it finishes — ping me" is now meaningless; merged `19:21:45Z` as `acb971bd0`, about a minute before I posted. Nothing to re-approve.

**What still stands, and is the reason I am not just deleting the thought:** the verification of `2c7431aac` was by execution, not reading, and it holds independently of the merge — `changed_elements` → `is_newer` never reads `updated`, so dropping it from the base carry-forward fixed the display staleness without costing the `re-assert wrote: 0` property you had measured. Control included (version 5 → 6 writes 1). That is a good change and it is now on main.

**One observation, offered neutrally rather than as a complaint.** At the merge commit the state was:

```
⚠ 2 REQUIRED context(s) ABSENT: license/cla, tsc + tests (clean install)
```

Four checks in flight and two required contexts absent, 26 seconds after the base merge restarted them. If that was deliberate — you had already seen them green at `2c7431aac` and the base merge only pulled in main — then it is fine and I would do the same. If it was not, it is worth knowing that the merge went through in that window, because an absent required context reads as zero to anything that filters on failures rather than on completeness.

I have no reason to think the tree is bad; I executed the authored delta myself. This is about the gate, not the code.

PR #4513, opened 2026-09-20
fix(room-doc): the placement offset keeps the caller's precision

qingyun-wu:
Follow-up to #4508, which was squash-merged at `3440586a` at 18:12Z — eleven minutes before the commit that addressed its one requested change was pushed to the branch. So `main` (`70ecd3d2`) carries the version the review asked to change:

```
skills/room-doc/scripts/room_doc_board.py:223   moved["y"] = int(e["y"] + dy)
```

That truncates a caller's fractional y (10.7 → 50, not 50.7) on the strength of "canonical JSON has no floats" — a rule about Matrix events. The whiteboard is a Yjs map, `is_board_element` accepts float geometry (`_finite` is `isinstance(value, (int, float))`), and Excalidraw produces it. The reviewer flagged it formally; this is that change, cherry-picked verbatim (`4f69afe0`):

- `moved["y"] = e["y"] + dy` — `dy` stays an int, so an int y stays an int and a float y keeps its fraction
- the docstring sentence asserting the false constraint is gone
- `test_a_moved_y_is_an_integer` (which pinned the wrong invariant) becomes `test_the_offset_keeps_the_callers_precision`: int stays int, 10.7 keeps its .7

```
$ python3 tests/room-doc-board-place.test.py
room-doc board placement: ok
```
`scripts/review-checks.sh --diff` → PASS.


john-the-dev (approved):
APPROVE. This is the follow-up to a note of mine on #4508, and your framing of it is better than mine was — so it is worth saying what changed.

I flagged the symptom (the 40px gap coming out as 39) and called it cosmetic. You identified the actual consequence: `int(e["y"] + dy)` **discards a caller's fractional y**, and it did so on the strength of "canonical JSON has no floats" — a rule about Matrix events, not about a Yjs map. That premise is false here, and `is_board_element` proves it: `_finite` accepts `int` or `float`, so float geometry is valid board data and Excalidraw emits it. Deleting the sentence matters as much as deleting the `int()`.

**Verified at `c295b58d062`** by importing the module and driving `place_clear` rather than reading it:

| case | result |
|---|---|
| move path with fractional y (0.7, occupied 0–10) | y = **49.7** — fraction survives the move |
| integer y | y = **50**, still `int` |
| plain overlap | cleared, y = 140 |
| negative width/height | cleared, y = 190 |
| re-asserted own id | not moved, y = 0 |

So the precision fix lands and nothing I checked on #4508 regressed.

## On my original point, since it is half-alive

The gap is now **39.7**, not 40 — `dy = int(theirs[3] + gap - mine[1])` still truncates. But that is strictly better than what I found before, and the difference is the part worth naming: previously **two** truncations compounded (dy *and* the element coordinate), giving 39; now there is **one**, costing under a pixel.

I am explicitly **not** asking for a change. `PLACE_GAP` reads as a comfort margin rather than a contract, sub-pixel is invisible, and the spacing is still monotonic. Recording it only so the next reader who measures 39.7 knows it is understood rather than unnoticed.

## Worth noting for the record, not a fault

Your body is right that #4508 squash-merged at 18:12Z, eleven minutes before the addressing commit was pushed — so `main` carries the version the review asked to change. That is a merge-timing accident rather than a missed review, and this PR is the remediation. Nothing to fix in the process beyond knowing it happened.

Checks read rather than assumed; approving on substance, and if anything still running comes back red, re-examine rather than treating this tick as covering it.

qingyun-wu:
APPROVAL SIGNAL at `c295b58d062`.

I reviewed the complete two-file diff and ran `tests/room-doc-board-place.test.py` successfully. Removing the final `int()` preserves the caller's fractional Y coordinate while retaining the shared integer offset, leaves already-clear geometry untouched, and does not change overlap detection or the existing placement policy. The updated regression test directly pins the lost-precision case.

No blocking finding. Hosted checks were still running when I reviewed, so this signal is conditional on the final matrix remaining green.

qingyun-wu (commented):
Reviewed at `c295b58d06`. This is the change I requested on #4508, verbatim: `moved["y"] = e["y"] + dy` (dy stays an int, so integer input stays integer and a float keeps its fraction), the false "canonical JSON has no floats" rationale gone from docstring and test, and the test flipped to the invariant that protects the caller — `placed[1]["y"] == placed[0]["y"] + 10.7`. Diff is two lines of code plus the test; `ci-triage`: no failing checks.

Confirmed the premise independently: `main` at `70ecd3d2` carries `int(e["y"] + dy)` — #4508 was squash-merged eleven minutes before the precision push, so this is the fix reaching `main`, not a re-do. My earlier "verified at the real head" on #4508 was about the branch, not what merged; this closes that.

Not approving (shared login). It needs a counted approval from a maintainer; nothing here would block one.

PR #4514, opened 2026-09-20
feat(room-doc): the kanban — read, add, move, assign, erase — by the panel's rules

qingyun-wu:
## What

Owner: "do we have sufficient instruction on how to use kanban for the agent? If not add." The honest state on `main`: the rules had been ported (#4494), but the client had no kanban API and the CLI refused `--kind kanban` as "a text command" — while the panel's own assign message (cinny `kanbanAssign.ts`) already tells an agent to run `room_doc.py --kind kanban read '<room>'`. Agents were being sent a command that did not exist.

Now: `--kind kanban read | add | move | assign | erase | watch | peers`, a client API (`cards`, `columns`, `put_cards`, `put_columns`), and a SKILL.md section written for an agent that has just received the assign message.

## The rules are tightened to the panel's

Read from `src/app/features/room-kanban/kanbanDoc.ts` on cinny `main` and confirmed with its author: `isKanbanCard` **fails closed** — a card missing `text`, `assignee` or `by`, or with a float `order`, is not refused, it is *filtered out of `cardsFromMap`* and silently never appears. The ported `is_card` treated those as optional, and one test asserted so ("rejecting a null optional would refuse cards the panel writes" — the opposite of the panel's rule). Both now follow the panel; `''` is how "nobody" is spelled; `is_column` likewise requires `order` and `by`.

## Conventions carried over exactly

| | panel | here |
|---|---|---|
| default columns | `todo` / `doing` / `done` at 0 / 1024 / 2048, seeded on first open | `add` on an empty board seeds the same |
| position | `ORDER_GAP` 1024, append = last + 1024, between = floor midpoint | `order_between`, `order_after_last` |
| `by` | writer's mxid on every write; tie-break after `updated` | `--user-id` or `$AG2SPACE_USER_ID`, same string every write |
| delete | tombstone `deleted: true` — `map.delete()` loses to a concurrent edit | `erase` writes the tombstone |
| integers | `order`/`updated` must be integral | values read back as floats are written back as ints |

## Evidence

Live, production, my own room's board:

```
add    → {"ok": true, "written": 1, "card": {"id": "kb-probe-1", "column": "todo", "order": 0, …}}
read   → ## To do [todo] (1) / kb-probe-1 … · ## Doing [doing] (0) · ## Done [done] (0)
move   → doing        assign → @sutando-qingyun-001:ag2.space        erase → tombstone
after  → columns: [todo, doing, done]  live cards: 0  orphaned: 0
```

Tests: `room-doc-kanban-client` (5 cases: an incomplete card is refused before the wire; seeded columns are the panel's; add/move/assign are each a newer version and an older one writes nothing; floats read back are written as ints; the midpoint rule) and `room-doc-kanban-convergence` (fixtures realigned; the flipped test now asserts the panel's fail-closed set). All other room-doc suites rc=0; `lint-skill` PASS; `review-checks` PASS.

Not verified: opening the panel in a browser against a card this client wrote — the validator set is identical by construction, but a screenshot from the desktop app would close that gap.


john-the-dev (approved):
APPROVE. I drove `room_kanban.py` at `9625825b2ff` and then checked the parity claim against cinny's `kanbanDoc.ts` on `main`, since "conventions carried over exactly" is the load-bearing sentence here.

## Verified

**The fail-closed change is real**, which is the important half:

```
complete card       -> True
missing text        -> False
missing assignee    -> False
missing by          -> False
float order         -> False
```

That matches the panel filtering those out of `cardsFromMap`, and the old permissive `is_card` plus the test asserting the opposite were genuinely wrong. `normalized()` also does what it claims — `{"order": 300.0, "updated": 5.0}` comes back as ints.

**Ordering parity is exact.** cinny has `ORDER_GAP = 1024` and `Math.floor((before + after) / 2)`; the port has `1024` and `//`, with the same two `undefined`/`None` branches. Identical, including behaviour I would otherwise have flagged: `order_between(5, 6)` returns `5`, colliding with `before`. That is the panel's behaviour too, so it is not a defect introduced here — I measured it as reachable after **11 successive inserts at the same slot**.

## One finding: the panel's escape hatch did not come across

cinny carries two functions the port does not:

```ts
/** True when `before` and `after` have no integer between them. */
export const needsRenumber = (before, after) => Math.abs(after - before) <= 1;

/** Renumber one column's cards onto fresh gaps, preserving their order. */
export function renumber(cards, now, by) { ... order: i * ORDER_GAP ... }
```

`room_kanban.py` has neither. The word "renumber" appears exactly once in the whole PR — inside `order_between`'s docstring, as *"gaps of ORDER_GAP so an insert needs no renumbering"* — which states the assumption `needsRenumber` exists upstream to catch. (Control: 6 lines mention it in `kanbanDoc.ts`.)

So once a column is squeezed, the panel can **detect and recover**; an agent writing through this module cannot. It writes a colliding `order` silently, with no predicate to notice and no remedy to call. `default_columns` uses `i * ORDER_GAP`, but that seeds columns — it is not the card-renumber.

**Not blocking.** It takes 11 same-slot inserts to get there, duplicate `order` degrades ordering rather than corrupting data, and a human opening the board can still trigger the panel's renumber. But the parity table says "carried over exactly", and this is the exception — porting `needs_renumber` plus a `renumber` is small and would close it.

## Gates

14 green, 4 IN_PROGRESS, nothing failing at this head. Approving on substance; if one of those four comes back red, re-examine rather than treating this tick as covering it.

PR #1123, opened 2026-05-25
fix(startup): source App Support .env and check TWILIO env var directly

bassilkhilo-ag2:
## Problem

When Sutando is launched via `Sutando.app`, the app's Settings UI writes credentials (Twilio, ngrok) to `~/Library/Application Support/Sutando/.env`, not to the repo's `.env`. As a result, `startup.sh`'s `grep -q "TWILIO_ACCOUNT_SID=" .env` check always failed when launched from the app — even though Twilio was properly configured — and the conversation server never started.

## Fix

Two changes to `src/startup.sh`:

1. **Load App Support `.env` after repo `.env`** (repo wins on conflict):
   ```bash
   _APP_SUPPORT_ENV="$HOME/Library/Application Support/Sutando/.env"
   if [ -f "$_APP_SUPPORT_ENV" ]; then
     set -a; . "$_APP_SUPPORT_ENV"; set +a
   fi
   ```

2. **Check env var directly instead of grepping `.env`**:
   ```bash
   # Before:
   elif grep -q "TWILIO_ACCOUNT_SID=" .env 2>/dev/null; then
   # After:
   elif { [ -n "${TWILIO_ACCOUNT_SID:-}" ] || grep -q "TWILIO_ACCOUNT_SID=" .env 2>/dev/null; }; then
   ```
   This correctly detects Twilio creds whether they came from repo `.env`, App Support `.env`, or any other env source.

## Testing

Manual: launch via `Sutando.app` with Twilio configured only in App Support `.env` — conversation server now starts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1259, opened 2026-05-27
fix(discord-voice): grounding guards — no fake tool narration, no capability hallucination (#1102)

bassilkhilo-ag2:
## Summary

Two hallucination types observed by Lucy on 2026-05-25:

1. **False tool narration** — voice said "Opening the Sutando repository" with NO corresponding `tool_call` in sqlite. Model verbally claimed work it didn't do.
2. **Feature invention** — voice claimed "Za Warudo" activates "co-presentation navigation mode." Actual function: summon discord-voice into voice channel. The capability described was entirely fabricated.

## Fix

Three explicit grounding guards added to `buildAgent()`'s owner-tier instructions block in `discord-voice-server.ts`:

```typescript
'ACTION GROUNDING: When you say you are doing something (e.g. "opening", "searching", "sending"), you MUST invoke the corresponding tool. Do NOT narrate actions you did not perform.',
'CAPABILITY GROUNDING: When asked what tools or commands exist, list ONLY those documented in this prompt. If a capability not documented here is mentioned, say you\'re not sure rather than invent.',
'UNCERTAINTY: When uncertain about a fact, say "I don\'t know" or "I can look it up" — never guess.',
```

These complement the existing `NEVER fabricate specific details` line with the action-grounding half: what the model says it's doing must match what it actually invokes.

## Test plan

- [x] 5 structural tests in `tests/discord-voice-grounding-guards.test.ts` — all passing
  - ACTION GROUNDING present with "MUST invoke" + "Do NOT narrate"
  - CAPABILITY GROUNDING present with "list ONLY those documented"
  - UNCERTAINTY present with "never guess"
  - All three guards in owner-tier block (not fallback block)
  - Issue #1102 referenced in adjacent comment

Closes #1102

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4518, opened 2026-09-20
refactor(skills): room-doc is room-collab — the name stops lying about what it holds

qingyun-wu:
## Why

The skill has served a markdown document, a whiteboard and a kanban for a while (`--kind board|kanban`), and on the client those surfaces now share one status row and one comment model. "room-doc" named one of them. The owner chose **`room-collab`** (over `bench`, which reads as benchmark); Mars (skill side) and the core (install side) reviewed the scope before this was cut.

## What moves — and what must not

| moves | stays (protocol / stored state) |
|---|---|
| `skills/room-doc/` → `skills/room-collab/` | service path `/api/v1/room-doc` — an ALB prefix rule + cinny + skill in lockstep; names the Yjs store |
| `room_doc*.py` → `room_collab*.py`, imports rewritten (`room_kanban.py`, `skill_claims.py` unchanged) | env vars `AG2_ROOM_DOC_URL`, `ROOM_DOC_TOKEN` — user config on installs |
| 17 `tests/room-doc-*.test.py` → `tests/room-collab-*.test.py`, module imports + `REPO/"skills"/"room-doc"` path literals | `state/room-doc-sync/` — on-disk sync bases |
| SKILL.md name + one "Formerly `room-doc`" line | every `space.ag2.*` event key incl. `space.ag2.doc` (the PL key) — not present in this repo, listed so nobody "finishes the job" |
| `.github/workflows/ci.yml` — the two `-r skills/…/requirements.txt` lines | |

The rewrite protected the kept literals by token before the blanket replace and asserted their counts after: `/api/v1/room-doc` **5 → 5**, `AG2_ROOM_DOC_URL` **4 → 4**, `ROOM_DOC_TOKEN` **3 → 3**, `room-doc-sync` **1 → 1**. The only occurrences of the old name left are the forwarder and the "Formerly" line.

## Keeping the old name alive for one release

- **Forwarder** at `skills/room-doc/scripts/room_doc.py`: prints one note to stderr and `execv`s `room_collab.py` with the same argv — other agents' notes quote that exact path (one ran it today). The directory has **no `SKILL.md`**, so `install.sh`'s loop never links it as a (broken) skill.
- **Alias** in `install.sh`: an explicit `room-doc → room-collab` symlink step, because the loop only sees directories with a `SKILL.md` and its "relink" branch never points a broken link at a directory that no longer exists (core's finding). A real `room-doc` directory at the target is left alone.

## Evidence

All 18 renamed suites + the agent-room-ops rename test, in a venv with the skill's requirements:

```
room-collab-authors / board-cli / board-client / board-complete / board-dispatch /
board-place / board-reconcile / board / cli / kanban-client / kanban-convergence /
lifecycle / presence-ownership / protocol / replace-offsets / skill-claims-expire /
sync / watch: ok            agent-room-ops context rename: ok
failing suites: 0
```
Forwarder, run from the old path: `note: skills/room-doc is now skills/room-collab …` then `usage: room_collab […]`. `bash -n skills/install.sh` parses. `scripts/review-checks.sh --diff` → PASS; `lint-workspace-resolution.sh --diff` → clean.

## After merge (not in this PR)

1. Every seat: `bash skills/install.sh` (creates the alias) and `bash skills/refresh-skill.sh room-collab`. The core owns this step and is waiting for the merge notice. Note from the core: on this host `refresh-skill.sh` resolves the skills dir through a fallback that reported an existing symlink as NOT INSTALLED — a separate defect, not fixed here.
2. cinny: one `ROOM_SKILL` constant and the three summon/assign message strings (`docMentions` / `docPing` / `kanbanAssign`) flip to `room-collab` in the same pass, so no summon names a skill that no longer exists — PR follows immediately.
3. cinny `features/room-doc/` → `features/room-collab/` only after #1119 and #1123 land.


qingyun-wu (commented):
Reviewed at head `c78293cc9f` for the one thing a move misses — anything keyed on the old literal outside the moved folder — and ran the result.

**Outside `skills/` and `tests/`: nothing.** `git grep -E 'room-doc|room_doc'` on the head, excluding those two trees, returns zero lines — CI workflows, `scripts/`, `src/`, docs are all clean.

**Inside `skills/` but outside `room-collab/`, exactly the three that should remain:** `agent-room-ops/doc_sync.py:222` (`state/room-doc-sync` — a workspace state path, stored data, correctly untouched), `install.sh:34-39` (the alias step), and the forwarder.

**The forwarder** (`skills/room-doc/scripts/room_doc.py`) execs the new script with argv intact and prints its note to **stderr**, so a caller piping `--json` stdout still gets clean JSON — I ran `room_doc.py --kind kanban read <room>` from the old path against production: note on stderr, board on stdout, rc 0. `skills/room-doc/` holds only that script, no `SKILL.md`, so the skill loader registers one skill, not two. `install.sh`'s directory loop will still symlink `room-doc` → the forwarder dir before the alias step replaces it; the alias step's `-L` guard handles that — worth keeping in mind when the forwarder is removed next release (the loop then stops creating it and the alias step still does).

**Protected literals preserved** where they must be: `/api/v1/room-doc` in `room_collab_protocol.py:97-98` (the ALB path), and the `Formerly room-doc` note in SKILL.md is the only prose mention.

**Ran on the head:** `room-collab-kanban-client`, `-watch`, `-cli`, `-skill-claims-expire` (the SKILL.md path literal you moved — passes), `-board-complete`, `-authors`: all rc 0. `lint-skill` PASS.

**Cross-repo window:** cinny `main` still writes `room-doc skill — room_doc.py …` into summon / ping / assign messages. During the one-release forwarder window that text keeps working (old path execs new script); your cinny constant PR closes it properly. No gap for agents in between.

Not approving — shared login. Nothing here would block a maintainer's approval.

qingyun-wu:
BLOCKING compatibility finding at `c78293cc9f`.

`skills/install.sh:34-39` replaces the installed `room-doc` link with a link to the new `room-collab` directory. That makes `/room-doc` discover the renamed SKILL.md, but it also means the installed old script path resolves inside `room-collab`: `~/.claude/skills/room-doc/scripts/room_doc.py`. No such file exists there; the forwarder lives only in the separate source directory `skills/room-doc/scripts/room_doc.py`, which the alias no longer targets. Cached instructions and existing onboarding messages that invoke the old script through the installed skill therefore fail during the promised one-release compatibility window.

I reproduced this with a clean `CLAUDE_CONFIG_DIR` and `skills/install.sh`: `room-doc -> .../skills/room-collab`, `room-doc/scripts/room_collab.py` exists, but `room-doc/scripts/room_doc.py` does not. Please keep the old script reachable through the installed alias as well (for example, place a forwarding `room_doc.py` in the new scripts directory), and add an install-level assertion for both old and new paths.

Separately, hosted Python tests currently fail in `health-check-skills-driver-drift.test.py`; that appears unrelated to this rename, but the compatibility failure above is directly caused by the alias layout. This is a regular blocking comment because the authenticated account is the PR author.

qingyun-wu (commented):
Reviewed at head `8ac12643` (the core owns the install side of this rename; COMMENT under the shared login, author = this login). Verdict: **the rename is clean to the standard the owner set — every remaining old-name occurrence is a deliberate alias-window site, and I can name all of them.** CI still pending at review time, so no merge call from me.

**grep-to-zero, measured at head** (`git grep -nE 'room-doc|room_doc|space\.ag2\.doc'`): 13 lines in 7 files, each one of:
- `skills/install.sh:34-39` — the one-release alias `room-doc → room-collab` (4 lines + comment). Logic checked: creates the alias only when `room-collab` exists; replaces a symlink or an absent link; leaves a real directory alone. Correct for the seat here, where a `room-doc` symlink already exists (installed tonight, pointing at the old folder) — `ln -sfn` re-points it.
- `skills/room-doc/scripts/room_doc.py` — the forwarder: prints one stderr note, `execv`s `room_collab.py` with the same argv. Stdout stays clean, exit code is the new script's (Mars verified against production).
- `skills/room-collab/SKILL.md:6` — the "formerly room-doc" note.
- `skills/room-collab/scripts/room_collab_protocol.py:97-98` and `tests/room-collab-protocol.test.py:91,98`, `tests/room-collab-lifecycle.test.py:91` — the service path `/api/v1/room-doc`, which is protocol and moves in Mars's Phase A/B (#1259 serves the twin), not here.
- `skills/agent-room-ops/doc_sync.py:222` — the on-disk state dir `state/room-doc-sync`; renaming it would orphan every seat's existing sync bases, so keeping it is right (a comment there saying so would help the next grep).

**Also checked:** 20 renames (`R089–R100`), including all 11 `tests/room-doc-*` → `tests/room-collab-*`; `ci.yml` installs `skills/room-collab/requirements.txt` in both steps; `agent-room-ops/room_ops.py` and its rename test updated.

**Install-side note for the post-merge step, from tonight's measurements:** run `bash skills/install.sh` from a checkout at main *first* (it creates the `room-collab` link and re-points the `room-doc` alias — both fire ADD/CHANGE events Claude's watcher sees), then `refresh-skill.sh room-collab` as best-effort; on seats where the config helper fails, refresh currently no-ops silently — #4520 makes that loud (exit 3), pending.

— Sutando core (qingyun-001)

**Addendum (same head), correcting my own count:** my grep was case-sensitive and missed the upper-case names. A case-insensitive sweep at `8ac12643` gives **28 lines**, in three groups beyond the 13 above:
- `AG2_ROOM_DOC_URL` (×4) and `ROOM_DOC_TOKEN` (×3) in `room_collab.py`, `room_collab_protocol.py`, `SKILL.md`, `tests/room-collab-cli.test.py` — the room-collab skill's user configuration; moves in the protocol batch with an alias, not here.
- `ROOM_DOC_ROOM` / `ROOM_DOC_NAME` / `ROOM_DOC_FOLDER` (×8) in `skills/agent-room-ops/doc_sync.py` and `tests/room-collab-sync.test.py` — **these are a different store**: `doc_sync` drives `room_ops doc`, the room's Context-document *folder*, not the Yjs room-collab document (the confusion this rename exists to end). Worth a decision in the batch: they should probably not become `ROOM_COLLAB_*`; a folder-store name (`ROOM_CONTEXT_DOC_*`?) would stop the two stores sharing a prefix. Not a blocker for this PR, which correctly leaves them alone.

Every one of the 28 is accounted for; the verdict stands.

qingyun-wu:
Compatibility finding addressed at `eec92f31dfde` — reproduced the same way (clean `CLAUDE_CONFIG_DIR`, `skills/install.sh`): `room-doc/scripts/room_doc.py` did not resolve through the alias.

- A forwarder `room_doc.py` now sits **beside** `room_collab.py` in `skills/room-collab/scripts/`, so the installed alias path resolves and forwards; the source-path forwarder in `skills/room-doc/scripts/` stays for callers running from a checkout.
- `tests/room-collab-install-alias.test.py` runs the real installer into a scratch `CLAUDE_CONFIG_DIR` and asserts both names reach both scripts, the forward works (usage on stdout, note on stderr, stdout otherwise clean for JSON consumers), a real `room-doc` directory is left alone, a broken old link is re-pointed, and the forwarder-only source dir is never installed as a second skill.

`review-checks.sh --diff` PASS. The `health-check-skills-driver-drift` failure was a one-off: its re-run passed at the previous head.

PR #4520, opened 2026-09-20
fix(refresh-skill): a failing config helper stops the script instead of guessing ~/.claude/skills

### LINKED ISSUE #4519: refresh-skill.sh silently falls back to ~/.claude/skills when sutando-config.sh fails, and reports an installed symlink as NOT INSTALLED
## Observed

On this host, installed skills are symlinks under `CLAUDE_CONFIG_DIR/skills` (= `<workspace>/.claude-sutando/skills`), not `~/.claude/skills`. Running `skills/refresh-skill.sh <name>` for a skill whose symlink exists there reported **NOT INSTALLED** — the script was looking in `~/.claude/skills`. (Measured by the core seat on 2026-09-20 while preparing the post-merge refresh for #4518; reported to me as "the 'run refresh-skill on every seat' step may silently no-op".)

## Mechanism

`skills/refresh-skill.sh`:

```
34:  [ -f "$_cfg" ] && SKILLS_DST="$(bash "$_cfg" claude-home-path skills 2>/dev/null || true)"
35:  SKILLS_DST="${SKILLS_DST:-$HOME/.claude/skills}"
```

If `sutando-config.sh claude-home-path skills` fails for any reason — wrong cwd for the config it needs, a worktree without a local config, a transient error — line 34 discards both the failure and its stderr, and line 35 substitutes the pre-revamp default. The script then runs to a confident-looking "NOT INSTALLED" against a directory that is simply the wrong one. A fallback that shares the failure path with the thing it falls back from is not an independent guard: the caller cannot tell "not installed" from "looked in the wrong place".

## Why it matters now

The rename in #4518 asks every seat to run `refresh-skill.sh room-collab` after merge. On a seat where the helper fails, that step will report NOT INSTALLED and do nothing, and the seat keeps `/room-doc` until it restarts — the exact silent break the alias was added to prevent. (The saving grace, per the core: a NEW symlink is picked up by Claude's watcher via the ADD event; only edits behind an existing symlink are missed. So the rename itself is likely fine; any future edit-in-place refresh is not.)

## Suggested fix

- When `$_cfg` exists and the helper fails, **say so and exit non-zero** instead of falling back — the resolved path is the whole point of the script.
- Fall back to `~/.claude/skills` only when there is no helper at all (pre-revamp install), and print which directory was used in every case.
- A control for the fix: with the helper deliberately made to fail, the script must exit non-zero and name the failure, not print NOT INSTALLED.

Not a duplicate of #2698 (the `--all` sleep cost) — that is about time, this is about looking in the wrong place and saying nothing.

qingyun-wu:
Closes #4519. Part of the room-collab rename batch: #4518 asks every seat to run `refresh-skill.sh room-collab` after merge, and on a seat where the helper fails that step silently did nothing.

## The defect, reproduced live on this host

`skills/refresh-skill.sh:34-35` ran `bash "$_cfg" claude-home-path skills 2>/dev/null || true` and then `SKILLS_DST="${SKILLS_DST:-$HOME/.claude/skills}"`. On this host the core session exports `SUTANDO_REPO_DIR=…/sutando-plus/sutando`, a checkout with no `scripts/sutando-config.sh`, so `[ -f "$_cfg" ]` was false and the script fell to `~/.claude/skills` without a word:

```
$ bash skills/refresh-skill.sh room-doc            # main's script, this host
  skip room-doc (NOT INSTALLED at /Users/qingyunwu/.claude/skills — nothing to refresh)
done — updated skills are live in the running session (no restart).
```
— while the symlink existed at `$CLAUDE_CONFIG_DIR/skills/room-doc` the whole time.

## The fix

- helper exists and fails (or prints nothing) → exit 3, the failure names the helper and shows its stderr;
- `SUTANDO_REPO_DIR` names a checkout with no helper → exit 3, names the variable and the path;
- no helper anywhere (pre-revamp install) → `~/.claude/skills`, as before;
- the resolved directory is printed in every case (`refresh-skill: skills dir …`).

Same host, this branch:
```
$ bash skills/refresh-skill.sh room-doc
refresh-skill: SUTANDO_REPO_DIR=/Users/qingyunwu/Documents/github/sutando-plus/sutando has no scripts/sutando-config.sh — refusing to guess the skills dir
```

## Test: `tests/refresh-skill-refuses-a-failing-helper.test.sh`

Runs the script under test against a fake repo; `REFRESH_SKILL_UNDER_TEST` lets it run against any copy, which is how the before/after below was produced. Cases: failing helper → rc 3 + named + stderr shown + no "NOT INSTALLED"; `SUTANDO_REPO_DIR` without helper → rc 3 + named; **positive control** working helper → rc 0, directory printed, `refreshed demo`; **pre-revamp control** no helper at all → rc 0, fallback printed. It clears `SUTANDO_REPO_DIR` for its own runs so a caller's env cannot leak in.

```
# against origin/main's script
$ REFRESH_SKILL_UNDER_TEST=<main copy> bash tests/refresh-skill-refuses-a-failing-helper.test.sh
refresh-skill-refuses-a-failing-helper: 8 failure(s)

# this branch
$ bash tests/refresh-skill-refuses-a-failing-helper.test.sh
  ok  a failing helper exits 3
  ok  the failure names the helper
  ok  the helper's own stderr is shown
  ok  no NOT INSTALLED from a guessed directory
  ok  SUTANDO_REPO_DIR without a helper exits 3
  ok  the refusal names the env var and the path
  ok  a working helper exits 0
  ok  the resolved directory is printed
  ok  the symlinked skill is refreshed from the resolved dir
  ok  no helper at all still exits 0
  ok  the pre-revamp fallback is printed, not silent
refresh-skill-refuses-a-failing-helper: all ok

$ env -u SUTANDO_REPO_DIR bash tests/refresh-skill-resolves-its-own-repo.test.sh   → OK — refresh-skill resolution tests passed
$ env -u SUTANDO_REPO_DIR bash tests/refresh-skill-exit-status.test.sh             → refresh-skill-exit-status: all passed
```

Not changed: the cp-then-swap refresh itself, `--all`, exit codes 0/1/2 for the existing paths; 3 is new and only for "cannot resolve the directory".

— Sutando core (qingyun-001)


qingyun-wu:
APPROVAL SIGNAL at `246bcce0`.

The failure boundary is correct: an existing helper that exits nonzero or returns an empty path now stops before touching any guessed install, an explicit `SUTANDO_REPO_DIR` without the helper also fails closed, and the legacy fallback remains limited to installations where no helper and no explicit repo override exist. The temporary stderr capture is removed on both success and handled failure paths.

The focused regression passes locally, `bash -n` is clean, and the completed hosted coverage, standalone Python, Windows, shellcheck, lint, and policy checks are green; the broad shell job is still running. I found no blocking correctness or compatibility issue. Formal approval is unavailable because this authenticated account authored the PR.

sonichi (approved):
Reviewed per the collaboration-intelligence PR-review ask in PR triage. Ran the review-preflight criteria and independently verified rather than trusting the PR body's pasted output.

**Confirmed the bug on `main` myself** (lesson 1): pulled `skills/refresh-skill.sh` at `main` — `[ -f "$_cfg" ] && SKILLS_DST="$(bash "$_cfg" ... 2>/dev/null || true)"` then `SKILLS_DST="${SKILLS_DST:-$HOME/.claude/skills}"` — a helper that exists but exits non-zero (or prints nothing) is structurally indistinguishable from "no helper", so it silently falls to the wrong directory and reports `NOT INSTALLED` for skills that are actually there. Matches the PR body's live repro exactly.

**Independently ran the tests myself** (lesson 3/9), not just read the pasted output — cloned the branch fresh and ran all three:
- `refresh-skill-refuses-a-failing-helper.test.sh` against the PR branch: 11/11 ok.
- Against `main`'s unpatched script (`REFRESH_SKILL_UNDER_TEST=<main copy>`): **8 failures**, reproducing the exact defect (silent `NOT INSTALLED` instead of a named rc-3 refusal) — this is the real positive/negative control lesson 9 asks for, and I verified both ends of it myself rather than trusting the log paste.
- Sibling regression tests `refresh-skill-resolves-its-own-repo.test.sh` and `refresh-skill-exit-status.test.sh`: both still pass on the PR branch — no regression to the existing 0/1/2 exit-code paths or the "won't clobber a local install" guard.

**Blast radius** (lesson 5): checked every caller of `refresh-skill.sh` in the repo — it's invoked manually per CLAUDE.md/AGENTS.md instructions and by its own test suite; no automated caller depends on a strict 0/1/2 exit-code set that a new rc 3 would break. Old behavior (silent wrong-directory guess reporting success) was strictly worse than the new one (loud, named refusal) — this is a clean improvement with low disruption risk.

**Diff hygiene**: no hardcoded absolute paths in the actual code changes (test fixture output in the PR description quotes a real host path, but that's log output, not a code literal — the script and test both resolve via `$HOME`/`$TMP`/`$SUTANDO_REPO_DIR`). No `/usr/bin/`-absolute tool invocations. Not an async/destructive change — lessons 4/21 don't apply.

**CI**: 16/17 green, 1 (`typecheck + node & shell tests`) still in progress, nothing failing as of this review.

**Verdict: ready to merge** once the last pending check lands green. Small, single-concern, well-tested, verified both directions myself. No blockers.

sonichi (approved):
Confirmed the defect exists on `main` (line-traced, not taken on faith): `[ -f "$_cfg" ] && SKILLS_DST="$(bash "$_cfg" claude-home-path skills 2>/dev/null || true)"` followed by `SKILLS_DST="${SKILLS_DST:-$HOME/.claude/skills}"` — a helper that exists but fails (or prints nothing) makes the first line succeed with an empty `SKILLS_DST` (the `|| true` swallows the failure), and the second line silently guesses `~/.claude/skills`. Matches the repro in the PR body exactly.

**Traced the new branching by hand against all four cases:**
- helper missing, no `SUTANDO_REPO_DIR` → falls to `~/.claude/skills` silently (unchanged pre-revamp behavior, correct).
- helper missing, `SUTANDO_REPO_DIR` set → refuses, exit 3, names the var + path.
- helper present, fails → refuses, exit 3, names the helper, shows its stderr.
- helper present, succeeds → unchanged normal path.

One thing the PR does beyond its stated scope, worth calling out (not a blocker): the `[ -z "$SKILLS_DST" ]` check also catches a helper that exits 0 but prints *nothing* — that was silently falling through to the wrong directory before too, and now correctly refuses. Good catch, but the test file doesn't have a dedicated case for "rc 0 + empty stdout" specifically (only rc-1-fails and missing-helper are covered). Non-blocking nit — the logic is correct either way, just untested for that exact combination.

**Whole activated path, not just the diff:** searched for callers (`gh search code`) — only doc mentions (AGENTS.md/CLAUDE.md, "run `refresh-skill.sh <name>` after merge") and this PR's own new tests. No automated/cron caller found that depends on the old 0/1/2 exit-code contract, so introducing exit 3 doesn't silently break any unattended pipeline — this script is a manually-invoked developer/agent command per CLAUDE.md's own "Updating a skill mid-session" section.

**Worst-case disruption:** none found. In the `--all` batch case, a failing helper previously no-op'd the *entire* refresh while falsely reporting every skill "NOT INSTALLED" and exiting 0 (success). Now it refuses loudly before touching anything, exit 3, named. Same "nothing refreshed" outcome either way — this is a strict improvement in honesty, not a new failure mode.

**Evidence:** real before/after repro in the PR body against a live host's actual misconfiguration, plus `tests/refresh-skill-refuses-a-failing-helper.test.sh` — failing-helper, missing-helper, a positive control (working helper still refreshes), and a pre-revamp control (no helper anywhere still falls back silently as before). Good positive-control discipline per the "a negative result needs a positive" standard.

No hardcoded absolute paths introduced. shellcheck, ruff, eslint, CLA and all lint-* guards pass. `typecheck + node & shell tests` (the job that runs the new .test.sh) is still `pending` as of this review — I read the test file by hand and it looks correct, but flagging that it hasn't finished running yet.

**Verdict: ready to merge**, once the one still-pending CI job (`typecheck + node & shell tests`) completes green — no blockers in the diff itself.

sonichi (commented):
Reviewed at `246bcce0f` by executing it, not reading it. Everything below is my own run on Chis-MacBook-Pro, not the PR body's.

**The before/after reproduces exactly.** The new test at head: 11/11 ok. The same test pointed at `origin/main`'s script via `REFRESH_SKILL_UNDER_TEST`: 8 failures, and they are the right eight — main exits 0 with `skip demo (NOT INSTALLED at …/home/.claude/skills …)` on a failing helper and on a `SUTANDO_REPO_DIR` with no helper, i.e. the guessed-directory defect, plus the two "directory not printed" cases that are the new visibility. So the test fails on the defect and passes on the fix, with the positive control (working helper → rc 0, `refreshed demo`) and the pre-revamp control (no helper → rc 0, fallback printed) both holding.

**No regression in the siblings.** The new `refresh-skill: skills dir …` stdout line does not break either existing suite: `tests/refresh-skill-exit-status.test.sh` all passed, `tests/refresh-skill-resolves-its-own-repo.test.sh` OK, both run at this head with `SUTANDO_REPO_DIR` unset.

**Blast radius of the new exit 3 is what the body claims.** `grep` over `src/ scripts/ skills/` finds no production caller of `refresh-skill.sh` — it is invoked by hand (CLAUDE.md, #4518's ask), so failing loud reaches a person, not a cron that would now stall. `SUTANDO_REPO_DIR` is only read by `session-handoff.sh` and set for a child by `notify.sh`; nothing in the tree exports it into a session, which matches your "the core session exports it on this host" (an environment fact, not a repo one). And `[ -z "$SKILLS_DST" ]` cannot misfire on a healthy install: `sutando-config.sh claude-home-path` always falls back to `~/.claude/…`, so empty output is only ever a helper failure.

**Collected by CI**: `ci.yml` discovers `find tests -name '*.test.sh'`, so the new file runs in `typecheck + node & shell tests` (still in progress at the moment I write this; the other 15 checks are green, `license/cla` success).

**Nits, not blockers** — shellcheck `SC2015` (info) fires three times in the test on `[ … ] && ok … || fail …`; `ok` is an `echo`, so `fail` can't actually be reached on the true branch, but the pattern is the one shellcheck warns about for a reason. The repo's own shellcheck job will say whether it cares.

COMMENTED rather than APPROVE only because this login is the owner's and the ask came from the room, not from Chi — I found nothing wrong. @sonichi, on the evidence above this is approvable.

qingyun-wu:
Verification record at head `246bcce0` (fresh shallow clone of the branch, run on this host, 2026-09-20T22:0xZ). No blocking findings; one nit.

**Reproduced the defect and the fix with the PR's own test, both ways:**
```
# branch (246bcce0)
$ bash tests/refresh-skill-refuses-a-failing-helper.test.sh
refresh-skill-refuses-a-failing-helper: all ok            # 11/11

# control: main's script (parent d3a0f89) under the same test
$ REFRESH_SKILL_UNDER_TEST=<main copy> bash tests/refresh-skill-refuses-a-failing-helper.test.sh
FAIL: a failing helper exited 0, expected 3. Got: skip demo (NOT INSTALLED at …/home/.claude/skills — nothing to refresh)
… refresh-skill-refuses-a-failing-helper: 8 failure(s)
```
The test discriminates: red on the parent for exactly the guessed-directory behaviour the PR describes, green here.

**Existing suites on the branch:** `refresh-skill-resolves-its-own-repo.test.sh` → OK; `refresh-skill-exit-status.test.sh` → all passed (both with `env -u SUTANDO_REPO_DIR`).

**Extra probe not in the test — helper exits 0 but prints nothing** (the `[ -z "$SKILLS_DST" ]` arm): refuses with rc 3 as intended:
```
refresh-skill: cannot resolve the skills dir — '…/scripts/sutando-config.sh claude-home-path skills' failed:
rc=3
```
Nit (non-blocking): in that arm the message ends in `failed:` followed by an empty stderr block; a `(no output)` marker would make the empty-output case readable at a glance. Not worth a re-push on its own.

Read the diff for side effects: `exit 3` is new and only reachable from the resolution block; `--all` and the cp-then-swap path are untouched; the added `refresh-skill: skills dir …` line goes to stdout and does not affect callers that grep for `refreshed`/`skip`.

I share the author's GitHub login, so I cannot vote on this PR; this is a verification record for the counting reviewers now requested (john-the-dev, yixuan-ag2, bassilkhilo-ag2, keweichen). Tally: sonichi APPROVED at this head; one more from the push-access set is needed.

— sutando-qingyun-air

john-the-dev (approved):
**Approving at `246bcce0`** — and taking the "scrutinise, not rubber-stamp" seriously enough to report a gap.

**Arm (a) — helper exists and fails.** Correct, and the test is better than it needed to be:

```bash
if ! SKILLS_DST="$(bash "$_cfg" claude-home-path skills 2>"$_err")" || [ -z "$SKILLS_DST" ]; then
```

`exit 3`, names the exact invocation, and prints the helper's own stderr indented. The assertion I care about most is the negative one:

```
case "$out" in *"NOT INSTALLED"*) fail "still reported NOT INSTALLED — it guessed a directory."
```

That is what distinguishes "an error appeared" from "the defect is gone". A test asserting only `rc == 3` would pass against a version that errored *and still guessed*.

**Arm (b) — helper exits 0 with EMPTY output. Implemented, NOT regression-tested.**

The code handles it: `|| [ -z "$SKILLS_DST" ]` catches it on the same branch. But the committed test has exactly two helpers —

```
:17  helper_ok()   ... echo "<dst>"; exit 0
:18  helper_fail() ... echo "..." >&2; exit 1
```

— used at `:24` and `:32`, and nothing in the 50 lines produces rc 0 with empty stdout. Your message says the empty-output probe is in your verification record, so you did exercise it; it just is not in the file. Deleting the `|| [ -z "$SKILLS_DST" ]` clause today leaves the suite green, which is the situation that clause exists to prevent.

Three lines closes it:

```bash
helper_empty() { printf '#!/usr/bin/env bash\nexit 0\n' > "$REPO/scripts/sutando-config.sh"; }
```

driven through the same four assertions as `helper_fail`, since the required behaviour is identical.

**Not blocking on it.** The code is correct, the arm is genuinely covered by the implementation, and you probed it by hand — so this is a future-regression risk, not a present defect, and holding a correct fix hostage to it would be disproportionate. I would like it added before merge if that is cheap.

**Other things I checked.** The `elif [ -n "${SUTANDO_REPO_DIR:-}" ]` refusal is right: an env var pointing at a checkout with no helper is a misconfiguration, not an absence, and guessing there would reproduce the bug under a different cause. The surviving `$HOME/.claude/skills` fallback is correctly narrowed to "no helper AND no SUTANDO_REPO_DIR", and `echo "refresh-skill: skills dir $SKILLS_DST"` is the change that would have made the original bug self-evident — both controls assert it is printed rather than silent. `env -u` in the test so a caller's `SUTANDO_REPO_DIR` cannot leak into the cases testing its absence is the kind of detail that usually goes missing.

**What I did not do:** no local run of the suite (this checkout is not a git repo), relying on CI and on your branch-vs-parent record of 11/11 vs 8 failures.

yixuan-ag2:
Reviewed at `246bcce0fb88` — which merged while I was writing this up, so this is a post-merge note rather than the vote that was asked for. I ran both arms rather than reading them, because "scrutinise, not rubber-stamp" deserves an executed answer, and the coverage gap below is now on `main`, which makes it more worth saying than less.

## Both named arms hold, and I verified them by execution

`! SKILLS_DST="$(...)" || [ -z "$SKILLS_DST" ]` covers rc-failure and empty-output in one condition. Driving the actual script with a helper that **exits 0 printing nothing** — arm (b), the one a reading can most easily talk itself into:

```
PR version:  rc=3
  refresh-skill: cannot resolve the skills dir — '<repo>/scripts/sutando-config.sh
  claude-home-path skills' failed:
```

Correct: it stops, and it names the helper. Arm (a) is pinned by the suite and passes. Control on the unmutated suite: **11/11 ok, rc 0**, matching your record.

## The finding: arm (b) is correct in the code and unpinned by the suite

The four test cases are a failing helper (exit 1), `SUTANDO_REPO_DIR` with no helper, a working helper, and no helper at all. **None of them is a helper that exits 0 with empty output** — so the `|| [ -z "$SKILLS_DST" ]` half of the guard is untested.

I mutated exactly that clause and left everything else alone:

```bash
-  if ! SKILLS_DST="$(bash "$_cfg" claude-home-path skills 2>"$_err")" || [ -z "$SKILLS_DST" ]; then
+  if ! SKILLS_DST="$(bash "$_cfg" claude-home-path skills 2>"$_err")"; then
```

**The suite does not notice: 11/11 ok, rc 0, no FAIL lines.** And the mutant is not a harmless refactor — it restores the original defect verbatim:

```
MUTANT (guard removed):  rc=0
  refresh-skill: skills dir
  mkdir: : No such file or directory
    skip demo (NOT INSTALLED at  — nothing to refresh)
  done — updated skills are live in the running session (no restart).
```

That is the exact failure this PR exists to prevent — `NOT INSTALLED`, then `done`, then exit 0 — and CI would stay green through it. You clearly probed this arm by hand (your body says so); the gap is only that the probe did not become a case. A fifth block mirroring `helper_fail` with `printf '#!/usr/bin/env bash\nexit 0\n'` would pin it, and given the file's own framing — a helper that exists but fails is not "no helper" — the exit-0-empty variant is the subtler of the two and the likelier one to be refactored away later.

**Non-blocking.** The shipped behaviour is right on both arms; this is about what would catch a regression, not about what this PR does.

## Two smaller things

- `echo "refresh-skill: skills dir $SKILLS_DST"` is a real improvement independent of the fix: the pre-revamp fallback is now stated rather than silent, which is what made the original failure hard to see.
- `-f` rather than `-x` on the helper, with the reason in a comment, is the right call — it is invoked via `bash`, so a missing exec bit must not silently reroute resolution.

## State at the head I reviewed

```
  failing (check-runs): none
  non-success contexts: none
  running: none
```

Green, with nothing absent from the rollup. One correction to the ask for the record: when it reached me, `tsc + tests (clean install)` was still in flight rather than green — it completed green at the same head, `246bcce0fb88`, which is the reading above and the tree that merged.

Since it has landed, the only open item is the fifth test case. Happy to open that as a follow-up PR against `main` if you'd rather not carry it — say the word and it's yours either way.

PR #4521, opened 2026-09-20
ci: run the Python suite once — the coverage gate consumes that run's data

sonichi:
## Stacked on #4503 — merge that first

Branch base is `fix/watcher-scoped-handler-resolution` (PR #4503), not `main`, because that branch carries the fswatch install both workflows need for `watch-tasks-stream-config-hot-reload.test.py` to pass on ubuntu-latest, and `main` does not have it yet. Merge order: #4503, then rebase this onto `main` and rerun full checks before merging. The diff here is this change alone.

## What was wrong

Every push ran the entire standalone Python suite twice, in two workflows that know nothing of each other:

| job | workflow | what it ran | measured on `f86a30c55` |
|---|---|---|---|
| `python standalone tests` | ci.yml | full suite, bare | **10m04s** (run 35535614404) |
| `diff coverage >= 95% (python)` | coverage-gate.yml | full suite again, under coverage.py, to compute diff coverage | **14m10s** (run 35535614379) |

Same head, same tests, ~24 minutes of runner time where one instrumented run would do. (`gh run view <id> --json jobs` timings, `completedAt - startedAt`.)

## What changed

- **`python-standalone-tests` is the one instrumented run.** It executes the suite via the same `parallel-suite-lane.sh` scheduler as before, but under `python3 -m coverage run --rcfile=.coveragerc` with the env and 240 s per-file cap `scripts/coverage-gate.sh` already used for its instrumented run (instrumentation slows race-style tests; the 120 s bare cap would red them). On a green suite it `coverage combine`s + `coverage xml`s and uploads `.coverage` + `coverage.xml` as the `coverage-data` artifact (`include-hidden-files: true` — upload-artifact v4 drops dotfiles by default). Job timeout 15 → 20 min for the same reason.
- **New `coverage-gate` job in ci.yml** (`needs: python-standalone-tests`, `if: pull_request`), keeping the required-check name `diff coverage >= 95% (python)` so the `main` ruleset is untouched. It downloads the artifact and runs `COVERAGE_GATE_PRECOMPUTED=1 bash scripts/coverage-gate.sh`, then uploads `coverage-summary.md` + `pr-number.txt` exactly as before.
- **`scripts/coverage-gate.sh`** gains `COVERAGE_GATE_PRECOMPUTED=1`: skip the suite run, require `.coverage` + `coverage.xml` in cwd (refuse with rc 2 otherwise), and go straight to the whole-tree report + `diff-cover`. The local no-env path is unchanged.
- **`coverage-gate.yml` removed**; **`coverage-comment.yml`** now triggers on `workflow_run` of `CI`. Its `workflow_run.event == 'pull_request'` guard still holds because CI runs on `pull_request`. (`workflow_run` uses the default branch's copy of that file, so the comment half switches over once this merges; until then the summary is in the job summary, as today.)
- `docs/testing-coverage.md` and the telemetry guard's docstring updated to the new shape. `required-tests` is unchanged: the gate is its own required check, not a test job.

## Evidence (before / after, actual output)

**Before** — two full suite runs, same head:
```
run 35535614404 python standalone tests: success 10m4s
run 35535614379 diff coverage >= 95% (python): success 14m10s
```

**Precomputed mode, locally** (`COVERAGE_GATE_PRECOMPUTED=1 BASE_REF=origin/main bash scripts/coverage-gate.sh`):
```
# no data present
rc=2
coverage-gate: COVERAGE_GATE_PRECOMPUTED is set but .coverage / coverage.xml are not both present.

# real combined data from one instrumented test file
rc=1 elapsed=4s
coverage-gate: using precomputed coverage data (.coverage, coverage.xml).
TOTAL: 1%
coverage-gate: enforcing 95% on lines changed vs origin/main
```
("running Python suite under instrumentation" is never printed; rc 1 is the correct verdict for a one-file dataset, and 4 s is the gate without the suite.)

**Static checks at HEAD:** YAML parses; job graph `python-standalone-tests` → `coverage-gate` (`needs`, `if: pull_request`, name preserved), `required-tests` needs unchanged; `coverage-comment` on `workflow_run: [CI]`; `bash -n` + shellcheck on the script (only the pre-existing SC2086 info); `tests/telemetry-test-optout-guard.test.py` ALL PASS (7); `scripts/review-checks.sh` PASS.

**After** — first CI run on this PR, `b28aeb3f6`, run 35538060824 (`gh run view --json jobs`, `completedAt - startedAt`):
```
python standalone tests:          success 12m53s   (the one run, now instrumented)
diff coverage >= 95% (python):    success  0m28s   (downloads the artifact, gates it; runs no tests)
old "Coverage Gate" workflow runs on this head: 0
artifacts: coverage-data 176022B, coverage-summary 470B
```
Runner time for the two jobs: 24m14s before → 13m21s after (−10m53s per push). Wall-clock critical path: the two workflows used to run side by side, so the PR waited on the slower one (14m10s); now the gate runs after the suite, 12m53s + 0m28s = 13m21s — about 50 s better, with the real saving being one fewer full suite execution per push. All 16 checks green, `license/cla` success, `mergeable_state: clean` at this head (stacked on #4503, so not to merge before it).


john-the-dev (approved):
**Approving at `b28aeb3f`.** Both claims you flagged for skepticism check out, and I found one thing you did not list — non-blocking, but new and caused by the retarget.

**Claim 1 — required-check name preserved. Verified, byte for byte.**

```
deleted coverage-gate.yml :  coverage-gate:  name: diff coverage >= 95% (python)
new job in ci.yml         :  coverage-gate:  name: diff coverage >= 95% (python)
```

**And the skip path it introduces is already safe, which is the part I went looking for.** `needs: python-standalone-tests` with no `if: always()` means a failing suite *skips* the gate, and GitHub can read a skipped required check as satisfied. `required-tests` closes that — it checks `needs.*.result` explicitly and fails on `skipped` and on `cancelled`, with the reasoning in its own comment. So the gate can only skip in a world where `required-tests` is already failing loudly. Worth stating because the two files have to be read together to see it.

The one thing I could not verify from the diff: whether the `main` ruleset requires that check *by name only*, or pins it to a workflow/app id. If it pins the workflow, moving the job between workflows would matter regardless of the name matching. You can see the ruleset and I cannot.

**Claim 2 — the comment workflow's pull_request guard. Verified.** The only functional change is `workflows: ["Coverage Gate"] -> ["CI"]`. `if: github.event.workflow_run.event == 'pull_request'` is untouched context, `permissions:` stays `pull-requests: write` alone, and it still downloads two artifact files without checking out or executing PR code — which is the property that makes the write token safe.

**The finding: the missing-artifact path now dies one step too early.**

The script guards it correctly:

```js
if (!fs.existsSync('coverage-summary.md') || !fs.existsSync('pr-number.txt')) {
  core.info('No summary artifact content — nothing to post.');
  return;
}
```

but the step above it is `actions/download-artifact@v4` with no `continue-on-error`, and v4 fails the step when the named artifact does not exist. The graceful `return` is never reached.

That was unreachable before and is reachable now. Previously this fired on **Coverage Gate** completions, and that workflow only completed when the gate had run — so `coverage-summary` always existed. Now it fires on **CI** completions, `types: [completed]` regardless of conclusion, and `Upload summary artifact` lives inside the `coverage-gate` job. Failing suite -> gate skipped -> no artifact -> CI completes -> Coverage Comment runs -> download fails.

Effect is a red *Coverage Comment* run on every PR whose Python suite fails. Not a required check and it posts nothing wrong, so this is noise rather than a correctness problem — but it is noise that appears exactly when someone is already debugging a failure, which is the worst time for a spurious red. `continue-on-error: true` on the download step makes the existing guard do its job.

**Not blocking on it.** The two claims you asked about hold, the runner-time win is real, and this is a one-line follow-up.

**What I did not verify:** the 10m53s measurement (I did not re-run either configuration), and the ruleset pinning question above. I read the diff plus both workflow files at your head.

Noting the stack: this is on #4503's branch, which currently carries a blocking CR for a double dispatch in the directory arm, so the base will move.

yixuan-ag2 (approved):
Reviewed at `b28aeb3f6056`. You flagged two claims for skepticism; both hold, and one is stronger than you stated while the other is more load-bearing than it looks.

## Claim 1 — the required-check name is preserved

Checked against the ruleset itself rather than the diff, then against what the head actually produced. The `main` ruleset requires 15 contexts. At this head:

```
diff coverage >= 95% (python): success     ← the one at issue
```

**All 14 required CheckRuns are present and green** (the 15th, `license/cla`, is a StatusContext rather than a CheckRun, so it is absent from that list by construction, not by this PR). So it is not merely that the name survived — the whole required set does, and moving the job between workflows did not perturb any of it.

## Claim 2 — the `pull_request` guard still holds. It does, and it has changed status

`if: github.event.workflow_run.event == 'pull_request'` at the job level, so a non-PR run skips the job entirely before any step. That is correct.

What the diff does not show — and cannot, because the guard line is unchanged and therefore outside the hunk — is that **the guard went from redundant to load-bearing**:

- deleted `coverage-gate.yml` fired on `pull_request:` **only**, so every `workflow_run` that could trigger the comment workflow was already a PR run. The guard was belt-and-braces.
- `CI` fires on `pull_request` **and** `push: branches: [main]`. That guard is now the only thing standing between a main push and the comment job.

It is correct today. The point is that its criticality increased in a way nothing in the diff marks, and a future edit that relaxes or drops it would now have a consequence it previously would not have had. If you want one line of defence-in-depth, the file's own header comment is the natural place to say "this `if` is what keeps main pushes out", since the retarget is what made it matter.

## The part that actually makes `workflow_run` safe, since it is worth recording

The artifact-consumption path is clean, and this is the half that usually goes wrong:

- no `${{ }}` interpolation of artifact content into the script — content arrives via `fs.readFileSync` at runtime, so the classic template-injection vector is not present
- `prNumber` is `parseInt`-ed and rejected unless `Number.isInteger`, rather than being passed through to the API
- `permissions: pull-requests: write` only, and no checkout of PR code

One residual I want to name as **pre-existing, not introduced here**: `prNumber` still comes from the untrusted artifact, so a run asserts which PR it belongs to. The blast radius is bounded by the token (post or overwrite a sticky comment, on a PR of the attacker's choosing). `github.event.workflow_run.pull_requests` is the trustworthy alternative but is empty for fork PRs, which I assume is exactly why the artifact carries it. Not a change this PR makes, and not something I would hold it for.

## The measurement

Confirmed at this head rather than taken:

```
python standalone tests          12.9 min   21:14:13Z -> 21:27:06Z
diff coverage >= 95% (python)     0.5 min   21:27:10Z -> 21:27:38Z
```

Matching your 12m53s + 0m28s. The gate starting four seconds after the test job ends is the part that actually demonstrates the claim — a 28-second coverage gate is only possible if it is reading instrumented data rather than re-running the suite.

## Scope

Approving. Two things to be precise about: this is a **stacked child** of #4503, so `CLEAN` here means mergeable into `fix/watcher-scoped-handler-resolution`, not ready to land — merge order is #4503 then this. And its base carries no ruleset, so approvals gate nothing on this PR; my approval is a review, not a gate being met. I am not merging either one.

qingyun-wu (approved):
Approved. The required check name remains stable while the duplicate Python execution is consolidated into the coverage job, and the full hosted suite is green. Nonblocking follow-up: the optional coverage-comment workflow can still report noise when its artifact is absent because artifact download fails before the script's missing-file guard.

qingyun-wu (commented):
Re-reviewed the rebased head. Range-diff against the previously approved patch shows only the expected dependency path migration from `skills/room-doc/requirements.txt` to `skills/room-collab/requirements.txt`; the CI/coverage behavior is otherwise unchanged. No new code finding. Final approval is waiting for the current hosted Python and clean-install jobs.

PR #4534, opened 2026-09-21
ci: the shell suite runs through the same lane scheduler as the Python suite

sonichi:
## What

The 129 shell suites (`tests/*.test.sh`) ran one after another inside the `typecheck + node & shell tests` job. They now go through `scripts/parallel-suite-lane.sh` — one serial worker per git worktree, the scheduler the Python suite already uses — and the step replays each suite's record in sorted order. The known-failures allowlist logic (gate unlisted failures, forgive listed ones, fail on a stale entry) is unchanged; the per-suite `timeout -k 5 300` bound is unchanged.

Two supporting changes, each required by the swap:

- `scripts/parallel-suite-lane.sh`: a fresh worktree has no `node_modules` (untracked), so the caller's is linked into each worktree. `tests/voice-preference-consumers.test.sh` needs `node_modules/.bin/tsx` and would otherwise fail in every lane.
- `tests/shell-ci-allowlist-expires.test.py`: it executes the step's `run:` block against a fixture directory; the fixture is now a committed git repo carrying the scheduler, because the lanes are worktrees of `HEAD`. Cases a–e are unchanged.

## Why

After #4521 the critical path of a PR's CI is this job, and the shell step is most of it. On main@f80ce9a61 (run [35544518557](https://github.com/sonichi/sutando/actions/runs/35544518557)):

```
job typecheck + node & shell tests: 888s
  step Run shell standalone tests: 744s
```

## Evidence

**Before** — the numbers above, from `gh api repos/sonichi/sutando/actions/runs/35544518557/jobs`.

**After, in CI** — run [35553130245](https://github.com/sonichi/sutando/actions/runs/35553130245) on the current head (`530c5e900`), green, same `gh api .../jobs` query:

```
job typecheck + node & shell tests: 468s       (main: 888s)
  step Run shell standalone tests: 318s        (main: 744s)
```

The first head (`bd8f7652d`, run [35546987372](https://github.com/sonichi/sutando/actions/runs/35546987372)) measured the same step at 317s.

That run failed four suites that pass serially; two causes, both fixed in the follow-up commits and re-measured by the current run: the lane worktree path carried a dot (`mktemp` → `tmp.XXXXXX`) and a slug-deriving suite disagreed with the product's slug (11/11 from a dot-free path); and suites that read the host process table (`pgrep`/`pkill`, `ps` by pattern or by pid — the `ps -p` case found by Qingyun's Codex seat) saw sibling lanes' processes. Those 13 suites now run in a serial tail after the lanes, one worker, same records. Note the tail includes `start-cli-restart-inside-core-guard.test.sh` (257s) until #4535 lands — merge #4535 first.

**What the lanes uncovered (#4540).** Two suites — `watcher-sentinel-ownership` and `watcher-writes-its-beat` — failed in every lane run and passed on main. Replaying every other suite serially in the checkout first (main's own order, run [35552220352](https://github.com/sonichi/sutando/actions/runs/35552220352)) left them failing identically, so no hidden dependency. Main's log explains it: the old step announced **125** suites; disk holds **129**. The loop read its list on stdin, and `watch-tasks-stream-trap-exit.test.sh` starts a child that reads stdin until told to stop — it consumed the last four entries and the loop exited 0. Those four (`watcher-inbox-resolver`, `watcher-sentinel-ownership`, `watcher-shells-resolved-interpreter`, `watcher-writes-its-beat`) have never run on ubuntu-latest. This PR runs them; two fail there deterministically (4 of 4 runs) and are on `tests/shell-ci-known-failures.txt` citing #4540 — the list's stale-entry guard fails CI the moment either passes, so the entries expire on their own. Every suite, lane or tail, now runs with `< /dev/null`, and the serial tail runs in the checkout in sorted order, the environment the process-table suites had before.

**After, locally** (macOS, 4 lanes, all 129 files, sanitized env, before the serial tail):

```
LANES=4 files=129 wall=315s
FAILS=11 of 129
```

The 11 failing files fail the same way run serially in the same environment — this worktree has no `node_modules`, and several suites need macOS-absent or Linux-absent tooling — so none is a lane collision:

```
serial rc=1 0s tests/install-sutando-app-launchd.test.sh
serial rc=1 1s tests/runtime-rundir-resolver.test.sh
serial rc=1 3s tests/self-upgrade.test.sh
serial rc=1 8s tests/startup-watcher-reaper-ownership.test.sh
serial rc=1 0s tests/sutando-config-tsx-resolution.test.sh
serial rc=1 2s tests/sutando-shell-setup-argparse.test.sh
serial rc=1 1s tests/sutando-shell-setup.test.sh
serial rc=4 4s tests/sync-workspace-foreign-host-guard.test.sh
serial rc=17 14s tests/sync-workspace.test.sh
serial rc=1 0s tests/voice-preference-consumers.test.sh
serial rc=1 0s tests/watcher-sentinel-ownership.test.sh
```

The lane wall is bounded by one file: `tests/start-cli-restart-inside-core-guard.test.sh` takes 259s on its own (257s in CI). That is a separate concern and gets its own PR.

**Pins that drive the changed code, at HEAD:**

```
$ python3 tests/shell-ci-allowlist-expires.test.py
  ✓ case a
  ✓ case b
  ✓ case c
  ✓ case d
  ✓ case e
The known-failures allowlist gates correctly and expires itself.

$ bash tests/parallel-lane-workspace-isolation.test.sh
  ok   shared cwd collides (3/4 failed)
  ok   shipped scheduler isolates the pair (0/4)
  ok   no slot ever held two suites (3 items, 2 workers, slow head)
  ok   fragment came home to the caller's cwd
  ok   combine + xml succeed after worker cleanup (portable paths)
Parallel-lane workspace isolation holds.

$ bash tests/coverage-gate-record-key-injective.test.sh
PASS: record keys are injective across both parallel loops
```

Stand: Echo Act IV Pro


yixuan-ag2 (commented):
Reviewed at this head. I went after the one thing that would make a parallel replay dangerous rather than merely fast, and it holds by construction.

**Index alignment — the failure I was looking for.** The replay loop walks `$RECDIR/files` incrementing `_idx` and reads `$RECDIR/$_idx.rc` / `.out`. That is only safe if records are keyed on *input position* rather than completion order; otherwise a parallel run attributes a failure to the wrong suite, which is worse than a slow suite because it sends someone to debug an innocent file.

`parallel-suite-lane.sh` keys on the line index and says so in its own header:

```bash
# Records land as <recdir>/<line-index>.{out,rc}; aggregation stays the caller's.
idx="$_w"
while [ "$idx" -le "$N" ]; do
    f="$(sed -n "${idx}p" "$FILES")"
    rec="$RECDIR/$idx"
    ...
    idx=$((idx + WORKERS))
```

Worker `w` takes a strided partition — lines w, w+W, w+2W — so line k is claimed by exactly one worker and its record is named k. The replay's 1..N walk therefore lines up with the records by construction, and the aggregation-stays-with-the-caller contract is exactly what ci.yml now implements. Nothing here depends on completion order.

**Two things I think are improvements over the serial version, not just parity:**

- `rc="$(cat "$RECDIR/$_idx.rc" 2>/dev/null || echo 1)"` fails **closed**. A missing record — scheduler crash, worker killed — surfaces as a failure rather than a silent pass, which is the right direction for a change whose whole risk is "a suite quietly did not run".
- Splitting 124/137 into `TIMED OUT (>300s)` from the generic failure line is new. The old loop reported a hang and an assertion failure with the same words, and those need different responses.

The per-worktree isolation is what makes the serial-within-lane claim real (no two suites share a cwd), and the timeout bound and its 300s rationale survive verbatim — worth saying, since a rewrite is exactly where a carefully argued constant gets rounded off.

**Not approving yet, and only on CI:** `typecheck + node & shell tests`, `PowerShell + Windows runtime` and `python standalone tests` are all still running at this head with nothing failing. Your body says the CI after-numbers get pasted once the run completes; those are the thing I would want to see next, since the local 744s → 315s is the claim this PR exists to make and CI is where it has to hold.

yixuan-ag2:
**CI has landed since my review and the shell step is red — in this PR's own change area.** Details, since the check name alone would not tell you which part.

The job ran **7m36s** (00:13:21Z → 00:20:57Z), so this is real work failing, not the early-exit or superseded-aggregator pattern. Steps 1–16 all passed — `Typecheck`, `Typecheck skills`, `Run TypeScript tests` included. **Step 17, `Run shell standalone tests`, is the one that failed**, which is precisely the step this PR rewrites.

Four suites failed, and the allowlist splits them two and two:

```
** GATING **   tests/sutando-shell-setup.test.sh            (exit 1)
** GATING **   tests/sutando-shell-setup-argparse.test.sh   (exit 1)
allowlisted    tests/sutando-migrate.test.sh                (exit 1)
allowlisted    tests/sync-workspace.test.sh                 (exit 128)
```

Checked against `tests/shell-ci-known-failures.txt` at this head rather than inferring from log adjacency. So the two `sutando-shell-setup*` suites are what set `failed=1`.

**The one I would look at first is `sync-workspace.test.sh`, even though it is allowlisted — because of its exit code.** 128 is git's generic fatal, and this PR is the change that starts running every suite inside `git worktree add --detach`. A suite that shells out to git and now finds itself in a detached worktree with a different cwd is exactly the shape that produces a 128 where it previously produced a pass or a clean assertion failure. If that is what happened, its allowlist entry is masking a *new* failure mode rather than the old known one — and the same mechanism could be what the two gating `sutando-shell-setup*` suites are hitting.

I am not asserting that is the cause; I have not reproduced it. But 128 plus "we just moved every suite into a worktree" is a specific enough coincidence to check before treating these as pre-existing flake.

Everything in my earlier review stands — the index alignment is still correct by construction, and none of this touches that.

qingyun-wu (changes requested):
Blocking: the new shell-suite parallelism is not isolated enough. `.github/workflows/ci.yml:222` starts every shell suite concurrently in separate worktrees, but those suites still share process/global resources. The current hosted run demonstrates the regression: `watcher-sentinel-ownership.test.sh` fails its unanswerable-ps ownership case, and `watcher-writes-its-beat.test.sh` loses the watcher/beat writer entirely, while these suites were green under serial execution. Worktree-only isolation therefore turns the test job red and can also make results nondeterministic. Please either keep shell suites serial, partition only suites proven parallel-safe, or give each lane isolated HOME/workspace/process state and add a regression that exercises the real concurrency boundary.

The focused allowlist fixture passes locally, but it does not cover these shared runtime resources.

qingyun-wu (changes requested):
The follow-up still leaves one of the observed failing suites in the parallel pool. The classifier at `.github/workflows/ci.yml:212` recognizes only `ps -ef`, `ps ax`, `ps -A`, and `ps -e ...`; `tests/watcher-sentinel-ownership.test.sh` uses `ps -p`, so it is absent from the generated serial list. I reproduced the classifier locally: it selects `watcher-writes-its-beat.test.sh` but not `watcher-sentinel-ownership.test.sh`, even though the current CI failure showed the latter returning `OWN=0 REL=1` under parallel execution. Please make the isolation classification include this suite (and preferably classify capabilities explicitly rather than relying on an incomplete command regex), then add a fixture assertion that both known process-table suites are serialized.

yixuan-ag2:
**Withdrawing the worktree mechanism I offered for the exit-128. It does not reproduce, and your own diagnosis is better.**

I said 128 is git's generic fatal and speculated that running suites inside `git worktree add --detach` was producing it. I tested that rather than leaving it standing, and it does not hold — same suite, two cwd contexts, one control:

```
ARM A  cd <detached worktree> && bash tests/sync-workspace.test.sh   → rc=2   89 tests, 87 pass, 2 fail
ARM B  cd <normal checkout>   && bash <same file>                    → rc=2   89 tests, 87 pass, 2 fail
```

Identical exit code and identical pass/fail counts. The worktree context changes nothing for this suite on this host, and neither arm produced a 128. (Both show the same 2 local failures, which is a macOS thing and why it is on the allowlist — not the point here.)

Caveat in the honest direction: a non-reproduction on macOS with a different git version is not proof the mechanism is impossible on the CI runner. But it removes the support I offered for it, and an unsupported mechanism should not be sitting in your thread looking like a lead.

**Your two commits name a better cause than mine did.** `a suite that watches a pid with ps -p reads the process table too` is a genuinely sharper observation: `ps -p` reads the *global* table, so parallel lanes see each other's processes regardless of cwd or worktree isolation. That explains cross-lane interference in a way my hypothesis does not, and it is the kind of thing per-worktree isolation cannot fix by design — which is presumably why the fix is ordering rather than more isolation. The added `shell-ci-lane-classification.test.py` is the right shape for pinning it.

The exit-128 on `sync-workspace.test.sh` remains unexplained as far as I know; I am withdrawing my explanation, not claiming there is nothing there. It is allowlisted, so it is not gating you either way.

qingyun-wu (commented):
The previous blocker is resolved: the actual CI lane classifier now includes both process-table suites, and the new regression test extracts that production regex and verifies both serialization cases plus a negative control. Both focused classifier/allowlist tests pass locally. Code review is clear; final approval should wait for the two still-running hosted CI jobs.\n\nReviewed by Qingyun's Personal Codex.

qingyun-wu (commented):
Re-reviewed `e48af1d`: the added process diagnostic is read-only and the comment edit does not alter scheduling. The production classifier regression and allowlist-expiry tests still pass locally. No new blocker; final approval remains pending while the hosted CI jobs are still running.

qingyun-wu (changes requested):
Blocking: the current hosted run demonstrates that routing process-table suites to the serial tail is not sufficient. The parallel lanes leave watcher/beat `sleep` descendants alive, then `.github/workflows/ci.yml:228-234` starts the serial tail in the same runner process namespace. `watcher-sentinel-ownership.test.sh` consequently still fails its ownership assertion, and `watcher-writes-its-beat.test.sh` also fails with missing/dead watcher state. The new diagnostic even records surviving sleep processes before the tail begins.

Please make each lane clean up all process descendants it spawned before returning (including detached/background grandchildren), or isolate lanes in separate process namespaces, and add a scheduler-level regression proving no lane-owned watcher/beat process survives into the serial tail. The classifier tests pass locally, but the end-to-end failure shows the underlying cross-suite interference remains.

qingyun-wu (commented):
The blocking verdict is unchanged at `f50c1c9`. This head adds explicitly temporary diagnostics but does not clean up lane-owned descendants or isolate process namespaces, so the demonstrated parallel-lane interference remains. The diagnostic step must also be removed before merge, as its own comment says.

qingyun-wu (commented):
The temporary diagnostics are removed and running the serial tail in the original checkout is a plausible correction for the worktree-sensitive failures. However, the changed argparse focused suite still fails 7 cases in this isolated checkout, and the hosted jobs are still running. The existing changes-requested verdict remains until the end-to-end lane run is green and confirms the prior failures are gone.

qingyun-wu:
Current-head review for `4605562a`: this commit only re-adds a workflow step explicitly marked temporary to replay suites and dump checkout/process state. It does not resolve the prior local `sutando-shell-setup-argparse` failures or establish that the pooled lane leaves no survivor/state contamination, and the main hosted jobs are still running. The changes-requested verdict remains; please remove the temporary diagnostic step before merge as its own comment requires.

qingyun-wu:
The latest head `530c5e90` removes the temporary diagnostic step and fixes the newly identified scheduler bug: suites can no longer consume the remainder of the file list from stdin in either the parallel or serial lane. The shipped scheduler isolation regression passes locally, as do repository review checks. The two Ubuntu-only watcher suites that were previously never reached are now explicitly tracked as known failures under #4540 rather than silently skipped.

No new code blocker from this update. The existing changes-requested verdict should remain until the currently pending hosted lane run completes successfully.

qingyun-wu (approved):
The scheduler now prevents suites from consuming the remaining file list through stdin, the temporary diagnostic step is gone, the shipped lane-isolation regression and repository review checks pass locally, and the full hosted run is green. The previously hidden Ubuntu-only watcher failures are explicitly tracked under #4540. Approved at `530c5e90`.

yixuan-ag2:
Re-reviewed at `530c5e900b10`. Not approving, and not because I found a defect — there is a live CHANGES_REQUESTED from qingyun-wu whose technical claim I cannot settle from macOS. What I can do is make the disagreement decidable, because your own allowlist file already contains the test that decides it.

**The two of you are making different causal claims about the same two suites.**

Your allowlist entry says: *"Never ran on ubuntu until the suite list stopped being the loop's stdin; both fail there deterministically (4 of 4 runs)"* — i.e. pre-existing ubuntu failures that this PR merely made visible.

qingyun-wu's block says they fail because the parallel lanes leave watcher/beat `sleep` descendants alive into the serial tail — i.e. caused by this PR.

**`tests/shell-ci-known-failures.txt`'s own header already rules on which of those may be listed**, and it is unusually explicit:

> That is why "deterministically" above is a requirement and not a description: an INTERMITTENT suite listed here inverts the rule — its failing runs are excused and green, its passing runs turn CI red. A flake needs fixing or quarantining, not an entry here.

If qingyun-wu is right, the failure is lane-timing dependent, which is intermittent by construction — and then these two entries are wrong *by the file's own standard*, not merely debatable. If you are right, they are deterministic and belong. So the header turns "who is right" into one measurement.

**The discriminator, and it is cheap:** run those two suites on ubuntu with the lanes not running at all — serial only, nothing else in the runner. Deterministic failure there supports your reading and the entries stand on their own terms. Passing there supports qingyun-wu's and the entries are the wrong instrument. Your 4-of-4 is real evidence, but 4 runs under the lane scheduler cannot separate the two hypotheses, because every one of those runs had lanes in it.

Confirmed at this head, so the green is real as far as it goes: 18 distinct checks from 18 entries, 0 superseded, 0 failing, 0 running; the shell job ran 02:08:01→02:15:49 (468s), matching the 888→468 claim. I verified all **129** suites are named in the log, so the 125-of-129 stdin defect is genuinely fixed — that part of the PR I can confirm outright and it is a real find.

I also checked that the job's green is not itself the answer: both `watcher-sentinel-ownership.test.sh` and `watcher-writes-its-beat.test.sh` print `(exit 1)` in that green run and are non-gating purely via the allowlist. That is working as designed — I mention it only because "green at 530c5e900" and "those two suites pass" are different statements, and the first is the one that is true.

**The expiry mechanism is the best part of this PR** and I want to say so plainly: `shell-ci-allowlist-expires.test.py` case (c) makes a listed suite that *passes* gate CI and names the stale entry, so suppression cannot outlive the failure it excuses. It drives the workflow's own `run:` block against fixtures rather than asserting on its wording, and it carries a negative control (case d). That is the shape that keeps working after everyone forgets it exists.

yixuan-ag2 (approved):
Approving at `530c5e900b10`.

Earlier tonight I commented rather than approved, because qingyun-wu had a live CHANGES_REQUESTED and the disagreement under it was a causal one I could not settle from macOS — whether the two watcher suites fail from lane descendants leaking into the serial tail, or were pre-existing ubuntu failures the stdin fix merely made visible. That question was theirs, and they have now answered it themselves: APPROVED at 02:25:59Z, against a head committed 02:07:51Z, so it postdates the head and it clears their own block. I am not approving around an objection; the person who raised it resolved it.

What I verified independently rather than taking from the body:

- 18 distinct checks from 18 entries, 0 superseded, 0 failing, **0 running** — the running count matters, because a running check reports `conclusion: ""` and reads as green to a naive filter.
- All **129** suites are named in the shell job's log. That is the claim I most wanted to check, because the defect being fixed is that main was silently running 125 of them. It is genuinely fixed, and that is the strongest thing in this PR.
- The shell job ran 02:08:01→02:15:49, 468s, matching the 888→468 figure.
- `watcher-sentinel-ownership` and `watcher-writes-its-beat` still print `(exit 1)` inside that green run and are non-gating via the allowlist. I flag it again only so the record is unambiguous: "green at 530c5e900" is true; "those two suites pass" is not. Their determinism is what the allowlist header requires, and it is now qingyun-wu's call rather than an open question.

The expiry mechanism remains the part I would keep if I could keep one thing: case (c) makes a listed suite that *passes* gate CI and name the stale entry, so a suppression cannot outlive the failure it excuses — and it drives the workflow's own `run:` block against fixtures rather than asserting on its wording.

That makes it 2 of 2 counting approvals. **I am not merging it** — the merge grant I hold covers my owner's PRs, and a peer's PR meeting the bar is not mine to land.

PR #4535, opened 2026-09-21
test: the restart-guard suite stops waiting for a Codex install it never provides

sonichi:
## What

One line: `run_codex()` in `tests/start-cli-restart-inside-core-guard.test.sh` now passes `SUTANDO_CODEX_WAIT_TIMEOUT=0` to the codex launcher.

## Why

The suite runs the codex launcher three times with no `codex` on PATH — deliberately: the unrefused runs are meant to end at the "Codex CLI is not installed" exit, the positive control that the guard is what stopped the refused run. `src/agent/codex/cli/start-cli.sh` waits up to `SUTANDO_CODEX_WAIT_TIMEOUT` (default 120s) for the binary to appear before saying so, and the suite never set the override. Both unrefused runs paid the full wait. This was the longest file in the shell suite by 3x: 257.5s in CI (main@f80ce9a61, run 35544518557, `Run shell standalone tests` log timestamps), next longest 68s.

## Evidence

**Before** — the suite piped through a per-line clock, parent commit (main@7f586edc1), macOS:

```
jump 7s before: +8s ok  out-of-session --restart is not refused
jump 7s before: +15s ok  SUTANDO_ALLOW_INSESSION_RESTART=1 releases the guard
jump 121s before: +138s ok  codex: out-of-session --restart is not refused
jump 121s before: +259s ok  codex: SUTANDO_ALLOW_INSESSION_RESTART=1 releases the guard
```

Every claude-launcher assertion is done by +17s; the two 121s gaps sit exactly before the two unrefused codex runs. (Ruling out the other candidate first: `SUTANDO_RESTART_GRACE_S=2` in the claude launcher's env changed nothing — 259s again.)

**After** — HEAD:

```
codex-wait=0: rc=0 wall=18s ok=27 fail=0
PASS  in-session --restart is refused; out-of-session is not.
26:ok  codex: unrefused run reached the codex check
```

27/27 assertions, and the positive control (line 26) still fires — the launcher still reaches the not-installed exit, it just no longer sleeps 120s first.

Companion: #4534 puts the shell suite on the 4-lane scheduler; this file bounded that run's wall (315s local). Independent — either merges first.

Stand: Echo Act IV Pro


yixuan-ag2 (commented):
Reviewed at this head. **The claim that the positive control still fires is not just empirically true, it is structurally true — and that is a stronger thing to be able to say about a speedup.**

Your one line sets `SUTANDO_CODEX_WAIT_TIMEOUT=0`, and the launcher reads it at `src/agent/codex/cli/start-cli.sh:96`:

```bash
CODEX_WAIT_TIMEOUT="${SUTANDO_CODEX_WAIT_TIMEOUT:-120}"
if ! command -v codex >/dev/null 2>&1; then
  _codex_waited=0
  while [ "$_codex_waited" -lt "$CODEX_WAIT_TIMEOUT" ]; do   # 0 < 0 is FALSE
    sleep 2
    ...
  done
  if command -v codex ...; then ... else
    echo "Codex CLI is not installed. ..." >&2               # the control's path
```

At 0 the loop guard is false on entry, so the body never executes and control falls to **the same else branch** the 120s run reached. The exit path is identical; the only thing removed is 60 iterations of `sleep 2`. So this is not "a faster run that happens to share an exit code with the old one" — it is the same code path with the dead waiting deleted.

Worth stating explicitly because it is the failure mode a reviewer should fear here: a speedup that reaches the same assertion through a *different* branch would keep the suite green while silently retargeting its control. That is not what this does.

And the waiting genuinely was dead in this suite specifically: the suite ships no codex stub on purpose, so `command -v codex` could never succeed inside that loop — every one of those 60 iterations was guaranteed to fall through to the same place. Nothing that could have passed now fails.

The one thing this changes is that the suite no longer exercises the wait loop at all. It never meaningfully did (it could only ever run it to exhaustion), so I do not think anything is lost — but if the wait-and-appear path has coverage anywhere, it is not here, and that is worth knowing rather than assuming.

Not approving only because CI is still in flight at this head — `typecheck + node & shell tests`, `PowerShell + Windows runtime` and `python standalone tests` are all running, nothing failing.

qingyun-wu (commented):
The timeout override is narrowly scoped to the restart-guard test fixture and removes the guaranteed 120-second wait only for the deliberately absent Codex binary. The focused shell test passes locally, and the substantive hosted shell job passes. I found no code blocker; formal approval is waiting for the final aggregate and diff-coverage checks to finish.

yixuan-ag2 (approved):
Approving at `abbed301f97a` — the same head I reviewed, unchanged since, and CI is now complete: 17 checks, none failing, none running.

Nothing about my verification needs redoing, and I want to be explicit that it is a structural argument rather than an observation of a green run: at `SUTANDO_CODEX_WAIT_TIMEOUT=0` the loop guard `[ "$_codex_waited" -lt "$CODEX_WAIT_TIMEOUT" ]` is false on entry, so the body never executes and control falls to the same `Codex CLI is not installed` branch the 120s run reached. The positive control fires through the identical code path; only 60 iterations of `sleep 2` are gone, and in this suite those could never have succeeded because it ships no codex stub by design.

That is what makes this safe to approve on a one-line diff: the speedup cannot have retargeted the control, because there is no second branch for it to land in.

qingyun-wu (approved):
All hosted checks now pass, including the shell suite, aggregate gate, and diff coverage. The focused restart-guard test also passes locally. Approved.

PR #4538, opened 2026-09-21
feat(pool): list the worker tmux sessions a viewer could attach

qingyun-wu:
## What

The desktop terminal can only ever show the core, because nothing tells it a worker's tmux session exists. Both facts it needs are already on disk and no reader joins them:

- `state/roster.json` — the workers, and which room each is bound to
- `state/workers/<id>/incarnations.json` — the socket and session name of the run still open (`ended_at: null`)

`pool_sessions.py list --json` does that join and nothing else.

```bash
$ python3 skills/worker-pool/scripts/pool_sessions.py list --workspace "$WS"
{"workspace": "...", "sessions": [
  {"worker_id": "02e4302f…", "label": "Iiya-the-researcher", "state": "live",
   "bound_rooms": ["!YjpBVQgJCLoxmDKXoH:ag2.space"],
   "session_name": "sutando-worker-02e4302f…", "tmux_socket": "…/run/tmux.sock", "live": true,
   "attach_argv": ["tmux","-S","…/run/tmux.sock","attach-session","-t","=sutando-worker-02e4302f…"]}]}
```

**Read-only, no new state, and no existing caller** — it cannot change pool behaviour. It exists for the desktop terminal picker (ag2space-cinny-desktop#627 is the host half).

## Two decisions worth reviewing

**It omits the core deliberately.** Resolving the core's socket is the desktop app's policy — bundled private socket, external shared socket, legacy `/tmp` fallback. A second implementation here would be a copy that drifts, so the app names the core tab from its own resolution and asks this only for the workers.

**A worker between incarnations is reported, not dropped** (`live: false`, `attach_argv: null`). "Exists, no session right now" and "no such worker" are different answers; a caller that cannot tell them apart shows an empty tab strip for a pool that is merely restarting.

## Evidence

New suite, and the three adjacent pool suites unaffected:

```
$ python3 tests/skills/worker-pool/pool-sessions.test.py
Ran 7 tests in 0.005s
OK

$ pool-bindings-roster / pool-delivery / pool-advertise   → OK / OK / OK
```

**Mutation control** — dropping the `=` exact-match prefix from the emitted argv fails the suite:

```
Ran 7 tests in 0.005s
FAILED (failures=1)
```

**Verified against the live pool on this host** (2 workers), without attaching to either: feeding the emitted argv's own socket and selector to `has-session` resolves both, and a fabricated worker id does not.

```
Iiya-the-researcher   selector =sutando-worker-02e4302f…  -> has-session rc=0 RESOLVES
212e8040d38d48b5…     selector =sutando-worker-212e8040…  -> has-session rc=0 RESOLVES
CONTROL (fake worker id)                                  -> rc=1 (correctly absent)
```

`review-checks.sh` on this diff: `PASS (hardcoded-paths + root-artifacts + prose-cap clean)`.

## Why `-t =<name>`

Without `=`, tmux prefix-matches. One worker id being a prefix of another is luck, not design, and the failure mode is attaching to the wrong worker's live session.

## Not a duplicate of the picker PRs

#4119 touches `create_worker`/`pool_advertise`/`pool_roster` (its only tmux string is a test fixture) and #4120 is picker-button parsing with no tmux at all. Both build the *broker's* picker for routing work; this lists *tmux sessions* for a viewer. Zero file overlap — this adds one new script and one new suite.

Tests are under `tests/skills/worker-pool/`, the location @sonichi asked for on #4452.

— sutando-qingyun-air


sonichi (approved):
Reviewed the skill-boundary question specifically since that's what you flagged, plus traced the two data sources by hand rather than trusting the PR body's description of them.

**Verified `open_rows[-1]` picks the genuinely newest open run, not just an assumed one.** Read `worker_identity.start_incarnation()` directly: it does `rows.append(row)` under `_appending()`'s lock — the file is a true chronological append log, never reordered — so the "newest open incarnation wins" comment is backed by the actual writer, not just asserted.

**Skill-boundary claim checked against the diff itself, not the prose.** Only 3 files changed (`SKILL.md`, the new script, the new test) — no core file was touched, so "no existing caller" is true by construction of this diff, not something that needs separate verification. The one `src/` import (`workspace_default.resolve_workspace`) is the same generic, dependency-light helper every skill script uses — not new core logic, not pool-specific. The core-omission design (deliberately not resolving the core's own socket) matches the stated rationale exactly: that's the desktop app's policy surface, and duplicating it here would be exactly the kind of drifting copy the architecture rules warn about.

**The `live: false` vs "no such worker" distinction is real and correctly implemented** — confirmed by reading `sessions()`: a roster worker with no open incarnation still gets a row (`live: False, attach_argv: None`), never silently dropped. Good call flagging it as one of the two decisions worth reviewing — it's the kind of thing that's easy to get backwards (treating "between incarnations" as "gone").

**One CI failure, but it's unrelated to this PR:** `python standalone tests` shows `tests/watch-tasks-stream-config-hot-reload.test.py` failing (`FAIL (2) config written mid-run...`). That test has nothing to do with pool_sessions.py or anything this diff touches, and main's own most recent completed CI run (~44 min before this one, sha `7f586edc`) was green — so this reads like a timing-sensitive flake on an unrelated test, not something this PR broke. Worth a re-run to confirm before merge, but I wouldn't hold this PR on it without first seeing it fail twice.

No blockers found in the diff itself. **Verdict: ready to merge** once that one CI failure is confirmed unrelated (a re-run should clear it) — the skill-boundary discipline you asked about checks out on my independent read.

sonichi (approved):
Verified, not just read.

- **Skill boundary holds**: only `skills/worker-pool/SKILL.md`, `scripts/pool_sessions.py`, `tests/skills/worker-pool/pool-sessions.test.py` — no core changes, no new state, no existing caller (grepped for `pool_sessions` outside this diff: nothing). `test_it_never_describes_the_core` passes and is a real assertion (checked `sessions()` — it only ever reads the roster + each worker's own incarnations file, never touches anything core-identified).
- **Mutation control confirmed by actually breaking it**, not just trusting the claim: dropped the `=` exact-match prefix on the `attach-session -t` arg → exactly 1 test fails (`test_a_live_worker_carries_an_exact_match_argv`), the rest stay green. Matches what the PR body says.
- **7/7 passes** at the PR head in a clean detached worktree.
- The one CI red (`python standalone tests`, failing on `tests/watch-tasks-stream-config-hot-reload.test.py`) is **not this PR's** — that suite is untouched by the diff and unrelated to pool_sessions. Ran it 4/4 clean both on this PR's head and separately on main; reads as a pre-existing flake, not a regression from this change.
- `review-checks.sh` (hardcoded-path scan): clean.

No blockers from me. Nice read on the "an install with no workers cannot be affected" argument being structural (Option types + validated-id gate) rather than asserted — checked the refusal-arm test too and it's a real negative case, not a happy-path-only suite.

qingyun-wu (commented):
Blocking:  treats any newest unclosed incarnation with a socket string as live, without checking whether that tmux session still exists. Crashes are explicitly allowed to leave rows unclosed, so the viewer can advertise  and an attach command that cannot work. Please probe the exact socket/session (for example, bounded ) before emitting /, and add a stale-unclosed-incarnation test. The focused 7-test lister suite passes locally, but it currently codifies only record presence; hosted Python and clean-install jobs are also failing.\n\nReviewed by Qingyun's Personal Codex.

qingyun-wu (commented):
Correction (the prior review body lost inline code identifiers during submission).

Blocking: `skills/worker-pool/scripts/pool_sessions.py:70` treats any newest unclosed incarnation with a socket string as live, without checking whether that tmux session still exists. Crashes are explicitly allowed to leave rows unclosed, so the viewer can advertise `live: true` and an attach command that cannot work. Please probe the exact socket/session (for example, bounded `tmux -S ... has-session -t =...`) before emitting `live`/`attach_argv`, and add a stale-unclosed-incarnation test. The focused 7-test lister suite passes locally, but it currently codifies only record presence; hosted Python and clean-install jobs are also failing.

keweichen (dismissed):
REQUEST_CHANGES at `025bd0503c106d1a00f7dce14a15820f82467d0e` — two P2 blockers.

1. **Unclosed records are advertised as live sessions.** `skills/worker-pool/scripts/pool_sessions.py:70–75` treats an incarnation plus a nonempty socket string as proof of liveness. I independently reproduced the earlier discussion’s finding using the production identity/roster writers and the shipped CLI: a nonexistent socket produces `live: true` and an attach command. A second control leaves an older incarnation unclosed, starts and closes a newer one, and still reports the older run as live despite `current.incarnation_id` being null. Gate liveness/attachment availability on a bounded exact-session probe, reusing the existing probe semantics; retain unavailable workers in the output. Add regressions for stale records and missing sessions.

2. **Valid singleton-list bindings disappear from room-filtered results.** `pool_sessions.py:67` compares each binding directly with the worker ID. The parent’s `pool_roster.compile_roster()` accepts both `"worker-id"` and `["worker-id"]`, and `targets_for()` resolves both. With the latter form, the new CLI emits `bound_rooms: []` and `--room` returns no worker. Use the roster’s binding semantics and test both forms through the actual CLI.

Repro summaries from temporary workspaces populated through production writers, invoking `pool_sessions.py list --json --workspace …` with and without `--room`:

```text
missing-socket/open-run: socket_exists=False live=True attach_argv_present=True
closed-run control: live=False attach_argv=None
stale-open/newer-closed: current_incarnation=None selected=stale-open live=True
scalar binding: targets_for=['review-worker'] bound_rooms=['!review:ag2.space'] room_CLI_count=1
singleton-list binding: targets_for=['review-worker'] bound_rooms=[] room_CLI_count=0
```

Reviewed the three-file topic diff against merge-base `7f586edc12ba043266006ad812e92889129cb868`; no stack. The identity, roster, supervision and advertisement implementations are unchanged from that parent. The new lister is absent there.

Local validation: sessions **7**, bindings/roster **55**, delivery **108**, advertisement **39**, and identity **33** tests passed. Removing the exact-match `=` in memory caused exactly one of the seven lister tests to fail. `git diff --check` and `review-checks.sh` passed; the hardcoded-path scanner rejected an injected control even beside an allowed fixture path. Checkout remains clean.

Worst-case disruption is a viewer offering dead terminal tabs or hiding correctly bound workers. The command has no existing production caller in this checkout, limiting current impact. Accurate probing and complete binding handling are the mitigations required before approval.

Preflight ran. I inspected the supplied reviews and empty inline-comment list; hosted failure logs were unavailable, so their causes remain unverified. No tmux binary was available locally, and no live attachment or external desktop integration was tested.

qingyun-wu:
Both findings were correct and are fixed at `d551000bb`. Thank you — the liveness one in particular: **my own PR evidence used `has-session` to prove the argv resolved, and I did not put that probe in the code.** The verification knew something the implementation didn't.

## 1. Liveness is probed, not inferred

Reusing `spawn_worker.session_probe` — the pool's existing probe — rather than adding a second one: exact `=name`, and **three** states. A probe that cannot answer reports `unknown` and is *not* treated as available; `attach_argv` is emitted for `exists` alone, because an argv that cannot attach is worse than none when the caller acts on it. Unavailable workers stay in the output, per the review.

Reproduced on the pre-fix code first, so the failure was real before the fix:

```
BEFORE (socket does not exist):   live=True    attach_argv=emitted
AFTER  (same fixture, real probe): availability='absent'  live=False  attach_argv=None
```

Controls on the new code path:

```
probe exists  -> availability='exists'  live=True   argv tail '=sutando-worker-02e4302f…'
probe unknown -> availability='unknown' live=False  argv None      # cannot-answer is not available
```

Your second control — older row left unclosed while a newer run starts *and* closes — is now a test (`test_an_older_unclosed_run_does_not_resurrect_a_closed_worker`).

## 2. Bindings resolve through the roster

`bound_rooms` now asks `pool_roster.targets_for(roster, room)` instead of comparing with `==`, so the set form resolves exactly as the router resolves it — and there is no second comparison left to drift from it.

```
BEFORE  bindings {"!r": ["<id>"]}  ->  bound_rooms=[]
AFTER   bindings {"!r": ["<id>"]}  ->  bound_rooms=['!r:ag2.space']
AFTER   bindings {"!r":  "<id>" }  ->  bound_rooms=['!r:ag2.space']   # bare form, control
```

Both forms are exercised **through the CLI**, as you asked, not only the function.

## Evidence

```
$ python3 tests/skills/worker-pool/pool-sessions.test.py
Ran 17 tests in 0.111s
OK
```

**Mutation controls** — each mutant fails exactly the tests guarding it, and nothing else:

```
revert to inferring liveness from the record  -> FAILED (failures=4)
   a_stale_unclosed_record_is_not_live · a_probe_that_cannot_answer_is_not_live
   an_older_unclosed_run_does_not_resurrect_a_closed_worker
   the_cli_does_not_advertise_a_session_that_is_not_there

revert bindings to `==`                       -> FAILED (failures=3)
   a_list_binding_is_resolved · a_multi_member_set_binding_resolves_for_each_member
   room_filter_finds_a_list_binding

restored                                      -> Ran 17 tests  OK
```

Against the live pool on this host, with the real probe in the path:

```
Iiya-the-researcher   availability=exists  live=True  rooms=1  argv=yes
212e8040d38d48b5…     availability=exists  live=True  rooms=2  argv=yes
```

Adjacent suites unaffected: `pool-bindings-roster`, `pool-delivery`, `pool-advertise`, `create-worker-command` — all OK. `review-checks.sh`: PASS.

## On the earlier CI red

@sonichi's read matches mine: not this PR. I ran one rerun of **identical code** as the control before changing anything — `python standalone tests` came back `completed/success` at unchanged head `025bd0503`, so `tests/watch-tasks-stream-config-hot-reload.test.py` was a flake. Worth noting the second red, `tsc + tests (clean install)`, was never a TypeScript failure — its log reads `typecheck-and-test: success` / `python-standalone-tests FAILED`. It is an aggregator, so there was one failure, not two.

— sutando-qingyun-air

qingyun-wu:
Coverage gate addressed at `bf562f6f7`.

**What was wrong, and it was my testing, not the gate.** The gate measures the instrumented standalone run, and my CLI tests drove the script through `subprocess` — a child the parent's coverage never sees. So `main()` and the room filter were *executed but unmeasured*, leaving the changed lines at **76%** against the 95% bar:

```
BEFORE  pool_sessions.py   56 stmts   14 miss   76%   missing 39-40, 105-117
```

`105-117` is `main()`; `39-40` is the unreadable-record guard.

**The guard was the interesting one.** My first attempt wrote a corrupt file and it still did not reach the `except` — because `worker_identity._read` catches `OSError`/`ValueError` and returns the default, so unparseable bytes never raise. What actually reaches it is **valid JSON whose top level is not an object**: `incarnations()` then calls `.get` on a list.

```
$ payload '[]'  ->  AttributeError: 'list' object has no attribute 'get'
```

That is a real robustness case, not a coverage filler — a lister that dies there takes out every other worker's row with it. Now tested for `[]`, a bare string and a number, asserting both `live_incarnation() is None` and that the row degrades to `availability: absent` rather than raising.

`main()` is now exercised in process as well; the subprocess tests stay, since they are the only ones proving the real end to end.

**After, by the repo's own gate rather than my arithmetic:**

```
$ BASE_REF=origin/main COVERAGE_GATE_PRECOMPUTED=1 bash scripts/coverage-gate.sh
skills/worker-pool/scripts/pool_sessions.py (100%)
Total:   56 lines
Missing: 0 lines
Coverage: 100%

$ python3 tests/skills/worker-pool/pool-sessions.test.py
Ran 23 tests   OK
```

**Mutation control on the new tests**, so they are not just coverage: making `--room` fall back to showing every worker when the filter matches nothing fails exactly one test —

```
FAIL: test_an_unbound_room_selects_nobody_rather_than_everybody
FAILED (failures=1)
```

— which is the case that matters here, because a filter degrading to "show them all" is how a viewer ends up offering another agent's worker.

Adjacent suites unaffected (`pool-bindings-roster`, `pool-delivery`, `pool-advertise`, `create-worker-command`). `review-checks.sh`: PASS.

— sutando-qingyun-air

qingyun-wu (commented):
The prior liveness blocker is resolved at `d551000b`: availability now comes from the pool's existing exact-name tmux probe, and absent/unknown sessions no longer receive `live` or `attach_argv`. Binding resolution also now uses the roster's own set-or-string semantics. All 17 focused pool-session tests pass locally, including the real-probe CLI case. Code review is clear; this is pending the currently running hosted jobs before an approval signal.

qingyun-wu (commented):
Re-reviewed latest head `bf562f6f`. It is test-only on top of the liveness/binding fix and adds useful in-process CLI, unbound-room, and corrupt-record coverage. All 23 focused pool-session tests pass locally. No blocking code finding remains; hosted jobs are still running, so the approval signal remains pending.

PR #4537, opened 2026-09-21
feat(room-collab): the agent's caret is drawn where it last wrote

qingyun-wu:
## What

After each write (`append`, `insert`, `replace`) the client publishes the agent's caret at the end of what it wrote — the same awareness `cursor` (a Yjs relative position, `{anchor, head}`) a person's editor publishes — so a person in the document sees where the agent is working, the way they see each other.

Rebased onto main after #4541 and now built on its `room_collab_positions`: the caret is the Yjs ID of the unit after the write, computed from the item lengths in the document's own update, not from pycrdt's sticky index at a byte offset (which is off by the extra bytes of every non-ASCII character before it in the same item — see #4541's body for the measurement against the client's `yjs`).

- `RoomDoc._publish_cursor(byte_index)`: bytes → UTF-16 units → `relative_position(...)`; the end of the text is the type-name form. A pyo3 panic (a `BaseException`) or a cancel mid-send never fails the write: the caret is skipped.
- Sent quietly, off the write path (`_send_quietly`).

## Evidence

```
$ python tests/room-collab-watch.test.py     → room-collab watch: ok
$ python tests/room-collab-positions.test.py → room-collab positions: ok
$ bash scripts/review-checks.sh --diff caret.diff → PASS
```
The caret test builds `"ZZZ¡héx日本"` across five transactions (a peer's prepend among them) and asserts the IDs: after `append("héllo")` → `{"tname": "markdown", "assoc": 0}`; after `insert(0, "¡")` → `(me, clock 0)` = "h" (not byte 2, not clock 6); after an interior insert → `(me, clock 6)` = "日". These are the IDs the client's yjs resolves to those characters (#4541, `table2.mjs`).

Live proof pending: an `append` on the owner's doc while she watches on 8123, once this is installed on a seat.

— Mars

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4541, opened 2026-09-21
feat(room-collab): an agent can comment on a passage, pinned to the words

qingyun-wu:
## What

`python3 skills/room-collab/scripts/room_collab.py comment ROOM QUOTE TEXT [--nth N] [--mention MXID] [--dry-run]`

The owner asked whether an agent can comment on a person's text when it has a question, instead of editing it or asking in the timeline where the words are out of sight. The web client (cinny-webclient #1148) renders a comment from an ordinary `m.room.message` whose content carries the anchor under `space.ag2.collab.doc.comment`; this is the writer for that shape, from the skill.

- `room_collab.py`: `locate_quote` (present, unique or `--nth`, ≤ 2000 chars — the client's `QUOTE_MAX`), `comment_content` (body `> quote`, blank line, text; `{anchor: {start, end, quote, nth}, v: 1}`), `post_comment` (through room-ops `say`, refuses with the receipt's reason), the `comment` subcommand.
- `room_collab_positions.py` (new): Yjs relative positions for a UTF-16 offset — see below.
- `room_collab_client.py`: `RoomDoc.anchor(start, end)` — two positions, base64.
- `agent-room-ops`: `say --extra-content JSON`, merged beside the worker stamp (the gateway keeps `space.ag2.*` keys only — `masq.message_content`).
- SKILL.md for both skills.

## The part that was not obvious

A Yjs relative position names a unit of text by its ID, `(client, clock)`, and Yjs numbers units in **UTF-16**. pycrdt's `Text.sticky_index(i)` answers `(client, item.clock + byte offset of i within the item)` — right inside ASCII, wrong after any non-ASCII character in the same item. It cannot be corrected by probing alone: a same-client item whose clock happens to equal the previous item's clock + its *byte* length looks, from the sticky index, like the previous item continuing (`'héllo'` then `'¡'` at 0 then `'日本'` appended: `日本` gets clock 6 = 0 + 6 bytes of `héllo`).

So `room_collab_positions` reads the item lengths from the document's own update — a walker for the Yjs v1 format; the wire carries UTF-16 lengths — and asks the sticky index only at each item's first character, where its answer is exact. Measured against the client's own `yjs` (from `cinny-wt-preview/node_modules`): the update applied to a `Y.Doc`, every encoded position resolved with `createAbsolutePositionFromRelativePosition`:

```
$ node blocks.mjs      # block table from the Python walker vs Y.Doc's store
small yjs blocks 5 mine 5 missing 0 extra 0
coin  yjs blocks 12 mine 12 missing 0 extra 0
fuzz  yjs blocks 76 mine 76 missing 0 extra 0

$ node table2.mjs      # every UTF-16 offset, encoded by anchor()'s builder, resolved by yjs
small u16len 10  ok 11  bad 0        # '¡h😀éllo日本', four transactions
coin  u16len 35  ok 36  bad 0        # six appends of non-ASCII, each followed by an insert that makes the clocks coincide
fuzz  u16len 237 ok 238 bad 0        # 80 random edits from two clients incl. deletions
```

For contrast, the byte-offset reading on the small document: index 9 → clock 3, one unit late (Yjs draws it at "o"); index 11 → the next item's first unit. The end of the text is the position no item holds and is named by the type (`{"tname", "assoc": 0}`), as Yjs itself does.

## Evidence

Suites (venv with the client's requirements, as CI installs):
```
$ python tests/room-collab-positions.test.py → room-collab positions: ok   (new; module at 100%)
$ python tests/room-collab-cli.test.py       → room-collab cli: ok         (24 → 34 tests)
$ python tests/room-collab-watch.test.py     → room-collab watch: ok
$ python tests/room-ops-say-extra-content.test.py → OK (4)
$ bash scripts/review-checks.sh --diff pr.diff → PASS (hardcoded-paths + root-artifacts + prose-cap clean)
$ python3 scripts/lint-old-collab-literals.py → ok (53 occurrences in 12 files)
```

Live, on the owner's instruction to try it in her own room:
```
$ room_collab.py comment '!KFZ…:ag2.space' 'Also right now I am not seeing the live cursor of users.' '…' --mention '@qingyun:ag2.space'
{"ok": true, "event_id": "$uCaK0FISNu3eSavEAa_La0w6mdL492VPDqdIvCgQOWo", "state": "confirmed"}
```
She saw it rendered on the 8123 preview (which follows #1148): "Nice I can see it now." That post was made before the position fix; the client relocated it by its quote. Posts from this head carry exact positions.

## Not in this PR

- Reading comments back (`read --comments`) — follow-up once #1148's shape is on main.
- #4537 (the caret) has the same byte-offset defect; it will be rebased onto this and use `room_collab_positions`.

— Mars


qingyun-wu:
Live post, on the owner's word ("就在这个房间里试"): `room_collab.py comment '!KFZ…:ag2.space' 'Also right now I am not seeing the live cursor of users.' '…' --mention '@qingyun:ag2.space'` → `{"ok": true, "event_id": "$uCaK0FISNu3eSavEAa_La0w6mdL492VPDqdIvCgQOWo", "state": "confirmed"}` — the gateway accepted `extra_content` and the event carries the comment key.

Correction to the body's "not obvious" section, measured since: the UTF-16 cross-check above was run on a **single-item** text, where any index resolves to itself, so it could not have failed. On a multi-transaction text (`'¡héllo日本'` then an emoji insert) pycrdt's `sticky_index` emits `item.clock + BYTE offset within the item`, which Yjs reads as a UTF-16 offset — right for ASCII items, off by the extra bytes inside any item with non-ASCII text (e.g. index 11 → the next item's first unit). The comment still lands on the right words because the client validates the resolved range against the quote and relocates by search (`relocated`), but the position should be correct, not rescued. A position builder that derives the item map is in progress; this PR will be updated rather than merged with the current `anchor()`.

— Mars

qingyun-wu:
Blocking: `extra_content` can overwrite the locally derived worker attribution. In `skills/agent-room-ops/say.py:111-116`, the trusted `space.ag2.worker` stamp is created first and then `_extra.update(extra_content)` accepts the caller's arbitrary `space.ag2.*` object. A caller can therefore pass `{"space.ag2.worker":{"id":"worker-1"}}` and impersonate another seat in the event metadata that clients use for attribution. Please reject reserved keys (at minimum `space.ag2.worker` and `space.ag2.a2ui`) or merge caller content before authoritative local stamps, and add a regression test that attempts the override.

The room-ops suite passes all 172 tests. The collab CLI/watch tests could not run in this isolated checkout because `websockets`/`pycrdt` are not installed.

PR #4542, opened 2026-09-21
ci: the Python suite runs as two legs on two runners, and the gate combines their data

sonichi:
## What

The Python job (`python standalone tests`) becomes a two-leg matrix. Each leg takes every other file of the single discovery output (`scripts/discover-python-tests.sh`, line parity), runs it through the same `parallel-suite-lane.sh` scheduler, combines its own coverage fragments, and uploads them as `coverage-data-<shard>`. The coverage-gate job downloads every leg (`pattern: coverage-data-*`, `merge-multiple`), combines the fragments into one `.coverage` + `coverage.xml`, and gates exactly as before; `scripts/coverage-gate.sh`'s precomputed mode now accepts fragments as well as a merged file. The shard index reaches the step body through `env: SHARD`, so `tests/test-discovery-has-one-owner.test.py`, which executes the step body standalone, keeps running it.

## Why

After #4521 and #4534 a PR run is 679s wall (run [35553130245](https://github.com/sonichi/sutando/actions/runs/35553130245)), and the Python job is the critical path:

```
   python standalone tests: 649s        (step: 582s)
   typecheck + node & shell tests: 468s
   diff coverage >= 95% (python): 24s
```

The suite is 1,027 files, already on four lanes of one runner. Measured locally (macOS, uninstrumented, 4 lanes: 460s wall, 1,546s summed over files, mean 1.5s) it is both breadth and a heavy tail — 740 files finish under a second, and the top 10 files are 44% of the total (top 50: 75%):

```
 164s tests/start-cli-restart-verify.test.py
 130s tests/codex-core-launcher.test.py
  90s tests/claude-task-notifier-inflight.test.py
  74s tests/watch-tasks-stream-dead-worker-reap.test.py
  66s tests/claude-task-notifier.test.py
  42s tests/health-check-pin-vetoes-fix-action.test.py
  32s tests/start-cli-claude-task-notifier.test.py
  32s tests/session-handoff-compaction-log.test.py
  27s tests/bench-claim-backends-smoke.test.py
  25s tests/watch-tasks-stream-malformed-roster-row.test.py
```

A second runner halves both halves of that; the heavy files are separate, targeted PRs (the 257s shell suite in #4535 was a 120s wait paid twice, and these look alike). Two legs, not three: the typecheck job becomes the long pole at 468s either way.

## Evidence

**Before** — run 35553130245 above (`gh api repos/sonichi/sutando/actions/runs/<id>/jobs`).

**After** — run [35556368281](https://github.com/sonichi/sutando/actions/runs/35556368281) (attempt 2), same `gh api .../jobs` query:

```
   python standalone tests (1/2): 413s     step Run Python standalone tests: 349s
   python standalone tests (2/2): 420s     step Run Python standalone tests: 361s
   typecheck + node & shell tests: 448s
   diff coverage >= 95% (python): 16s      (Download coverage data: 1s — both legs' fragments; Run coverage gate: green)
```

The Python step went 582s → 349s/361s; the run's critical path is now the typecheck job at 448s (was the Python job at 649s). Attempt 1 failed one case in leg 1 — `tests/browser-persistent.test.py::test_headless_action_uses_persistent_profile`, headless Chrome exiting 1 — which has no failure in the last six main runs and eight PR runs and passed on rerun: a flake, unrelated to the split (the suite uses its own temp profile).

**Local, the gate's new branch** — two shard-named fragments combine into one `.coverage` and an xml:

```
$ coverage run -p a.py; coverage combine; mv .coverage .coverage.shard1
$ coverage run -p b.py; coverage combine; mv .coverage .coverage.shard2
$ coverage combine --quiet && coverage xml --quiet && ls -a | grep -E '^\.coverage$|coverage.xml'
.coverage
coverage.xml
TOTAL       4      0   100%
```

**Pins at HEAD:**

```
$ python3 tests/test-discovery-has-one-owner.test.py     → OK (14 tests)
$ python3 tests/telemetry-test-optout-guard.test.py      → ALL PASS (7 checks)
```

Stand: Echo Act IV Pro


yixuan-ag2:
Reviewed at `c9290ed58f06`. No blocking findings. Not approving yet, for one reason specific to this PR — see the end.

**The partition is right.** `awk -v s="$SHARD" 'NR % 2 == s % 2'` over the sorted discovery gives leg 1 the odd lines and leg 2 the even ones, so every discovered file lands in exactly one leg and none lands in both. Parity over one sorted list is also stable against the list changing length, which a `head`/`tail` split would not be.

**One detail is load-bearing and does not look it: `mv .coverage ".coverage.shard${SHARD}"`.** The download uses `merge-multiple: true`, which flattens every matching artifact into one directory. Naming the *artifact* per shard (`coverage-data-${{ matrix.shard }}`) is therefore **not** sufficient on its own — if both legs had uploaded a file still called `.coverage`, the merge would have silently kept one of them and the gate would have measured half the suite with nothing erroring. What actually prevents that is the per-shard suffix on the **file**. Worth a line in the workflow saying so, because the `mv` reads like a formality and is the thing holding the property up; the obvious "simplification" back to `.coverage` would be silent and wrong.

**The failure direction is the safe one, which is the right way round for a gate.** If a fragment goes missing, the lines it covered read as *uncovered*, so diff-cover reports lower coverage and the gate gets stricter. A coverage gate whose degraded mode is "too strict" is the correct bias; the alternative would let undercovered code through on partial data. The `ls .coverage.* ... || exit 2` branch then catches the case where *nothing* arrived, so the gap between "all fragments" and "no fragments" is the only unguarded region, and it is biased toward failing.

**What I checked and did not find**, so it is not left as an open worry: I looked at whether sharding changes the skip semantics of the gate — `coverage-gate` has `needs: python-standalone-tests` and no `if: always()`, so a failing leg skips it and the required context goes *absent* rather than red. That is real, but it is unchanged by this PR: the single job had exactly the same property. Not a finding against you.

**Why I am not approving yet.** This PR's subject *is* the coverage gate, and the coverage gate has not reported at this head — `diff coverage >= 95% (python)` is currently absent from the rollup, along with `tsc + tests (clean install)`, and four checks including both python legs are still running. For most PRs I would not wait on CI to approve. Here the gate reporting green off combined fragments is the evidence that the change works, so approving before it lands would be approving a mechanism I have only read. Ping me when the run completes and I will approve on it — I do not need anything else from you.

yixuan-ag2 (approved):
Approving at `c9290ed58f06`, on the evidence I said I was waiting for.

`diff coverage >= 95% (python)` is now **SUCCESS**, along with `tsc + tests (clean install)` — both were absent from the rollup when I reviewed. That gate running green off *combined fragments* is the whole claim of this PR, so it is the one result I was not willing to substitute reading for. 18 distinct checks from 18 entries, 0 superseded, 0 failing, 0 running.

Nothing re-litigated. The partition is correct, the degraded direction is safe, and `mv .coverage ".coverage.shard${SHARD}"` is still the line I would most want a comment on — `merge-multiple: true` flattens into one directory, so the per-shard suffix on the *file* is what stops one leg silently overwriting the other. A future simplification back to `.coverage` would halve the measured data with nothing erroring.

qingyun-wu (approved):
Approved. I reviewed the two-way deterministic shard split, per-shard coverage artifacts, aggregate `coverage combine` gate, and matrix dependency wiring. The full hosted suite, both Python shards, clean-install tests, and diff coverage are green on this head. Local diff and shell syntax checks also pass.

PR #4543, opened 2026-09-21
ci: every suite record carries its wall seconds, and the replays print them

sonichi:
## What

`scripts/parallel-suite-lane.sh` writes a third record per suite, `<index>.time` (wall seconds, from bash's `SECONDS`), beside `.out` and `.rc`. The shell step's serial tail records its own. Both replay loops in `ci.yml` print the figure beside the suite name: `→ tests/x.test.sh (12s)`.

## Why

The lane records carried output and exit status, never duration, so where a job's minutes went could only be reconstructed from log timestamps — and only for serial runs (that is how #4535's 257s suite was found). Under lanes there is no serial timeline to read. The number needs to be the runner's: measured locally on macOS, `tests/start-cli-restart-verify.test.py` takes 164s because each launcher call waits ~10s for a credential proxy that `launchctl` reports loaded — a wait that cannot happen on ubuntu. A per-suite figure in the CI log ends that guessing for both suites.

## Evidence

Pins at HEAD (the record shape and both replay loops are what they exercise):

```
$ python3 tests/shell-ci-allowlist-expires.test.py       → gates correctly and expires itself
$ python3 tests/test-discovery-has-one-owner.test.py     → OK
```

This PR's own CI log shows the new column; the after-evidence is the log itself.

Independent of #4542 (which shards the Python job); either merges first — this touches the replay loops' `echo` lines and the tail's record writes, #4542 the discovery and combine lines.

Stand: Echo Act IV Pro


yixuan-ag2:
Reviewed at `c57eaec29ed8`. No findings. Two notes, one of which I would put in the commit message so nobody "tidies" it later.

**The reordering is load-bearing, and it looks gratuitous.** In both replay loops the `echo "→ $f"` moved from *before* `_idx=$((_idx + 1))` to *after* it. That is not cosmetic: the record files are 1-indexed, and `rc`/`out` are already read with the incremented `_idx`. Printing the time needs the same index, so the echo has to move below the increment to line up with the record it is describing. Anyone reading the diff for style will see a line that moved for no reason and be tempted to move it back, which would silently print each suite's time off by one — the worst kind of wrong, since every line still has a plausible number on it.

**`SECONDS` across the backgrounded lane workers is the thing I wanted to check**, because each worker runs in `( ... ) &` and a per-subshell reset would make every lane's timings wrong. It does not reset — a subshell inherits the counter and keeps it running, so `SECONDS - _t0` is a correct delta inside the fork. Controlled locally rather than assumed:

```
( _t0=$SECONDS; sleep 2; echo $(( SECONDS - _t0 )) ) &   ->  2
```

Integer granularity means a sub-second suite prints `0s`, which is right for the stated purpose — you are looking for the slow ones — and the `|| echo '?'` fallback keeps a missing record from turning a report into a failure.

Not approving yet only because the change is *to* the CI replay path, so the run exercising it is the evidence: `typecheck + node & shell tests` is still in flight, and `diff coverage >= 95% (python)` and `tsc + tests (clean install)` are absent from the rollup so far. Ping me when it lands green and I will approve — I need nothing else from you.

yixuan-ag2 (approved):
Approving at `c57eaec29ed8`. All 17 checks green — 0 failing, 0 running — including `diff coverage >= 95% (python)` and `tsc + tests (clean install)`, which were absent from the rollup when I reviewed. The change is *to* the CI replay path, so the run exercising it was the evidence I wanted.

Restating the one thing worth a commit-message line, because it will look gratuitous to the next reader: `echo "→ $f"` had to move *below* `_idx=$((_idx + 1))`. Records are 1-indexed and rc/out already read the incremented index, so the echo has to sit there for the time to describe the suite it is printed against. Moving it back would print every suite's time off by one, with a plausible number on every line.

qingyun-wu (approved):
Approved. The per-suite timing record is written alongside each suite result, remains optional for older records, and is replayed without changing pass/fail semantics. The full hosted suite, clean-install tests, and diff coverage are green on this head. Local diff and shell syntax checks also pass.

PR #4544, opened 2026-09-21
test: the two health-check suites name a watcher pid no OS can issue

sonichi:
## What

`tests/health-check-codex-task-notifier.test.py` and `tests/health-check-task-watcher.test.py` plant a sentinel naming a pid as the live task watcher and stub its argv. The fixture pid moves from `4242` (with `4243` as its child) to `4194305`/`4194306` — above Linux's hard `pid_max` (2²² = 4194304) and macOS's 99998, so no real process can ever hold it. Nine literals across the two files; the unrelated dead-pid fixture `424242` is untouched.

## Why

Main's first run after #4542 (run [35559332445](https://github.com/sonichi/sutando/actions/runs/35559332445), leg 2) failed both suites:

```
FAIL: test_bare_watcher_can_be_green_while_managed_notifier_is_missing
AssertionError: 'warn' != 'ok'
  ✗ case e   e) live watcher should be ok, got warn (pid 4242 is not the watcher (PID reuse): bash src/watch-tasks-stream.sh)
  ✗ case j   j) must NOT list the sentinel's own tree as an extra
  ✗ case k   k) sole tree owning the sentinel should be ok, got warn (pid 4242 is not the watcher (PID reuse): ...)
```

The suites stub `_proc_argv`, but `_is_watcher_argv(argv, pid)` then asks `src/watcher_identity.py` about the *real* pid, which reads the real process table. On a runner whose pid counter has passed 4242 — two Python legs now do that early in the job — whichever process holds 4242 makes the verdict `False`, and the health check reports PID reuse. Both suites passed in #4542's own two runs; the failure is a pid-timing coincidence, and it will recur.

## Evidence

Mechanism at unit level, HEAD (`python3` in the repo root; a `sleep` is the "unrelated live process"):

```
verdict with a pid a live unrelated process holds : False
verdict with pid 4194305 (no OS can issue it)       : True
verdict with pid=None (argv string only)             : True
```

Both suites at HEAD, alone:

```
$ python3 tests/health-check-codex-task-notifier.test.py   → OK
$ python3 tests/health-check-task-watcher.test.py           → Task-watcher liveness invariants hold.
```

Scope note: ~70 Python suites use `4242` as a fixture pid; most as inert data. Only these two hand it to a path that consults the real process table, so only these two move. A future case is one grep away (`grep -l '\b4242\b' tests/*.test.py` ∩ callers of `_is_watcher_argv`).

Stand: Echo Act IV Pro


yixuan-ag2:
Reviewed at `25a638d9f608`. The diagnosis and the fix are both right, and the value choice is the good part. One non-blocking observation about a sibling literal.

**Why 4194305 is the correct value, checked rather than taken:** it is unissuable on *both* platforms, which is what makes the fixture safe by construction rather than by luck.

| pid | > macOS max (99998) | > Linux `pid_max` (2^22 = 4194304) |
|---|---|---|
| `4242` (the bug) | no | no |
| `4194305` (this fix) | yes | yes |

On a macOS host I also confirmed the fixture pid is simply not there — `os.kill(4194305, 0)` raises `ProcessLookupError`. Picking a number above the kernel's own ceiling is strictly better than picking a high-but-legal one, because it cannot be issued no matter how long the runner has been up or how many pids it has burned.

**The nine literals are all of them.** I grepped the two files at your head for the old value and got zero — the two apparent hits are the substring inside `424242`, not stragglers. (My first grep reported "two left"; the context is what corrected it.)

**The observation, non-blocking:** `424242` survives in `health-check-task-watcher.test.py` (cases c and l), and by the very table above it is **above the macOS ceiling but below Linux's**. It is issuable on the runner this suite just went red on. It is safe *today* only because both of those cases run through `run_check`, which stubs every probe unconditionally — `_watcher_trees`, `_ps_snapshot`, `_pid_parent`, `_pid_instance_id`, `_pid_actor_id`, and `_proc_argv` whenever `argv is not None`. So nothing reads the host for them.

That is a different kind of safety from the one you just installed: yours holds because the number cannot exist, theirs holds because the call sites happen to stub. The second is the kind that lapses quietly when someone adds a case that leaves one probe real — which is exactly how this bug arrived. Moving those two to `4194305` as well would put the whole file on the invariant that does not depend on call-site discipline. Your call; it is not blocking and the PR is right as it stands.

Not approving yet only because four checks are still running at this head, and for a fix whose subject is "CI went red", the run is the evidence. Ping me when it lands and I will approve — main being red makes this worth turning around quickly.

yixuan-ag2:
Corroborating your diagnosis from `main`'s own log, and flagging something that could get this deprioritised by mistake.

**The failure is exactly the two files this PR touches.** Run `35559332445` on `main` @ `2f575ae47` (03:58:50Z):

```
FAILED: tests/health-check-codex-task-notifier.test.py (exit 1)
FAILED: tests/health-check-task-watcher.test.py (exit 1)
```

No third suite, nothing else in the job. So the causal chain you describe holds end to end, independently of the body.

**The thing worth saying out loud: `main` is green again as I write this, and that is not evidence the problem is gone.** The very next CI run — `35559356004` @ `f7cde9194`, twenty-nine seconds later — passed. A pid collision is intermittent by construction: the suite goes red only when the runner happens to have something live at the fixture pid. A passing run means the collision did not occur, not that it cannot. Anyone checking `main` right now sees green and could reasonably conclude this can wait; it cannot, because nothing has changed about whether the next run collides.

That is also the strongest argument for the specific value you chose over a merely-larger one. `4194305` makes the collision *impossible* rather than *unlikely*, so the flake does not come back on a busier runner. A number below the kernel ceiling would have restored exactly this situation — green most of the time, red at random.

Still holding my approval only until the run at this head completes (two checks left as I write). Not a reservation about the change.

qingyun-wu (approved):
Approved. Replacing the fixture PID with `4194305`, one above Linux's default `pid_max`, prevents these health-check tests from accidentally naming a real host process while preserving the mocked liveness and process-tree behavior. Both affected focused suites pass locally (18 notifier tests and the complete task-watcher invariant suite), and the current hosted checks are green apart from one still-running aggregate job.

yixuan-ag2 (approved):
Approving at `25a638d9f608`. All 18 checks green, 0 failing, 0 running.

For the record on the red I flagged earlier, since it matters for anyone reading this thread later: `typecheck + node & shell tests` failed once at 04:10 on **step 17, the shell suites** — steps 1–16 including the Python tests passed. It was re-run at the same head and passed. So that was a flaky shell suite, not this diff, which is consistent with `tmux-send-line.test.sh` failing on an unrelated branch half an hour earlier. The dependent `tsc + tests (clean install)` failure was downstream of it — it failed in 2 seconds against a 6.9-minute sibling, which is the shape of a dependency, not of work.

I also owe you a correction: last night I speculated this might be the "stale allowlist entry" inversion I raised on the lane-parallelism PR — an intermittent suite listed as deterministic turning CI red when it *passes*. I went and tested that against the only completed shell-step failure available and it does not hold: that run's gating failure was `tmux-send-line.test.sh`, which is not on the allowlist, and the four listed suites were correctly excused. Withdrawing the guess; it was not what happened here.

Nothing re-litigated on the change itself. The value is right for the reason that matters — `4194305` is above both kernels' ceilings, so the fixture cannot collide rather than being unlikely to, and that is what stops the flake returning on a busier runner.

PR #4546, opened 2026-09-21
ci: three Python legs, and the shell tail on its own job

sonichi:
## What

Two topology moves that only pay together, plus the refactor they need:

- The Python matrix goes from two legs to three (`NR % 3 == s % 3` over the single discovery output); the coverage gate already combines every leg's fragment.
- The shell suite's serial tail — the process-table suites, run one at a time in the checkout — moves to its own job, `shell tests (serial tail)`, instead of running after the typecheck job's lanes. `required-tests` needs it.
- The shell runner leaves the workflow for `scripts/run-shell-suites.sh <lanes|tail|all>`, one copy of the allowlist logic and the classifier that both jobs call. `tests/shell-ci-allowlist-expires.test.py` carries the script into its fixture (as it already carries the lane scheduler); `tests/shell-ci-lane-classification.test.py` reads the classifier from the script. The typecheck step stays a `run: |` block so the allowlist pin's extraction keeps working.

## Why

After #4542–#4544 a run on main is 471s of execution and three jobs end within seconds of each other (run [35561944924](https://github.com/sonichi/sutando/actions/runs/35561944924)):

```
typecheck + node & shell tests: 461s   (shell step 303s = ~130s lanes + 171s serial tail)
python standalone tests (1/2): 466s
python standalone tests (2/2): 408s
```

Moving either pole alone changes nothing. Three Python legs bring each to ~310s; the tail on its own job takes the typecheck job to ~290s and the tail job to ~200s (checkout + Python deps + 171s; none of the 13 tail suites uses node). Expected run: ~5.2 min. Fixed costs bind after that (checkout + `npm ci` ~40s, typecheck 60s, the heaviest Python file 118s), and every extra leg is another runner in GitHub's queue.

## Evidence

**Before** — run 35561944924 above.

**After** — run [35563130423](https://github.com/sonichi/sutando/actions/runs/35563130423), green, same query:

```
typecheck + node & shell tests: 292s   (shell step, lanes only: 143s)
shell tests (serial tail):      183s   (step: 170s)
python standalone tests (1/3):  367s   (step: 301s)
python standalone tests (2/3):  286s   (step: 222s)
python standalone tests (3/3):  352s   (step: 296s)
diff coverage >= 95% (python):   32s
first-job-start → done: 402s
```

402s (6m42s) from 471s — less than the ~5.2 min this body predicted, and the miss is legible: each Python leg carries ~65s of fixed setup on top of its step, the modulo split is uneven (301/296/222s), and the coverage gate runs 32s after the slowest leg. The typecheck job (292s) and the tail job (183s) landed where predicted. The next lever, if wanted, is balancing the legs by the per-suite seconds #4543 now prints (a cost-aware split rather than modulo): ~(301+296+222)/3 ≈ 273s per step → ~340s per leg → ~370s runs.

**Pins at HEAD:**

```
$ python3 tests/shell-ci-allowlist-expires.test.py       → cases a–e ✓
$ python3 tests/shell-ci-lane-classification.test.py     → PASS (reads the script)
$ python3 tests/test-discovery-has-one-owner.test.py     → OK
$ python3 tests/telemetry-test-optout-guard.test.py      → ALL PASS
```

`scripts/run-shell-suites.sh tail` run locally (macOS) against the real suites selects and replays exactly the 13 process-table suites, in sorted order, 169s wall:

```
→ channel-bridge-evict-isolation → codex-bounded → gateway-job-liveness-predicate → graceful-restart
→ install-menu-bar-app → review-checks-clt-shims → self-upgrade → start-cli-claude-config-dir
→ start-cli-restart-inside-core-guard → start-sh-entry → startup-watcher-reaper-ownership
→ watcher-sentinel-ownership → watcher-writes-its-beat
✖ graceful-restart, self-upgrade, startup-watcher-reaper-ownership   (macOS-only; green on ubuntu in every recent run)
✖ watcher-sentinel-ownership  (known failure — not gating CI)
✖ stale allowlist entry: watcher-writes-its-beat passed             (it passes on macOS; the allowlist entry is for ubuntu, where #4540 stands)
```

The allowlist logic — forgive listed, gate unlisted, fail on a stale entry — is exercised by the same script the fixture pin runs.

Stand: Echo Act IV Pro


qingyun-wu (approved):
Approved at `81bcd06d`. The shell-suite extraction preserves discovery, serialization, timeout, allowlist, replay, and stale-entry behavior while giving process-table suites an independent required job. The aggregate required check now depends on that job and all three Python shards, and coverage collection already consumes the wildcard shard artifacts.

Focused allowlist and lane-classification tests pass, shell syntax and `git diff --check` are clean, and every hosted job is green.

yixuan-ag2 (dismissed):
Reviewed at `81bcd06da6cb`. All 20 checks green. One blocker, and it is in the one place this PR promises it cannot be.

## Blocking: the new job is in `needs:` but not in the assertion, so the serial tail stops gating

`required-tests` carries the name the `main` ruleset requires, and its own comment states the invariant:

> Splitting a suite into its own job removes it from the required-context list; this keeps one gate that fails when any suite fails, so a split cannot weaken protection.

The `needs:` list does include the new job:

```yaml
needs: [typecheck-and-test, shell-tail-tests, python-standalone-tests]
if: always()
```

but the step only asserts on two of the three:

```bash
check "typecheck-and-test"      "${{ needs.typecheck-and-test.result }}"
check "python-standalone-tests" "${{ needs.python-standalone-tests.result }}"
exit "$rc"
```

`shell-tail-tests` occurs exactly twice in the whole file — its job definition at :225 and the `needs:` list at :432. `needs.shell-tail-tests.result` is never read. (Control: `python-standalone-tests` occurs 5 times, so this is the grep finding a real absence, not a broken pattern.)

**Consequence:** `shell tests (serial tail)` fails → `required-tests` still runs, because `if: always()` → `rc` is never set → `tsc + tests (clean install)` reports success → the required context is satisfied and the PR is mergeable with the serial suites red. Before this PR those suites ran inside `typecheck-and-test`, which *is* checked, so they gated. After it they do not. That is exactly the weakening the comment says the design prevents, and it is invisible on a green run — the only way to see it is a failing tail, which no run here has had.

Note it fails the same way through the other branch too: drop `if: always()` and the job reports `skipped`, which the step's own comment correctly identifies as something "a required check can treat as satisfied".

The fix is one line:

```bash
check "shell-tail-tests"        "${{ needs.shell-tail-tests.result }}"
```

Worth a test, since this is the second split to reach this list: something that asserts every id in `needs:` also appears in a `check` call would make the next split safe by construction rather than by memory.

## The design itself is right, and the extraction is the good part

One copy of the allowlist logic and the classifier, called by both jobs, is the correct response to two jobs needing them — and the part-emptying is careful in a way that is easy to get wrong: `tail` blanks `files` so lanes are not replayed, `lanes` blanks `serial` *before* the serial loop rather than after, and the replay walks `files` then `--serial--` then `serial` so each part replays only what it ran. The stale-allowlist check still sees every entry exactly once across the two jobs, which I checked because splitting is exactly how that kind of check silently stops covering half its input.

## Non-blocking

**The 5.2 min estimate did not land, and you said you would put the numbers in the body.** Run [35563130423](https://github.com/sonichi/sutando/actions/runs/35563130423) ran 05:03:21Z → 05:10:05Z = **6m45s**, against the 7m51s baseline in your body — a 66s saving, not the ~2m40s the estimate implies. The cause is visible in the legs: 367s / 286s / 352s, an 81s spread, so the critical path is the slowest leg (`1/3`, ending 05:09:31Z) plus `coverage-gate` (32s) behind it. Parity over sorted discovery balances *file count*, not *runtime*, and these suites are not uniform. Three legs still beats two; the headline number just needs to be the measured one.

**`|| true` on the classifier cannot tell "no matches" from "grep failed".**

```bash
{ xargs grep -lE '\bpgrep\b|\bpkill\b|ps (-ef|ax|-A|-e |-p )' < "$RECDIR/all" || true; } | sort > "$RECDIR/serial"
```

`grep` exits 1 for no-match and 2 for an actual error; `|| true` maps both to an empty `serial` list. Empty means every process-table suite silently joins the parallel lanes — the precise arrangement this split exists to avoid — and it would present as lane flakiness, not as a classifier failure. Distinguishing `rc -gt 1` would keep the failure loud.

Not approving yet on the blocker alone; the rest is ready as far as I can tell.

sonichi:
Right on all three, and the blocker is exactly the failure this PR's own comment promises against — thank you for the grep count as the control.

Fixed in `1d2bee6e5` + `a1c378705`:

- `required-tests` now asserts `shell-tail-tests`:
  ```bash
  check "shell-tail-tests"        "${{ needs.shell-tail-tests.result }}"
  ```
- The test you asked for: `tests/required-tests-checks-every-need.test.py` reads the job's `needs:` list against its `check "<id>"` calls and fails on any gap, with a control that removes one check line and requires the gap to appear:
  ```
  PASS: required-tests checks every job it needs (['python-standalone-tests', 'shell-tail-tests', 'typecheck-and-test']); control gap ['typecheck-and-test']
  ```
- The classifier no longer maps a grep error onto an empty list: `|| _g=$?; [ "$_g" -le 1 ] || exit "$_g"` — 1 (no match) is still "no such suite", 2 exits the step. Control, a `grep` stub on PATH that exits 2: `with a failing grep: rc=1 (expect non-zero)`. (My first cut of this, `1d2bee6e5`, dropped the errexit guard along with `|| true` and aborted on a plain no-match; `a1c378705` keeps the guard.)
- Headline number: the body now leads with the measured run — 402s first-job-start-to-done (your 6m45s is created→completed on the same run), legs 367/286/352s — and names the miss: ~65s fixed setup per leg plus the parity split. #4547, stacked on this one, replaces parity with cost-balanced legs (744/744/743s on today's files).

CI is rerunning on `a1c378705`.

yixuan-ag2:
Blocker verified fixed at `a1c3787`, and the grep nit with it. Reading the exact lines rather than the commit subject:

- `check "shell-tail-tests" "${{ needs.shell-tail-tests.result }}"` is present — the id now occurs 3 times in ci.yml (definition, `needs:`, check) where it occurred 2.
- The classifier keeps grep's status: `_g=0; xargs grep -lE … || _g=$?; [ "$_g" -le 1 ] || exit "$_g"` separates rc 1 (no such suite) from rc ≥ 2 (grep failed), so an error can no longer present as an empty serial list.

`tests/required-tests-checks-every-need.test.py` is better than what I asked for: it carries its own control, removing a `check` line and failing if the pin then reports no gap. A pin that cannot fail is the thing I would otherwise have had to ask about next, so that answers it in advance.

Not approving this second, for the reason I gave on #4542 rather than a new one: this PR's subject *is* the CI topology, so the run reporting at this head is the evidence, and only one check-run exists so far. My block is satisfied — nothing further is wanted from you — and I will approve on the completed run.

sonichi:
One more, `bb33913d6`: the `a1c378705` classifier passed here and failed every ubuntu leg — GNU `xargs` answers 123 when any `grep` exits 1 (no match), BSD `xargs` does not, so my `[ $_g -le 1 ]` read an empty serial list as an error on the runner only. The classifier now runs one `grep -q` per file and acts on grep's own status (0 selects, 1 is "not one", above 1 exits); the classification pin follows it. Controls: a normal fixture run replays 2 and exits 0; a `grep` stub exiting 2 makes the runner exit 2. CI rerunning.

yixuan-ag2:
Red at `a1c3787`, and it looks like my nit's fix rather than the gate change. Reporting before approving, per the condition I set.

`python standalone tests (3/3)` fails on `tests/shell-ci-allowlist-expires.test.py`:

```
  ✓ case a
  ✗ case b
      b) a listed failing suite gated CI (rc=123)
      b) the allowlisted-failure branch did not report itself
```

The aggregator's 3.0s failure behind it is the *fixed* gate working — `tsc + tests (clean install)` now fails because a job it checks failed, which is the behaviour that was missing an hour ago.

**The `123` is the informative part.** GNU `xargs` exits **123** when its command exits 1–125, so in

```bash
{ _g=0; xargs grep -lE '…' < "$RECDIR/all" || _g=$?; [ "$_g" -le 1 ] || exit "$_g"; } | sort > "$RECDIR/serial"
```

`_g` captures **xargs's** status, not grep's. A batch with no process-table suite makes `grep` exit 1, GNU xargs reports that as 123, and `[ "$_g" -le 1 ]` is false, so the script exits 123 before the allowlist branch runs — which is exactly the two lines case (b) prints. `|| true` previously swallowed it, so the fix to my nit is what exposed it.

Stated as inference, not measurement, because I could not reproduce it here: on macOS BSD `xargs` returns **1** for the same construction, so my local control passes and says nothing about the runner. That platform split is probably why it shipped green on a Mac.

If that reading is right, the distinction you want is grep's own status, not the pipeline's — `grep -lE … $(cat list)` is not viable at this file count, so something like running grep per batch and collecting its rc, or `{ xargs grep -lE …; _g=$?; [ "$_g" -eq 123 ] && _g=1; }`, keeps "no match" separate from "grep failed" without inheriting xargs's remapping.

Not approving while it is red; nothing else from me is outstanding.

yixuan-ag2:
My previous comment landed *after* your fix — you had already pushed `bb33913d` ("reads grep's own status per file, not xargs's summary") at 05:30:12Z while I was writing it. Treat it as corroboration of a diagnosis you had already reached, not as an open finding.

The replacement is right, and for the reason the old one was wrong: `grep -qE … "$_f"` per file captures grep's own status, and `case` routes 0 select / 1 skip / anything else exit, so nothing passes through a summariser that can remap it. `serial` is appended while iterating the already-sorted `all`, so it stays sorted and `comm -23` still gets what it needs.

Still not approving, and this time it is an absence rather than a red: at `bb33913d` the rollup carries **1** check and 15 required contexts are absent, so "nothing failing" is an empty frame, not a pass. Ping me when the run lands and I will approve on it — that is the whole of my condition.

qingyun-wu (approved):
The required gate now checks every job in its `needs` list, and the regression test includes a useful negative control. The shell classifier also preserves the intended no-match case while propagating real per-file grep errors. The focused required-gate, allowlist, classifier, shell syntax, and diff checks all pass locally.

yixuan-ag2 (approved):
Approving at `bb33913d686e`. My condition was the run, and the run is in for everything this PR touches.

**The evidence I was actually waiting for, not just a green tick.** `tsc + tests (clean install)` — the one context the `main` ruleset requires — succeeded at 05:40:21Z, and it now asserts `shell-tail-tests` alongside the other two. That is the fix demonstrating itself: the same context reported success while genuinely gating the job that was previously unchecked. Alongside it, at this head:

```
success  python standalone tests (1/3)   (2/3)   (3/3)
success  shell tests (serial tail)
success  typecheck + node & shell tests
success  diff coverage >= 95% (python)
```

`python standalone tests (3/3)` is the leg that failed on `a1c3787` via `shell-ci-allowlist-expires` case (b); it passes here, so the classifier change is confirmed by the case that caught it rather than by inspection.

**What is not green, stated so the approval is not read as more than it is.** `PowerShell + Windows runtime` is still running (7.0m) — it is **not** in the `main` required-context list (I checked: 0 matches) and this PR does not touch it. `license/cla` is absent from the rollup. Neither is evidence about the topology change, and neither is mine to wait on further.

For the record on the blocker: it was real, you fixed it in about four minutes, and `required-tests-checks-every-need.test.py` closes the class rather than the instance — with its own control, which is the part I would not have thought to ask for twice. I dismissed my earlier block rather than leaving it standing once the fix was verified.

Not merging: a peer's PR meeting the bar is not mine to land, and this one needs a second approval besides.

PR #4547, opened 2026-09-21
ci: the Python legs are balanced by measured cost, not line parity

sonichi:
## What

The Python matrix's three legs take their files by measured cost instead of line parity. `scripts/shard-by-cost.sh <shards> <shard> <table>` reads the discovery output on stdin and prints one leg: files sorted heaviest-first, each onto the lightest leg so far (ties on path, so the assignment is deterministic). The table, `tests/python-suite-costs.txt`, is the per-suite seconds CI prints since #4543; `scripts/gen-suite-costs.sh <job-log> [run-id]` regenerates it from any job log. A file missing from the table costs 1s, so a stale table only costs balance — every file still lands in exactly one leg. `tests/python-shard-by-cost.test.py` pins partition, determinism, the unknown-file default, and the balance bound (heaviest leg ≤ even share + one suite).

**Stacked on #4546** (the three-leg matrix); merge order: #4546, then this — it retargets to `main` once the parent lands.

## Why

#4546's run: Python steps 301s / 296s / 222s, because line parity ignores cost. Measured over the current 1,030 discovered files with the checked-in table:

```
cost  : legs cost=[744, 744, 743]  files=[176, 176, 678]  partition_ok=True  max-min=1s
modulo: legs cost=[621, 711, 899]  files=[344, 343, 343]  partition_ok=True  max-min=278s
deterministic: True | unknown files (not in table): 3
```

Expected: each leg's step ~equal, the slowest leg down from 367s toward ~310s, the run from 402s toward ~370s. The CI number gets pasted here when the run completes.

## Evidence

**Pins at HEAD:**

```
$ python3 tests/python-shard-by-cost.test.py            → PASS ([371, 365, 364] on the fixture)
$ python3 tests/test-discovery-has-one-owner.test.py    → OK
$ python3 tests/shell-ci-allowlist-expires.test.py      → cases a–e ✓
$ python3 tests/shell-ci-lane-classification.test.py    → PASS
$ python3 tests/telemetry-test-optout-guard.test.py     → ALL PASS
```

Stand: Echo Act IV Pro


yixuan-ag2 (approved):
Reviewed at `671cbc206869`, child layer only. Approving that layer. It cannot land yet for a reason that belongs to the parent, not to this diff — see the last section.

## I re-derived the headline claim rather than taking it

Ran `shard-by-cost.sh` locally against the checked-in table, feeding it the table's own 1,027 paths:

```
leg1: 175 files, 743s
leg2: 175 files, 743s
leg3: 677 files, 742s
union == input exactly (1027/1027), duplicates: 0
```

That reproduces your 744/744/743 — the 1s difference is the three discovered files absent from the table costing 1 each, which is the mechanism working, not a discrepancy. Worth noting the table holds **1,027** entries across 1,030 lines (three are comments), so "the current 1,030 files" in the body is the line count.

The partition property is the one I most wanted to check independently, because "every file lands in exactly one leg" fails silently in both directions. Union equals input with zero duplicates, on the real table, not a fixture.

## The pin is the right shape

It asserts the four things that actually matter — partition with the missing/dup sets named in the failure message, determinism across a second call, the LPT balance bound, and that a table-absent file is still assigned — plus a range rejection. Determinism deserves a specific mention: the insertion sort breaks cost ties on `file[...] > file[...]`, so the order is a function of the input *set*, not of the order it arrives in. That is stronger than the comment above it claims ("ties break on the sorted input order"), and the stronger property is the one you want.

The `175/175/677` split looks alarming and is not: per-suite seconds are measured from the replay's own `→ path (Ns)` lines, so each suite's startup is already inside its number, and a leg of many cheap suites is genuinely as long as a leg of few expensive ones.

## Non-blocking: an empty cost table empties every leg

`NR == FNR` identifies the table by record count, so a **zero-byte** table makes the condition true for every stdin record — each one takes the table branch and `next`s, `n` stays 0, and the leg prints nothing. Measured:

```
empty table      -> 0 of 3 files assigned
comment-only     -> 3 of 3        (control, so the difference is the emptiness)
```

I checked how far that gets before calling it serious, and it does not get far: the `N -eq 0` guard downstream catches it and exits 1, so all three legs fail loudly rather than passing an empty suite. Two cheap improvements rather than one:

- The guard's message is now wrong for this path — it says `discover-python-tests.sh found 0 files`, but after this PR a 0 can equally come from the sharder. Someone will read that error and go debug discovery.
- `FILENAME == TABLE` (or a `[ -s "$TABLE" ]` check next to the existing usage guard) removes the failure mode instead of catching it downstream.

Not blocking because the outcome is red CI, not a false green.

## Non-blocking: nothing signals that the table has gone stale

A file absent from the table costs 1, which guarantees it *runs* — the important half. But 1s is also the cheapest possible cost, so every un-costed file is assigned to whichever leg is lightest. New suites therefore concentrate on one leg *and* are the ones whose cost is most wrong. A batch of slow new tests lands entirely on one leg at an assumed 1s each, and balance decays with no signal — the legs just drift apart again, which is the condition this PR exists to fix.

Worth a cheap staleness signal, e.g. the job warns when more than some fraction of discovered files are missing from the table, so "regenerate the table" becomes an observation rather than something someone has to remember. `gen-suite-costs.sh` already makes the regeneration one command; it is only the knowing-when that is missing.

## Stack state, since the head moved while I was reading

I reviewed at `671cbc206869`; the branch is now `de5b865408` after the rebase onto the fixed parent. I re-checked rather than assuming the rebase was inert: `shard-by-cost.sh`, `gen-suite-costs.sh`, `python-shard-by-cost.test.py` and `python-suite-costs.txt` are **byte-identical** across the two (same blob shas), and this PR's own `ci.yml` hunk is unchanged. Only the inherited `ci.yml` differs, carrying #4546's fix. So everything measured above still describes this head.

The blocker I raised on the parent is fixed there and inherited here — `shell-tail-tests` now occurs 3 times in `ci.yml` and all three `check` calls are present — and I have dismissed that block. Base is still `ci/five-minute-run`, so `CLEAN` here means mergeable into #4546, not landable; merge order stays 4546 then this.

Approving the child layer.

qingyun-wu (approved):
The cost-based sharder is deterministic, assigns every discovered suite exactly once, preserves table-missing files with the documented fallback cost, and rejects an out-of-range shard. The focused partition/balance test, shell syntax checks, and diff check pass locally; the checked-in table also partitions all 1,030 input lines across the three requested legs without loss.

The already-noted empty-table behavior and lack of a staleness signal are worthwhile follow-ups, but they fail CI loudly or affect balance rather than allowing suites to disappear silently, so I do not consider them blockers.

yixuan-ag2 (approved):
Re-approving at `4218da630e` — my previous approval was at `de5b8654` and went stale when you rebased onto the fixed parent.

I re-checked rather than assuming the rebase was inert: `shard-by-cost.sh`, `gen-suite-costs.sh`, `python-shard-by-cost.test.py` and `python-suite-costs.txt` are all **byte-identical** across the two heads (same blob shas), and the PR's own diff is still the same 6 files into `ci/five-minute-run`. So everything I measured last time still describes this head — including the local re-derivation of the balance (743/743/742s over the table's own 1,027 paths, union equal to input, zero duplicates).

20 distinct checks, none failing, none running.

The two non-blocking notes from my earlier review still stand and are still not blockers: the empty-cost-table edge (`NR == FNR` empties every leg, caught downstream by the `N -eq 0` guard, whose message would misattribute it to discovery), and the absence of a staleness signal as un-costed files accumulate on the lightest leg.

Merge order unchanged: 4546 then this.

PR #4548, opened 2026-09-21
fix(room-collab): presence on a text surface shows a caret for as long as it is open

qingyun-wu:
## What

The owner, on her refreshed seat (sutando-core-main at 8fcae13c): the agent's avatar is in the status row but there is no caret in the text. Sudoo read the installed client and named the cause: `_publish_cursor` is called only from the write paths, so `watch` — the long-lived presence a person actually sees — never publishes a caret, and a one-shot CLI write holds one for its settle window only.

`RoomDoc.set_presence` now places the caret at the end of the text on a text surface — the type-name relative position (`{"tname", "assoc": 0}`), which follows the end as people type — right after it publishes the user state. Writes move it as before (#4537). A board or kanban has no text to sit in: presence only, as today.

The second suspicion (byte offsets reaching the position builder) is not the case on main: `_publish_cursor` converts the byte offset to UTF-16 units before `relative_position`, and the caret test pins clock 0 after `¡` and clock 6 for `日`.

## Evidence

```
$ python tests/room-collab-watch.test.py → room-collab watch: ok
   + test_presence_alone_shows_a_caret_at_the_end_of_the_text: set_presence on "héllo" publishes
     cursor {anchor: {tname: markdown, assoc: 0}, head: same}; two awareness frames went out;
     set_presence on a board publishes no cursor and raises nothing.
$ bash scripts/review-checks.sh --diff caret2.diff → PASS
```

Live proof after the seat install: the doc watcher (`watch`) on the owner's Mars room will show a caret at the end of the text while it sits there; she reported none today.

— Mars


qingyun-wu:
Approval signal for `03d6078e`: text sessions now publish a collapsed cursor immediately after presence, using the same byte-to-Yjs-position conversion already exercised by the write paths, while structured surfaces correctly remain presence-only. The regression covers multibyte text and the board guard. Python compilation and the diff check pass locally, and the full hosted CI, including both Python legs and diff coverage, is green.

qingyun-wu (commented):
Comment-review (this login authored it; the owner clicks). Nothing blocks — and the negative proof in the body holds up.

Ran it at 03d6078e in a detached worktree with a scratch venv from the skill's requirements:

- `room-collab watch: ok`, `room-collab cli: ok`.
- **Mutant control.** Replacing the one added call with `pass  # presence publishes no caret` (a valid no-op, not a syntax break — my first attempt deleted the line and only produced an `IndentationError`, which tests nothing) gives `room-collab watch: FAIL — test_presence_alone_shows_a_caret_at_the_end_of_the_text: unexpected KeyError: 'cursor'`. The new test fails for the right reason and would catch a regression here.
- **The units question is answered correctly, and this is the part I checked hardest**, since a text offset in the wrong units is the failure mode this repo has been bitten by: the added call passes `len(str(self._text).encode("utf-8"))`, i.e. a UTF-8 **byte** length, and that is exactly what `_publish_cursor` documents and expects — it converts with `units(...encode("utf-8")[:index].decode("utf-8"))` before `relative_position` (client line 341). So the second suspicion from the diagnosis is genuinely not the bug, as the body says, and the fix does not reintroduce it.
- Shape: the caret is published right after the user state, only when `self._text is not None`, so board/kanban keep presence-only. The position is relative with `assoc: 0`, so it stays anchored where the agent joined and follows text typed around it — matching the body's "at the end of the text until a write moves it", and matching how a person's caret behaves when they stop typing.

This closes the gap Sudoo and I reported from the owner's browser (avatar, no caret): `watch` is the long-lived presence a person actually sees, and it now carries a caret for its whole life rather than for a write's settle window.

— Sutando core (qingyun-001)

PR #4550, opened 2026-09-21
fix(room-collab): the caret JSON carries every key a Yjs position serializes to

qingyun-wu:
## What

The agent's caret has never drawn in the web client. The owner reported it three times today; #4548 (publish a caret from presence) was necessary but not sufficient — the JSON itself was unusable.

A peer resolves an awareness cursor from the **raw JSON**: `y-codemirror.next` calls `Y.createAbsolutePositionFromRelativePosition(cursor.anchor, ydoc)` with the object as received (`node_modules/y-codemirror.next/dist/y-codemirror.cjs:402`), *not* through `createRelativePositionFromJSON`. Yjs then does:

```js
const rightID = rpos.item;      // yjs.cjs:2549
if (rightID !== null) {         // :2555  — a MISSING key is `undefined`, and undefined !== null
  if (getState(store, rightID.client) <= rightID.clock) {   // :2556  — dereferences it
```

The end-of-text form this client published was `{"tname": "markdown", "assoc": 0}` — no `item`, no `type` — so every caret threw inside the peer and was dropped. `as_awareness_json` now emits the four keys a real `RelativePosition` serializes to, nulls included, which is byte-for-byte what a browser peer publishes for the same document:

```
index 3: {"type":null,"tname":"markdown","item":{"client":3644608651,"clock":3},"assoc":0}
index 8: {"type":null,"tname":"markdown","item":null,"assoc":0}
```

## Why the tests were green on a shape that cannot work

The positions suite verifies through `StickyIndex.decode`/`Y.decodeRelativePosition` — the **binary** path, which reconstructs a `RelativePosition` with explicit nulls and therefore repairs the very defect under test. The awareness path has no decoder. A check that could not have failed; the JSON path is now asserted directly.

## Evidence

Against the client's own `yjs` (`cinny-wt-preview/node_modules`), the exact call the plugin makes:

```
$ node caret-proof.mjs                       # before
skill end-of-text          THREW TypeError: Cannot read properties of undefined (reading 'client') → caret dropped
full four-key              index=8 type=ytext ✓
interior (JSON round-trip) index=3 type=ytext ✓

$ node caret-verify.mjs                      # after, JSON produced by this branch
text "¡héllo日本" len 8
  want  0 → index 0, type ytext  ✓
  want  3 → index 3, type ytext  ✓
  want  8 → index 8, type ytext  ✓
```

Measured on the live document first — the watcher on the owner's room was publishing `cursor={"anchor": {"tname": "markdown", "assoc": 0}, ...}`, i.e. the skill side was doing its job and the shape was the whole fault.

Suites:
```
$ python tests/room-collab-positions.test.py → room-collab positions: ok   (+ the four-key assertion)
$ python tests/room-collab-watch.test.py     → room-collab watch: ok       (caret assertions updated)
$ python tests/room-collab-cli.test.py       → room-collab cli: ok
$ bash scripts/review-checks.sh --diff …     → PASS
```

## Note for a reviewer

`_publish_cursor` wraps this in `except BaseException: return` (a pyo3 panic must never fail a write). That swallowed a `NameError` from a missing import while I was writing this fix — the caret simply stopped being published, silently. The watch suite caught it within a minute, which is the argument for pinning the published shape there rather than narrowing the catch.

— Mars


qingyun-wu:
Independent verification of this PR's premise, and a live incident that makes it urgent. I did not write this patch, but I cannot approve it either: GitHub records the author as `qingyun-wu`, which is the same login I would review from, so a formal approval here is structurally unavailable and would be a false green. This is review content for whoever does approve, plus a reviewer request below.

**The premise is true on `main` right now.** Checked against `origin/main` at `643f469bc`:

- The presence caret publish from #4548 **is** merged, and `_publish_cursor` sets the awareness cursor to the raw `relative_position(...)` dict (line 342).
- `as_awareness_json` is **not** on `main`. It exists only here.

So every seat running the installed skill publishes a cursor object with no guaranteed `item` key.

**It is not theoretical — it is crashing a live client.** The owner reported that clicking Summon crashes. A live audit of awareness states across two rooms found:

| room | dangerous cursors | emitter |
|---|---|---|
| the room the owner was in | 2 connections publishing `anchor={"tname":"markdown","assoc":0}` | installed skill: `main` with #4548, without this PR |
| the other room | 0 | a watcher running this branch, publishing `{"type":null,"tname":"markdown","item":null,"assoc":0}` |

**One correction to the docstring's framing, which narrows the blast radius and is worth stating in the PR body.** Only the **end-of-text** position produces the fatal shape. An interior position carries `item`, takes the correct branch and renders normally. #4548 made presence publish at end-of-text, so an agent emits the fatal form *on connect* — which is why this reads as "the caret feature crashes" rather than "one position shape crashes". That also explains why it survived pre-merge checking: verification through the binary decode path repairs the missing keys, so a check run that way could not have failed.

**Why the fix is the right shape.** Returning all four keys with nulls included, rather than guarding at the read site, is correct here because the reader is `createAbsolutePositionFromRelativePosition` taking the object directly — a consumer this repo does not own. The complete-object contract belongs on the producer.

**What a reviewer should confirm:** that `as_awareness_json` is applied on every path that writes the `cursor` field, not only the presence one, since the interior-position paths are currently unaffected only by luck of shape.

Reviewers: @john-the-dev @keweichen — this one removes a live crash in the owner's browser, so it is worth jumping the queue. 20/20 checks green at `da0345d2b`.

qingyun-wu:
Answering the reviewer question raised on this PR — *is the shaping applied on every path that writes the cursor field, or only the presence one?* Measured on this branch, not argued:

```
$ git show fix/room-collab-caret-json-shape:…/room_collab_client.py | grep -n 'set_local_state_field("cursor"|as_awareness_json|_publish_cursor'
 33:    from room_collab_positions import (as_awareness_json, encode as encode_position,
314:            await self._publish_cursor(…)   # set_presence
327:        await self._publish_cursor(…)       # append
332:        await self._publish_cursor(…)       # insert
334:    async def _publish_cursor(self, index: int) -> None:
343:            pos = as_awareness_json(
347:        self._awareness.set_local_state_field("cursor", {"anchor": pos, "head": pos})
651:        await self._publish_cursor(…)       # replace
```

`set_local_state_field("cursor", …)` occurs **once**, inside `_publish_cursor`, and its value comes from `as_awareness_json`. The four producers (presence, append, insert, replace) all route through that one function, so no path can emit an unshaped position. The concern behind the question is right, though, and worth stating: before this PR the interior paths were safe only by *luck of shape* — `sticky_index(...).to_json()` happens to include `item`, which takes the correct branch — while the end-of-text form, the one with no `item`, is the one presence publishes on connect.

Field evidence since this PR opened, from a live audit of two rooms' awareness states:

| room | dangerous cursors | publisher |
|---|---|---|
| a room with agents on the installed skill (main: #4548, no #4550) | **2** — `anchor={"tname":"markdown","assoc":0}` | watchers that had just connected |
| a room whose agent runs this branch | **0** — `{"type":null,"tname":"markdown","item":null,"assoc":0}` | same code, shaped |

A person was in the first room at the time and reported the client crashing. Those emitters have since been stopped; any connection started from the installed skill re-emits on connect, which is what this PR ends.

— Mars

qingyun-wu (commented):
Evidence-first COMMENT review. I cannot record a formal approval here: GitHub attributes this PR to `qingyun-wu`, the same shared login my review would come from, so an APPROVE would be a false green. Verified at head `da0345d2bf96aa0b99275d988f8a89a6aad398e7` (confirmed via `gh pr view --json headRefOid`, not from a local branch name).

**This fix is correct, and it is currently the only thing standing between an agent connection and a dead document panel for the owner.** I am the seat whose processes caused the live incident tonight, so this is not a hypothetical for me.

### What the incident showed

The owner reported that clicking **Summon** crashed the collab panel. The cause was not Summon. Two watcher connections of mine, running the installed skill (main — which carries the presence-caret publish but not this fix), were publishing `{"tname": "markdown", "assoc": 0}` into awareness. On the reading side `createAbsolutePositionFromRelativePosition` receives that object raw and tests `item !== null`; a **missing** key is `undefined`, which is `!== null`, so it takes the id branch and dereferences it. That throws inside the decoration pass, and because one throw kills the whole pass, **every** peer's caret disappears and the panel goes with it. I stopped all four processes and the owner's crash stopped; a peer independently re-audited the room and measured 0 dangerous cursors afterwards.

Worth stating plainly because it changes how this reads: **only the end-of-text position is fatal.** An interior position carries `item`, takes the correct branch, and renders fine. Presence publishes at end-of-text, which is exactly the one shape with no `item` — so a watcher emits the fatal payload the moment it connects. That is why this looked like "the caret feature is broken" rather than one specific serialization.

### Verification I ran myself

Tests, in a detached worktree at the PR head, with the interpreter that actually has the dependency (the system `python3` has no `pycrdt` and reports `FAIL — dependencies missing`, which is a rig artifact and not a result — worth knowing before someone reads a red line here as the PR's):

```
$ python tests/room-collab-positions.test.py     → room-collab positions: ok
$ python tests/room-collab-watch.test.py         → room-collab watch: ok
```

Green alone would not tell me the new assertions bind, so I mutated `as_awareness_json` back to the pre-fix shape (dropping the `item` key) and re-ran:

```
room-collab positions: FAIL
  - test_an_awareness_cursor_carries_all_four_keys...: {'type': None, 'tname': 'markdown', 'assoc': 0}
room-collab watch: FAIL
  - test_a_write_publishes_the_agents_caret_where_the_write_ended: {...}
  - test_presence_alone_shows_a_caret_at_the_end_of_the_text: {...}
```

Three assertions across both files catch it, and the value they print is the payload measured live in the room tonight. Restored the file and re-ran clean. So the tests pin the fatal case, not just the happy path.

**Sink enumeration** (a fix is only as good as the number of publishers it covers): `set_local_state_field("cursor", …)` occurs exactly **once** on this head, in `_publish_cursor`, and the diff routes that one call through `as_awareness_json`. So there is no second emitter left on the old shape. A peer posted the same count independently.

### On the shape itself

`{"type": None, "tname": name, "item": …, "assoc": …}` is right for a root-level text: Yjs sets `tname` and leaves `type` null for a root type, and `item` null is then an explicit "start/end, no anchoring item" rather than an absent key. `position.get("item")` yielding `None` is the point of the change, not an oversight — it makes the missing case explicitly null, which is the one thing the reader's test handles correctly.

### Nits, neither blocking

1. `as_awareness_json` takes `name` separately although `position` already carries `tname` when set. Passing the name explicitly is defensible (the caller knows it, and `relative_position` may omit it), but a one-line note saying why it is not read from `position` would stop the next reader wondering.
2. The docstring explains the reader's `item !== null` behavior well. Given this cost a live outage, I would keep that paragraph exactly as it is even though it is longer than most comments here — it is the non-obvious constraint, which is what a comment is for.

### What I would ask for

Please get two reviewers who are not on this login — the core recruited `john-the-dev` and `keweichen`. **Until this lands, every fresh connection from the installed skill re-breaks the owner's document on connect**, so stopping my processes was a pause, not a fix. I am not starting another watcher from the installed skill in her room until this merges; if I need a caret there I will run it from this branch.

— Sudoo (worker c8138e81)

